PR #2, opened 2016-05-04
Initial support for Metal Shading Language.

billhollings:


billhollings:
The original pull request seems to have been nullified because of the move from private to public repos, which caused my fork to detach. I've reworked and submitted this new pull request.

Brenwill has already signed the CLA.

HansKristian-ARM:
And finally I can merge this :D

PR #13, opened 2016-05-18
Flexible array member is not supported in C++11, so emit an array of …

syoyo:
…size 1.

### DIFF
--- a/spirv_cpp.cpp
+++ b/spirv_cpp.cpp
@@ -36,6 +36,64 @@ void CompilerCPP::emit_buffer_block(const SPIRVariable &var)
 	statement("");
 }
 
+string CompilerCPP::type_to_array_cpp(const SPIRType &type)
+{
+	string res;
+	for (auto &size : type.array)
+	{
+		res += "[";
+		if (size)
+			res += convert_to_string(size);
+		else
+			res += "1"; // Flexible array member is prohibited in C++11.
+		res += "]";
+	}
+	return res;
+}
+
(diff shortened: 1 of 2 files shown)


spirv_cpp.cpp
```diff
@@ -36,6 +36,64 @@ void CompilerCPP::emit_buffer_block(const SPIRVariable &var)
...
+	uint64_t memberflags = 0;
+	auto &memb = meta[type.self].members;
+	if (index < memb.size())
+		memberflags = memb[index].decoration_flags;
+
+	return join(layout_for_member(type, index), flags_to_precision_qualifiers_glsl(membertype, memberflags),
+	            type_to_glsl(membertype), " ", to_member_name(type, index), type_to_array_cpp(membertype));
+}
+
+void CompilerCPP::emit_struct(const SPIRType &type)
```
HansKristian-ARM: Is there anything C++ specific about this method? It's not marked as override or virtual, so it looks like it can be removed?

syoyo:
The purpose of this PR is just emit `[1] ` when size is zero  in `type_to_array_glsl`, so you can find a way to fix it more efficient way. It may better to introducing some flag args  in `type_to_array_glsl`. This will  reuse CompilerGLSL code as much as possible.

HansKristian-ARM:
It should be easier to add a flag to CompilerGLSL::BackendVariations. That struct is designed to handle tiny variations like this without having to reimplement virtuals.

HansKristian-ARM:
Implemented it with a BackendVariations check instead.

syoyo:
Cool! Thanks!

PR #32, opened 2016-07-25
For MSL, convert use of global vars inside functions to function args on both callee and caller.

billhollings:
MSL does not support non-constant global vars. GLSL and SPIR-V allow globals to be referenced from within secondary functions. These are converted to function args and passed between functions and function calls.


HansKristian-ARM:
Are you able to run the clang-format script?

spirv_common.hpp
```diff
@@ -708,6 +708,7 @@ struct Meta
 	struct Decoration
 	{
 		std::string alias;
+        std::string qualified_alias;
```
HansKristian-ARM: What is this used for?

spirv_glsl.cpp
```diff
@@ -3908,6 +3917,26 @@ void CompilerGLSL::emit_instruction(const Instruction &instruction)
...
+// to extract static global variables referenced from within the function and convert
+// them to function arguments. This is necessary for shader languages that do not
+// support global access to shader input content from within a function (eg. Metal).
+// Each additional function args uses the name of the global var. Function nesting
+// will modify the functions and calls all the way up the nesting chain.
+string CompilerGLSL::static_func_args(const SPIRFunction &func, uint32_t index)
+{
+    string static_args;
+    auto& args = func.arguments;
+    uint32_t arg_cnt = (uint32_t)args.size();
```
HansKristian-ARM: Prefer C++ style cast uint32_t(x) over C-style.

spirv_msl.cpp
```diff
@@ -138,6 +165,92 @@ void CompilerMSL::localize_global_variables()
...
+    std::set<uint32_t> added_arg_ids;
+    std::set<uint32_t> processed_func_ids;
+    extract_global_variables_from_function(execution.entry_point, added_arg_ids, global_var_ids, processed_func_ids);
+}
+
+// MSL does not support the use of global variables for shader input content.
+// For any global variable accessed directly by the specified function, extract that variable,
+// add it as an argument to that function, and the arg to the added_arg_ids collection.
+void CompilerMSL::extract_global_variables_from_function(uint32_t func_id,
+                                                         std::set<uint32_t>& added_arg_ids,
```
HansKristian-ARM: Does this need to be ordered? std::unordered_set would be better here.

spirv_msl.cpp
```diff
@@ -138,6 +165,92 @@ void CompilerMSL::localize_global_variables()
...
+    {
+        auto &b = get<SPIRBlock>(block);
+        for (auto &i : b.ops)
+        {
+            auto ops = stream(i);
+            auto op = static_cast<Op>(i.op);
+
+            switch (op) {
+                case OpLoad:
+                case OpAccessChain: {
```
HansKristian-ARM: What about OpStore?

HansKristian-ARM:
Overall, I think this approach makes sense, and should be reusable for HLSL backend, assuming it has the same restrictions.

HansKristian-ARM:
This PR has been sitting here for a month without reply, what is the status on this?

billhollings:
Sorry...it's been a busy summer!

This is on my TODO list. I'll deal with it within the next week or two.

...Bill

billhollings:
Is the delay holding anything else up?

...Bill

HansKristian-ARM:
No problem. Just wanted to make sure the PR hadn't been forgotten. It's not holding up anything, so no rush.

spirv_msl.cpp
```diff
@@ -398,6 +510,19 @@ void CompilerMSL::emit_header()
 	statement("");
 	statement("using namespace metal;");
 	statement("");
+    emit_msl_defines();
+}
+
+void CompilerMSL::emit_msl_defines()
+{
+    statement("// Standard GLSL->MSL redefinitions");
+    statement("#define discard discard_fragment()");
```
HansKristian-ARM: I think these things can be done by making CompilerGLSL::emit_instruction virtual in CompilerGLSL. Then MSL can implement the opcodes it needs to handle special, and fallback to parent class implementation for everything else.

spirv_msl.cpp
```diff
@@ -398,6 +510,19 @@ void CompilerMSL::emit_header()
...
+    statement("#define dFdy dfdy");
+    statement("#define dFdx dfdy");
+    statement("#define atan(x,y) atan2((y),(x))");
+    statement("inline bool greaterThan(float2 a,float2 b) { return all(a>b) ? true : false; }");
```
HansKristian-ARM: This seems extremely hackish. greaterThan(float2, float2) should return bool2 here.

spirv_common.hpp
billhollings: MSL disallows inputs and outputs (incl uniforms) to be defined as globals. All input/output vars must be passed as structure members. The qualified_alias iVar allows a type to be assigned an alternate name that includes the input/output structure it has been moved to.

spirv_glsl.cpp
billhollings: Will fix.

spirv_msl.cpp
billhollings: Uncertain. Haven't encountered need yet. Will leave for now.  OpStore might require different arg binding.

spirv_msl.cpp
billhollings: These are remapping defines, not ops passed in from SPIRV. Not clear how you think emit_instruction will be useful here. Can you elaborate?

spirv_msl.cpp
billhollings: Agreed. Being added for specific customer requirements right now. A better approach to mapping. GLSL->MSL functions needs to be developed.

spirv_msl.cpp
```diff
@@ -1276,6 +1432,38 @@ string CompilerMSL::argument_decl(const SPIRFunction::Parameter &arg)
...
+    // Get name and strip any underscore prefix
+    string mbr_name = to_member_name(type, index);
+    size_t startPos = mbr_name.find_first_not_of("_");
+    mbr_name = (startPos != std::string::npos) ? mbr_name.substr(startPos) : "";
+    return join(to_name(type.self), "_", mbr_name);
+}
+
+// Ensures that the specified struct member name is permanently usable by prepending
+// an alpha char if the first chars are _ and a digit, which indicate a transient name.
+string CompilerMSL::ensure_member_name(string mbr_name)
```
billhollings: spirv_msl creates dynamic structs to convert global vars to function arg structs. The name of global is used for struct member. However, if no alpha name is provided and default numeric name is used, when emitting the struct, emit_struct() clears the numeric names and regenerates based on member position in struct. The original global ID form is lost. This causes trouble maintaining references between struct member declarations and accesses. Forcing an alpha name preserves the original numeric names.

spirv_msl.cpp
```diff
@@ -398,6 +515,20 @@ void CompilerMSL::emit_header()
 	statement("");
 	statement("using namespace metal;");
 	statement("");
+	emit_msl_defines();
+}
+
+void CompilerMSL::emit_msl_defines()
+{
+	statement("// Standard GLSL->MSL redefinitions");
+	statement("#define discard discard_fragment()");
```
HansKristian-ARM: Discard would be a terminator, so it indeed does not hit emit_instruction.
However, these kinds of overrides would neatly fit into the BackendVariations, as it's intended to resolve trivial differences between target languages.

Adding a `std::string discard = "discard";` there could solve this. MSL can then override that to `discard_fragment()`.

For the other cases, these are all handled by emit_instruction.

billhollings:
I've updated the pull request to cover the bulk of your requests, including discard behaviour in emit_msl_defines(), and running clang-format.

Regarding emit_msl_defines(), can you clarify and provide some guidance on what you think an spirv_msl override of emit_instruction() would do? Are you thinking emit_instruction()->case OpFunctionCall would handle these kinds of remappings? There are several use-cases covered by these defines:

- Some are simple name mappings (eg. #define dFdx dfdy). Would case OpFunctionCall look for certain GLSL function names and swap them out for an MSL equivalent name?
- Some are restructurings (eg. #define greaterThan(a,b) ((a)>(b))). Would case OpFunctionCall look for these function names and hard code the corresponding output code?
- Some are new functions (eg. imageSize()). Would Would case OpFunctionCall hardcode an inline function somewhere at the top? Or spit out the body of the new function on each invocation?

The primary purpose of the MSL output is to be submitted to an MSL compiler. So we've used #defines here to make use of the MSL compiler to handle these remappings automatically. However, I recognize this degrades readability and adds a bit of overhead to MSL compilation. Some guidance on how you think an emit_instruction() override would behave would help.

PR #36, opened 2016-08-10
Do not try to pump OpLine instructions into a block

RobDangerous:
When the SPIR-V includes metadata for debugging
(aka OpLine instructions) the CompilerError at
line 1588 was eventually triggered.


HansKristian-ARM:
Thanks for the find.

PR #39, opened 2016-08-13
Change variable names beginning with gl_ in GLSL

RobDangerous:
Using old-school GLSL as input containing code ala
"gl_FragColor = whatever" resulted in illegal
declarations ala "out vec4 gl_FragColor;".

### DIFF
--- a/spirv_glsl.cpp
+++ b/spirv_glsl.cpp
@@ -1026,7 +1026,7 @@ void CompilerGLSL::replace_illegal_names()
 			auto &var = id.get<SPIRVariable>();
 			auto &type = get<SPIRType>(var.basetype);
 
-			if (!is_builtin_variable(var))
+			if (!is_builtin_variable(var) && !var.remapped_variable)
 			{
 				auto &m = meta[var.self].decoration;
 				if (m.alias.compare(0, 3, "gl_") == 0)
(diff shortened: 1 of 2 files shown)


spirv_glsl.cpp
```diff
@@ -1017,6 +1017,27 @@ void CompilerGLSL::emit_uniform(const SPIRVariable &var)
...
 }
 
+void CompilerGLSL::replace_illegal_names()
+{
+	for (auto &id : ids)
+	{
+		if (id.get_type() == TypeVariable)
+		{
+			auto &var = id.get<SPIRVariable>();
+			auto &type = get<SPIRType>(var.basetype);
```
HansKristian-ARM: Unused variable?

spirv_glsl.cpp
```diff
@@ -1017,6 +1017,27 @@ void CompilerGLSL::emit_uniform(const SPIRVariable &var)
...
+
+			if (!is_builtin_variable(var))
```
HansKristian-ARM: You also need to check for var.remapped_variable. Remapped variables are not declared, and they must be able to have "reserved" GL names, because they aren't declared.

spirv_glsl.cpp
RobDangerous: Aye. Fixed.

spirv_glsl.cpp
RobDangerous: Interesting feature. OK, works now.

PR #41, opened 2016-08-17
Fix gl_FragData transformation

RobDangerous:
When reading legacy GLSL and writing legacy GLSL
gl_FragData[0] was transformed to gl_FragData[0][0].

This pull request feels like a hack on top of a hack and
there might be edge cases where it fails so I'd be happy
to find a better solution. It fixes an assumably very
common situation though.

### DIFF
--- a/spirv_glsl.cpp
+++ b/spirv_glsl.cpp
@@ -2496,7 +2496,19 @@ string CompilerGLSL::access_chain(uint32_t base, const uint32_t *indices, uint32
 {
 	string expr;
 	if (!chain_only)
+	{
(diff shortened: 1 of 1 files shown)


HansKristian-ARM:
Ye, this solution is kinda hacky. I added a PR which should be cleaner. It solves the problem on my end at least.
https://github.com/KhronosGroup/SPIRV-Cross/pull/42

I'll merge that instead if it solves the issue on your end.

RobDangerous:
Confirmed, works here as well, let's go ahead with your fix.

PR #45, opened 2016-08-26
Implement dead variable elimination.

HansKristian-ARM:
Adds a way to remove unused variables from consideration in reflection and cross-compilation. Example:

```
auto active = compiler->get_active_interface_variables();
res = compiler->get_shader_resources(active);
compiler->set_enabled_interface_variables(move(active));
```


hiddenbitious:
That was super quick :)

Tried a quick test with a couple of trivial shaders and seems to work fine. Redundant variable is still registered though in the generated .cpp, but I think that this is a minor issue, as it will never be used in shader.

```
                VertexResources::init(s);
                s.register_stage_input(a_pos__, 0);
                s.register_stage_input(a_pos2__, 0);

```

HansKristian-ARM:
Sounds trivial to fix.

HansKristian-ARM:
It seems to work for me. --remove-unused-variables removed dead variables for me in C++.

hiddenbitious:
Sorry my mistake. I merged in a local branch that had some conflicting changes. Works good.

PR #56, opened 2016-09-18
Do not add layout qualifiers to variables for legacy GLSL versions

h3xl3r:
Layout qualifiers were added to GLSL ES < 300 and GLSL < 330


spirv_glsl.cpp
```diff
@@ -804,6 +804,9 @@ bool CompilerGLSL::ssbo_is_std430_packing(const SPIRType &type)
 
 string CompilerGLSL::layout_for_variable(const SPIRVariable &var)
 {
+	if (is_legacy_es() || (!options.es && options.version < 330))
```
HansKristian-ARM: I'm a bit unsure about the desktop check here. The 420 pack is used on desktop, so it should work in anything >= 130 with that. Would it make sense to check this only for legacy ES and < 130 desktop?

h3xl3r: Indeed... I was reading up on the 420 pack in the meantime and realize now that this is too general.

Problem is on OS X there's no 420 pack (and probably never will be) - so anything below 330 won't like layout qualifiers and this check is actually just fine. 

How about a compiler option to enable/disable the use of the 420 pack?

HansKristian-ARM: Ye, that makes sense, the 420 define is already added as a conditional define, so it's kinda ugly. I suppose the check could be. If 420 is enabled (default), the legacy check is done, if not, check for core support.

One problem really is that layout() on uniform buffers cannot be used unless 420 is enabled, so this check must be quite variable in what it will pass.

h3xl3r: It's taking some time getting my head around the implications of (optionally) disabling the 420 pack, so I've modified the condition to handle the obvious case for now. 

In that case invalid GLSL was emitted, so this is an improvement even if it doesn't tackle the problem of >= 130 and no 420 pack available (e.g. OS X)

HansKristian-ARM:
Merged, but we'll need a better solution eventually.

PR #61, opened 2016-10-05
Fix compilation on some old C++11 standard libraries.

Kangz:
I would completely understand if you reject the PR, this is for Goobuntu's ancient GCC stdlib.


spirv_glsl.cpp
```diff
@@ -5354,7 +5354,7 @@ string CompilerGLSL::emit_continue_block(uint32_t continue_block)
 	for (auto &s : statements)
 	{
 		if (!s.empty() && s.back() == ';')
-			s.pop_back();
+			s.erase(s.size()-1, 1);
```
HansKristian-ARM: Add spaces: `size() - 1`

Kangz:
Addressed comments, PTAL again.

Issue #70, opened 2016-11-18
Python3 detection not working on Windows


cdwfs:
I get a "Could not find python3" warning when I run cmake for this project on Windows, despite having Python 3.5.2 installed. It seems to be looking for a python3.exe, and I don't see one in the installation directory (just python.exe).

It's not a fatal error, but is there any way to eliminate the spurious warning? Some way to disable testing before the Python3 check, perhaps?

HansKristian-ARM:
Probably makes most sense to add a cmake option to disable/enable testing.

cdwfs:
Should I put together a pull request, or would you like to handle it among yourselves? My cmake-fu is barely above white-belt, but it would be good practice :)

HansKristian-ARM:
I'd prefer a PR. :)

cdwfs:
This PR doesn't add an option to disable tests, but it should at least fix the existing Python test.

PR #72, opened 2016-11-18
Fixed python3 detection for spirv-cross-test.

cdwfs:
Issue #70: Windows Python 3.x installations do not include a python3.exe; this
caused the old test to fail and issue a spurious warning.


HansKristian-ARM:
Thanks.

PR #81, opened 2016-12-05
Defer parenthesis generation until needed.

HansKristian-ARM:
Previously, we would generate parentheses proactively when generating
binary ops, however, this leads to uglier code and hits warnings in
compilers when used as a conditional.


billhollings:
This PR does indeed fix the double parens on if() statements, and avoids the warnings.

However...it causes errors elsewhere. What was originally:

    out.ViewNormal = float3((ciBlock0.ciViewMatrixInverse * float4(out.Normal, 0.0)).xyz);

is now:

    out.ViewNormal = float3(ciBlock0.ciViewMatrixInverse * float4(out.Normal, 0.0).xyz);

which results in the following error:

    program_source:86:58: error: can't convert between vector and non-scalar values ('float4x4' (aka 'matrix<float, 4, 4>') and 'vector_float3' (vector of 3 'float' values))

because the order of the matrix multiplication and `xyz` swizzle have changed.

HansKristian-ARM:
Ok, this is probably a case of a missing to_enclosed_expression somewhere when dealing with swizzles, should be easy to fix.

HansKristian-ARM:
Do you have the GLSL/SPIR-V of a repro case? Tried reproing this with same kind of code, but seems to work as expected for me.

billhollings:
Attached is a ZIP containing a file that includes SPIR-V decompiled text, MSL extracted from SPIR-V, and GLSL extracted from SPIR-V...plus a file of what I believe is the original GLSL.

[SPIRV-Cross-missing_swizzle_parens.zip](https://github.com/KhronosGroup/SPIRV-Cross/files/634316/SPIRV-Cross-missing_swizzle_parens.zip)

HansKristian-ARM:
Ah, ye, it doesn't handle OpCompositeConstruct optimizations correctly, should be a simple fix.

HansKristian-ARM:
Fixed that particular case.

billhollings:
That fixed the issues.

MSL code gen is still triggering:

    program_source:53:70: error: invalid suffix '.xy' on floating constant
        float3 diffuseRgb = uDiffuseTex.sample(uDiffuseTexSmplr, tc * 0.1.xy).xyz;
                                                                     ^
    program_source:54:71: error: invalid suffix '.xy' on floating constant
        float3 normalsRgb = (uNormalsTex.sample(uNormalsTexSmplr, tc * 0.1.xy).xyz * 2.0) - float3(1.0);

But this is caused by `CompilerMSL::emit_texture_op()`, which adds swizzle to tex coords. This error can be fixed by modifying `spirv_msl.cpp:926` from:

    auto coord_expr = to_expression(coord);

to:

    auto coord_expr = to_enclosed_expression(coord);

Either you can add this patch before you commit the PR...or I can add it after.

[SPIRV-Cross-missing_swizzle_parens_2.zip](https://github.com/KhronosGroup/SPIRV-Cross/files/634509/SPIRV-Cross-missing_swizzle_parens_2.zip)

HansKristian-ARM:
Fixed that, and fixed some potential redundant parens being generated in OpCompositeConstruct.

billhollings:
Thanks. It tests out on MSL now.

So...from an MSL perspective, you can merge this branch now.

HansKristian-ARM:
Excellent. I'll write some test shaders which catch the issues you had in MSL before merging tomorrow.

billhollings:
Great. Thanks for getting this done so quickly!

Issue #79, opened 2016-12-03
flatten_interface_block is too limited


zeux:
This library includes a function for flattening UBO for legacy GL targets (flatten_interface_block), but it only works if all members are of the same primitive type (so would work if all of them are float3 for example).

I'd like to change this - currently using SPIRV-Cross on our shader codebase isn't really practical for legacy GLES targets because of this (and even on GLES 3 there are bugs in some Adreno drivers that make using UBO problematic).

My understanding is that the biggest issue with this is that right now flattening can only rewrite the UBO declaration, which requires that SPIRV access chains are compatible between old and new layout. This may be solvable to an extent for simple cases, but ideally this would be generic.

I see two possible approaches for this:

1) Extend flatten_interface_block to be able to rewrite SPIRV access chains. I'm a bit wary of this approach but mostly because I don't have a good understanding of SPIRV yet and am not sure of what rewriting parts of SPIRV code involves.

2) Change flatten_interface_block to merely *mark* UB as "flattened", and extend GLSL generator to rewrite access chains. I did something along these lines once before, in an AST-to-GLSL translator:

https://github.com/zeux/hlslparser/commit/5a2106a754ac6b08fc4f017f26a3faea6b590a20#diff-8fa46791ec7a30a112f1575798cebc19R1311 (click "Load diff" to go to the important part)

The diff is a bit messy because I initially tried to generate a GLSL struct from a uniform buffer but it became apparent that supporting arrays is too inefficient with that approach even with glsl-optimizer running on the result - but it should give you the idea of what roughly I am aiming for. This approach results in generic UBO support that can respect whatever packing rules the UBO in question has, and is trivial to do immediate constant folding (https://github.com/zeux/hlslparser/commit/6b6b919ca3cf7eff9a602536bc6d8e3dfc2f00ba) on so that you can convert accesses like

    Lights[index].data.radius

to

    LightBuffer[5 * index + 4].w

for something like:

    struct Light {
        float4x4 transform;
        struct Data {
            float3 color;
            float radius;
        } data;
    };

Looking for some guidance & discussion on the above - does it seem like a worthwhile feature for SPIRV-Cross to adopt, what approach seems more promising, am I missing some complexities that this would involve due to the nature of SPIRV, etc.

HansKristian-ARM:
[...]

PR #82, opened 2016-12-07
Implement buffer block flattening

zeux:
Legacy GLSL targets do not support uniform buffers, and as such require
some sort of emulation. There are two alternatives - one is to represent
a uniform buffer as a uniform struct, and another one is to flatten it
into an array of primitive vector types (vec4).

Uniform struct have two disadvantages that make using them prohibitive
in some applications:

- The location assignment for struct members is arbitrary which means
the application has to set each struct member one by one
- Some Android drivers fail to link shader programs if both vertex and
fragment shader use the same uniform struct

Because of this, we need to support flattening uniform buffers into an
array. This is not just important for legacy GLSL but also is sometimes
useful for ESSL 3.0 where some Android drivers do not have stable UBO
support.

The way flattening works is the entire buffer is represented as a vec4
array; each access chain is rewritten into a combination of array
accesses, swizzles and data type constructors. Specifically:

- Extracting a vector or a scalar requires indexing into the array with
an optional swizzle, for example CB0[13].yz for reading vec2
- Extracting a matrix or a struct requires extracting each individual
vector or struct member and then combining them into the resulting
object
- Extracting arrays is not supported, mostly because the resulting
construct is very inefficient and ESSL 1.0 does not support array
constructors.

Additionally, while we try to constant-fold each individual indexing
operation, there are cases where we have to use dynamic index
computation (specifically for indexing arrays with non-constants); so
the general form of the primitive array extraction expression is:

    buffer[stride0*index0+...+strideN*indexN+offset]

Where stride/offset are integer literals and index represents variables.

### DIFF
--- a/spirv_glsl.cpp
+++ b/spirv_glsl.cpp
@@ -129,6 +129,23 @@ static uint32_t pls_format_to_components(PlsFormat format)
 	}
 }
 
+static const char* vector_swizzle(int vecsize, int index)
+{
+	static const char* swizzle[4][4] =
+	{
+		{ ".x", ".y", ".z", ".w" },
+		{ ".xy", ".yz", ".zw" },
+		{ ".xyz", ".yzw" },
+		{ "" }
+	};
+
+	assert(vecsize >= 1 && vecsize <= 4);
+	assert(index >= 0 && index < 4);
+	assert(swizzle[vecsize - 1][index]);
+
+	return swizzle[vecsize - 1][index];
+}
+
 void CompilerGLSL::reset()
 {
 	// We do some speculative optimizations which should pretty much always work out,
@@ -3277,13 +3294,10 @@ string CompilerGLSL::access_chain(uint32_t base, const uint32_t *indices, uint32
 {
 	if (flattened_buffer_blocks.count(base))
 	{
-		std::string expr;
-		flattened_access_chain(expr, base, indices, count, target_type, 0);
-
 		if (need_transpose)
 			*need_transpose = false;
 
-		return expr;
+		return flattened_access_chain(base, indices, count, target_type, 0);
 	}
 	else
 	{
@@ -3291,7 +3305,7 @@ string CompilerGLSL::access_chain(uint32_t base, const uint32_t *indices, uint32
 	}
 }
 
-void CompilerGLSL::flattened_access_chain(std::string &expr, uint32_t base, const uint32_t *indices, uint32_t count, const SPIRType &target_type, uint32_t offset)
+std::string CompilerGLSL::flattened_access_chain(uint32_t base, const uint32_t *indices, uint32_t count, const SPIRType &target_type, uint32_t offset)
 {
 	if (!target_type.array.empty())
 	{
@@ -3299,20 +3313,22 @@ void CompilerGLSL::flattened_access_chain(std::string &expr, uint32_t base, cons
 	}
 	else if (target_type.basetype == SPIRType::Struct)
 	{
-		flattened_access_chain_struct(expr, base, indices, count, target_type, offset);
+		return flattened_access_chain_struct(base, indices, count, target_type, offset);
 	}
 	else if (target_type.columns > 1)
 	{
-		flattened_access_chain_matrix(expr, base, indices, count, target_type, offset);
+		return flattened_access_chain_matrix(base, indices, count, target_type, offset);
 	}
 	else
 	{
-		flattened_access_chain_vector_scalar(expr, base, indices, count, target_type, offset);
+		return flattened_access_chain_vector_scalar(base, indices, count, target_type, offset);
 	}
 }
 
-void CompilerGLSL::flattened_access_chain_struct(std::string &expr, uint32_t base, const uint32_t *indices, uint32_t count, const SPIRType &target_type, uint32_t offset)
+std::string CompilerGLSL::flattened_access_chain_struct(uint32_t base, const uint32_t *indices, uint32_t count, const SPIRType &target_type, uint32_t offset)
 {
+	std::string expr;
+
 	expr += type_to_glsl(target_type);
 	expr += "(";
 
@@ -3324,14 +3340,18 @@ void CompilerGLSL::flattened_access_chain_struct(std::string &expr, uint32_t bas
 		const SPIRType &member_type = get<SPIRType>(target_type.member_types[i]);
 		uint32_t member_offset = type_struct_member_offset(target_type, uint32_t(i));
 
-		flattened_access_chain(expr, base, indices, count, member_type, offset + member_offset);
+		expr += flattened_access_chain(base, indices, count, member_type, offset + member_offset);
 	}
 
 	expr += ")";
+
+	return expr;
 }
 
-void CompilerGLSL::flattened_access_chain_matrix(std::string &expr, uint32_t base, const uint32_t *indices, uint32_t count, const SPIRType &target_type, uint32_t offset)
+std::string CompilerGLSL::flattened_access_chain_matrix(uint32_t base, const uint32_t *indices, uint32_t count, const SPIRType &target_type, uint32_t offset)
 {
+	std::string expr;
+
 	expr += type_to_glsl(target_type);
 	expr += "(";
 
@@ -3340,51 +3360,48 @@ void CompilerGLSL::flattened_access_chain_matrix(std::string &expr, uint32_t bas
 		if (i != 0)
 			expr += ", ";
 
-		flattened_access_chain_vector_scalar(expr, base, indices, count, target_type, offset + i * 16);
+		expr += flattened_access_chain_vector_scalar(base, indices, count, target_type, offset + i * 16);
 	}
 
 	expr += ")";
+
+	return expr;
 }
 
-void CompilerGLSL::flattened_access_chain_vector_scalar(std::string &expr, uint32_t base, const uint32_t *indices, uint32_t count, const SPIRType &target_type, uint32_t offset)
+std::string CompilerGLSL::flattened_access_chain_vector_scalar(uint32_t base, const uint32_t *indices, uint32_t count, const SPIRType &target_type, uint32_t offset)
 {
 	if (target_type.basetype != SPIRType::Float)
 		SPIRV_CROSS_THROW("Access chains that use non-floating-point base types can not be flattened");
 
+	auto result = flattened_access_chain_offset(base, indices, count, offset);
+
+	assert(result.second % 4 == 0);
+	uint32_t index = result.second / 4;
+
 	auto buffer_name = to_name(expression_type(base).self);
+
+	std::string expr;
 
 	expr += buffer_name;
 	expr += "[";
-
-	uint32_t result = flattened_access_chain_offset(expr, base, indices, count, offset);
-	assert(result % 4 == 0);
-
-	uint32_t index = result / 4;
-
+	expr += result.first; // this is a series of N1*k1+N2*k2+... that is either empty or ends with a +
 	expr += convert_to_string(index / 4);
 	expr += "]";
 
-	static const char* swizzle[4][4] =
-	{
-		{ ".x", ".y", ".z", ".w" },
-		{ ".xy", ".yz", ".zw" },
-		{ ".xyz", ".yzw" },
-		{ "" }
-	};
+	expr += vector_swizzle(target_type.vecsize, index % 4);
 
-	assert(target_type.vecsize >= 1 && target_type.vecsize <= 4);
-	assert(swizzle[target_type.vecsize - 1][index % 4]);
-
-	expr += swizzle[target_type.vecsize - 1][index % 4];
+	return expr;
 }
 
-uint32_t CompilerGLSL::flattened_access_chain_offset(std::string &expr, uint32_t base, const uint32_t *indices, uint32_t count, uint32_t offset)
+std::pair<std::string, uint32_t> CompilerGLSL::flattened_access_chain_offset(uint32_t base, const uint32_t *indices, uint32_t count, uint32_t offset)
 {
 	const auto *type = &expression_type(base);
 	uint32_t type_size = 0;
 
 	// For resolving array accesses, etc, keep a local copy for poking.
 	SPIRType temp;
+
+	std::string expr;
 
 	for (uint32_t i = 0; i < count; i++)
 	{
@@ -3431,13 +3448,12 @@ uint32_t CompilerGLSL::flattened_access_chain_offset(std::string &expr, uint32_t
 		// Matrix -> Vector
 		else if (type->columns > 1)
 		{
-			if (ids[index].get_type() == TypeConstant)
-			{
-				auto &c = get<SPIRConstant>(index);
-				offset += c.scalar() * 16;
-			}
-			else
+			if (ids[index].get_type() != TypeConstant)
 				SPIRV_CROSS_THROW("Cannot flatten dynamic matrix indexing!");
+
+			index = get<SPIRConstant>(index).scalar();
+
+			offset += index * 16;
 
 			// We have to modify the type, so keep a local copy.
 			if (&temp != type)
@@ -3448,13 +3464,12 @@ uint32_t CompilerGLSL::flattened_access_chain_offset(std::string &expr, uint32_t
 		// Vector -> Scalar
 		else if (type->vecsize > 1)
 		{
-			if (ids[index].get_type() == TypeConstant)
-			{
-				auto &c = get<SPIRConstant>(index);
-				offset += c.scalar() * 4;
-			}
-			else
+			if (ids[index].get_type() != TypeConstant)
 				SPIRV_CROSS_THROW("Cannot flatten dynamic vector indexing!");
+
+			index = get<SPIRConstant>(index).scalar();
+
+			offset += index * 4;
 
 			// We have to modify the type, so keep a local copy.
 			if (&temp != type)
@@ -3466,7 +3481,7 @@ uint32_t CompilerGLSL::flattened_access_chain_offset(std::string &expr, uint32_t
 			SPIRV_CROSS_THROW("Cannot subdivide a scalar value!");
 	}
 
-	return offset;
+	return std::make_pair(expr, offset);
 }
 
 bool CompilerGLSL::should_forward(uint32_t id)
--- a/spirv_glsl.cpp
+++ b/spirv_glsl.cpp
@@ -3316,7 +3316,7 @@ void CompilerGLSL::flattened_access_chain_struct(std::string &expr, uint32_t bas
 	for (size_t i = 0; i < target_type.member_types.size(); ++i)
 	{
 		if (i != 0)
-			expr += ",";
+			expr += ", ";
 
 		const SPIRType &member_type = get<SPIRType>(target_type.member_types[i]);
 		uint32_t member_offset = type_struct_member_offset(target_type, uint32_t(i));
@@ -3335,7 +3335,7 @@ void CompilerGLSL::flattened_access_chain_matrix(std::string &expr, uint32_t bas
 	for (uint32_t i = 0; i < target_type.columns; ++i)
 	{
 		if (i != 0)
-			expr += ",";
+			expr += ", ";
 
 		flattened_access_chain_vector_scalar(expr, base, indices, count, target_type, offset + i * 16);
 	}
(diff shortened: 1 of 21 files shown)


zeux:
This is a WIP change that addresses #79. I want to refactor & submit the part that moves type computation primitives from MSL to GLSL class separately, and the flattening code is missing column/row major matrix handling, base type verification and some other small details.

HansKristian-ARM:
Looking good so far :) This is complex enough that we'll need some dedicated testing I think, so something like:
- In test_shader.py, check for .flatten. as part of the shader.
- Add a --flatten-buffer-blocks to main.cpp and use that in test_shader.py for shaders which contain .flatten.
- Add some dedicated test shaders for testing corner cases in legacy GLSL support.

I can do the plumbing here, but some test shaders with good coverage would be appreciated.

zeux:
Yup, I'm planning to add tests - this is why I started looking at the test framework and setting up Travis build :)

HansKristian-ARM:
Hi, what is the status of this PR?
I just added the default behavior of UBOs being emitted as uniform structs in https://github.com/KhronosGroup/SPIRV-Cross/pull/96 at least.

zeux:
Didn't have a chance to work on this and then the holidays happened; I'll try to finish the remaining tasks soon. #96 looks like it would help this PR as well because of needs_transpose - which will make one the items I had yet to solve easier.

HansKristian-ARM:
Cool, sounds good.

zeux:
Looking at get_declared_type_size more closely, I now realize that there are two separate implementations of get_declared_struct_member_size - one in spirv_cross.cpp and one in spirv_msl.cpp that overrides the base impl. Looking at the code, it's very very close. I think the MSL version has a bug in get_declared_struct_member_size where it doesn't handle arrays at all.

I'm wondering why this is done this way. My inclination is to remove the virtual dispatch from get_declared_struct_member_size and just add get_declared_type_size to base Compiler class - that seems more consistent and eliminates a source of bugs. However, there aren't any tests for MSL so I can't really validate that this doesn't break anything, *and* there must have been a reason for this since AFAICT this virtual dispatch was introduced when MSL was introduced in the first place. Maybe @billhollings can shed some light on this/

zeux:
Also looked a bit into row/column major handling. Doing it for matrices is pretty simple, unfortunately there are corner cases that are tricky - specifically, if you access a specific column (say, 1) in a matrix and the matrix is in a row-major format in a flattened buffer, what needs to happen is smth like:

    vec4(CB[0].y, CB[1].y, CB[2].y, CB[3].y)

or maybe (this doesn't work in GL2/GLES2):

    transpose(mat4(CB[0], CB[1], CB[2], CB[3]))[0]

Which isn't very easy in the current code structure - basically this has to be special cased (for access chains that go "through" a row major matrix but don't stop there). I'll see how much complexity this adds, there's always an option of detecting this and failing to compile the access chain for now.

zeux:
I've added tests. I think this is ready modulo a) rebasing after #96 gets merged (which has multiple conflict points), b) figuring out what to do with MSL changes (this could be merged as is though), c) row-major support (currently flattening explicitly fails for row-major so it's a missing feature)

I'm currently considering doing b) & c) post-merge separately so removing the WIP tag.

zeux:
When writing more tests, I've stumbled across one more issue with arrays. I believe now I understand the code structure a little more, but not the design decisions.

Let's say you have this uniform buffer:

    layout(std140) uniform UBO
    {
        vec4 Data[3][5];
    };

Here's how it's represented in SPIRV:

               6:             TypeFloat 32
               7:             TypeVector 6(float) 4
              13:             TypeInt 32 0
              14:     13(int) Constant 5
              15:             TypeArray 7(fvec4) 14
              16:     13(int) Constant 3
              17:             TypeArray 15 16
         18(UBO):             TypeStruct 17

Note that this has a simple recursive structure - UBO has one member which is an array of an array of a vector of floats. Every single item on this list is a separate id that can have metadata. Indeed, here's the relevant decorations from the shader:

                              Decorate 15 ArrayStride 16
                              Decorate 17 ArrayStride 80

So you can see that the array vec4[5] is annotated to have a stride of 16, and the array vec4[3][5] is annotated to have a stride of 80. It all makes sense.

Now unfortunately, SPIRV-Cross for some reason flattens out the array structure (I initially assumed SPIRV uses this type of specification, but it's not the case - it's just an artificial construct from SPIRV-Cross) with this code in Compiler::parse:

	case OpTypeArray:
	{
		uint32_t id = ops[0];

		auto &base = get<SPIRType>(ops[1]);
		auto &arraybase = set<SPIRType>(id);

		arraybase = base;

		auto *c = maybe_get<SPIRConstant>(ops[2]);
		bool literal = c && !c->specialization;

		arraybase.array_size_literal.push_back(literal);
		arraybase.array.push_back(literal ? c->scalar() : ops[2]);
		// Do NOT set arraybase.self!
		break;

As a result the struct member points (via an indirection) to vec4 directly, thereby resulting in the following structure:

        18: struct type, members: [17]
        17: vec4 type (id 7), array size: [3, 5]

Note that "15" is completely lost - it's now impossible to get the stride of the inner array. The function get_declared_type_size I copied from MSL source also doesn't work for much the same reasons. Technically maybe it shouldn't, since the type wasn't declared. Additionally, there's very important difference between "member_types" and "self" - specifically, if you get a type by id 17, self will be 7! Also the link between matrix->vector and vector->scalar types is destroyed as well, but at least it's not all flattened into "float"...

Now that I understand all this I can conclude that this structure is what led to most of the pain while implementing this feature. You can also see this structure poisoning the rest of the code - e.g. access_chain (and now flattened_access_chain) has this fun way to try to change the types while iterating to reconstruct the recursive structure that *was* there in the first place and got erased by the SPIRV parser. Similarly, in SPIRV-Cross there's a distinction between types that were declared and types that weren't, and querying information from struct members that doesn't work on non-struct-members - SPIRV does not have this but if you try to use the information from SPIRV (via type->self) you may find it was lost.

I've found a way to work around my issue which also means I can revert the changes I made to MSL, but overall all of this is far from ideal.

HansKristian-ARM:
Ah, ok, I think I see what you mean. Yes, so the main reason for why the "self" and the type ID do not match, is that SPIR-V decorations and the type ID are kind of decoupled.

Only the base type (e.g. a plain OpTypeStruct) will have all the reflection information like offsets, names, etc, while the qualifiers like pointers/array did not add any significant information. To avoid a lot of special cases scattered around for qualified types vs plain types, I opted for this style from the beginning.

Up till now, only the outermost array stride was interesting, which could be fetched from the type ID itself. Unfortunately, all information about which type IDs compose the array type are lost, so array strides for partial arrays are not possible to get atm, I didn't consider that scenario.

I think it does make sense to add this information to SPIRType. Along with the "array" list, we should probably store the type IDs which ultimately compose the final array type so we can perform array stride reflection on them, that should help a bit when traversing the access chain.

I will look at your other PR first however.

HansKristian-ARM:
Merged the conflicting PR, should I review this after you've resolved merge conflict?

spirv_glsl.cpp
```diff
@@ -1037,6 +1037,17 @@ void CompilerGLSL::emit_buffer_block(const SPIRVariable &var)
...
 
+void CompilerGLSL::emit_flattened_buffer_block(const SPIRVariable &var)
+{
+	auto &type = get<SPIRType>(var.basetype);
+
+	// Block names should never alias.
+	auto buffer_name = to_name(type.self, false);
+	size_t buffer_size = (get_declared_struct_size(type) + 15) / 16;
+
+	statement("uniform vec4 ", buffer_name, "[", buffer_size, "];");
```
HansKristian-ARM: This doesn't handle flattened blocks of int/uint, but this can be done after the PR.

spirv_glsl.cpp
```diff
@@ -3221,6 +3239,199 @@ string CompilerGLSL::access_chain(uint32_t base, const uint32_t *indices, uint32
...
+
+		return expr;
+	}
+	else
+	{
+		return access_chain(base, indices, count, false);
+	}
+}
+
+void CompilerGLSL::flattened_access_chain(std::string &expr, uint32_t base, const uint32_t *indices, uint32_t count, const SPIRType &target_type, uint32_t offset)
```
HansKristian-ARM: This could probably return a string instead to be more in line with the rest of the methods.

spirv_glsl.cpp
```diff
@@ -3221,6 +3239,199 @@ string CompilerGLSL::access_chain(uint32_t base, const uint32_t *indices, uint32
...
+{
+	if (!target_type.array.empty())
+	{
+		SPIRV_CROSS_THROW("Access chains that result in an array can not be flattened");
```
HansKristian-ARM: I suppose that access chains of access chains in general won't easily work here, right?

spirv_glsl.cpp
```diff
@@ -3221,6 +3239,199 @@ string CompilerGLSL::access_chain(uint32_t base, const uint32_t *indices, uint32
...
+
+void CompilerGLSL::flattened_access_chain_struct(std::string &expr, uint32_t base, const uint32_t *indices, uint32_t count, const SPIRType &target_type, uint32_t offset)
+{
+	expr += type_to_glsl(target_type);
+	expr += "(";
+
+	for (size_t i = 0; i < target_type.member_types.size(); ++i)
+	{
+		if (i != 0)
+			expr += ",";
```
HansKristian-ARM: ", ", should be a space after the comma.

spirv_glsl.cpp
HansKristian-ARM: We could also check if the buffer block is read only, flattening SSBOs is just not going to happen :)

spirv_glsl.cpp
```diff
@@ -3221,6 +3239,199 @@ string CompilerGLSL::access_chain(uint32_t base, const uint32_t *indices, uint32
...
+
+void CompilerGLSL::flattened_access_chain_matrix(std::string &expr, uint32_t base, const uint32_t *indices, uint32_t count, const SPIRType &target_type, uint32_t offset)
+{
+	expr += type_to_glsl(target_type);
+	expr += "(";
+
+	for (uint32_t i = 0; i < target_type.columns; ++i)
+	{
+		if (i != 0)
+			expr += ",";
```
HansKristian-ARM: Space after comma.

spirv_glsl.cpp
```diff
@@ -3221,6 +3239,199 @@ string CompilerGLSL::access_chain(uint32_t base, const uint32_t *indices, uint32
...
+
+		flattened_access_chain_vector_scalar(expr, base, indices, count, target_type, offset + i * 16);
```
HansKristian-ARM: There is a MatrixStride decoration for stride between columns. I think 16 would be correct here, but probably a good idea to not hardcode it.

spirv_glsl.cpp
```diff
@@ -3221,6 +3239,199 @@ string CompilerGLSL::access_chain(uint32_t base, const uint32_t *indices, uint32
...
+
+	for (uint32_t i = 0; i < count; i++)
+	{
+		uint32_t index = indices[i];
+
+		// Arrays
+		if (!type->array.empty())
+		{
+			// We have to modify the type, so keep a local copy.
+			if (&temp != type)
```
HansKristian-ARM: Ok, so instead of this hack, it would probably make sense to save the type of the "previous" array type in OpTypeArray to fix this hack.

```
type = get<SPIRType>(type->previous_array_type);
```
or something similar.

spirv_glsl.cpp
```diff
@@ -3221,6 +3239,199 @@ string CompilerGLSL::access_chain(uint32_t base, const uint32_t *indices, uint32
...
+			temp.array.pop_back();
+
+			assert(type_size > 0);
+			assert(type_size % array_size == 0);
+
+			uint32_t array_stride = type_size / array_size;
+			assert(array_stride % 16 == 0);
+
+			expr += to_expression(index);
+			expr += "*";
```
HansKristian-ARM: Spaces.

spirv_glsl.cpp
```diff
@@ -3221,6 +3239,199 @@ string CompilerGLSL::access_chain(uint32_t base, const uint32_t *indices, uint32
...
+			expr += convert_to_string(array_stride / 16);
+			expr += "+";
```
HansKristian-ARM: Spaces

spirv_glsl.cpp
```diff
@@ -3221,6 +3239,199 @@ string CompilerGLSL::access_chain(uint32_t base, const uint32_t *indices, uint32
...
+		// Matrix -> Vector
+		else if (type->columns > 1)
+		{
+			if (ids[index].get_type() == TypeConstant)
+			{
+				auto &c = get<SPIRConstant>(index);
+				offset += c.scalar() * 16;
+			}
+			else
+				SPIRV_CROSS_THROW("Cannot flatten dynamic matrix indexing!");
```
HansKristian-ARM: Isn't dynamically indexing here very similar to arrays?

spirv_glsl.cpp
```diff
@@ -3221,6 +3239,199 @@ string CompilerGLSL::access_chain(uint32_t base, const uint32_t *indices, uint32
...
+		// Vector -> Scalar
+		else if (type->vecsize > 1)
+		{
+			if (ids[index].get_type() == TypeConstant)
+			{
+				auto &c = get<SPIRConstant>(index);
+				offset += c.scalar() * 4;
+			}
+			else
+				SPIRV_CROSS_THROW("Cannot flatten dynamic vector indexing!");
```
HansKristian-ARM: Isn't dynamically indexing here very similar to arrays?

spirv_glsl.cpp
```diff
@@ -3221,6 +3239,199 @@ string CompilerGLSL::access_chain(uint32_t base, const uint32_t *indices, uint32
...
+
+			// We have to modify the type, so keep a local copy.
+			if (&temp != type)
```
HansKristian-ARM: Hm, seeing a pattern here, might make sense to just store the "previous" type. OpTypeVector also has a reference to the scalar type, OpTypeMatrix has a reference to the vector type, etc, etc ...

spirv_glsl.cpp
```diff
@@ -5331,6 +5542,32 @@ void CompilerGLSL::require_extension(const string &ext)
...
+	auto &var = get<SPIRVariable>(id);
+	auto &type = get<SPIRType>(var.basetype);
+	auto name = to_name(type.self, false);
+	auto flags = meta.at(type.self).decoration.decoration_flags;
+
+	if (!type.array.empty())
+		SPIRV_CROSS_THROW(name + " is an array of UBOs.");
+	if (type.basetype != SPIRType::Struct)
+		SPIRV_CROSS_THROW(name + " is not a struct.");
+	if ((flags & (1ull << DecorationBlock)) == 0)
```
HansKristian-ARM: Ok, seems like the UBO vs SSBO check is done here.

spirv_glsl.cpp
```diff
@@ -5331,6 +5542,32 @@ void CompilerGLSL::require_extension(const string &ext)
...
+		SPIRV_CROSS_THROW(name + " is not a block.");
+	if (type.member_types.empty())
+		SPIRV_CROSS_THROW(name + " is an empty struct.");
+
+	uint64_t member_flags = 0;
+	for (uint32_t i = 0; i < type.member_types.size(); i++)
+		member_flags |= combined_decoration_for_member(type, i);
+
+	if (member_flags & (1ull << DecorationRowMajor))
```
HansKristian-ARM: We could probably support this if the access chain terminates at the matrix, the access chain expression can just get "need_transpose" in that case.

spirv_glsl.cpp
zeux: The major limitation here actually is just that legacy GLSL (at least on GLES I think?) does not support array literals - the natural translation for something like this is `Light[4](Light(UBO[0].xyz, ...`. It's probably possible to work around this if there is a way to emit extra variables but TBH I haven't really seen a shader that copies an array out of the uniform buffers so the value is questionable...

spirv_glsl.cpp
zeux: Yup - if this was the case it would be possible to basically reconstruct the original SPIRV hierarchy of types which would make this simpler because access chains traverse this hierarchy.

spirv_glsl.cpp
zeux: Yeah - I'll need to do this because apparently glslang defaults to row major matrices in HLSL and flips the shader multiplication order, so this PR as it is only works on GLSL sources.

spirv_glsl.cpp
zeux: Yeah, I guess this can work. Not sure how important this is - indexing matrices with non-literals may not be very common...

spirv_glsl.cpp
zeux: Vector indexing is translated to swizzles so this has to be compile-time. It's theoretically possible to synthesize complex expressions such as `dot(UBO[4], lookUpTable[i])` with `vec4 lookUpTable[4]` but it seems like an overkill.

HansKristian-ARM: Dynamic vector lookup is possible in GLSL though, vec4 a; float b = a[index], but might complicate things a lot?

spirv_glsl.cpp
HansKristian-ARM: Not important enough to block a PR.

spirv_glsl.cpp
HansKristian-ARM: iirc, glslang will actually copy out the array element by element anyways in this case as the storage types and "stack" types are different and aliasing is not allowed.

spirv_glsl.cpp
zeux: 16 is basically the only valid value for std140 (GL spec 7.6.2.2 Standard Uniform Block Layout), and the code assumes this in some other places, specifically for array elements. The problem with handling other strides (say, for std430) is that it can result in vectors being "split" across individual elements, which is not supported by the code right now - so it's not enough to just use another value here.

spirv_glsl.cpp
zeux: Looks like ESSL 1.0 is undecided:

> Should dynamic indexing of vectors and matrices be removed from the specification?
> RESOLUTION: Keep in main specification. Support is not mandated.

I'm, uh, not sure what this means exactly :) "use at your own peril" maybe.

HansKristian-ARM: Uh ... ye, nvm then. :p

spirv_glsl.cpp
zeux: I was mostly wary of excessive string copies, and I'd need to return smth like a std::pair from flattened_access_chain_offset. I can do that though if this is more consistent with the rest of the code.

zeux:
Updated with whitespace fixes, code cleanup and row-major support for access chains that end in a matrix. Extra extensions after this PR could include:

- Support for int buffer flattening
- Support for dynamic matrix indexing
- Support for row major matrix indexing (that is, access chains with row-major matrices in the middle)

From my perspective only the last item is interesting (I don't have use cases for the first two); I mocked up an implementation for and it's pretty simple but it can probably be in a separate PR. Please let me know if any other revisions are necessary!

spirv_glsl.cpp-f102ece2
```diff
@@ -0,0 +1,6394 @@
+/*
```
HansKristian-ARM: Commited a temp file?

spirv_glsl.cpp
```diff
@@ -3252,6 +3290,208 @@ string CompilerGLSL::access_chain(uint32_t base, const uint32_t *indices, uint32
...
+	expr += result.first; // this is a series of N1*k1+N2*k2+... that is either empty or ends with a +
+	expr += convert_to_string(index / 4);
+	expr += "]";
+
+	expr += vector_swizzle(target_type.vecsize, index % 4);
+
+	return expr;
+}
+
+std::pair<std::string, uint32_t> CompilerGLSL::flattened_access_chain_offset(uint32_t base, const uint32_t *indices, uint32_t count, uint32_t offset, bool *need_transpose)
```
HansKristian-ARM: Should have clarified, for these functions which returned offset, the expr as a param was fine, it was the access chain ones which returned void which could return string.

spirv_glsl.cpp-f102ece2
zeux: Oops. Too much rebasing. Removed and updated.

spirv_glsl.cpp
zeux: It's actually a bit nicer that way since I can easily reuse flattened_access_chain_offset in an arbitrary location at the call site and then use the resulting expression to make the final one later.

HansKristian-ARM: Ah, ok, let's keep it then.

HansKristian-ARM:
Thanks for this PR.

PR #83, opened 2016-12-07
Add Travis build

zeux:
Builds on both Linux and OSX and tests all shaders for regressions. The
shader set currently does not include any Metal shaders but once it does
we should be able to also validate the resulting MSL using Xcode metal
compiler.

Also fix reference shader.comp and use Python 2 in the regression
testing script because Python 3 does not seem to be readily available on
Travis CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,21 @@
+language: cpp
+os:
+  - linux
+  - osx
+
+# Use Ubuntu 14.04 LTS (Trusty) as the Linux testing environment.
+sudo: required
+dist: trusty
+
+before_script:
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install python3; fi
+  - git clone https://github.com/KhronosGroup/glslang.git glslang
+  - git clone https://github.com/KhronosGroup/SPIRV-Tools SPIRV-Tools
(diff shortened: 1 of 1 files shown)


zeux:
I don't have access to KhronosGroup Travis account but if this gets merged it would be great to enable Travis builds on this repository so that all PRs get regression testing and we don't get mismatching test output.

test_shaders.py
```diff
@@ -1,5 +1,5 @@
-#!/usr/bin/env python3
-
+#!/usr/bin/python
```
HansKristian-ARM: If python is actually python3, will this still work?

HansKristian-ARM: Also, CMakeLists.txt would also have to be updated with python 2 checks instead if both python2/3 will work.

zeux:
Updated to keep Python version at 3 (the file could compile for both versions but it won't be very easy to maintain this if we only test 2...) and instead install Python 3 on Travis for OSX.

HansKristian-ARM:
Updated the shared.comp reference in parallel from last PR. glslang sometimes changes its output. We might want to check out a fixed commit of glslang in Travis and update that at regular intervals to avoid mismatchs.

HansKristian-ARM:
Added the build status to the README, seems to be working :)

PR #84, opened 2016-12-11
Support row-major matrices in access chain when not natively supporte…

billhollings:
…d by backend (MSL).

This has only been applied to OpAccessChain so far, which should cover most cases of incoming row-major matrices. However, I'm not clear if this will cover all requirements to transpose, or whether we will need to transpose during other types of matrix access operations.

### DIFF
--- a/spirv_glsl.cpp
+++ b/spirv_glsl.cpp
@@ -3040,6 +3040,7 @@ string CompilerGLSL::access_chain(uint32_t base, const uint32_t *indices, uint32
 	SPIRType temp;
 
 	bool access_chain_is_arrayed = false;
+	bool mtx_needs_transp = matrix_needs_transposition(base);
 
 	for (uint32_t i = 0; i < count; i++)
 	{
@@ -3092,11 +3093,18 @@ string CompilerGLSL::access_chain(uint32_t base, const uint32_t *indices, uint32
 				expr += ".";
 				expr += to_member_name(*type, index);
 			}
+			mtx_needs_transp = member_matrix_needs_transposition(*type, index);
 			type = &get<SPIRType>(type->member_types[index]);
 		}
 		// Matrix -> Vector
 		else if (type->columns > 1)
 		{
+			if (mtx_needs_transp)
+			{
+				expr = transpose(expr);
+				mtx_needs_transp = false;
+			}
+
 			expr += "[";
 			if (index_is_literal)
 				expr += convert_to_string(index);
@@ -3140,6 +3148,9 @@ string CompilerGLSL::access_chain(uint32_t base, const uint32_t *indices, uint32
 		else
 			throw CompilerError("Cannot subdivide a scalar value!");
 	}
+
+	if (mtx_needs_transp)
+		expr = transpose(expr);
 
 	return expr;
 }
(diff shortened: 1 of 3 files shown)


spirv_glsl.cpp
```diff
@@ -3092,11 +3093,18 @@ string CompilerGLSL::access_chain(uint32_t base, const uint32_t *indices, uint32
 				expr += ".";
 				expr += to_member_name(*type, index);
 			}
+			mtx_needs_transp = member_matrix_needs_transposition(*type, index);
```
HansKristian-ARM: Please use full name here "matrix".

spirv_glsl.cpp
```diff
@@ -3141,6 +3149,9 @@ string CompilerGLSL::access_chain(uint32_t base, const uint32_t *indices, uint32
 			throw CompilerError("Cannot subdivide a scalar value!");
 	}
 
+	if (mtx_needs_transp)
+		expr = transpose(expr);
+
```
HansKristian-ARM: The other case that's missing here is OpLoad from input variables I'd assume.

spirv_glsl.hpp
```diff
@@ -223,6 +223,11 @@ class CompilerGLSL : public Compiler
 	void add_local_variable_name(uint32_t id);
 	void add_resource_name(uint32_t id);
 	void add_member_name(SPIRType &type, uint32_t name);
+
+	bool matrix_needs_transposition(uint32_t id);
+	bool member_matrix_needs_transposition(const SPIRType &type, uint32_t index);
+	virtual std::string transpose(std::string exp_str);
```
HansKristian-ARM: Does this need to be virtual?

spirv_glsl.hpp
```diff
@@ -244,6 +249,7 @@ class CompilerGLSL : public Compiler
 		bool flexible_member_array_supported = true;
 		bool explicit_struct_type = false;
 		bool use_initializer_list = false;
+		bool transpose_row_major_matrices = false;
```
HansKristian-ARM: I think supports_row_major_matrices is a bit clearer, as transposition is an implementation detail.

spirv_glsl.cpp
HansKristian-ARM: Another case is that transposing in the access chain doesn't support OpStore to a transposed matrix.
Although, I'm not sure how important this case is. I'd like to have an exception being thrown that unless the access chain is accessing storage class Input (read-only), UniformConstant (UBO) or PushConstant, row-major is flagged as currently not supported.

billhollings:
I've added OpLoad support, throw an error on non-square matrices, modified the syntax to align with your suggestions, and documented the intention of the virtual function.

I could not come up with a reliable mechanism for detecting when the LHS of an OpStore would violate the row_major format, and throw an exception...so I have left that unsolved for now.

Use cases include including a row_major matrix in a Uniform, but not writing to it...so throwing an error on any writable row_major uniform would be overkill. I think we need some reliable mechanism for detecting when a LHS expression includes a non-writable syntax...such as the embedded transpose() function call that results from this enhancement.

I think a generic mechanism for detecting whether a LHS expression is valid would be a useful addition...but I couldn't come up with such a mechanism.

HansKristian-ARM:
Yup, I have some ideas for how to improve on this, but it's good enough for now.

PR #85, opened 2016-12-13
Add the option to build without exceptions

godlikepanos:


CMakeLists.txt
```diff
@@ -56,17 +62,23 @@ target_include_directories(spirv-cross-core PUBLIC ${CMAKE_CURRENT_SOURCE_DIR})
 if (NOT "${MSVC}")
   set(spirv-compiler-options -std=c++11 -Wall -Wextra -Werror -Wshadow)
   set(spirv-compiler-defines __STDC_LIMIT_MACROS)
+
+  set(spirv-exe-options ${spirv-compiler-options})
```
HansKristian-ARM: I think it makes sense to propagate -fno-exceptions even to executable to be able to test this more easily.

CMakeLists.txt
```diff
@@ -16,6 +16,12 @@ cmake_minimum_required(VERSION 2.8)
 project(SPIRV-Cross)
 enable_testing()
 
+option(SPIRV_CROSS_EXCEPTIONS_TO_ASSERTIONS "Instead of throwing exceptions assert. Only applies to the library targets" OFF)
+
+if(SPIRV_CROSS_EXCEPTIONS_TO_ASSERTIONS)
+	add_definitions(-DSPIRV_CROSS_EXCEPTIONS_TO_ASSERTIONS)
```
HansKristian-ARM: Makes sense to document this in README as well.

Makefile
```diff
@@ -15,7 +15,7 @@ CXXFLAGS += -std=c++11 -Wall -Wextra -Wshadow -D__STDC_LIMIT_MACROS
 ifeq ($(DEBUG), 1)
 	CXXFLAGS += -O0 -g
 else
-	CXXFLAGS += -O2 -g
+	CXXFLAGS += -O2 -DNDEBUG
```
HansKristian-ARM: A trivial ifeq ($(EXCEPTIONS_TO_ASSERTIONS), 1) CXXFLAGS += -D ... -fno-exceptions endif could work fine too.

CMakeLists.txt
godlikepanos: Considered that as well. That will require some refactoring of main.cpp to remove the exceptions from there as well. Happy to do it.

HansKristian-ARM: Ye, makes sense.

spirv_common.hpp
```diff
@@ -17,13 +17,27 @@
...
 #include <functional>
 #include <sstream>
-#include <stdio.h>
-#include <string.h>
 
 namespace spirv_cross
 {
+
+#ifdef SPIRV_CROSS_EXCEPTIONS_TO_ASSERTIONS
+[[noreturn]] inline void report_and_abort(const std::string &msg)
```
HansKristian-ARM: [[noreturn]] breaks MSVC it seems, so an #ifdef is probably in order here.

PR #87, opened 2016-12-19
MSL enhancements

billhollings:
Several enhancements to CompilerMSL for MSL output.

### DIFF
--- a/spirv_msl.cpp
+++ b/spirv_msl.cpp
@@ -51,6 +51,7 @@ string CompilerMSL::compile(MSLConfiguration &msl_cfg, vector<MSLVertexAttr> *p_
 
 	set_enabled_interface_variables(get_active_interface_variables());
 
+	register_custom_functions();
 	extract_builtins();
 	localize_global_variables();
 	add_interface_structs();
@@ -83,6 +84,7 @@ string CompilerMSL::compile(MSLConfiguration &msl_cfg, vector<MSLVertexAttr> *p_
 
 		emit_header();
 		emit_resources();
+		emit_custom_functions();
 		emit_function_declarations();
 		emit_function(get<SPIRFunction>(entry_point), 0);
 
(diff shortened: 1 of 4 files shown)


spirv_msl.cpp
```diff
@@ -96,6 +100,25 @@ string CompilerMSL::compile()
 	return compile(default_msl_cfg, nullptr, nullptr);
 }
 
+// Register the need to output any custom functions.
+void CompilerMSL::register_custom_functions()
```
HansKristian-ARM: traverse_reachable_opcodes would be a bit better here as here we'd emit mod() even if a function isn't actually used.

spirv_msl.hpp
```diff
@@ -144,10 +147,11 @@ class CompilerMSL : public CompilerGLSL
 	uint32_t get_ordered_member_location(uint32_t type_id, uint32_t index);
 	uint32_t pad_to_offset(SPIRType &struct_type, bool is_indxd_vtx_input, uint32_t offset, uint32_t struct_size);
 	SPIRType &get_pad_type(uint32_t pad_len);
-	size_t get_declared_type_size(const SPIRType &type) const;
-	size_t get_declared_type_size(const SPIRType &type, uint64_t dec_mask) const;
+	size_t get_declared_type_size(uint32_t type_id) const;
+	size_t get_declared_type_size(uint32_t type_id, uint64_t dec_mask) const;
 
 	MSLConfiguration msl_config;
+	std::unordered_set<uint32_t> custom_function_ops;
```
HansKristian-ARM: I think the ordering matters here as helper-functions will be emitted in arbitrary order, which would break regression runs, as they need the output to be exact every time. Emitting to a std::vector sounds appropriate here as I doubt we'd have that many custom function ops.

spirv_msl.cpp
billhollings: Changed to use traverse_all_reachable_opcodes().

spirv_msl.hpp
billhollings: Changed to use std::set instead of std::unordered_set, to retain ordering...but also only record the need for each function only once. Using std::vector results in the potential output of multiple copies of the declaration and definition of a custom function.

PR #90, opened 2017-01-05
Fixes for casting\truncation

gwihlidal:


main.cpp
```diff
@@ -235,7 +235,7 @@ static void print_resources(const Compiler &compiler, const char *tag, const vec
 		if (mask & (1ull << DecorationInputAttachmentIndex))
 			fprintf(stderr, " (Attachment : %u)", compiler.get_decoration(res.id, DecorationInputAttachmentIndex));
 		if (is_sized_block)
-			fprintf(stderr, " (BlockSize : %u bytes)", block_size);
+			fprintf(stderr, " (BlockSize : %zu bytes)", block_size);
```
HansKristian-ARM: %zu doesn't work on Windows unless recent msvcrt is used, I'd keep it as unsigned.

spirv_cross.cpp
```diff
@@ -1018,6 +1018,9 @@ uint32_t Compiler::get_decoration(uint32_t id, Decoration decoration) const
 		return dec.input_attachment;
 	case DecorationSpecId:
 		return dec.spec_id;
+	case DecorationNonWritable:
+	case DecorationNonReadable:
+		return 1;
 	default:
```
HansKristian-ARM: The default case should just return 1 here.
This needs to be implemented for get_member_decoration as well.
The header should be updated to mention this new behavior.

gwihlidal: Sounds great; making the changes now.

main.cpp
```diff
@@ -217,7 +217,7 @@ static void print_resources(const Compiler &compiler, const char *tag, const vec
 
 		uint32_t block_size = 0;
 		if (is_sized_block)
-			block_size = compiler.get_declared_struct_size(compiler.get_type(res.base_type_id));
+			block_size = (uint32_t)compiler.get_declared_struct_size(compiler.get_type(res.base_type_id));
```
HansKristian-ARM: Minor nit, but uint32_t(x) instead of (uint32_t)x.

PR #94, opened 2017-01-11
Use the classic locale when converting floats to string

eoma:
When used in a locale-sensitive setting, eg. GUI application, the float to string conversions will necessarily use the current locale. This may result in floating point numbers with eg. commas as a decimal mark.

This patch attempts to avoid these locale specific conversions by temporarily overriding the current locale with a classic locale (C locale).

I am not sure if this is the best way.

### DIFF
--- a/spirv_common.hpp
+++ b/spirv_common.hpp
@@ -20,6 +20,7 @@
 #include <cstdio>
 #include <cstring>
 #include <functional>
+#include <locale>
 #include <sstream>
 
 namespace spirv_cross
@@ -905,6 +906,22 @@ struct Meta
 // name_of_type is the textual name of the type which will be used in the code unless written to by the callback.
 using VariableTypeRemapCallback =
     std::function<void(const SPIRType &type, const std::string &var_name, std::string &name_of_type)>;
+
+class ClassicLocale
+{
+public:
+	ClassicLocale()
+	{
+		old = std::locale::global(std::locale::classic());
+	}
+	~ClassicLocale()
+	{
+		std::locale::global(old);
+	}
+
+private:
+	std::locale old;
+};
 }
 
 #endif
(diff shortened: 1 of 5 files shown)


spirv_common.hpp
```diff
@@ -107,7 +107,11 @@ inline std::string convert_to_string(float t)
 	// std::to_string for floating point values is broken.
 	// Fallback to something more sane.
 	char buf[64];
+
+	auto old = std::locale::global(std::locale::classic());
```
HansKristian-ARM: This seems a bit inefficient to do for every convert_to_string. I'd have this in ::compile(), wrap it in an RAII class and then go back to the existing locale in the destructor.

HansKristian-ARM: Are you sure std::locale is the right thing to do though? This is a C function, does std::locale affect it?

eoma:
Addresses comments. Added a simple RAII-style ClassicLocale style class. Added an instance to each Compile*::compile() function.

spirv_common.hpp
eoma: According to [the docs](http://en.cppreference.com/w/cpp/locale/locale/global), std::local::global will set the C locale as well if the supplied locale has a name.

eoma:
According to [std::local::global](http://en.cppreference.com/w/cpp/locale/locale/global), it will set the locale if the locale has a (valid) name.

HansKristian-ARM:
Hm, http://en.cppreference.com/w/cpp/locale/locale/global says it changes the C locale if the locale has a name ... whatever that means. Can you confirm this fixes locale issues on locales which use comma instead of point for you?

eoma:
Yes, this fixes my issue.

HansKristian-ARM:
Alright.

PR #96, opened 2017-01-13
Legacy GLSL improvements

HansKristian-ARM:
This improves handling of transpose in targets which do not support row_major layouts natively. The common use-case of row_major_matrix * vector will now be written as vector * matrix instead to avoid the transpose().

UBOs are now emitted as uniform structs by default.

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -27,7 +27,7 @@
 #include <unordered_set>
 
 #ifdef _MSC_VER
-#pragma warning(disable: 4996)
+#pragma warning(disable : 4996)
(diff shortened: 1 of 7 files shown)


HansKristian-ARM:
@billhollings does this help for MSL? It seems like MSL supports the distinction between matrix * vector and vector * matrix.

HansKristian-ARM:
Merging this. Unblocks https://github.com/KhronosGroup/SPIRV-Cross/pull/82 and it's a fairly minor change.

PR #97, opened 2017-01-15
Metal additions/fixes

RobDangerous:
This includes some Metal. As per the author:

Implemented:
- uniform struct
- increase attribute index to identify each vertex input
- shader signature matching (e.g. [[user(locn0)]])

It was all done by @CatCuddler - I think she can comment some more on Monday. Also calling @billhollings to have a look - there was some repeated work, didn't expect you to get back to it right when we started.


spirv_msl.hpp
```diff
@@ -156,10 +162,13 @@ class CompilerMSL : public CompilerGLSL
 	MSLResourceBinding next_metal_resource_index;
 	uint32_t stage_in_var_id = 0;
 	uint32_t stage_out_var_id = 0;
+	uint32_t stage_uniforms_var_id = 0;
 	std::string qual_pos_var_name;
 	std::string stage_in_var_name = "in";
 	std::string stage_out_var_name = "out";
+	std::string stage_uniform_var_name = "uniforms";
 	std::string sampler_name_suffix = "Smplr";
+	std::vector<std::string> reserved_names = { "kernel", "bias" };
```
HansKristian-ARM: Any reason this is not a static const unordered set or something?

spirv_cross.cpp
```diff
@@ -998,6 +998,12 @@ uint64_t Compiler::get_decoration_mask(uint32_t id) const
 	return dec.decoration_flags;
 }
 
+bool Compiler::is_decoration_set(uint32_t id, spv::Decoration decoration) const
```
HansKristian-ARM: get_decoration_mask does something very similar, do we need another interface to check decorations?

spirv_msl.cpp
```diff
@@ -19,6 +19,8 @@
 #include <algorithm>
 #include <numeric>
 
+#include <iostream>
```
HansKristian-ARM: Where are iostreams used?

spirv_msl.cpp
```diff
@@ -309,6 +325,11 @@ uint32_t CompilerMSL::add_interface_block(StorageClass storage)
 		break;
 	}
 
+	case StorageClassUniformConstant:
+	{
+		ib_var_ref = stage_uniform_var_name;
```
HansKristian-ARM: break?

spirv_msl.cpp
```diff
@@ -372,13 +397,17 @@ uint32_t CompilerMSL::add_interface_block(StorageClass storage)
...
+			{
+				set_member_decoration(ib_type.self, ib_mbr_idx, DecorationLocation, locn);
+			}
 			mark_location_as_used_by_shader(locn, storage);
 
 			// Mark the member as builtin if needed
 			if (is_builtin_variable(*p_var))
 			{
 				set_member_decoration(ib_type_id, ib_mbr_idx, DecorationBuiltIn, dec.builtin_type);
+				is_builtin_variable(*p_var);
```
HansKristian-ARM: Redundant?

spirv_msl.cpp
```diff
@@ -1004,9 +1052,11 @@ void CompilerMSL::emit_fixup()
 	{
 		if (options.vertex.fixup_clipspace)
 		{
-			const char *suffix = backend.float_literal_suffix ? "f" : "";
+			/*const char *suffix = backend.float_literal_suffix ? "f" : "";
 			statement(qual_pos_var_name, ".z = 2.0", suffix, " * ", qual_pos_var_name, ".z - ", qual_pos_var_name,
-			          ".w;", "    // Adjust clip-space for Metal");
+			          ".w;", "    // Adjust clip-space for Metal");*/
+			statement(qual_pos_var_name, ".z = (", qual_pos_var_name, ".z + ", qual_pos_var_name,
```
HansKristian-ARM: If I read it right, this compacts [-1, 1] range to [0, 1] instead of the opposite, any particular reason this change is done?

spirv_msl.hpp
CatCuddler: Since i simply added a prefix to all reserved names, this vector can be a const. Or, like you suggested, i change it to a unordered_map.

spirv_cross.cpp
CatCuddler: Sorry, i didn't see this function before, that's why i implemented a new one. Of course, we don't need two similar functions.

spirv_msl.cpp
CatCuddler: I used it to debug... not needed anymore.

spirv_msl.cpp
CatCuddler: OMG such a stupid mistake. I'm ashamed. :S I will add a break and hopefully never forget it again.

spirv_msl.cpp
CatCuddler: Something strange happened when i tried to merge... I will delete this line.

spirv_msl.hpp
HansKristian-ARM: This could just be moved into the member that does the checking, just like GL backend as a static const.

spirv_msl.cpp
CatCuddler: I hope i understood your question correctly. Metal normalized device coordinate is a cuboid [-1,1]x[-1,1]x[0,1].  For x and y the coordinates are in range between -1 and 1. But the z coordinate is in range between 0 and 1.

billhollings:
Since this is a fairly substantial change to MSL functionality, please delay merging this PR until I have had a chance to review it in more detail, and to test it against MoltenVK.

As Rob mentions, at first glance, there seems to be some duplication of...or changes to...existing behaviour, which I'd like to explore in more detail. This week is a busy one for me...but I'll try to get a full review done in the next few days.

To help with that...if possible, can either @RobDangerous or @CatCuddler add a comment to describe the intended behaviour being added, beyond Rob's point-form above, please?

HansKristian-ARM:
@billhollings Ack.
Also @CatCuddler, you need to sign the CLA, or we cannot merge this PR.

CatCuddler:
Ok, here is what i did:

1. Create a uniform structure: this is done the same way as the implementation of the input and output structures (simple check if a structure of a var is equal to StorageClassUniformConstant...). When a structure has 0 members, it wont be added.
2. Increase attribute index: Actually I just check if DecorationLocation is set. Previously the location was always set to 0, even when the location mask was not set. Thus there was always [[attribute(0)]].
3. Set main function name. Actually this is a duplication. My idea was to extract the filename of vertex or fragment shader (e.g. billboard.frag.glsl) and to create the function name (so, something like billboard_frag_main). If the function name is not explicitly set, then the main function will be overriden by main0.
4. Change buffer index in vertex shader. To access uniform in vertex shader we need [[buffer(1)]].
5. Normalize z value. Because metal normalized device coordinate is a cuboid [-1,1]x[-1,1]x[0,1], we have to normalize the z coordinate between 0 and 1.
6. Flexible pairing between vertex and fragment: simply sort variables alphabetically so that the signature will match. 
7. Check if a function argument is a uniform pointer. If we have a custom function and one argument is a uniform variable, then we dont need a pointer for this one variable.
8. Check variable name. Name such as “kernel” or “basis” are reserved. Add a prefix 'm'.
9. Check function name. I removed this already, because it was a duplication.
10. Global vars: this was already done, but there was a small mistake. Because metal does not allow global variables, we have to declare all in main function. This global, actually now local vars (which are accessed by a function) also have to be added to custom functions. There is one more thing: because one function is processed only once, it could happen that not all variables are added. For this reason i save global vars of each processed function and return it later if needed.

zeux:
> Change buffer index in vertex shader. To access uniform in vertex shader we need [[buffer(1)]].

This assumes that the ObjC code performs the same adjustment and there's just one vertex buffer bound to slot 0 - neither of those is going to be the case for every app. This can be exposed as a setting (e.g. an offset that shifts *all* buffer bindings - not just 0!), but in general I recommend binding the vertex buffers with an offset that's big enough so that it's not likely to overlap with explicit buffers - that is, you can bind the vertex buffer at a slot 30 instead of slot 0 (UE4 binds from 30 down, so vertex stream 1 goes to slot 29 IIRC; in our app we bind from 16 up).

HansKristian-ARM:
Normalizing the Z coord for Metal doesn't seem to make much sense at all considering Vulkan is already  [0, 1] range for depth anyways.

billhollings:
@zeux 

That 30-descending mapping is exactly how MoltenVK handles vertex attribute buffers as well.

`CompilerMSL` supports passing a `MSLResourceBinding` structure that tells the shader converter how to map the descriptor sets and bindings to Metal resource indexes for uniform buffers, textures and shaders, in order to support alignment with the corresponding assignments within the app.

RobDangerous:
We don't use this to map from Vulkan to Metal. We map from OpenGL style GLSL to Metal and therefore need an option for the Z coord normalization.

billhollings:
@RobDangerous 

I think your logic is correct. The existing logic was just an early port of the GLSL version, has the wrong (in fact opposite) logic, and was never used.

When using this, how do you set `fixup_clipspace` to activate that conversion. It is initialized to `false` in `CompilerMSL`, and never changed.

RobDangerous:
That happens outside of SPIRV-Cross, it's in https://github.com/Kode/krafix/blob/master/Sources/MetalTranslator2.cpp (we treat SPIRV-Cross as a library).

billhollings:
@CatCuddler 

Thanks for your contributions...and the design explanations. I'll make some comments here, and maybe more specific comments in the PR code sections online.

1. The SPIR-V spec identifies only `StorageClassInput` and `StorageClassOutput` variables in `OpEntryPoint`. Can you clarify how a struct with`StorageClassUniformConstant` arrives in `OpEntryPoint`? Do you have example shader snippets you can share?

2. This breaks MoltenVK. The test needs to be `& (1ull << DecorationLocation)` instead of `== DecorationLocation`. And I'm curious...under what conditions are you not getting the location from SPIR-V?

3. For consistency, can we pass the alternate main function name in `MSLConfiguration`?

4. Per @zeux and my comments above, we need more flexibility than this. This flexibility is provided by the incoming vector of `MSLResourceBinding` struct.

5. As I mention above to @RobDangerous, you're right, the existing logic was wrong (and never used). Can you clean up the old commented-out code, please?

6. While there is nothing wrong with sorting alphabetically, Metal won't use that to match vertex output and fragment input (it matches names regardless of order (maybe you've encountered an MSL bug?)...and the names may be different if coming from different sources). `CompilerMSL` uses the `[[user(locnN)]]` qualifier to ensure matching. See (2) above...SPIR-V should be providing this location value (or you could set it earlier in the processing).

7. Thanks.

8. I've seen this occasionally too in var names...and with functions. Can we handle this in the same way as `func_name_overrides`?
	- Make the collection a component of the instance so that it can be modified by any possible subclasses or an API (unless we know the list is fixed and we've caught them all).
	- Append `0` instead of prepending `m`. For consistency...and the `0` makes it more obvious that it is a system-generated mod (the prepended `m` for numeric-only members is because they are members...I guess we can rename `ensure_valid_name` to make this clearer).

9. Is this to avoid conflict with MSL standard library functions? If so...which were you trying to avoid? The `func_name_overrides` map is meant to handle this.

10. Thanks. I haven't seen this...but if you have...then this is a good change.

Misc:
- Remove unused `MSLConfiguration::vtx_attr_stage_in_binding` member.
- Replace `ib_type.self` with `ib_type_id` in `add_interface_block`.

spirv_cross.cpp
```diff
@@ -2378,8 +2378,10 @@ SPIREntryPoint &Compiler::get_entry_point()
 bool Compiler::interface_variable_exists_in_entry_point(uint32_t id) const
 {
 	auto &var = get<SPIRVariable>(id);
-	if (var.storage != StorageClassInput && var.storage != StorageClassOutput)
-		SPIRV_CROSS_THROW("Only Input and Output variables are part of a shader linking interface.");
+	if (var.storage != StorageClassInput && var.storage != StorageClassOutput &&
+	    var.storage != StorageClassUniformConstant)
```
billhollings: The SPIR-V spec identifies only `StorageClassInput` and `StorageClassOutput` variables in `OpEntryPoint`. Can you clarify how a struct with`StorageClassUniformConstant` arrives in `OpEntryPoint`? Do you have example shader snippets you can share?

spirv_msl.cpp
```diff
@@ -1394,9 +1458,17 @@ string CompilerMSL::to_qualified_member_name(const SPIRType &type, uint32_t inde
 string CompilerMSL::ensure_valid_name(string name, string pfx)
 {
 	if (name.size() >= 2 && name[0] == '_' && isdigit(name[1]))
+	{
 		return join(pfx, name);
+	}
+	else if (std::find(reserved_names.begin(), reserved_names.end(), name) != reserved_names.end())
```
billhollings: Can we handle this in the same way as `func_name_overrides`?
	- Make the collection a component of the instance so that it can be modified by any possible subclasses or an API (unless we know the list is fixed and we've caught them all).
	- Append `0` instead of prepending `m`. For consistency...and the `0` makes it more obvious that it is a system-generated mod (the prepended `m` for numeric-only members is because they are members...I guess we can rename `ensure_valid_name` to make this clearer).

spirv_msl.cpp
```diff
@@ -372,7 +405,9 @@ uint32_t CompilerMSL::add_interface_block(StorageClass storage)
 			// Copy the variable location from the original variable to the member
 			auto &dec = meta[p_var->self].decoration;
 			uint32_t locn = dec.location;
-			set_member_decoration(ib_type_id, ib_mbr_idx, DecorationLocation, locn);
+			if (get_decoration_mask(p_var->self) == DecorationLocation) {
```
billhollings: This test should be `& (1ull << DecorationLocation)` instead of `== DecorationLocation`. And I'm curious...under what conditions are you not getting the location from SPIR-V?

spirv_msl.cpp
billhollings: This change is correct. The existing logic was wrong (and never used). Can you clean up the old commented-out code, please?

spirv_msl.cpp
```diff
@@ -1336,6 +1394,8 @@ uint32_t CompilerMSL::get_metal_resource_index(SPIRVariable &var, SPIRType::Base
 	switch (basetype)
 	{
 	case SPIRType::Struct:
+		if (execution.model == ExecutionModelVertex && next_metal_resource_index.msl_buffer == 0)
+			next_metal_resource_index.msl_buffer = 1;
```
billhollings: This is a hard-coded special case that will break in the general sense. What is the issue you are trying to fix with this change?

See also the MSLResourceBinding API struct for the general mechanism for configuring resource indexes.

spirv_msl.cpp
```diff
@@ -393,6 +428,15 @@ uint32_t CompilerMSL::add_interface_block(StorageClass storage)
...
 	memberSorter.sort();
 
+	// Sort input or output variables alphabetical
+	auto &execution = get_entry_point();
+	if ((execution.model == ExecutionModelFragment && storage == StorageClassInput) ||
+	    (execution.model == ExecutionModelVertex && storage == StorageClassOutput))
+	{
+		MemberSorter memberSorter(ib_type, meta[ib_type.self], MemberSorter::Alphabetical);
+		memberSorter.sort();
+	}
```
billhollings: While there is nothing wrong with sorting alphabetically, Metal won't use that to match vertex output and fragment input (it matches names regardless of order (maybe you've encountered an MSL bug?)...and the names may be different if coming from different sources). `CompilerMSL` uses the `[[user(locnN)]]` qualifier to ensure matching. See (2) above...SPIR-V should be providing this location value (or you could set it earlier in the processing).

spirv_msl.hpp
```diff
@@ -29,8 +29,9 @@ namespace spirv_cross
 // Options for compiling to Metal Shading Language
 struct MSLConfiguration
 {
-	bool flip_vert_y = true;
-	bool flip_frag_y = true;
+	uint32_t vtx_attr_stage_in_binding = 0;
```
billhollings: Remove unused `MSLConfiguration::vtx_attr_stage_in_binding` member.

spirv_msl.hpp
```diff
@@ -92,6 +93,8 @@ class CompilerMSL : public CompilerGLSL
 	// Compiles the SPIR-V code into Metal Shading Language using default configuration parameters.
 	std::string compile() override;
 
+	void set_entry_point_name(std::string func_name);
+
```
billhollings: For consistency, can we pass an alternate main function name in `MSLConfiguration` instead?

spirv_msl.cpp
```diff
@@ -372,7 +405,9 @@ uint32_t CompilerMSL::add_interface_block(StorageClass storage)
 			// Copy the variable location from the original variable to the member
 			auto &dec = meta[p_var->self].decoration;
 			uint32_t locn = dec.location;
-			set_member_decoration(ib_type_id, ib_mbr_idx, DecorationLocation, locn);
+			if (get_decoration_mask(p_var->self) == DecorationLocation) {
+				set_member_decoration(ib_type.self, ib_mbr_idx, DecorationLocation, locn);
+			}
```
billhollings: Replace `ib_type.self` with `ib_type_id` in `add_interface_block()`.

spirv_cross.cpp
CatCuddler: Because `add_interface_block(StorageClassUniformConstant)` is called to create uniform structure. Then in the `add_interface_block` method, `interface_variable_exists_in_entry_point` is called and thus compiler error thrown.

spirv_msl.cpp
CatCuddler: Hmm, thats exactly why i sort alphabetically. I don’t know how exactly members are sorted now. But in `CompilerGLSL::emit_struct` an index i is used to set the position, and since members in vertex and fragment shaders are not in the same order, this won’t work. Try it with [painter-text.frag.glsl](https://github.com/Kode/krafix_tests/blob/a5aefcd66fb1ddc575cfb5a38e24587319ff49d4/in/painter-text.frag.glsl) and [painter-text.vert.glsl](https://github.com/Kode/krafix_tests/blob/a5aefcd66fb1ddc575cfb5a38e24587319ff49d4/in/painter-text.vert.glsl). Without sorting alphabetically i get for vertex:
```
struct painter_text_vert_main_out
{
    float2 texCoord [[user(locn0)]];
    float4 fragmentColor [[user(locn1)]];
};
```

and fragment:
struct painter_text_frag_main_in
```
{
    float4 fragmentColor [[user(locn0)]];
    float2 texCoord [[user(locn1)]];
}
```

The fragment input does not match the corresponding vertex shader output. Perhaps i should better look how members are ordered until now.

spirv_msl.cpp
CatCuddler: Ok i fixed this.

spirv_msl.cpp
CatCuddler: Fixed.

spirv_msl.cpp
CatCuddler: When i did this, i did not saw that i can pass `MSLResourceBinding` to set buffer index. I removed this two lines.

spirv_msl.cpp
CatCuddler: I changed this, like you suggested.

spirv_msl.hpp
CatCuddler: Main function name can be now set as `MSLConfiguration.entry_point_name`.

spirv_msl.hpp
CatCuddler: I don't know where this came from. I removed it.

billhollings:
@CatCuddler 

Thanks for your updates. I'll review later today.

Github is reporting that a conflict needs resolution.

spirv_cross.cpp
billhollings: Thanks.

spirv_msl.cpp
billhollings: The value of _`N`_ in the `[[user(locnN)]]` qualifier is determined by the `get_ordered_member_location()` function from the `DecorationLocation` decoration as specified in the incoming SPIR-V.

If no `DecorationLocation` decoration has been specified, then the `get_ordered_member_location()` function uses the index of the member in the structure.

Sorting alphabetically seems like quite a fragile design that depends on very deliberate agreement between the exact order of the members of the vertex output structure and the vertex input structure.

In the absence of `DecorationLocation` decorations, the alphabetic sorting will only work in the special case where the member indexes align between the vertex output and fragment input structures. In your example, fortunately, the builtins are sorted so that they appear at the end of the structure, otherwise the `gl_Position` member would interfere with the member index of the `texCoord` member, because a strict alphabetic sorting would result in:

```
struct painter_text_vert_main_out
{
    float4 fragmentColor [[user(locn0)]];
    float2 texCoord [[user(locn2)]];
};
```
Notice that `texCoord ` is now set to `[[user(locn2)]]`, because it now sits at member index 2 in this structure.

So...in the general case, any variation between the content of the two structures ahead of the member you are trying to assign a `[[user(locnN)]]` qualifier to will cause mismatches. This might occur, for example, if the vertex output structure outputs more than the fragment input structure uses...such as when a general vertex shader is used with several different fragment shaders.

Is there some reason why your SPIR-V is not specifying a `DecorationLocation` decoration for these inputs and outputs?

In the absence of an explicit `DecorationLocation` decoration, a  more reliable method of matching might be to let Metal perform the matching using names.

We could do this by modifying the `get_ordered_member_location()` function to return an indication that a location has not been defined (maybe return `-1`), and then have the `member_attribute_qualifier()` function return an empty string for that member. Metal will then revert to matching by member names between the two structures.

billhollings:
I have added a lengthy comment to the alphabetic sorting change, because, in the general sense, I don't believe it will solve the issue it is meant to address.

However...if location info is available (which I expect is the typical case), there is nothing inherently wrong with sorting alphabetically...so I have accepted the changes, so that we can move forward with this rather large PR.

However...we SHOULD address my design concerns at some point. If no location info is defined for an input/output variable...then it would be better to avoid outputting the `[[user(locnN)]]` qualifier altogether...rather than relying on a generally fragile alphabetic-plus-index alignment.

billhollings:
And of course...the version conflicts need to be addressed too before the PR can be merged.

RobDangerous:
Addressed the conflicts.

spirv_msl.cpp
```diff
@@ -158,10 +170,23 @@ void CompilerMSL::extract_global_variables_from_functions()
...
+			{
 				global_var_ids.insert(var.self);
+			}
 		}
 	}
 
+	// Local vars that are declared in the main function and accessed directy by a function
+	auto &entry_func = get<SPIRFunction>(entry_point);
+	auto iter = entry_func.local_variables.begin();
+	while (iter != entry_func.local_variables.end())
```
HansKristian-ARM: Pretty sure this can be simplified to
```
for (auto &var : entry_func.local_variables)
   global_var_ids.insert(var);
```

spirv_msl.cpp
```diff
@@ -1232,12 +1288,15 @@ string CompilerMSL::entry_point_args(bool append_comma)
 			auto &type = get<SPIRType>(var.basetype);
 
 			if ((var.storage == StorageClassUniform || var.storage == StorageClassUniformConstant ||
-			     var.storage == StorageClassPushConstant) &&
-			    !is_hidden_variable(var))
```
HansKristian-ARM: Why is the is_hidden_variable check removed?

spirv_msl.cpp
```diff
@@ -1353,8 +1413,12 @@ string CompilerMSL::argument_decl(const SPIRFunction::Parameter &arg)
 	auto &type = expression_type(arg.id);
 	bool constref = !type.pointer || arg.write_count == 0;
 
+	// TODO: Check if this arg is an uniform pointer
+	bool pointer = type.storage == StorageClassUniformConstant;
```
HansKristian-ARM: Why does UniformConstant storage class imply pointer here?

spirv_msl.hpp
```diff
@@ -115,22 +116,26 @@ class CompilerMSL : public CompilerGLSL
...
 	                             uint32_t grad_y, uint32_t lod, uint32_t coffset, uint32_t offset, uint32_t bias,
 	                             uint32_t comp, uint32_t sample, bool *p_forward) override;
-	std::string clean_func_name(std::string func_name) override;
 
 	void preprocess_op_codes();
 	void emit_custom_functions();
 	void localize_global_variables();
 	void extract_global_variables_from_functions();
+
+	std::unordered_map<uint32_t, std::unordered_set<uint32_t>> function_global_vars;
```
HansKristian-ARM: If these are unordered maps, will that ensure that MSL output is the same, even with a different std::unordered_map implementation?

spirv_msl.cpp
HansKristian-ARM: I haven't looked at this in detail, so I might be repeating stuff, but if a shader does declare DecorationLocation on varyings/attributes, this should always be honored.

I guess the question here is what the MSL backend should do in case that there is no information in the shader itself?

spirv_msl.cpp
CatCuddler: Because `storage_class_is_interface` returns true for `StorageClassUniformConstant`. Since uniform storage is a hidden variable, it won’t be added as an main function arguments.

spirv_msl.hpp
CatCuddler: I use unordered_map, because i need a pure lookup. I think output should always be the same, since i always check the key (no need to traverse the map in a certain sequence).

CatCuddler:
Ok, i hope we can close now this PR. I think you all would implement this additions faster … sorry that i caused so much discussion. But thank you @HansKristian-ARM and @billhollings, with your help i certainly improved my c++ skills.

billhollings:
@CatCuddler 

Thanks for all your work...and for making the mods suggested by the reviewers.

...Bill

PR #98, opened 2017-01-17
Fix get_declared_struct_member_size for struct members

zeux:
When a member of a struct is a struct, get_declared_struct_member_size
instead returned the size of the entire outer struct because it added
the offset of the last field to the size of the last field.

Restructure the function so that it handles all arrays in the same way
(by using array stride) and for the rest reuses get_declared_struct_size
if possible - this simplifies the function and fixes the issue.

This is a prerequisite for merging #82.


HansKristian-ARM:
I'll try to review this today.

PR #100, opened 2017-01-18
Add basic setup for regression testing Metal output.

HansKristian-ARM:
This PR sets up the basic system for how we can have regression tests for Metal as well.
While shaders we test specifically are found in shaders-msl, we would ideally reuse the same shaders to test Metal as well.


HansKristian-ARM:
@billhollings This should be a decent starting point. Would it be possible for you to contribute test shaders which are relevant for Metal output?

HansKristian-ARM:
Merging this to unblock other PRs.

PR #101, opened 2017-01-18
Fix compilation on some old C++11 standard libraries.

Kangz:
Hey all, this is another fix for compiling spirv_cross with the old C++11 stdlib used by Chrome.


Kangz:
Thank you!

PR #102, opened 2017-01-18
spirv_msl: Put writable SSBOs in the 'device' address space

Kangz:
Hey all, this fixes writable SSBOs being put in the "constant" address spaces causing compile errors when their are written to. This is a quick-fix and I would understand it if more work is required. If that's the case please kindly provide guidance on what should be done.

### DIFF
--- a/spirv_msl.cpp
+++ b/spirv_msl.cpp
@@ -1253,7 +1253,16 @@ string CompilerMSL::entry_point_args(bool append_comma)
 				case SPIRType::Struct:
 					if (!ep_args.empty())
 						ep_args += ", ";
-					ep_args += "constant " + type_to_glsl(type) + "& " + to_name(var.self);
+					if ((meta[type.self].decoration.decoration_flags & (1ull << DecorationBufferBlock)) != 0 &&
+					    (meta[type.self].decoration.decoration_flags & (1ull << DecorationNonWritable)) == 0)
+					{
+						ep_args += "device ";
+					}
+					else
+					{
+						ep_args += "constant ";
+					}
+					ep_args += type_to_glsl(type) + "& " + to_name(var.self);
 					ep_args += " [[buffer(" + convert_to_string(get_metal_resource_index(var, type.basetype)) + ")]]";
 					break;
 				case SPIRType::Sampler:


spirv_msl.cpp
```diff
@@ -1253,7 +1253,16 @@ string CompilerMSL::entry_point_args(bool append_comma)
 				case SPIRType::Struct:
 					if (!ep_args.empty())
 						ep_args += ", ";
-					ep_args += "constant " + type_to_glsl(type) + "& " + to_name(var.self);
+					if ((meta[type.self].decoration.decoration_flags & (1ull << DecorationBufferBlock)) != 0 ||
+						(meta[type.self].decoration.decoration_flags & (1ull << DecorationNonWritable)) == 0)
```
HansKristian-ARM: The test should be more in tune of
if (BufferBlock && NonWritable == 0) device; else constant;

Regular UBOs are never writable, the decoration is kind of meaningless for them.

I'm not sure about the semantics for device vs. constant however in Metal. Does constant mean readonly, or does it imply that the buffer is found in special constant memory (UBO)?

Kangz: Just to be sure, you are suggesting changing the || in a && because only BufferBlock can be writable?

From the MSL spec:

MSL 1.2 spec, section 4.2.3
> The constant address space name refers to buffer memory objects allocated from the device memory
pool but are read-only. 

MSL 1.2 spec, section 4.2.1
> The device address space name refers to buffer memory objects allocated from the device memory
pool that are both readable and writeable. 

MSL doesn't seem to make (or expose) a distinction between different memory types, constant is only used to tag something as readonly.

Kangz:
This PR should be covered by comp/ssbo-array.comp but I don't see a way to run tests against the MSL backend. When running testshaders.py, I only had cosmetic differences in the name of _NN variables maybe because I use ToT glslangValidator.

FYI here is the shader that requires this fix:
```
; SPIR-V
; Version: 1.0
; Generator: Khronos Glslang Reference Front End; 1
; Bound: 28
; Schema: 0
               OpCapability Shader
          %1 = OpExtInstImport "GLSL.std.450"
               OpMemoryModel Logical GLSL450
               OpEntryPoint GLCompute %4 "main"
               OpExecutionMode %4 LocalSize 1 1 1
               OpSource GLSL 450
               OpSourceExtension "GL_GOOGLE_cpp_style_line_directive"
               OpSourceExtension "GL_GOOGLE_include_directive"
               OpName %4 "main"
               OpName %8 "myBlock"
               OpMemberName %8 0 "a"
               OpMemberName %8 1 "b"
               OpName %10 "myStorage"
               OpMemberDecorate %8 0 Offset 0
               OpMemberDecorate %8 1 Offset 4
               OpDecorate %8 BufferBlock
               OpDecorate %10 DescriptorSet 0
               OpDecorate %10 Binding 0
          %2 = OpTypeVoid
          %3 = OpTypeFunction %2
          %6 = OpTypeInt 32 1
          %7 = OpTypeFloat 32
          %8 = OpTypeStruct %6 %7
          %9 = OpTypePointer Uniform %8
         %10 = OpVariable %9 Uniform
         %11 = OpConstant %6 0
         %12 = OpTypePointer Uniform %6
         %15 = OpConstant %6 1
         %17 = OpConstant %6 256
         %20 = OpTypePointer Uniform %7
         %23 = OpConstant %7 0.02
         %25 = OpConstant %7 1
          %4 = OpFunction %2 None %3
          %5 = OpLabel
         %13 = OpAccessChain %12 %10 %11
         %14 = OpLoad %6 %13
         %16 = OpIAdd %6 %14 %15
         %18 = OpSMod %6 %16 %17
         %19 = OpAccessChain %12 %10 %11
               OpStore %19 %18
         %21 = OpAccessChain %20 %10 %15
         %22 = OpLoad %7 %21
         %24 = OpFAdd %7 %22 %23
         %26 = OpFMod %7 %24 %25
         %27 = OpAccessChain %20 %10 %15
               OpStore %27 %26
               OpReturn
               OpFunctionEnd
```

From compiling the following with shaderc:

    #version 450
    layout(set = 0, binding = 0) buffer myBlock {
        int a;
        float b;
    } myStorage;
    void main() {
        myStorage.a = (myStorage.a + 1) % 256;
        myStorage.b = mod((myStorage.b + 0.02), 1.0);
    }

Kangz:
Addressed comment and ran format_all.sh.

HansKristian-ARM:
The commit seems good. To be sure, there is no distinction between UBOs and SSBOs in Metal?

Kangz:
There isn't, all buffer types are backed by the same object and are set through the same entrypoints (even vertex buffers use that entrypoint).

Kangz:
Thanks!

PR #104, opened 2017-01-21
Legacy UBO flattening improvements

HansKristian-ARM:
- Can flatten push constant blocks as well (they're technically UBOs too).
- Can flatten int/uint blocks.
- Precision qualifiers for flattened blocks is emitted.
- Propagate type hierarchy to not "guess" array strides and matrix strides.
- Can load row-major matrices from flattened UBOs with non-square matrices.
- Array indexing with constant index is folded into the overall constant offset.

### DIFF
--- a/spirv_cross.cpp
+++ b/spirv_cross.cpp
@@ -842,6 +842,10 @@ void Compiler::set_member_decoration(uint32_t id, uint32_t index, Decoration dec
 		dec.spec_id = argument;
 		break;
 
+	case DecorationMatrixStride:
+		dec.matrix_stride = argument;
+		break;
+
 	default:
 		break;
 	}
@@ -961,6 +965,10 @@ void Compiler::set_decoration(uint32_t id, Decoration decoration, uint32_t argum
 		dec.array_stride = argument;
 		break;
 
+	case DecorationMatrixStride:
+		dec.matrix_stride = argument;
+		break;
+
 	case DecorationBinding:
 		dec.binding = argument;
 		break;
@@ -1020,6 +1028,10 @@ uint32_t Compiler::get_decoration(uint32_t id, Decoration decoration) const
 		return dec.input_attachment;
 	case DecorationSpecId:
 		return dec.spec_id;
+	case DecorationArrayStride:
+		return dec.array_stride;
+	case DecorationMatrixStride:
+		return dec.matrix_stride;
 	default:
 		return 1;
 	}
@@ -1271,6 +1283,7 @@ void Compiler::parse(const Instruction &instruction)
 		vecbase = base;
 		vecbase.vecsize = vecsize;
 		vecbase.self = id;
+		vecbase.parent_type = ops[1];
 		break;
 	}
 
@@ -1285,6 +1298,7 @@ void Compiler::parse(const Instruction &instruction)
 		matrixbase = base;
 		matrixbase.columns = colcount;
 		matrixbase.self = id;
+		matrixbase.parent_type = ops[1];
 		break;
 	}
 
@@ -1302,6 +1316,7 @@ void Compiler::parse(const Instruction &instruction)
 
 		arraybase.array_size_literal.push_back(literal);
 		arraybase.array.push_back(literal ? c->scalar() : ops[2]);
+		arraybase.parent_type = ops[1];
 		// Do NOT set arraybase.self!
 		break;
 	}
@@ -1316,6 +1331,7 @@ void Compiler::parse(const Instruction &instruction)
 		arraybase = base;
 		arraybase.array.push_back(0);
 		arraybase.array_size_literal.push_back(true);
+		arraybase.parent_type = ops[1];
 		// Do NOT set arraybase.self!
 		break;
 	}
@@ -1370,6 +1386,8 @@ void Compiler::parse(const Instruction &instruction)
 
 		if (ptrbase.storage == StorageClassAtomicCounter)
 			ptrbase.basetype = SPIRType::AtomicCounter;
+
+		ptrbase.parent_type = ops[2];
 
 		// Do NOT set ptrbase.self!
 		break;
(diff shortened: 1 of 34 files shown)


HansKristian-ARM:
Ping @zeux.

spirv_cross.cpp
```diff
@@ -3008,3 +3041,43 @@ void Compiler::analyze_variable_scope(SPIRFunction &entry)
...
+{
+	auto &type = get<SPIRType>(var.basetype);
+
+	// Some flags like non-writable, non-readable are actually found
+	// as member decorations. If all members have a decoration set, propagate
+	// the decoration up as a regular variable decoration.
+	uint64_t base_flags = meta[var.self].decoration.decoration_flags;
+	uint64_t all_members_flag_mask = 0;
+	for (uint32_t i = 0; i < uint32_t(type.member_types.size()); i++)
+		all_members_flag_mask |= ~get_member_decoration_mask(type.self, i);
```
zeux: I feel like this would be slightly easier to understand if mask started as ~0ull and this used &= - that way the comment above explicitly reflects what the code is doing

spirv_cross.cpp
```diff
@@ -3008,3 +3041,43 @@ void Compiler::analyze_variable_scope(SPIRFunction &entry)
...
+
+uint64_t Compiler::get_buffer_block_flags(const SPIRVariable &var)
+{
+	auto &type = get<SPIRType>(var.basetype);
+
+	// Some flags like non-writable, non-readable are actually found
+	// as member decorations. If all members have a decoration set, propagate
+	// the decoration up as a regular variable decoration.
+	uint64_t base_flags = meta[var.self].decoration.decoration_flags;
+	uint64_t all_members_flag_mask = 0;
```
zeux: If the buffer block is empty (is this valid?) we'll end up getting all bits in the resulting mask set.

spirv_glsl.cpp
```diff
@@ -1092,7 +1095,34 @@ void CompilerGLSL::emit_buffer_block_flattened(const SPIRVariable &var)
...
 	size_t buffer_size = (get_declared_struct_size(type) + 15) / 16;
 
-	statement("uniform vec4 ", buffer_name, "[", buffer_size, "];");
+	SPIRType::BaseType basic_type;
+	if (get_common_basic_type(type, basic_type))
+	{
+		SPIRType tmp;
+		tmp.basetype = basic_type;
+
+		const char *flat_type = nullptr;
```
zeux: This could use type_to_glsl(tmp) if tmp.vecsize is set to 4; we'll still need to check if the type is one of float/int/uint but that can be handled in one if(...) THROW().

spirv_glsl.cpp
```diff
@@ -3414,42 +3457,65 @@ std::pair<std::string, uint32_t> CompilerGLSL::flattened_access_chain_offset(uin
...
-			uint32_t array_size = temp.array.back();
-			temp.array.pop_back();
+			uint32_t array_stride = get_decoration(current_type, DecorationArrayStride);
+			if (!array_stride)
+				SPIRV_CROSS_THROW("SPIR-V does not define ArrayStride for buffer block.");
 
-			assert(type_size > 0);
-			assert(type_size % array_size == 0);
-
-			uint32_t array_stride = type_size / array_size;
```
zeux: Does the new approach with parent_type work with 3-dimensional arrays? I think it should based on what I understand but maybe worth adding a test - for a two dimensional array you only need the parent type at the top level to be accurate, not sure they all are.

spirv_glsl.cpp
```diff
@@ -3396,13 +3440,12 @@ std::string CompilerGLSL::flattened_access_chain_vector_scalar(uint32_t base, co
...
-                                                                             bool *need_transpose)
+                                                                             bool *need_transpose,
+                                                                             uint32_t *out_matrix_stride)
 {
 	const auto *type = &expression_type(base);
-	uint32_t type_size = 0;
-
-	// For resolving array accesses, etc, keep a local copy for poking.
-	SPIRType temp;
+	uint32_t current_type = type->self;
```
zeux: Consider renaming to type_id and leaving a comment as to why it's different from type->self

spirv_glsl.cpp
```diff
@@ -3414,42 +3457,65 @@ std::pair<std::string, uint32_t> CompilerGLSL::flattened_access_chain_offset(uin
...
+			else
+			{
+				// Dynamic array access.
+				// FIXME: This will need to change if we support other flattening types than 32-bit.
+				const uint32_t word_stride = 16;
+				if (array_stride % word_stride)
+				{
+					SPIRV_CROSS_THROW(
+					    "Array stride for dynamic indexing must be divisible by the size of a 4-component vector. "
+					    "Likely culprit here is a float or vec2 array inside a push constant block. This cannot be "
```
zeux: Are push constant blocks std430 or something along these lines? If it's possible to specify std140 layout for them we can note that in the error.

spirv_glsl.cpp
```diff
@@ -3306,40 +3336,47 @@ std::string CompilerGLSL::flattened_access_chain(uint32_t base, const uint32_t *
...
-		uint32_t member_offset = type_struct_member_offset(target_type, uint32_t(i));
 
-		expr += flattened_access_chain(base, indices, count, member_type, offset + member_offset);
+		// The indices should be IDs, but to avoid creating new dummy SPIRConstant here,
+		// use MSB to indicate literals.
+		indices_copy.back() = uint32_t(i) | 0x80000000u;
+
+		bool need_transpose = false;
+		if (member_type.columns > 1)
+			flattened_access_chain_offset(base, indices_copy.data(), count + 1, offset, &need_transpose);
```
zeux: indices_copy.size() would be slightly cleaner

spirv_glsl.cpp
```diff
@@ -3306,40 +3336,47 @@ std::string CompilerGLSL::flattened_access_chain(uint32_t base, const uint32_t *
...
+
+		auto tmp = flattened_access_chain(base, indices_copy.data(), count + 1, member_type, offset);
```
zeux: indices_copy.size() would be slightly cleaner

spirv_glsl.cpp
```diff
@@ -3306,40 +3336,47 @@ std::string CompilerGLSL::flattened_access_chain(uint32_t base, const uint32_t *
...
 std::string CompilerGLSL::flattened_access_chain_struct(uint32_t base, const uint32_t *indices, uint32_t count,
                                                         const SPIRType &target_type, uint32_t offset)
 {
 	std::string expr;
 
-	expr += type_to_glsl(target_type);
+	expr += type_to_glsl_constructor(target_type);
 	expr += "(";
 
+	// Need to have an extra indices for the member so we can find matrix strides.
```
zeux: The comment should say "whether matrix members need transposing" - AFAICS matrix strides are not retrieved here

zeux:
So the only part of this review that I'm somewhat hesitant about is the struct code manufacturing the fake "literals". It's not too bad but if we didn't need it it would be easier. This is only used for figuring out if a specific member in struct type is row major or column major. You *do* have the target type in the function though - can you just call `combined_decoration_for_member` in `flattened_access_chain_struct`? If that *does* work you'll be able to remove fake literal handling, indices copy and a call to `flattened_access_chain_offset` so this seems like a win but I'm not sure if I'm missing a corner case that this would not work for.

HansKristian-ARM:
The row-major in struct did get kinda tricky, using combined_decoration_for_member in flattened_access_chain_struct might work indeed as the tacked on access chain index was there to be able to grab matrix stride and row-major information. I agree with the comments so far, so I'll fix it soon.

I'm pretty sure 3-dimensional arrays will work, but should be a test for it, yes.

PR #105, opened 2017-01-22
spirv_msl: Add support for gl_GlobalInvocationId

Kangz:
Hey all,
As the name implies this adds support for gl_GlobalInvocationID to the MSL output so that non-trivial compute shaders work. Unfortunately this commit is in the MSL backend that doesn't have tests but locally we have verified this works correctly and using gl_GlobalInvocationID works as expected.

### DIFF
--- a/reference/shaders-msl/comp/global-invocation-id-writable-ssbo-in-function.comp
+++ b/reference/shaders-msl/comp/global-invocation-id-writable-ssbo-in-function.comp
@@ -1,0 +1,31 @@
+#pragma clang diagnostic ignored "-Wmissing-prototypes"
+
+#include <metal_stdlib>
+#include <simd/simd.h>
+
+using namespace metal;
+
+struct myBlock
+{
+    int a;
+    float b[1];
+};
+
+// Support GLSL mod(), which is slightly different than Metal fmod()
+template<typename Tx, typename Ty>
+Tx mod(Tx x, Ty y)
+{
+    return x - y * floor(x / y);
+}
+
+float getB(device const myBlock& myStorage, thread const uint3& gl_GlobalInvocationID)
+{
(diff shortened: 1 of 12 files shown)


HansKristian-ARM:
I would like to get https://github.com/KhronosGroup/SPIRV-Cross/pull/100 in before merging this.

Kangz:
Sounds good, thanks for putting tests together.

reference/shaders-msl/comp/global-invocation-id-writable-ssbo-in-function.comp
```diff
@@ -0,0 +1,31 @@
...
+};
+
+// Support GLSL mod(), which is slightly different than Metal fmod()
+template<typename Tx, typename Ty>
+Tx mod(Tx x, Ty y)
+{
+    return x - y * floor(x / y);
+}
+
+float getB(device const myBlock& myStorage, thread const uint3& gl_GlobalInvocationID)
```
Kangz: myStorage shouldn't be const, and this is actually a bug. However fixing it nicely would require a huge refactor (was still untangling spaghetti after two hours).

Kangz:
Just saw the part about additional builtins.

Kangz:
PTAL again.

reference/shaders-msl/comp/global-invocation-id-writable-ssbo-in-function.comp
HansKristian-ARM: What is causing this bug? It doesn't seem too hard to not add "const" if "device" storage is used?

HansKristian-ARM:
Merged, and fixed the issue with device const&.

Kangz:
Sorry I didn't get back to this in time.

PR #107, opened 2017-01-24
HLSL Support

RobDangerous:
This adds basic HLSL support (vert, frag and early compute support). It's in active use for a while now in [Kha](https://github.com/Kode/Kha)/[Kore](https://github.com/Kode/Kore) via [krafix](https://github.com/Kode/krafix) and the code which actually runs it is in [HlslTranslator2.cpp](https://github.com/Kode/krafix/blob/master/Sources/HlslTranslator2.cpp) - there's also code for compiling the HLSL code in [D3D9Compiler.cpp](https://github.com/Kode/krafix/blob/master/Sources/D3D9Compiler.cpp) and [D3D11Compiler.cpp](https://github.com/Kode/krafix/blob/master/Sources/D3D11Compiler.cpp) - not sure about putting that into SPIRV-Cross.
I only just started testing and fixing the SPIRV-Cross shaders (basic.vert and basic.frag at least are working), so it's kinda early, but I'd still like to get the process started. The sooner I can get more people to see it and help out with it the better I think.


HansKristian-ARM:
Cool! Before reviewing in detail I'd like to see some basic regression testing set up.
I just added basic setup to regression test Metal output: https://github.com/KhronosGroup/SPIRV-Cross/pull/100. You could base it on that. I assume that we can use glslangValidator in HLSL mode to compile to SPIR-V and also to validate the HLSL output from SPIRV-Cross.

RobDangerous:
Will do. Can also implement an optional test mode using Microsoft's compiler (run only on Windows, load dll dynamically).

HansKristian-ARM:
Wouldn't it also work to just optionally call "fxc" from Python instead? I'd like to avoid Windows-specific test code if possible.

RobDangerous:
Sure, can as well call fxc.

RobDangerous:
All done.
There's a problem with line endings because spirv-cross on Windows outputs \r\n and regression_check just compares a file hash. We should probably change that and also add an appveyor script so the fxc test can actually run.

CMakeLists.txt
```diff
@@ -45,11 +45,16 @@ add_library(spirv-cross-msl STATIC
 		${CMAKE_CURRENT_SOURCE_DIR}/spirv_msl.hpp
 		${CMAKE_CURRENT_SOURCE_DIR}/spirv_msl.cpp)
 
+add_library(spirv-cross-hlsl STATIC
```
HansKristian-ARM: Missing MSVC project file update.

reference/shaders-hlsl/vert/basic.vert
```diff
@@ -0,0 +1,47 @@
+struct _UBO
+{
+    float4x4 uMVP;
+};
+
+cbuffer UBO
+{
+    _UBO _16;
+};
+uniform float4 gl_HalfPixel;
```
HansKristian-ARM: This is not HLSL, how will this work? Is this gl_HalfPixel there to handle the infamous DX9 window space offset?

reference/shaders-hlsl/vert/basic.vert
```diff
@@ -0,0 +1,47 @@
...
+
+static float4 gl_Position;
```
HansKristian-ARM: This is very different from how MSL does it. Is there any reason it's done differently here?

reference/shaders-hlsl/vert/basic.vert
```diff
@@ -0,0 +1,47 @@
...
+
+struct OutputVert
+{
+    float3 vNormal : TEXCOORD0;
+    float4 gl_Position : POSITION;
+};
+
+void vert_main()
+{
+    gl_Position = mul(transpose(_16.uMVP), aVertex);
```
HansKristian-ARM: Is a transpose needed here? How does this affect performance?

reference/shaders-hlsl/vert/basic.vert
```diff
@@ -0,0 +1,47 @@
...
+uniform float4 gl_HalfPixel;
+
+static float4 gl_Position;
+static float4 aVertex;
+static float3 vNormal;
+static float3 aNormal;
+
+struct InputVert
+{
+    float3 aNormal : TEXCOORD0;
```
HansKristian-ARM: How are semantics allocated? The test shader doesn't have layout() on it.

reference/shaders-hlsl/frag/basic.frag
```diff
@@ -0,0 +1,31 @@
+uniform sampler2D uTex;
```
HansKristian-ARM: This doesn't look like HLSL.

reference/shaders-hlsl/vert/basic.vert
RobDangerous: Aye and it needs the application to set gl_HalfPixel. It doesn't show up if shaders are compiled for D3D11.

spirv_hlsl.cpp
```diff
@@ -0,0 +1,1055 @@
+/*
+* Copyright 2016 Robert Konrad
```
HansKristian-ARM: 2017?

spirv_hlsl.cpp
```diff
@@ -0,0 +1,1055 @@
...
+*
```
HansKristian-ARM: Space before the '*' to align.

spirv_hlsl.cpp
```diff
@@ -0,0 +1,1055 @@
...
+		default:
+			return "???";
+		}
+	}
+	else if (type.vecsize > 1 && type.columns == 1) // Vector builtin
+	{
+		switch (type.basetype)
+		{
+		case SPIRType::Boolean:
+			return join("bfloat", type.vecsize);
```
HansKristian-ARM: bfloat?

spirv_hlsl.cpp
```diff
@@ -0,0 +1,1055 @@
...
+		++binding_number;
+	}
+}
+
+void CompilerHLSL::emit_resources()
+{
+	auto &execution = get_entry_point();
+
+	// Emit PLS blocks if we have such variables.
+	if (!pls_inputs.empty() || !pls_outputs.empty())
```
HansKristian-ARM: PLS is mostly a GLES backend thing, so this can be removed.

reference/shaders-hlsl/vert/basic.vert
RobDangerous: Matrices always have to be transposed to get the maths straight. ANGLE does the very same thing. Would love to find a better and faster solution though but I guess that's very tricky.

spirv_hlsl.cpp
```diff
@@ -0,0 +1,1055 @@
...
+			{
+				emit_push_constant_block(var);
+				emitted = true;
+			}
+		}
+	}
+
+	if (execution.model == ExecutionModelVertex && options.shader_model <= 30)
+	{
+		statement("uniform float4 gl_HalfPixel;");
```
HansKristian-ARM: Does HLSL really support naked uniforms like this?

reference/shaders-hlsl/vert/basic.vert
RobDangerous: inputs/outputs are sorted by name, then TEXCOORD semantics are iterated.

reference/shaders-hlsl/frag/basic.frag
RobDangerous: Hm, but I think it is.

spirv_hlsl.cpp
RobDangerous: Yes.

spirv_hlsl.cpp
```diff
@@ -0,0 +1,1055 @@
...
+
+	statement("cbuffer ", to_name(type.self));
+	begin_scope();
+	statement("_", to_name(type.self), " ", to_name(var.self), ";");
+	end_scope_decl();
+}
+
+void CompilerHLSL::emit_push_constant_block(const SPIRVariable &)
+{
+	statement("constant block");
```
HansKristian-ARM: placeholder? push constant blocks should just be regular constant buffers I'd think, but you have to be careful about buffer layouts.

RobDangerous: Yes, placeholder.

spirv_hlsl.cpp
```diff
@@ -0,0 +1,1055 @@
...
+	}
+
+	if (execution.model == ExecutionModelVertex)
+	{
+		if (options.shader_model <= 30)
+		{
+			statement("output.gl_Position.x = output.gl_Position.x - gl_HalfPixel.x * output.gl_Position.w;");
+			statement("output.gl_Position.y = output.gl_Position.y + gl_HalfPixel.y * output.gl_Position.w;");
+		}
+		statement("output.gl_Position.z = (output.gl_Position.z + output.gl_Position.w) * 0.5;");
```
HansKristian-ARM: This must be optional.

spirv_hlsl.cpp
RobDangerous: Hm, that looks plain wrong.

spirv_hlsl.cpp
```diff
@@ -0,0 +1,1055 @@
...
+	}
+
+	statement("return output;");
+
+	end_scope();
+}
+
+void CompilerHLSL::emit_texture_op(const Instruction &i)
```
HansKristian-ARM: There is a virtual interface which MSL and GLSL use to get the texture expression, any reason this is not used here?

reference/shaders-hlsl/vert/basic.vert
RobDangerous: I'm not deep into Metal things right now but I think Metal doesn't support globals. With globals things are easier.

spirv_hlsl.cpp
RobDangerous: Probably not...

reference/shaders-hlsl/frag/basic.frag
HansKristian-ARM: I recall stuff like:
```
Texture2D<float4> tex1[4] : register(t3)
```
Also, are layout()s handled in any way, or is it up to the application to remap before compiling the HLSL through some API?

reference/shaders-hlsl/vert/basic.vert
HansKristian-ARM: Ok, so interface variables do not need to be piped through function arguments always then?

reference/shaders-hlsl/vert/basic.vert
HansKristian-ARM: What if you invert the order there, does mul() always imply M * V, even if you pass in mul(V, M)?

reference/shaders-hlsl/vert/basic.vert
RobDangerous: Aye

reference/shaders-hlsl/frag/basic.frag
RobDangerous: There's different ways to handle textures - no expert on that, sampler2D decls work good enough for me. And yes, no layouts currently, apllications have to set up input element descriptions accordingly.

reference/shaders-hlsl/vert/basic.vert
HansKristian-ARM: And I suppose resource variables can always be declared in global scope (unlike Metal), so there is no need for hackery there either?

reference/shaders-hlsl/vert/basic.vert
```diff
@@ -0,0 +1,47 @@
+struct _UBO
+{
+    float4x4 uMVP;
+};
+
+cbuffer UBO
+{
+    _UBO _16;
```
HansKristian-ARM: Is there a reason the UBO block data is not inlined directly into the cbuffer? Is it possible to declare the cbuffer as

```
cbuffer UBO { float4x4 uMVP } _16;
```
instead?

reference/shaders-hlsl/vert/basic.vert
RobDangerous: According to https://msdn.microsoft.com/de-de/library/windows/desktop/bb509628(v=vs.85).aspx this should actually work. Will have to do some tests and think hard about what else could break.

reference/shaders-hlsl/vert/basic.vert
HansKristian-ARM: If there is a layout(location = N) will it use TEXCOORDN always then?

reference/shaders-hlsl/vert/basic.vert
RobDangerous: Yes, didn't come across anything in HLSL which can't be globlal.

reference/shaders-hlsl/vert/basic.vert
HansKristian-ARM: The other way is to unroll the multiply, but that leads to less clean though.

```
M * V = M[0] * V.x + M[1] * V.y + M[2] * V.z + M[3] * V.w
```
If the underlying matrix is column-major. HLSL is row-major, so it should just work reverse the multiplication order if that's doable in HLSL.

reference/shaders-hlsl/vert/basic.vert
RobDangerous: Didn't try this particular syntax yet...

reference/shaders-hlsl/vert/basic.vert
RobDangerous: Yes. Didn't do anything with layouts at all so far.

spirv_hlsl.cpp
```diff
@@ -0,0 +1,1055 @@
...
+void CompilerHLSL::emit_interface_block_in_struct(const SPIRVariable &var, uint32_t &binding_number, bool builtins)
+{
+	auto &execution = get_entry_point();
+	auto &type = get<SPIRType>(var.basetype);
+
+	const char *binding = "TEXCOORD";
+	bool use_binding_number = true;
+	if (execution.model == ExecutionModelFragment && var.storage == StorageClassOutput)
+	{
+		binding = "COLOR";
```
HansKristian-ARM: Isn't this SV_Target#?

spirv_hlsl.hpp
```diff
@@ -0,0 +1,78 @@
...
+#include <vector>
+
+namespace spirv_cross
+{
+class CompilerHLSL : public CompilerGLSL
+{
+public:
+	struct Options
+	{
+		uint32_t shader_model = 30; // TODO: map ps_4_0_level_9_0,... somehow
```
HansKristian-ARM: Has this been tested with DX10+ HLSL? Looking at various reference, it seems like it's mostly D3D9-style HLSL which is emitted here.

RobDangerous: Yes, has been tested with both. Started out as a D3D9 thing though.

test_shaders.py
```diff
@@ -240,13 +259,22 @@ def test_shader_msl(stats, shader, update, keep):
     regression_check(shader, msl, update, keep)
     os.remove(spirv)
 
+def test_shader_hlsl(stats, shader, update, keep):
+    joined_path = os.path.join(shader[0], shader[1])
+    print('Testing HLSL shader:', joined_path)
+    spirv, msl = cross_compile_hlsl(joined_path)
+    regression_check(shader, msl, update, keep)
```
HansKristian-ARM: Probably worthwhile to strip out Windows-style \r in regression_check before checksumming.

spirv_hlsl.cpp
RobDangerous: SV_Target only works in D3D11, COLOR works in D3D9 and 11.

test_shaders.py
RobDangerous: We should strip it out everywhere, currently tests on Windows are not working at all for all targets because of it.

HansKristian-ARM: Yes, if all text files are normalized to Unix-style line endings, it should be done on all text files and all OSes.

spirv_hlsl.hpp
HansKristian-ARM: Ok.

spirv_hlsl.cpp
HansKristian-ARM: Kinda expected some optional Y-flipping, considering DX Y-flips clip space to get to window space.

RobDangerous: The way I see it OGL flips it, so I didn't need it here - will add it.

HansKristian-ARM:
I assume more tests will be added soon? I assume compute/tess/geom isn't working at all?

RobDangerous:
Some compute works, tess/geom not yet. We'll have to talk about tess before I add it because the structure is somewhat different (separate function for controlling tess factors). Yes, will add more tests.

reference/shaders-hlsl/vert/basic.vert
RobDangerous: Sorry, doesn't work.

reference/shaders-hlsl/vert/basic.vert
RobDangerous: Switching arguments is working out fine.

RobDangerous:
All done.

spirv_hlsl.cpp
HansKristian-ARM: This is still wrong.

spirv_hlsl.cpp
```diff
@@ -0,0 +1,1061 @@
...
+#define BFOP(op) emit_binary_func_op(ops[0], ops[1], ops[2], ops[3], #op)
+#define BFOP_CAST(op, type, skip_cast) emit_binary_func_op_cast(ops[0], ops[1], ops[2], ops[3], #op, type, skip_cast)
+#define BFOP(op) emit_binary_func_op(ops[0], ops[1], ops[2], ops[3], #op)
+#define UFOP(op) emit_unary_func_op(ops[0], ops[1], ops[2], #op)
+
+	switch (opcode)
+	{
+	case OpMatrixTimesVector:
+	{
+		emit_binary_func_op_transpose_first(ops[0], ops[1], ops[2], ops[3], "mul");
```
HansKristian-ARM: Separate helper function for this isn't needed now, can just use emit_binary_func_op directly with swapped arguments.

spirv_hlsl.hpp
```diff
@@ -0,0 +1,80 @@
+/*
+* Copyright 2016 Robert Konrad
```
HansKristian-ARM: Space

RobDangerous:
All done once again.

spirv_hlsl.cpp
```diff
@@ -97,11 +97,11 @@ string CompilerHLSL::type_to_glsl(const SPIRType &type)
...
 		{
 		case SPIRType::Boolean:
-			return join("bfloat", type.vecsize);
+			return join("bvec", type.vecsize);
 		case SPIRType::Int:
-			return join("ifloat", type.vecsize);
+			return join("ivec", type.vecsize);
 		case SPIRType::UInt:
-			return join("ufloat", type.vecsize);
+			return join("uvec", type.vecsize);
```
HansKristian-ARM: Pretty sure this is bool2, bool4, int3, uint4, etc in HLSL.

RobDangerous: Doh! Corrected.

HansKristian-ARM:
Thanks for the contribution.

PR #108, opened 2017-01-24
Implement flattening of row major matrix indexing

zeux:
To extract a column from row-major matrix, we need to do a strided load one
component at a time. In this case flattened_access_chain_offset still returns
the offset to the first element, but the stride is equal to matrix stride
instead of vector stride.

For this to work, we need to pass matrix stride (and transpose flag) through,
similar to how matrix flattening works.

Additionally slightly clean up recursive flattened_access_chain structure -
specifically, instead of deciding mid-traversal that we need matrix stride
information, we can just pass the matrix stride through - for access chains
that end in matrix/vector this gets us what we need, and for access chains
that end in structs the flattened_access_chain_struct code will recompute
correct stride/transposition data to pass through further.

### DIFF
--- a/reference/shaders/flatten/matrixindex.flatten.vert
+++ b/reference/shaders/flatten/matrixindex.flatten.vert
@@ -14,6 +14,6 @@ void main()
     oB = vec4(UBO[4].y, UBO[5].y, UBO[6].y, UBO[7].y);
     oC = UBO[9];
     oD = vec4(UBO[10].x, UBO[11].x, UBO[12].x, UBO[13].x);
-    oE = vec4(UBO[1].z, float(UBO[6].y), UBO[9].z, float(UBO[12].y));
+    oE = vec4(UBO[1].z, UBO[6].y, UBO[9].z, UBO[12].y);
 }
 
--- a/reference/shaders/flatten/matrixindex.flatten.vert
+++ b/reference/shaders/flatten/matrixindex.flatten.vert
@@ -1,0 +1,19 @@
+#version 310 es
+
+uniform vec4 UBO[14];
+out vec4 oA;
+out vec4 oB;
+out vec4 oC;
+out vec4 oD;
+out vec4 oE;
+
+void main()
+{
+    gl_Position = vec4(0.0);
+    oA = UBO[1];
+    oB = vec4(UBO[4].y, UBO[5].y, UBO[6].y, UBO[7].y);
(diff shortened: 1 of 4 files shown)


reference/shaders/flatten/matrixindex.flatten.vert
```diff
@@ -0,0 +1,19 @@
...
+out vec4 oE;
+
+void main()
+{
+    gl_Position = vec4(0.0);
+    oA = UBO[1];
+    oB = vec4(UBO[4].y, UBO[5].y, UBO[6].y, UBO[7].y);
+    oC = UBO[9];
+    oD = vec4(UBO[10].x, UBO[11].x, UBO[12].x, UBO[13].x);
+    oE = vec4(UBO[1].z, float(UBO[6].y), UBO[9].z, float(UBO[12].y));
```
HansKristian-ARM: Any idea why this dummy float(UBO[6].y) happens?

spirv_glsl.cpp
```diff
@@ -3399,35 +3397,70 @@ std::string CompilerGLSL::flattened_access_chain_matrix(uint32_t base, const uin
...
+	if (need_transpose)
+	{
+		std::string expr;
 
-	expr += buffer_name;
-	expr += "[";
-	expr += result.first; // this is a series of N1 * k1 + N2 * k2 + ... that is either empty or ends with a +
-	expr += convert_to_string(index / 4);
-	expr += "]";
+		expr += type_to_glsl_constructor(target_type);
```
HansKristian-ARM: Don't need constructor for scalar results.

reference/shaders/flatten/matrixindex.flatten.vert
zeux: Ah, this is the constructor below I think. Will fix!

HansKristian-ARM:
Linux passed, OSX failed due to a Travis error, so I'll just merge it.

zeux:
Looks like it timed out. Travis OSX builds are much slower than Linux builds since they constantly have a backlog of builds :( We'll need to keep OSX builds if we want to test MSL thoroughly - I'm planning to submit a PR soon that validates MSL shaders using Metal compiler on OSX.

HansKristian-ARM:
Cool. I'd like to see the MSL test suite more fleshed out, so that's nice.

PR #109, opened 2017-01-24
CompilerMSL support OpImageFetch and Metal read() function.

billhollings:


spirv_msl.cpp
```diff
@@ -771,12 +782,58 @@ string CompilerMSL::to_function_args(uint32_t img, const SPIRType &imgtype, bool
...
+			if (is_fetch)
+				tex_coords = "uint3(" + tex_coords + ")";
+		}
+
+		alt_coord = ".w";
+
+		break;
+
+	case DimCube:
+		if (msl_config.flip_frag_y)
```
HansKristian-ARM: This flag still really confuses me. Why is this needed? Isn't flipping (if anything) a property of vertex shaders? How would you know that this particular texture being sampled needs flipping?

billhollings: You're right. It is normally set to false. It was added originally because of the difference between texture coordinate systems. So far it hasn't been needed...and at some point it should come out...but I'm not convinced we have hit all the test cases yet with MSL.

billhollings:
I've added a bunch of MSL tests cases...actually I've moved all of the existing GLSL test cases that convert GLSL->SPIRV->MSL successfully and compile successfully as MSL.

billhollings:
I've added a bunch of MSL tests cases...actually I've moved all of the existing GLSL test cases that convert GLSL->SPIRV->MSL successfully and compile successfully as MSL.

I've also beefed up the MSL testing process a bit.

billhollings:
I've added a bunch of MSL tests cases...actually I've moved all of the existing GLSL test cases that convert GLSL->SPIRV->MSL successfully and compile successfully as MSL.

I've also beefed up the MSL testing process a bit.

All tests check out (both MD5 checksums and MSL compilation) when running on OSX. But Travis seems to be failing at least one of the MSL shaders at the checksum stage.

What could be the reason for such a failure? Extra white-space in reference file? Line-end characters?

billhollings:
I've added a bunch of MSL tests cases...actually I've moved all of the existing GLSL test cases that convert GLSL->SPIRV->MSL successfully and compile successfully as MSL.

I've also beefed up the MSL testing process a bit.

All tests check out (both MD5 checksums and MSL compilation) when running on OSX.

However...Travis seems to be failing at least one of the MSL shaders at the checksum stage.

What could be the reason for such a failure? Extra white-space in reference file? Line-end characters?

All MSL reference shaders were generated together in a single run by clearing them all away first. So why some are passing MD5 and others are not is odd.

HansKristian-ARM:
One such possibility is that MSL output depends on ordering of unordered data structures. Travis might run a different STL implementation, I've had similar issues before.

billhollings:
@HansKristian-ARM 

Hmmm... in `CompilerMSL`, I've updated all uses of `unordered_set` and `unordered_map` to `set` and `map`, respectively...in all cases where an iteration occurs (as opposed to look-ups only).

No dice...still the same results in Travis.

Furthermore...the "non-optimized" `for(;;)` loop is being generated in the `CompilerGLSL` subclass, not `CompilerMSL`. I'm not familiar enough with the loop-generation logic to understand why that would be occurring under Travis and not OSX.

And currently...I can't see how I can debug this any further...since it does not seem to be related to unordered data structures within `CompilerMSL`...and it only occurs on a remote system that uses a different build configuration than I have access to.

How do you want to handle this kind of thing? Is there a way of replicating the Travis environment locally?

At this point...since the current MSL tests are a relatively arbitrary sub-selection of the GLSL tests...I think I will start iteratively removing any tests that fail in Travis...just so we can get this release committed. We can then work to solve this in a future release.

HansKristian-ARM:
Yes, it's a bit puzzling, but if it's just a few tests which fail, I'm okay leaving them out until we figure it out. I appreciate the addition of more tests, certainly.

spirv_msl.cpp
```diff
@@ -359,34 +364,44 @@ uint32_t CompilerMSL::add_interface_block(StorageClass storage)
...
-			for (auto &member : type.member_types)
+			for (auto &mbr_type_id : type.member_types)
 			{
-				// Add a reference to the member to the interface struct.
-				uint32_t ib_mbr_idx = uint32_t(ib_type.member_types.size());
-				ib_type.member_types.push_back(member); // membertype.self is different for array types
+				auto &mbr_type = get<SPIRType>(mbr_type_id);
+				if (is_matrix(mbr_type))
+				{
+					exclude_member_from_stage_in(type, mbr_idx);
```
HansKristian-ARM: I'm not sure what this is doing, is there a test for this?

spirv_msl.cpp
```diff
@@ -447,6 +469,169 @@ uint32_t CompilerMSL::add_interface_block(StorageClass storage)
...
+		return;
+
+	uint32_t mbr_type_id = var.basetype;
+	string mbr_name = ensure_valid_name(to_expression(var_id), "m");
+	uint32_t mbr_locn = get_decoration(var_id, DecorationLocation);
+	meta[var_id].decoration.qualified_alias = add_input_buffer_block_member(mbr_type_id, mbr_name, mbr_locn);
+}
+
+// Excludes the specified type member from the stage_in block structure.
+// Instead, the member is added to a block variable corresponding to a secondary MSL buffer.
```
HansKristian-ARM: Where is this "secondary" MSL buffer emitted?

reference/shaders-msl/flatten/push-constant.flatten.vert
```diff
@@ -0,0 +1,33 @@
+#include <metal_stdlib>
+#include <simd/simd.h>
+
+using namespace metal;
+
+struct PushMe
+{
+    float4x4 MVP;
+    float2x2 Rot;
```
HansKristian-ARM: How are the packing rules for Metal? GLSL has std140/std430 layouts which would change how e.g. mat2's are packed in memory.

reference/shaders-msl/flatten/multiindex.flatten.vert
```diff
@@ -0,0 +1,28 @@
...
+
+struct main0_in
+{
+    int2 aIndex [[attribute(0)]];
+};
+
+struct main0_out
+{
+    float4 gl_Position [[position]];
+    float gl_PointSize;
```
HansKristian-ARM: Does it hurt to always emit this?

reference/shaders-msl/comp/defer-parens.comp
```diff
@@ -0,0 +1,23 @@
+#include <metal_stdlib>
+#include <simd/simd.h>
+
```
HansKristian-ARM: Doesn't metal need workgroup size? Is it provided by the API user in Metal?

HansKristian-ARM:
Hm, OSX builds are failing due to this:
https://travis-ci.org/KhronosGroup/SPIRV-Cross/jobs/197022745#L2993

It it trying to validate shaders with xcrun there? Seems like Travis doesn't have the required stuff for that.

HansKristian-ARM:
Merged, added --force-no-external-validation, so it passes Travis now.

PR #114, opened 2017-02-03
Fix build when SPIRV_CROSS_EXCEPTIONS_TO_ASSERTIONS=On.

syoyo:
There is still some `throw` syntax, thus failed to compile when `SPIRV_CROSS_EXCEPTIONS_TO_ASSERTIONS=On`.

This simple patch fixes it.


HansKristian-ARM:
Thanks! Should probably build with exceptions off on Travis to catch stuff like this in the future.

PR #117, opened 2017-02-08
Suppress clang warnings

syoyo:
Suppress clang warnings which reveals when CXXFLAGS has strict warning level, i.e, `"-Weverything -Werror -Wno-c++98-compat-pedantic -std=c++11"`

Add data padding to classes and structs for preventing future potential memory/data align bugs(these could be found with the above clang CXXFLAGS)

There are some tests failing but this is also happen in recent `glslang`, `SPIRV-Tools` and `SPIRV-Cross` master branch combination, thus I'm not sure this patch gives some side effect to unit tests.

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -15,8 +15,8 @@
  */
 
 #include "spirv_cpp.hpp"
-#include "spirv_msl.hpp"
 #include "spirv_hlsl.hpp"
+#include "spirv_msl.hpp"
 #include <algorithm>
 #include <cstdio>
 #include <cstring>
@@ -29,6 +29,11 @@
 
 #ifdef _MSC_VER
 #pragma warning(disable : 4996)
+#endif
+
+#ifdef __clang__
+#pragma clang diagnostic ignored "-Wcovered-switch-default"
+#pragma clang diagnostic ignored "-Wswitch-enum"
 #endif
 
 using namespace spv;
@@ -54,6 +59,7 @@ struct CLICallbacks
 		callbacks[cli] = func;
 	}
 	unordered_map<string, function<void(CLIParser &)>> callbacks;
+	uint64_t pad0;
 	function<void()> error_handler;
 	function<void(const char *)> default_handler;
 };
@@ -120,7 +126,7 @@ struct CLIParser
 			THROW("Tried to parse uint, but nothing left in arguments");
 		}
 
-		uint32_t val = stoul(*argv);
+		uint32_t val = static_cast<uint32_t>(stoul(*argv));
 		if (val > numeric_limits<uint32_t>::max())
 		{
 			THROW("next_uint() out of range");
@@ -162,8 +168,10 @@ struct CLIParser
 
 	CLICallbacks cbs;
 	int argc;
+	uint32_t pad0;
 	char **argv;
 	bool ended_state = false;
+	uint8_t pad1[15];
 };
 
 static vector<uint32_t> read_spirv_file(const char *path)
@@ -176,11 +184,11 @@ static vector<uint32_t> read_spirv_file(const char *path)
 	}
 
 	fseek(file, 0, SEEK_END);
-	long len = ftell(file) / sizeof(uint32_t);
+	long len = ftell(file) / static_cast<long>(sizeof(uint32_t));
 	rewind(file);
 
-	vector<uint32_t> spirv(len);
-	if (fread(spirv.data(), sizeof(uint32_t), len, file) != size_t(len))
+	vector<uint32_t> spirv(static_cast<size_t>(len));
+	if (fread(spirv.data(), sizeof(uint32_t), static_cast<size_t>(len), file) != size_t(len))
 		spirv.clear();
 
 	fclose(file);
@@ -392,6 +400,7 @@ static void print_spec_constants(const Compiler &compiler)
 struct PLSArg
 {
 	PlsFormat format;
+	uint32_t pad0;
 	string name;
 };
 
@@ -400,6 +409,7 @@ struct Remap
 	string src_name;
 	string dst_name;
 	unsigned components;
+	uint32_t pad0;
 };
 
 struct VariableTypeRemap
@@ -421,6 +431,7 @@ struct CLIArguments
 	bool force_temporary = false;
 	bool flatten_ubo = false;
 	bool fixup = false;
+	uint8_t pad0[5];
 	vector<PLSArg> pls_in;
 	vector<PLSArg> pls_out;
 	vector<Remap> remaps;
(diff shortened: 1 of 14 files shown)


HansKristian-ARM:
First concern is, what problem is this trying to solve? It seems like a fairly intrusive change to remove warnings which are not enabled by default.

syoyo:
>  what problem is this trying to solve?

This does not solve any existing bugs but will increase the robustness of the code. Please see the following slide for details.

http://meetingcpp.com/index.php/tv14/items/19.html

main.cpp
```diff
@@ -54,6 +59,7 @@ struct CLICallbacks
 		callbacks[cli] = func;
 	}
 	unordered_map<string, function<void(CLIParser &)>> callbacks;
+	uint64_t pad0;
```
HansKristian-ARM: The manual padding here seems very strange and makes the code uglier for no apparent gain. What would this help? The obvious concern here is that struct layout has arbitrary padding, but none of the code relies on a particular struct memory layout.

syoyo: > What would this help?

HansKristian-ARM: Yes, I see the slide deck you mention put no-padded in there, but that is not a relevant warning for us.

main.cpp
```diff
@@ -31,6 +31,11 @@
 #pragma warning(disable : 4996)
 #endif
 
+#ifdef __clang__
+#pragma clang diagnostic ignored "-Wcovered-switch-default"
```
HansKristian-ARM: Can this be fixed simply by adding default: blocks where appropriate instead?

syoyo: In this case, `default` is not executed since all items are listed up in switch syntax. I have fixed it in this commit: https://github.com/KhronosGroup/SPIRV-Cross/pull/117/commits/5e5f93c2e7e1c0a521b97c2226244234839a9027#diff-118fcbaaba162ba17933c7893247df3aL350

main.cpp
syoyo: Removed data padding in this commit: https://github.com/KhronosGroup/SPIRV-Cross/pull/117/commits/5e5f93c2e7e1c0a521b97c2226244234839a9027

spirv_cfg.hpp
```diff
@@ -83,6 +83,9 @@ class CFG
 	std::vector<int> visit_order;
 	std::vector<uint32_t> post_order;
 
+public:
+	uint32_t pad0;
```
HansKristian-ARM: Padding is still here.

main.cpp
```diff
@@ -339,8 +344,6 @@ static void print_resources(const Compiler &compiler, const ShaderResources &res
 			CHECK_MODE(VecTypeHint);
 			CHECK_MODE(ContractionOff);
 
-		default:
```
HansKristian-ARM: Why remove default block?

main.cpp
```diff
@@ -31,6 +31,11 @@
 #pragma warning(disable : 4996)
 #endif
 
+#ifdef __clang__
+#pragma clang diagnostic ignored "-Wswitch-enum"
```
HansKristian-ARM: Is it necessary for these to be here? I assume they would only be relevant if you intentionally add -Weverything, but it is not applied in the build system.

spirv_cfg.cpp
```diff
@@ -130,7 +134,7 @@ bool CFG::post_order_visit(uint32_t block_id)
 	}
 
 	// Then visit ourselves. Start counting at one, to let 0 be a magic value for testing back vs. crossing edges.
-	visit_order[block_id] = ++visit_count;
+	visit_order[block_id] = static_cast<int>(++visit_count);
```
HansKristian-ARM: For integer static casts it's much nicer to just int(++visit_count) I think.

main.cpp
HansKristian-ARM: For these diagnostics to be relevant, the build system should have an option to build with the strictest possible warning level, or this will not be maintainable.

syoyo: Yes, you are better to add support building SPIRV-Cross with clang and gcc in Travis build, and add "-Weverything -Werror -Wno-c++98-compat-pedantic -std=c++11" when building with clang. My example in tinyobjloader may help  to setup compiler matrix:  https://github.com/syoyo/tinyobjloader/blob/master/.travis.yml

spirv_cfg.cpp
syoyo: You can use `int(++visit_count)`

spirv_glsl.hpp
```diff
@@ -66,6 +66,7 @@ class CompilerGLSL : public Compiler
 		// If true, Vulkan GLSL features are used instead of GL-compatible features.
 		// Mostly useful for debugging SPIR-V files.
 		bool vulkan_semantics = false;
+		uint8_t pad0[3];
```
zeux: From my perspective the result of fixing this warning is actively harmful to the codebase - it generates no value since SPIRV-Cross doesn't care about memory consumption (and even if it did, structs like this one just don't matter period), and changes to the code will change the amount of required padding which would invalidate the fixes and/or make the struct layout even worse compared to what it would have been without the padding fields.

zeux:
See comment above. In general I don't believe -Weverything is meaningful to support. I think it makes sense to enforce warning-free code on -Wall -Wextra (which we don't necessarily do right now because there's no -Werror), but I would not go beyond that unless there's a particular warning that's really beneficial.

syoyo:
> From my perspective the result of fixing this warning is actively harmful to the codebase 

I sometimes encountered runtime memory alignment bug due to bad struct padding in other projects thus I thought checking the padding would be helpful. 
But this is a bit too much and would not cause serious problem in the case for SPIRV-Cross project, so I can remove manual padding and disable padding check using `-Wno-padded`. Also I can relax warning level more if required.

HansKristian-ARM:
I agree that we should try to add the useful warnings on a per case basis, rather than removing the harmless ones by compiler pragmas. There is one such pragma currently regarding stdio.h with MSVC (that one is just a lost cause).

As far as I can see from the PR, it's mostly down to implicit integer promotion which triggers warnings?

Werror is a double-edged sword because users who compile SPIRV-Cross locally on other compilers might run into stray warnings which prevent them from building, but we can probably enable -Werror only in a strict mode and Travis could use that by default.

The risk here is that PRs will have been submitted with some warnings which Travis will complain about, but the developer making the PR might get frustrated (should be documented in README of course), but this is a better solution that always forcing -Werror.

HansKristian-ARM:
I will close this, as there hasn't been any activity, and I don't think the proposed PR will improve the code base.

PR #118, opened 2017-02-09
CompilerMSL add support for texel buffers.

billhollings:
Compiler MSL support DimBuffer as image dimension.
CompilerMSL check texture coordinate result type dimension before adding swizzles.
Update MSL reference shaders affected by this update.

### DIFF
--- a/reference/shaders-msl/frag/basic.frag
+++ b/reference/shaders-msl/frag/basic.frag
@@ -17,7 +17,7 @@ struct main0_out
 fragment main0_out main0(main0_in in [[stage_in]], texture2d<float> uTex [[texture(0)]], sampler uTexSmplr [[sampler(0)]])
 {
     main0_out out = {};
-    out.FragColor = in.vColor * uTex.sample(uTexSmplr, in.vTex.xy);
+    out.FragColor = in.vColor * uTex.sample(uTexSmplr, in.vTex);
     return out;
 }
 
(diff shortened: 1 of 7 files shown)


billhollings:
@HansKristian-ARM 

Would you mind having a look at the failed Travis build on `shaders-msl/frag/mix.frag` please?

There seems to be some platform-dependent misery going on here with `emit_mix_op()` statements.

Under OSX, the MSL shader is generated as in the `reference/shaders-msl/frag/mix.frag` file...with nice clean `?:` ternary statements everywhere. On the other hand, Travis is generating [this mess](https://travis-ci.org/KhronosGroup/SPIRV-Cross/jobs/200077036#L1441), which seems to be a combination of ternary and temp-var-based `if()` statements.

I've tried following the logic of `emit_mix_op()`...but I can't see any reason why the code would not always emit a nice ternary operator...and since it always does in my OSX environment, I am not able to effectively debug what's going on under the Travis environment.

HansKristian-ARM:
Will have a look tomorrow.

HansKristian-ARM:
Ok, I also get the if branchy variant. I think you're not using the same glslang checkout that the Travis script uses. It seems like that glslang version in Travis does not optimize the mix statements properly while perhaps your recent checkout does.

It probably makes sense to update the glslang revision now in the travis script and update all the reference shaders along with it.

HansKristian-ARM:
Indeed, it's a new glslang ... I'll update Travis to use latest glslang, it uncovered a lingering bug in emit_mix_op anyways.

HansKristian-ARM:
Merged it. Please see .travis.yml for the exact commit Travis is using for glslang.

billhollings:
@HansKristian-ARM 

Thanks for the fix. Travis is happy with this pull request now.

PR #122, opened 2017-02-15
Update to Xcode 8.2 on Travis and reenable Metal validation

zeux:
We need recent Xcode to support ios-metal1.2


HansKristian-ARM:
Thanks for the fix!

PR #134, opened 2017-03-06
CompilerMSL aligns MSL struct members with SPIR-V offsets

billhollings:
MSL and SPIR-V (std140) have slightly different structure member size and alignment rules.
This PR pads and packs MSL struct members to align them with the SPIR-V offsets.

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -432,7 +432,7 @@ struct CLIArguments
 
 	uint32_t iterations = 1;
 	bool cpp = false;
-	bool metal = false;
+	bool msl = false;
 	bool msl_pack_ubos = true;
 	bool hlsl = false;
 	bool vulkan_semantics = false;
@@ -443,10 +443,11 @@ struct CLIArguments
 static void print_help()
 {
 	fprintf(stderr, "Usage: spirv-cross [--output <output path>] [SPIR-V file] [--es] [--no-es] [--no-cfg-analysis] "
-	                "[--version <GLSL "
-	                "version>] [--dump-resources] [--help] [--force-temporary] [--cpp] [--cpp-interface-name <name>] "
-	                "[--metal] [--hlsl] [--shader-model] [--vulkan-semantics] [--flatten-ubo] [--fixup-clipspace] "
-	                "[--iterations iter] "
+	                "[--version <GLSL version>] [--dump-resources] [--help] [--force-temporary] "
+	                "[--vulkan-semantics] [--flatten-ubo] [--fixup-clipspace] [--iterations iter] "
+	                "[--cpp] [--cpp-interface-name <name>] "
+	                "[--msl] [--msl-no-pack-ubos] "
+	                "[--hlsl] [--shader-model] "
 	                "[--pls-in format input-name] [--pls-out format output-name] [--remap source_name target_name "
 	                "components] [--extension ext] [--entry name] [--remove-unused-variables] "
 	                "[--remap-variable-type <variable_name> <new_variable_type>]\n");
@@ -566,7 +567,8 @@ int main(int argc, char *argv[])
 	cbs.add("--iterations", [&args](CLIParser &parser) { args.iterations = parser.next_uint(); });
 	cbs.add("--cpp", [&args](CLIParser &) { args.cpp = true; });
 	cbs.add("--cpp-interface-name", [&args](CLIParser &parser) { args.cpp_interface_name = parser.next_string(); });
-	cbs.add("--metal", [&args](CLIParser &) { args.metal = true; });
+	cbs.add("--metal", [&args](CLIParser &) { args.msl = true; }); // Legacy compatibility
+	cbs.add("--msl", [&args](CLIParser &) { args.msl = true; });
 	cbs.add("--msl-no-pack-ubos", [&args](CLIParser &) { args.msl_pack_ubos = false; });
 	cbs.add("--hlsl", [&args](CLIParser &) { args.hlsl = true; });
 	cbs.add("--vulkan-semantics", [&args](CLIParser &) { args.vulkan_semantics = true; });
@@ -632,7 +634,7 @@ int main(int argc, char *argv[])
 		if (args.cpp_interface_name)
 			static_cast<CompilerCPP *>(compiler.get())->set_interface_name(args.cpp_interface_name);
 	}
-	else if (args.metal)
+	else if (args.msl)
 	{
 		compiler = unique_ptr<CompilerMSL>(new CompilerMSL(read_spirv_file(args.input)));
 
--- a/main.cpp
+++ b/main.cpp
@@ -433,6 +433,7 @@ struct CLIArguments
 	uint32_t iterations = 1;
 	bool cpp = false;
 	bool metal = false;
+	bool msl_pack_ubos = true;
 	bool hlsl = false;
 	bool vulkan_semantics = false;
 	bool remove_unused = false;
@@ -566,6 +567,7 @@ int main(int argc, char *argv[])
 	cbs.add("--cpp", [&args](CLIParser &) { args.cpp = true; });
 	cbs.add("--cpp-interface-name", [&args](CLIParser &parser) { args.cpp_interface_name = parser.next_string(); });
 	cbs.add("--metal", [&args](CLIParser &) { args.metal = true; });
+	cbs.add("--msl-no-pack-ubos", [&args](CLIParser &) { args.msl_pack_ubos = false; });
 	cbs.add("--hlsl", [&args](CLIParser &) { args.hlsl = true; });
 	cbs.add("--vulkan-semantics", [&args](CLIParser &) { args.vulkan_semantics = true; });
 	cbs.add("--extension", [&args](CLIParser &parser) { args.extensions.push_back(parser.next_string()); });
@@ -631,7 +633,14 @@ int main(int argc, char *argv[])
 			static_cast<CompilerCPP *>(compiler.get())->set_interface_name(args.cpp_interface_name);
 	}
 	else if (args.metal)
+	{
 		compiler = unique_ptr<CompilerMSL>(new CompilerMSL(read_spirv_file(args.input)));
+
+		auto *msl_comp = static_cast<CompilerMSL *>(compiler.get());
+		auto msl_opts = msl_comp->get_options();
+		msl_opts.pad_and_pack_uniform_structs = args.msl_pack_ubos;
+		msl_comp->set_options(msl_opts);
+	}
 	else if (args.hlsl)
 		compiler = unique_ptr<CompilerHLSL>(new CompilerHLSL(read_spirv_file(args.input)));
 	else
--- a/spirv_cross.cpp
+++ b/spirv_cross.cpp
@@ -920,6 +920,11 @@ uint64_t Compiler::get_member_decoration_mask(uint32_t id, uint32_t index) const
 	return m.members[index].decoration_flags;
 }
 
+bool Compiler::has_member_decoration(uint32_t id, uint32_t index, Decoration decoration)
+{
+	return (get_member_decoration_mask(id, index) & (1ull << decoration));
+}
+
 void Compiler::unset_member_decoration(uint32_t id, uint32_t index, Decoration decoration)
 {
 	auto &m = meta.at(id);
@@ -1015,6 +1020,11 @@ uint64_t Compiler::get_decoration_mask(uint32_t id) const
 {
 	auto &dec = meta.at(id).decoration;
 	return dec.decoration_flags;
+}
+
+bool Compiler::has_decoration(uint32_t id, Decoration decoration)
+{
+	return (get_decoration_mask(id) & (1ull << decoration));
 }
 
 uint32_t Compiler::get_decoration(uint32_t id, Decoration decoration) const
--- a/spirv_cross.cpp
+++ b/spirv_cross.cpp
@@ -920,9 +920,9 @@ uint64_t Compiler::get_member_decoration_mask(uint32_t id, uint32_t index) const
 	return m.members[index].decoration_flags;
 }
 
-bool Compiler::has_member_decoration(uint32_t id, uint32_t index, Decoration decoration)
+bool Compiler::has_member_decoration(uint32_t id, uint32_t index, Decoration decoration) const
 {
-	return (get_member_decoration_mask(id, index) & (1ull << decoration));
+	return get_member_decoration_mask(id, index) & (1ull << decoration);
 }
 
 void Compiler::unset_member_decoration(uint32_t id, uint32_t index, Decoration decoration)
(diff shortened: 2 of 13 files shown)


spirv_cross.cpp
```diff
@@ -920,6 +920,11 @@ uint64_t Compiler::get_member_decoration_mask(uint32_t id, uint32_t index) const
 	return m.members[index].decoration_flags;
 }
 
+bool Compiler::has_member_decoration(uint32_t id, uint32_t index, Decoration decoration)
+{
+	return (get_member_decoration_mask(id, index) & (1ull << decoration));
```
HansKristian-ARM: Redundant parens.

spirv_cross.cpp
```diff
@@ -1017,6 +1022,11 @@ uint64_t Compiler::get_decoration_mask(uint32_t id) const
 	return dec.decoration_flags;
 }
 
+bool Compiler::has_decoration(uint32_t id, Decoration decoration)
+{
+	return (get_decoration_mask(id) & (1ull << decoration));
```
HansKristian-ARM: Redundant parens.

spirv_cross.hpp
```diff
@@ -137,6 +137,9 @@ class Compiler
 	// I.e. (1ull << spv::DecorationFoo) | (1ull << spv::DecorationBar)
 	uint64_t get_decoration_mask(uint32_t id) const;
 
+	// Returns whether the decoration has been applied to the ID.
+	bool has_decoration(uint32_t id, spv::Decoration decoration);
```
HansKristian-ARM: Const member function.

spirv_cross.hpp
```diff
@@ -179,6 +182,9 @@ class Compiler
 	// Gets the decoration mask for a member of a struct, similar to get_decoration_mask.
 	uint64_t get_member_decoration_mask(uint32_t id, uint32_t index) const;
 
+	// Returns whether the decoration has been applied to a member of a struct.
+	bool has_member_decoration(uint32_t id, uint32_t index, spv::Decoration decoration);
```
HansKristian-ARM: Const member function.

spirv_glsl.cpp
```diff
@@ -504,9 +504,7 @@ void CompilerGLSL::emit_struct(SPIRType &type)
 	for (auto &member : type.member_types)
 	{
 		add_member_name(type, i);
-
-		auto &membertype = get<SPIRType>(member);
-		statement(member_decl(type, membertype, i), ";");
+		emit_stuct_member(type, member, i);
```
HansKristian-ARM: stuct typo

spirv_msl.hpp
```diff
@@ -67,6 +68,9 @@ struct MSLResourceBinding
 	bool used_by_shader = false;
 };
 
+// Tracks the type ID and member index of a struct member
+typedef uint64_t MSLStructMemberKey;
```
HansKristian-ARM: Prefer C++11 `using MSLStructMemberKey = uint64_t`.

spirv_msl.cpp
```diff
@@ -2088,36 +2125,91 @@ size_t CompilerMSL::get_declared_type_size(uint32_t type_id, uint64_t dec_mask)
...
-		else if ((dec_mask & (1ull << DecorationColMajor)) && vecsize == 3)
-			vecsize = 4;
+	case SPIRType::Unknown:
+	case SPIRType::Void:
+	case SPIRType::AtomicCounter:
+	case SPIRType::Image:
+	case SPIRType::SampledImage:
+	case SPIRType::Sampler:
+		SPIRV_CROSS_THROW("Querying alignment of opaque object.");
+		return 4; // A pointer
```
HansKristian-ARM: Misleading to return after throw.

spirv_msl.cpp
```diff
@@ -2088,36 +2125,91 @@ size_t CompilerMSL::get_declared_type_size(uint32_t type_id, uint64_t dec_mask)
 	case SPIRType::Image:
 	case SPIRType::SampledImage:
 	case SPIRType::Sampler:
-		SPIRV_CROSS_THROW("Querying size of object with opaque size.");
-	default:
-		break;
-	}
+		SPIRV_CROSS_THROW("Querying size of opaque object.");
+		return 4; // A pointer
```
HansKristian-ARM: Misleading to return after throw.

spirv_msl.cpp
```diff
@@ -577,61 +577,87 @@ uint32_t CompilerMSL::get_input_buffer_block_var_id(uint32_t msl_buffer)
...
+		// before the current member, or mark the previous member as packed, so that it will
+		// consume less space, and the current member can move forward.
+		// There is a situation where, in packing the previous member, the alignment of
+		// that member will change, necessitating the addition of a padding member before it.
+		// This is taken care of automatically because this compiler is multipass. The first
+		// pass will mark the packed member, and the second pass will insert a padding member.
+		// If we ever move to a single-pass design, this will break.
 		uint32_t mbr_offset = get_member_decoration(ib_type_id, mbr_idx, DecorationOffset);
-		uint32_t gap = mbr_offset - curr_offset;
+		int32_t gap = mbr_offset - curr_offset;
```
HansKristian-ARM: Cast mbr_offset and curr_offset to int32_t. Wrap-around is technically undefined, so the gap < 0 path might get optimized out.

spirv_glsl.cpp
```diff
@@ -3371,6 +3374,13 @@ string CompilerGLSL::access_chain_internal(uint32_t base, const uint32_t *indice
 				expr += ".";
 				expr += to_member_name(*type, index);
 			}
+
+			if (member_is_packed_type(*type, index))
+			{
+				auto &membertype = get<SPIRType>(type->member_types[index]);
+				expr = unpack_expression_type(expr, membertype);
```
HansKristian-ARM: This could get problematic when we need to support storing to a packed member, because this expression now has turned into l-value.

spirv_glsl.hpp
```diff
@@ -159,8 +159,8 @@ class CompilerGLSL : public Compiler
 	virtual void emit_texture_op(const Instruction &i);
 	virtual std::string type_to_glsl(const SPIRType &type);
 	virtual std::string builtin_to_glsl(spv::BuiltIn builtin);
-	virtual std::string member_decl(const SPIRType &type, const SPIRType &member_type, uint32_t member,
-	                                const std::string &qualifier = "");
+	virtual void emit_stuct_member(const SPIRType &type, const uint32_t member_type_id, uint32_t index,
```
HansKristian-ARM: const uint32_t seems odd.

spirv_glsl.cpp
```diff
@@ -5505,6 +5515,13 @@ bool CompilerGLSL::member_is_non_native_row_major_matrix(const SPIRType &type, u
 	return true;
 }
 
+// Checks whether the member is in packed data type, that might need to be unpacked.
+// GLSL does not define packed data types, but certain subclasses do.
+bool CompilerGLSL::member_is_packed_type(const SPIRType &type, uint32_t index)
+{
+	return has_member_decoration(type.self, index, DecorationCPacked);
```
HansKristian-ARM: Is CPacked a standard SPIR-V decoration? (quite convenient here :P)

spirv_cross.cpp
billhollings: I assume you mean the outer parens? I've removed them now.

spirv_cross.cpp
billhollings: I assume you mean the outer parens? I've removed them now.

spirv_cross.hpp
billhollings: Fixed for next push

spirv_cross.hpp
billhollings: Ditto.

spirv_glsl.cpp
billhollings: Good catch! Fixed for next push.

spirv_msl.cpp
billhollings: Fixed for next push.

spirv_msl.cpp
billhollings: Fixed for next push.

spirv_msl.cpp
billhollings: Hmmm...[some reading](http://stackoverflow.com/a/7221449/1678054) I had done before implementing this indicates that wrap-around of uint subtraction IS well defined...and would be cast correctly to the corresponding negative int value.

Nevertheless...I've added the int casts anyway...since it is unambiguous and clearer.

spirv_msl.hpp
billhollings: Changed.

spirv_glsl.hpp
billhollings: Don't know where that came from. Thanks. Fixed now.

spirv_glsl.cpp
billhollings: Yes! I was happy to find that decoration...because it meant not having to add an iVar to the type struct to track this.

spirv_glsl.cpp
billhollings: Yes. I wrestled with this...and actually explored passing the equivalent of need_transpose back...but ultimately aborted that out of concern about complexity.

As an l-value, there are two primary use cases for unpacking:
1) Allow assignment to a variable defined as the same vector type, but which would not be generally packed.
2) Allow component access using x, y... syntax instead of [0], [1]... Metal does not permit the former on packed types for some reason.
 
Regarding the potential use as an r-value...that is indeed problematic...but I am hoping very rare. This packing only applies to uniform buffer blocks...as input...and not to vertex or other array content that would be more commonly written to.

billhollings:
@HansKristian-ARM 

- I'll look into adding a small test case.

- I'm not sure what  you are asking regarding clarifying MSL packing. MSL alignment and packing rules can be found in Section 2.8-2.9 of the [MSL Spec](https://developer.apple.com/metal/metal-shading-language-specification.pdf).

- R-values are definitely a problem currently.

spirv_msl.cpp
HansKristian-ARM: Casting to int must not overflow. E.g. the unsigned subtraction might wrap, creating a value larger than INT_MAX. Compiler is free to assume that never happens, i.e. the subtraction result is always >= 0.

billhollings:
@HansKristian-ARM 

Fixes and UBO pack/pad test case posted to PR.

main.cpp
```diff
@@ -566,6 +567,7 @@ int main(int argc, char *argv[])
 	cbs.add("--cpp", [&args](CLIParser &) { args.cpp = true; });
 	cbs.add("--cpp-interface-name", [&args](CLIParser &parser) { args.cpp_interface_name = parser.next_string(); });
 	cbs.add("--metal", [&args](CLIParser &) { args.metal = true; });
+	cbs.add("--msl-no-pack-ubos", [&args](CLIParser &) { args.msl_pack_ubos = false; });
```
HansKristian-ARM: --metal-no-pack-ubos to be consistent with --metal.

spirv_msl.cpp
```diff
@@ -27,13 +27,20 @@ using namespace std;
 
 static const uint32_t k_unknown_location = ~0;
 
-CompilerMSL::CompilerMSL(vector<uint32_t> spirv_)
+CompilerMSL::CompilerMSL(vector<uint32_t> spirv_, vector<MSLVertexAttr> *p_vtx_attrs,
+                         std::vector<MSLResourceBinding> *p_res_bindings)
```
HansKristian-ARM: Inconsistent std:: usage.
Also, must the arguments be non-const?
It seems like CompilerMSL ends up storing pointers owned by these arguments. Is this behavior documented?

spirv_msl.cpp
```diff
@@ -125,10 +119,30 @@ string CompilerMSL::compile(MSLConfiguration &msl_cfg, vector<MSLVertexAttr> *p_
 	return buffer->str();
 }
 
-string CompilerMSL::compile()
+string CompilerMSL::compile(vector<MSLVertexAttr> *p_vtx_attrs, std::vector<MSLResourceBinding> *p_res_bindings)
```
HansKristian-ARM: Inconsistent std:: usage.
Can the arguments be const?

spirv_msl.hpp
```diff
@@ -27,14 +27,8 @@
...
-// Options for compiling to Metal Shading Language
-struct MSLConfiguration
-{
-	bool flip_vert_y = false;
-	bool flip_frag_y = false;
-	bool is_rendering_points = false;
-	std::string entry_point_name;
-};
+// Deprecated legacy syntax
+#define MSLConfiguration CompilerMSL::Options
```
HansKristian-ARM: Can this be a typedef instead ala using MSLConfiguration = CompilerMSL::Options?

main.cpp
billhollings: Before I make this change...the use of "metal" is inconsistent with the use of hlsl.

HansKristian-ARM: If the use of metal is basically only here, maybe add a new alias --msl (and keep --metal for compat) instead. All the CLI arguments atm are --metal, but in the code it's msl ...

billhollings: ...so should we change "metal" to "msl" instead?

So...

	./test_shaders.py --msl shaders-msl

instead of

	./test_shaders.py --metal shaders-msl

and then

	./test_shaders.py  --msl shaders-msl --msl-no-pack-ubos

HansKristian-ARM: Yup. No need to be backwards compat with the test scripts at least.

billhollings: This would also make it consistent with HLSL...

	./test_shaders.py --hlsl shaders-hlsl

where the equivalent of "metal" would be to use "directx" or "dx", etc.

HansKristian-ARM: Not quite the same imo, as it's Metal vs Metal Shading Language, but HLSL would be --high-level-shading-language, but anyways, seems sensible to use msl throughout the code base if we're going to choose between metal and msl.

Only thing that needs to remain is --metal in the CLI for compatibility (but help should say --msl, and --msl should be supported as well).
If using --metal option, an stderr message could say it's deprecated in favor of --msl.

spirv_msl.cpp
billhollings: Removed unnecessary use of std:: in cpp file.

Incoming vectors cannot be marked as `const` because the values in the vectors may be modified during compilation to have their `used_by_shader` flag set. This is documented in constructor API.

spirv_msl.cpp
billhollings: Removed unnecessary use of std:: in cpp file.

Incoming vectors cannot be marked as `const` because the values in the vectors may be modified during compilation to have their `used_by_shader` flag set. This is documented in constructor API.

spirv_msl.hpp
billhollings: Using typedef now.

main.cpp
billhollings: Deprecated `--metal` CLI option and replaced with `--msl` option. Original `--metal` still valid, but now undocumented.

billhollings:
Requested fixes are complete.

PR #135, opened 2017-03-07
Just fix a few minor warnings.

hrydgard:
The variable i was used as a uint32_t in multiple places, like type_member_for_offset, so changed its type.

I build SPIRV-Cross as part of my project so I see these quite a bit.

### DIFF
--- a/spirv_glsl.cpp
+++ b/spirv_glsl.cpp
@@ -3531,7 +3531,7 @@ std::string CompilerGLSL::flattened_access_chain_struct(uint32_t base, const uin
 	expr += type_to_glsl_constructor(target_type);
 	expr += "(";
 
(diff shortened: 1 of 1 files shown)


spirv_glsl.cpp
```diff
@@ -3531,7 +3531,7 @@ std::string CompilerGLSL::flattened_access_chain_struct(uint32_t base, const uin
 	expr += type_to_glsl_constructor(target_type);
 	expr += "(";
 
-	for (size_t i = 0; i < target_type.member_types.size(); ++i)
+	for (uint32_t i = 0; i < (uint32_t)target_type.member_types.size(); ++i)
```
HansKristian-ARM: Use C++ cast uint32_t(...).

hrydgard:
Style nit fixed, CLA signed!

PR #136, opened 2017-03-07
Rewrite HLSL backend handling of inputs and outputs

HansKristian-ARM:
Rewrote the implementation of HLSL inputs and outputs to be more robust and support more stuff.

- A traversal is done to find all statically accessed builtin variables in the shader. This is common to all APIs, so Metal might benefit from this as well.
- Separate out logic for builtins from the handling of user-defined inputs and outputs.
- Add support for the more common builtins like frag-coord, frag-depth, vertex-id, instance-id and MRT.
- Location decorations are now used always if the shader uses it. For varyings and attributes which are not using these, the locations (TEXCOORD[N]) are assigned based on their name, and failing that, their ID. All sensible SPIR-V shaders really should use locations, so these serve as crude fallbacks.
- I/O blocks are now supported by having extra in/out struct arguments from main(). This is to support geometry shaders and tessellation in the future ...
- VPOS in SM 3.0 gets a half-pixel offset applied to match all other APIs, not sure if this offset depends on the SM, or the API being used (like, D3D11 using D3D9 shaders, will that have half-pixel offset already applied?)
- Interpolation qualifiers are supported.

I need some review of this because while the test shaders pass glslangValidator HLSL mode, I'm not sure how HLSL really treats the vertex output/fragment input semantics, are they ignored, does ordering matter, and so on.

The HLSL linking interface between stages seems kinda shaky to me, so I'd like some clarification from someone who knows HLSL better.

### DIFF
--- a/reference/shaders-hlsl/frag/basic.frag
+++ b/reference/shaders-hlsl/frag/basic.frag
@@ -4,15 +4,15 @@ static float4 FragColor;
 static float4 vColor;
 static float2 vTex;
 
-struct InputFrag
+struct SPIRV_Cross_Input
 {
     float4 vColor : TEXCOORD0;
     float2 vTex : TEXCOORD1;
 };
 
-struct OutputFrag
+struct SPIRV_Cross_Output
 {
-    float4 FragColor : COLOR;
+    float4 FragColor : COLOR0;
 };
 
 void frag_main()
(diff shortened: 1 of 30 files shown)


HansKristian-ARM:
Ping @RobDangerous.

HansKristian-ARM:
Re-ping @RobDangerous.

RobDangerous:
Oh, sorry! Will check tomorrow.

RobDangerous:
This will work fine for krafix.

HansKristian-ARM:
Thanks.

Issue #146, opened 2017-03-22
Textures in vertex shaders in WebGL


RobDangerous:
Kind of a silly situation here as far as I can tell:
According to https://www.khronos.org/registry/spir-v/specs/1.1/SPIRV.html#OpImageSampleImplicitLod imlicit lods are disallowed in vertex shaders and glslang acts accordingly. On the other hand https://www.khronos.org/registry/webgl/extensions/EXT_shader_texture_lod/ is only defined for fragment shaders and browsers seem to act accordingly (see Kode/Kha#418 - could reproduce using Chrome 57 in Windows) while supporting regular texture2D calls just fine in vertex shaders.
So I guess we should change texture2DLodEXT back to texture2D when we are in legacy es and lod is 0 (glslang sets lod to 0 when it's not explicitly defined). Changes get a bit messy so wanted to discuss it first...

HansKristian-ARM:
Well, that's an odd ball.

The logic doesn't have to be so bad. You could check if you're emitting texture2DLodEXT and the stage is vertex, you could then check if the explicit LOD is a SPIRConstant with scalar value 0.0, if so, replace with texture2D.

RobDangerous:
Will do a PR using that exact strategy (started with that, have to change the interface a bit to pass the lod around...).

rdb:
The suggested fix is not right.  In fact, GLSL and ESSL have provided texture2DLod for vertex shaders from the beginning.  This is why the referenced extensions are only defined for fragment shaders—they are simply extending the functionality that was already available for vertex shaders to the fragment shader.

I have filed #1528 to address this.

PR #148, opened 2017-03-23
Prefer plain texture2D in legacy es vertex shaders

RobDangerous:
This is the fix for #146

### DIFF
--- a/spirv_glsl.cpp
+++ b/spirv_glsl.cpp
@@ -2413,11 +2413,19 @@ void CompilerGLSL::emit_quaternary_func_op(uint32_t result_type, uint32_t result
 // are not allowed in ES 2 vertex shaders. But SPIR-V only supports lod tex
 // functions in vertex shaders so we revert those back to plain calls when
 // the lod is a constant value of zero.
-bool CompilerGLSL::check_lod_allowed(uint32_t lod)
+bool CompilerGLSL::check_explicit_lod_allowed(uint32_t lod)
 {
 	auto &execution = get_entry_point();
-	auto *lod_constant = maybe_get<SPIRConstant>(lod);
-	return is_legacy_es() && execution.model == ExecutionModelVertex && lod_constant && lod_constant->scalar_f32() == 0.0f;
+	bool allowed = !is_legacy_es() || execution.model == ExecutionModelFragment;
+	if (!allowed)
+	{
+		auto *lod_constant = maybe_get<SPIRConstant>(lod);
+		if (!lod_constant || lod_constant->scalar_f32() != 0.0f)
+		{
+			SPIRV_CROSS_THROW("Explicit lod not allowed in legacy ES non-fragment shaders.");
+		}
+	}
+	return allowed;
 }
 
 string CompilerGLSL::legacy_tex_op(const std::string &op, const SPIRType &imgtype, uint32_t lod)
@@ -2448,13 +2456,13 @@ string CompilerGLSL::legacy_tex_op(const std::string &op, const SPIRType &imgtyp
 		break;
 	}
 
-	bool avoid_lod = check_lod_allowed(lod);
+	bool use_explicit_lod = check_explicit_lod_allowed(lod);
 
 	if (op == "textureLod" || op == "textureProjLod")
 	{
 		if (is_legacy_es())
 		{
-			if (!avoid_lod)
+			if (use_explicit_lod)
 				require_extension("GL_EXT_shader_texture_lod");
 		}
 		else if (is_legacy())
@@ -2465,18 +2473,20 @@ string CompilerGLSL::legacy_tex_op(const std::string &op, const SPIRType &imgtyp
 		return join("texture", type);
 	else if (op == "textureLod")
 	{
-		if (avoid_lod)
-			return join("texture", type);
-		else
+		if (use_explicit_lod)
 			return join("texture", type, is_legacy_es() ? "LodEXT" : "Lod");
+		else
+			return join("texture", type);
 	}
 	else if (op == "textureProj")
 		return join("texture", type, "Proj");
 	else if (op == "textureProjLod")
-		if (avoid_lod)
-			return join("texture", type);
-		else
+	{
+		if (use_explicit_lod)
 			return join("texture", type, is_legacy_es() ? "ProjLodEXT" : "ProjLod");
+		else
+			return join("texture", type);
+	}
 	else
 	{
 		SPIRV_CROSS_THROW(join("Unsupported legacy texture op: ", op));
@@ -2918,7 +2928,7 @@ string CompilerGLSL::to_function_args(uint32_t img, const SPIRType &, bool, bool
 
 	if (lod)
 	{
-		if (check_lod_allowed(lod))
+		if (check_explicit_lod_allowed(lod))
 		{
 			forward = forward && should_forward(lod);
 			farg_str += ", ";
--- a/spirv_glsl.cpp
+++ b/spirv_glsl.cpp
@@ -2409,7 +2409,7 @@ void CompilerGLSL::emit_quaternary_func_op(uint32_t result_type, uint32_t result
 	inherit_expression_dependencies(result_id, op3);
 }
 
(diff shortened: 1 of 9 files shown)


spirv_glsl.cpp
```diff
@@ -2437,22 +2437,32 @@ string CompilerGLSL::legacy_tex_op(const std::string &op, const SPIRType &imgtyp
...
-		else if (is_legacy())
+		else if (is_legacy() && !is_legacy_es())
 			require_extension("GL_ARB_shader_texture_lod");
 	}
 
 	if (op == "texture")
 		return join("texture", type);
-	else if (op == "textureLod")
-		return join("texture", type, is_legacy_es() ? "LodEXT" : "Lod");
+	else if (op == "textureLod") {
```
HansKristian-ARM: Brace placement should be here.

spirv_glsl.cpp
```diff
@@ -2437,22 +2437,32 @@ string CompilerGLSL::legacy_tex_op(const std::string &op, const SPIRType &imgtyp
 		break;
 	}
 
+	auto *lod_constant = maybe_get<SPIRConstant>(lod);
+	bool zero_lod = lod_constant && lod_constant->scalar_f32() == 0.0f;
```
HansKristian-ARM: Doesn't check if we're in vertex.

HansKristian-ARM: Add some comments here what we're trying to do.

spirv_glsl.cpp
```diff
@@ -2437,22 +2437,32 @@ string CompilerGLSL::legacy_tex_op(const std::string &op, const SPIRType &imgtyp
...
+
 	if (op == "textureLod" || op == "textureProjLod")
 	{
-		if (is_legacy_es())
+		if (is_legacy_es() && !zero_lod)
```
HansKristian-ARM: Probably cleaner to just nest the ifs to avoid the weird else if, if (is_legacy_es()) { if (!zero_lod) } ...

HansKristian-ARM:
See shaders/legacy for how to create test cases for legacy ESSL.

RobDangerous:
I'm a little confused because the legacy shaders are "#version 310 es" while is_legacy_es checks for < 300. Is that an error?

HansKristian-ARM:
If you have .legacy. in the file name of the shader, it will use ESSL 1.0 when cross-compiling instead of the original version.

spirv_glsl.cpp
```diff
@@ -2787,7 +2811,7 @@ void CompilerGLSL::emit_texture_op(const Instruction &i)
 // Returns the function name for a texture sampling function for the specified image and sampling characteristics.
 // For some subclasses, the function is a method on the specified image.
 string CompilerGLSL::to_function_name(uint32_t, const SPIRType &imgtype, bool is_fetch, bool is_gather, bool is_proj,
-                                      bool has_array_offsets, bool has_offset, bool has_grad, bool has_lod, bool)
+                                      bool has_array_offsets, bool has_offset, bool has_grad, bool has_lod, bool, uint32_t lod)
```
HansKristian-ARM: Instead of sending in an extra argument, can't has_lod be changed to take the uint32_t value instead?

spirv_glsl.cpp
```diff
@@ -2437,22 +2448,35 @@ string CompilerGLSL::legacy_tex_op(const std::string &op, const SPIRType &imgtyp
...
+		if (avoid_lod)
+			return join("texture", type);
+		else
+			return join("texture", type, is_legacy_es() ? "LodEXT" : "Lod");
+	}
 	else if (op == "textureProj")
 		return join("texture", type, "Proj");
 	else if (op == "textureProjLod")
-		return join("texture", type, is_legacy_es() ? "ProjLodEXT" : "ProjLod");
+		if (avoid_lod)
```
HansKristian-ARM: Brace around this if/else.

spirv_glsl.cpp
```diff
@@ -2437,22 +2448,35 @@ string CompilerGLSL::legacy_tex_op(const std::string &op, const SPIRType &imgtyp
 		break;
 	}
 
+	bool avoid_lod = check_lod_allowed(lod);
```
HansKristian-ARM: Negate the check?

spirv_glsl.cpp
```diff
@@ -2409,7 +2409,18 @@ void CompilerGLSL::emit_quaternary_func_op(uint32_t result_type, uint32_t result
...
-string CompilerGLSL::legacy_tex_op(const std::string &op, const SPIRType &imgtype)
+// EXT_shader_texture_lod only concerns fragment shaders so lod tex functions
+// are not allowed in ES 2 vertex shaders. But SPIR-V only supports lod tex
+// functions in vertex shaders so we revert those back to plain calls when
+// the lod is a constant value of zero.
+bool CompilerGLSL::check_lod_allowed(uint32_t lod)
+{
+	auto &execution = get_entry_point();
+	auto *lod_constant = maybe_get<SPIRConstant>(lod);
+	return is_legacy_es() && execution.model == ExecutionModelVertex && lod_constant && lod_constant->scalar_f32() == 0.0f;
```
HansKristian-ARM: This returns the opposite of what you want?

HansKristian-ARM: execution.model != ExecutionModelFragment is more clear. Only fragment supports implicit LOD.

spirv_glsl.cpp
HansKristian-ARM: Probably clearer to check something positive like `bool use_explicit_lod = check_explicit_lod_allowed(lod)` or something like that.

RobDangerous:
Thinking about it some more the whole lod == 0.0 check is kind of superfluous - if it's not 0 and is_legacy_es() && execution.model != ExecutionModelFragment we would output an illegal shader. Check, error and exit instead? Output a legal but kind of different shader?

HansKristian-ARM:
I would avoid emitting a legal shader, but wrong shader silently. A SPIRV_CROSS_THROW is also fine if we have a legacy non-fragment shader and explicit LOD != 0.0. Should simplify things a bit.

spirv_glsl.cpp
```diff
@@ -2409,7 +2409,26 @@ void CompilerGLSL::emit_quaternary_func_op(uint32_t result_type, uint32_t result
...
+// functions in vertex shaders so we revert those back to plain calls when
+// the lod is a constant value of zero.
+bool CompilerGLSL::check_explicit_lod_allowed(uint32_t lod)
+{
+	auto &execution = get_entry_point();
+	bool allowed = !is_legacy_es() || execution.model == ExecutionModelFragment;
+	if (!allowed)
+	{
+		auto *lod_constant = maybe_get<SPIRConstant>(lod);
+		if (!lod_constant || lod_constant->scalar_f32() != 0.0f)
```
HansKristian-ARM: Shouldn't this be more like `if (lod_constant && lod_constant->scalar_f32())` ?
If there is no explicit LOD (i.e. lod ID == 0), there wouldn't be any problem?

RobDangerous: I think it's correct as it is. If it is not a constant throw. If it is a constant which is not 0 also throw.

HansKristian-ARM: Hm, but it might be called if lod == 0, i.e. unused. Maybe if (!allowed && lod) then?

RobDangerous: Because it can throw if lod is 0 and not a constant? I think that's ok because in that case the original code did use an explicit lod value (which turned out to be 0).

HansKristian-ARM: I'm a bit confused. For lod here, the argument is an ID. If lod has ID 0, it means explicit was not used in the texture sampling operation.

RobDangerous: Ah, I'm talking about an actual float value of 0, that's what glslang outputs when no explicit lod was originally used in a vertex shader.

HansKristian-ARM:
There is still the issue of checking for lod != 0 before throwing, and replacing the has_lod argument with the uint32_t lod argument.

HansKristian-ARM:
Thanks.

PR #150, opened 2017-03-24
Workaround empty I/O variable declarations

HansKristian-ARM:


redorav:
That's very cool thanks! Will give it a go soon.

PR #156, opened 2017-04-01
Added a Compiler ctor that takes IR as raw array and count

kayru:
This avoids the need to construct a temporary std::vector on the application side just to create a Compiler instance if application itself doesn't use STL containers.

### DIFF
--- a/spirv_cross.cpp
+++ b/spirv_cross.cpp
@@ -70,6 +70,12 @@ Compiler::Compiler(vector<uint32_t> ir)
 	parse();
 }
 
+Compiler::Compiler(const uint32_t *ir, size_t word_count)
+    : spirv(ir, ir + word_count)
+{
+	parse();
+}
+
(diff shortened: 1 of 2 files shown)


spirv_cross.cpp
```diff
@@ -70,6 +70,12 @@ Compiler::Compiler(vector<uint32_t> ir)
 	parse();
 }
 
+Compiler::Compiler(const uint32_t* ir, size_t wordCount)
```
HansKristian-ARM: Asterisk binds to the right (Clang format script in format_all.sh).
word_count instead of wordCount.

spirv_cross.cpp
```diff
@@ -70,6 +70,12 @@ Compiler::Compiler(vector<uint32_t> ir)
...
+    : spirv(ir, ir + wordCount)
```
HansKristian-ARM: word_count

PR #157, opened 2017-04-01
Add vector-less IR construction to subclasses as well.

HansKristian-ARM:


spirv_msl.cpp
```diff
@@ -43,6 +41,20 @@ CompilerMSL::CompilerMSL(vector<uint32_t> spirv_, vector<MSLVertexAttr> *p_vtx_a
...
 }
 
+CompilerMSL::CompilerMSL(vector<uint32_t> spirv_, vector<MSLVertexAttr> *p_vtx_attrs,
+                         vector<MSLResourceBinding> *p_res_bindings)
+    : CompilerGLSL(move(spirv_))
+{
+	init(p_vtx_attrs, p_res_bindings);
+}
+
+CompilerMSL::CompilerMSL(const uint32_t *ir, size_t word_count, vector<MSLVertexAttr> *p_vtx_attrs,
```
HansKristian-ARM: Should we make p_vtx_attrs and p_res_bindings vector-less for this variants as well? It doesn't seem like these parameters *could* be raw pointers.

billhollings:
I've modified the vector-less CompilerMSL constructor to use vector-less attributes and resource bindings.

Issue #125, opened 2017-02-18
HLSL pixel shader: Even with SM5.0, old style texture sampling is used


hrydgard:
When choosing shader model 50, it still generates DX9-style texture sampling.

Input: 

```
#version 450
layout(binding = 0) uniform sampler2D sampler0;
layout(location = 0) in vec2 v_texcoord0;
layout(location = 0) out vec4 fragColor;
void main() {
  vec3 rgb = texture(sampler0, v_texcoord0.xy).xyz;
  float luma = dot(rgb, vec3(0.299, 0.587, 0.114));
  fragColor.rgb = vec3(luma, luma, luma);
  fragColor.a = 1.0;
}
```

    glslangValidator -H grayscale.frag -o grayscale.spv
    spirv-cross --hlsl --shader-model 50 grayscale.spv

```
uniform sampler2D sampler0;

static float2 v_texcoord0;
static float4 fragColor;

struct InputFrag
{
    float2 v_texcoord0 : TEXCOORD0;
};

struct OutputFrag
{
    float4 fragColor : COLOR;
};

void frag_main()
{
    float3 rgb = tex2D(sampler0, v_texcoord0).xyz;
    float luma = dot(rgb, float3(0.2989999949932098388671875f, 0.58700001239776611328125f, 0.114000000059604644775390625f));
    float3 _35 = float3(luma, luma, luma);
    fragColor = float4(_35.x, _35.y, _35.z, fragColor.w);
    fragColor.w = 1.0f;
}

OutputFrag main(InputFrag input)
{
    v_texcoord0 = input.v_texcoord0;
    frag_main();
    OutputFrag output;
    output.fragColor = fragColor;
    return output;
}
```

D3DCompile() does not accept tex2d when compiling for shader model 5.0. I would expect something like:

```
Texture2D<float4> sampler0_tex : register(t0);
SamplerState sampler0_samp : register(s0);
...
float3 rgb = sampler0_tex.Sample(sampler0_samp, v_texcoord0).xyz;
...
```

I suppose by default if the original sampler was in location 0, then the texture/sampler would be assigned to t0/s0, and so on, then you could override that by switching around the decorations if you like, having them share samplerstate or something.

Also I think all "uniform" prefixes should be dropped when generating HLSL, they are not normally used.

HansKristian-ARM:
Ping @RobDangerous. I noted this during the initial review, but I'm not sure how Robert got it working.

[...]

PR #161, opened 2017-04-18
Shader model 4/5 style texture sampling

RobDangerous:
For #125 aka no more need for D3DCOMPILE_ENABLE_BACKWARDS_COMPATIBILITY.
Also improves texture calls for shader model 2/3.

### DIFF
--- a/spirv_hlsl.cpp
+++ b/spirv_hlsl.cpp
@@ -1273,9 +1273,119 @@ void CompilerHLSL::emit_uniform(const SPIRVariable &var)
 {
 	add_resource_name(var.self);
 	auto &type = get<SPIRType>(var.basetype);
-	if (options.shader_model >= 40 && (type.basetype == SPIRType::Image || type.basetype == SPIRType::SampledImage))
+	if (options.shader_model >= 40 && type.basetype == SPIRType::SampledImage)
 	{
-		statement("Texture2D<float4> ", to_name(var.self), ";");
+		string format = "float4";
+		
+		switch (type.image.format) {
+		case ImageFormatUnknown:
+			format = "float4";
+			break;
+
+		case ImageFormatRgba32f:
+			format = "float4";
+			break;
+
+		case ImageFormatRgba16f:
+			format = "half4";
+			break;
+
+		case ImageFormatR32f:
+			format = "float";
+			break;
+
+		case ImageFormatRgba8:
+			format = "uint4";
+			break;
+		
+		case ImageFormatRg32f:
+			format = "float2";
+			break;
+
+		case ImageFormatRg16f:
+			format = "half2";
+			break;
+
+		case ImageFormatR11fG11fB10f:
+			format = "half3";
+			break;
+
+		case ImageFormatR16f:
+			format = "half";
+			break;
+
+		case ImageFormatRgba16:
+		case ImageFormatRgb10A2:
+			format = "uint4";
+			break;
+		
+		case ImageFormatRg16:
+		case ImageFormatRg8:
+			format = "uint";
+			break;
+
+		case ImageFormatR16:
+		case ImageFormatR8:
+			format = "uint";
+			break;
+
+		case ImageFormatRgba8Snorm:
+		case ImageFormatRgba16Snorm:
+			format = "snorm float4";
+			break;
+
+		case ImageFormatRg16Snorm:
+		case ImageFormatRg8Snorm:
+			format = "snorm float2";
+			break;
+
+		case ImageFormatR16Snorm:
+		case ImageFormatR8Snorm:
+			format = "snorm float";
+			break;
+
+		case ImageFormatRgba32i:
+		case ImageFormatRgba16i:
+		case ImageFormatRgba8i:
+			format = "int4";
+			break;
+
+		case ImageFormatRg32i:
+		case ImageFormatRg16i:
+		case ImageFormatRg8i:
+			format = "int2";
+			break;
+
+		case ImageFormatR32i:
+		case ImageFormatR16i:
+		case ImageFormatR8i:
+			format = "int";
+			break;
+
+		case ImageFormatRgba32ui:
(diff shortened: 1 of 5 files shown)


spirv_hlsl.cpp
```diff
@@ -1102,26 +1102,52 @@ void CompilerHLSL::emit_texture_op(const Instruction &i)
...
-			texop += "Grad";
-		if (lod)
-			texop += "Lod";
+		if (options.shader_model >= 40)
+		{
+			texop += to_expression(img);
+
+			if (gather)
+				texop += ".Gather";
+			else if (offset || coffsets || bias)
```
HansKristian-ARM: Is SampleBias() used even for coffsets?

spirv_hlsl.cpp
```diff
@@ -1102,26 +1102,52 @@ void CompilerHLSL::emit_texture_op(const Instruction &i)
...
+				texop += ".SampleBias";
+			else if (proj)
+				SPIRV_CROSS_THROW("textureProj is not supported in HLSL shader model 4/5.");
```
HansKristian-ARM: I think you'd have to manually divide xyz / w. That's basically how textureProj is defined anyways.

spirv_hlsl.cpp
```diff
@@ -1238,7 +1272,16 @@ void CompilerHLSL::emit_texture_op(const Instruction &i)
 void CompilerHLSL::emit_uniform(const SPIRVariable &var)
 {
 	add_resource_name(var.self);
-	statement(variable_decl(var), ";");
+	auto &type = get<SPIRType>(var.basetype);
+	if (options.shader_model >= 40 && (type.basetype == SPIRType::Image || type.basetype == SPIRType::SampledImage))
+	{
+		statement("Texture2D<float4> ", to_name(var.self), ";");
```
HansKristian-ARM: This only makes sense for OpTypeSampledImage (combined image sampler).
Is supporting separate image samplers TODO?

HansKristian-ARM: float4 could be int4 or uint4 depending on the image type.

HansKristian-ARM: Is it possible to move this code into emit_uniform()? iirc, CompilerHLSL overrides that.

RobDangerous: This is emit_uniform, did you mean something else?

HansKristian-ARM: nvm, brainfart.

spirv_hlsl.cpp
RobDangerous: Yes. Hm. Should coffsets be handled differently than offsets for some reason?

HansKristian-ARM: coffsets is just constant offset vs. offset which is dynamic, for purposes of code-gen, they are the same.

spirv_hlsl.cpp
```diff
@@ -1273,9 +1273,119 @@ void CompilerHLSL::emit_uniform(const SPIRVariable &var)
 {
 	add_resource_name(var.self);
 	auto &type = get<SPIRType>(var.basetype);
-	if (options.shader_model >= 40 && (type.basetype == SPIRType::Image || type.basetype == SPIRType::SampledImage))
+	if (options.shader_model >= 40 && type.basetype == SPIRType::SampledImage)
 	{
-		statement("Texture2D<float4> ", to_name(var.self), ";");
+		string format = "float4";
```
HansKristian-ARM: Hm, this "format" is for image load/store, for textures, you should look at image.type, to_glsl_type should suffice here.

RobDangerous: Ah, that's much easier. Actual i/usamplers do not work though because a Texture2D<int4> can not be sampled - HLSL only supports the Load method in that case which is pixel position based and does no interpolation. Not yet sure how to handle that situation.

spirv_hlsl.cpp
```diff
@@ -1146,6 +1193,12 @@ void CompilerHLSL::emit_texture_op(const Instruction &i)
 	// The IR can give us more components than we need, so chop them off as needed.
 	auto coord_expr = to_expression(coord) + swizzle(coord_components, expression_type(coord).vecsize);
 
+	if (options.shader_model >= 40 && proj)
+	{
+		requires_textureProj = true;
```
HansKristian-ARM: Shouldn't this trigger a recompile as well?

spirv_hlsl.cpp
HansKristian-ARM: o_O, guess you can throw if regular sampling is attempted on uint/int textures, texelFetch should work though.

spirv_hlsl.cpp
RobDangerous: It seems to work fine as it is in my tests.

HansKristian-ARM: Is force_recompile triggered some other way perhaps?

RobDangerous: Aye, something sets it to true regardless. Added it to be safe, also to OpFMod.

RobDangerous:
OK, want to continue with 3D textures, cube maps and image buffers right away though  - can continue here or put it in a new pull-request.

HansKristian-ARM:
If you want to continue on this PR, sure. I can also merge now and you can open another PR.

RobDangerous:
Let's continue here.

HansKristian-ARM:
k, I'll wait for more commits then.

HansKristian-ARM:
Is this ready for review now?

RobDangerous:
Not quite, want to do texelFetch and arrays on Monday (sorry, didn't originally plan to do everything at once, but now that I'm at it...).

RobDangerous:
Ready for review.

HansKristian-ARM:
Looks good.

HansKristian-ARM:
I'll have a look at separate image/samplers when I have time.

PR #165, opened 2017-04-25
HLSL: Support for separate image/samplers.

HansKristian-ARM:
This will get a bit tricky. Main complications:

- Sampler state needs to be SamplerState or SamplerComparisonState depending on how the sampler is used. We will need to fully traverse the call graph and see how OpSampledImage is used. If the result is used as part of Dref operations, we must use a comparison sampler.

- Just like MSL, we'll have to flatten combined samplers when passing them around as parameters.

It seems like this is a problem with GLSL as well. sampler vs. samplerShadow, and it's currently buggy.


RobDangerous:
Oh, didn't see this before. Guess I used a different strategy for passing combined samplers.

HansKristian-ARM:
Got the GLSL side of it sorted out, shouldn't be much work left now for HLSL.

PR #167, opened 2017-04-26
CompilerMSL round floating point tex coords for read() and tex array index.

billhollings:
CompilerMSL round floating point tex coords for read() and tex array index, to avoid toggling arbitrarily between two tex array layers due to rounding errors.
CompilerMSL remove option to flip fragment coordinates.


HansKristian-ARM:
Good to se flip_frag_y removed :)

PR #176, opened 2017-05-04
Proper types for sampled images in HLSL4/5

RobDangerous:
Fixes passing to functions,...

### DIFF
--- a/reference/shaders-hlsl/frag/basic.frag
+++ b/reference/shaders-hlsl/frag/basic.frag
@@ -1,5 +1,10 @@
-Texture2D<float4> uTex;
-SamplerState _uTex_sampler;
+struct SPIRV_Cross_Texture2D_float4_SamplerState
+{
+    Texture2D<float4> img;
+    SamplerState smpl;
+};
+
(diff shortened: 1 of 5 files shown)


HansKristian-ARM:
Using a struct to represent combined image samplers seems like the cleanest approach, but I'm a bit worried that this will not work in all cases.

Can you use `: register(t#)` and ` : register(s#)` for the texture and sampler in those structs?

How will combining the image and sampler work? Can you build a structure safely from a standalone Texture2D and SamplerState and use that to sample an image?

RobDangerous:
No luck with register declarations according to https://msdn.microsoft.com/en-us/library/windows/desktop/dn899207(v=vs.85).aspx
Have to flatten it. Should I wait for your other pull request?

HansKristian-ARM:
I will continue on that PR soon-ish. It is also needed to sort out some corner cases with shadow samplers as well, so I suggest we go with that approach instead.

RobDangerous:
Okidoki!

HansKristian-ARM:
Unfortunate about the duplicated work, but oh well.

PR #178, opened 2017-05-06
Implement more sophisticated check for point_size.

HansKristian-ARM:
This enables the point_size builtin based on the existing active_input/output_builtins variable.
Renames is_rendering_points to enable_point_size_builtin to be able to forcefully disable point_size when not using Point topology as before, but make the intention clearer.

### DIFF
--- a/spirv_msl.cpp
+++ b/spirv_msl.cpp
@@ -1380,7 +1380,14 @@ string CompilerMSL::member_attribute_qualifier(const SPIRType &type, uint32_t in
 				return " /* [[clip_distance]] built-in not yet supported under Metal. */";
 
 			case BuiltInPointSize: // Must output only if really rendering points
-				return options.is_rendering_points ? (string(" [[") + builtin_qualifier(builtin) + "]]") : "";
+				// SPIR-V might declare PointSize as a builtin even though it's not really used.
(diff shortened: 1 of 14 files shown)


billhollings:
Thanks...I'll do a review of this later this week.

spirv_msl.cpp
```diff
@@ -1380,7 +1380,14 @@ string CompilerMSL::member_attribute_qualifier(const SPIRType &type, uint32_t in
 				return " /* [[clip_distance]] built-in not yet supported under Metal. */";
 
 			case BuiltInPointSize: // Must output only if really rendering points
-				return options.is_rendering_points ? (string(" [[") + builtin_qualifier(builtin) + "]]") : "";
+				// SPIR-V might declare PointSize as a builtin even though it's not really used.
+				// In some cases PointSize builtin may be written without Point topology.
+				// This is not an issue on GL/Vulkan, but it is on Metal, so we also have some way to forcefully disable
+				// this builtin.
+				if (active_output_builtins & (1ull << BuiltInPointSize))
+					return options.enable_point_size_builtin ? (string(" [[") + builtin_qualifier(builtin) + "]]") : "";
```
billhollings: Maybe combine `if` and ternary tests into single ternary test.

Issue #171, opened 2017-05-03
[msl] bit casting


rextimmy:
The glsl bit casting functions like floatBitsToInt (and others like intBitsToFloat) are not getting translated with msl and therefore survive the the round trip. I believe the msl equivalent is as_type<type-id>()

billhollings:
I'm starting to look into this now.

@rextimmy Can you provide a SMALL sample GLSL file that we can add as an MSL test case for this please?

rextimmy:
@billhollings  Yeah no worries
[bitcasting.frag.txt](https://github.com/KhronosGroup/SPIRV-Cross/files/980547/bitcasting.frag.txt)

billhollings:
Fixed in PR #186.

PR #186, opened 2017-05-19
CompilerMSL map many GLSL functions to MSL functions

billhollings:
This started as a fix for #171, but extended to map a large number of GLSL functions to their Metal counterparts. 

See the commit note for details about the added functionality.

### DIFF
--- a/spirv_common.hpp
+++ b/spirv_common.hpp
@@ -276,6 +276,8 @@ struct SPIRType : IVariant
 		bool depth;
 		bool arrayed;
 		bool ms;
+		bool is_read = false;
+		bool is_written = false;
 		uint32_t sampled;
 		spv::ImageFormat format;
 	} image;
--- a/spirv_glsl.cpp
+++ b/spirv_glsl.cpp
@@ -2802,7 +2802,7 @@ void CompilerGLSL::emit_texture_op(const Instruction &i)
 	auto &imgtype = get<SPIRType>(type.self);
 
 	// Mark that this shader reads from this image
-	((SPIRType &)imgtype).image.is_read = true;
+	imgtype.image.is_read = true;
 
 	uint32_t coord_components = 0;
 	switch (imgtype.image.dim)
--- a/spirv_glsl.cpp
+++ b/spirv_glsl.cpp
@@ -3483,8 +3483,15 @@ string CompilerGLSL::access_chain_internal(uint32_t base, const uint32_t *indice
 			}
 			else
 			{
-				expr += ".";
-				expr += to_member_name(*type, index);
+				// If the member has a qualified name, use it as the entire chain
+				string qual_mbr_name = get_member_qualified_name(type->self, index);
+				if (!qual_mbr_name.empty())
+					expr = qual_mbr_name;
+				else
+				{
+					expr += ".";
+					expr += to_member_name(*type, index);
+				}
 			}
 
 			if (member_is_packed_type(*type, index))
--- a/spirv_glsl.cpp
+++ b/spirv_glsl.cpp
@@ -2769,6 +2769,7 @@ void CompilerGLSL::emit_texture_op(const Instruction &i)
 		break;
 
 	case OpImageFetch:
+	case OpImageRead: // Reads == fetches in Metal (other langs will not get here)
 		opt = &ops[4];
 		length -= 4;
 		fetch = true;
@@ -2787,7 +2788,13 @@ void CompilerGLSL::emit_texture_op(const Instruction &i)
 		break;
 	}
 
-	auto &imgtype = expression_type(img);
+	// Bypass pointers because we need the real image struct
+	auto &type = expression_type(img);
+	auto &imgtype = get<SPIRType>(type.self);
+
+	// Mark that this shader reads from this image
+	((SPIRType &)imgtype).image.is_read = true;
+
 	uint32_t coord_components = 0;
 	switch (imgtype.image.dim)
 	{
--- a/spirv_msl.cpp
+++ b/spirv_msl.cpp
@@ -87,7 +87,6 @@ string CompilerMSL::compile()
 	update_active_builtins();
 
 	// Preprocess OpCodes to extract the need to output additional header content
-	set_enabled_interface_variables(get_active_interface_variables());
 	preprocess_op_codes();
 
 	// Create structs to hold input, output and uniform variables
@@ -166,13 +165,20 @@ string CompilerMSL::compile(MSLConfiguration &msl_cfg, vector<MSLVertexAttr> *p_
 // Register the need to output any custom functions.
 void CompilerMSL::preprocess_op_codes()
 {
-	custom_function_ops.clear();
+	set_enabled_interface_variables(get_active_interface_variables());
+	spv_function_implementations.clear();
 
 	OpCodePreprocessor preproc(*this);
 	traverse_all_reachable_opcodes(get<SPIRFunction>(entry_point), preproc);
 
 	if (preproc.suppress_missing_prototypes)
-		add_header_line("#pragma clang diagnostic ignored \"-Wmissing-prototypes\"");
+		add_pragma_line("#pragma clang diagnostic ignored \"-Wmissing-prototypes\"");
+
+	if (preproc.uses_atomics)
+	{
+		add_header_line("#include <metal_atomic>");
+		add_pragma_line("#pragma clang diagnostic ignored \"-Wunused-variable\"");
+	}
 }
 
 // Move the Private global variables to the entry function.
@@ -687,7 +693,7 @@ bool CompilerMSL::is_member_packable(SPIRType &ib_type, uint32_t index)
 	return false;
 }
 
-// Simple combination of type ID and member index for use as hash key
+// Returns a combination of type ID and member index for use as hash key
 MSLStructMemberKey CompilerMSL::get_struct_member_key(uint32_t type_id, uint32_t index)
 {
 	MSLStructMemberKey k = type_id;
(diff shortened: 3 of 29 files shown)


HansKristian-ARM:
Thanks. I'll have a look when I get home from my trip.

spirv_common.hpp
```diff
@@ -296,6 +296,8 @@ struct SPIRType : IVariant
 		bool depth;
 		bool arrayed;
 		bool ms;
+		bool is_read = false;
```
HansKristian-ARM: What is this used for?

spirv_glsl.cpp
```diff
@@ -2796,7 +2797,13 @@ void CompilerGLSL::emit_texture_op(const Instruction &i)
 		break;
 	}
 
-	auto &imgtype = expression_type(img);
+	// Bypass pointers because we need the real image struct
+	auto &type = expression_type(img);
+	auto &imgtype = get<SPIRType>(type.self);
+
+	// Mark that this shader reads from this image
+	((SPIRType &)imgtype).image.is_read = true;
```
HansKristian-ARM: This looks a bit off, and the casting here is pretty ugly. What is this trying to accomplish?

spirv_msl.cpp
```diff
@@ -854,65 +1047,370 @@ void CompilerMSL::emit_instruction(const Instruction &instruction)
...
+		string img_exp = to_expression(img_id);                                             \
+		string expr = type_to_glsl(rslt_type) + "(" + img_exp + ".get_num_" #qrytype "())"; \
+		emit_op(rslt_type_id, id, expr, should_forward(img_id));                            \
+		break;                                                                              \
+	}
+
+	case OpImageQueryLevels:
+		ImgQry(mip_levels)
+
+		    case OpImageQuerySamples : ImgQry(samples)
```
HansKristian-ARM: The formatting here is off.

spirv_msl.cpp
```diff
@@ -854,65 +1047,370 @@ void CompilerMSL::emit_instruction(const Instruction &instruction)
...
+			uint32_t id = ops[1];
+			uint32_t arg = ops[2];
+
+			string exp;
+			auto &type = get<SPIRType>(result_type);
+
+			switch (type.vecsize)
+			{
+			case 1:
+				exp = join("float(half(", to_expression(arg), "))");
```
HansKristian-ARM: Is using half in Metal guaranteed to quantize?

spirv_msl.cpp
```diff
@@ -854,65 +1047,370 @@ void CompilerMSL::emit_instruction(const Instruction &instruction)
...
+
+	// OpOuterProduct
+
 	default:
 		CompilerGLSL::emit_instruction(instruction);
 		break;
 	}
 }
 
+void emit_atomic_func_op(uint32_t result_type, uint32_t result_id, const char *op, uint32_t mem_order_1,
```
HansKristian-ARM: Why this prototype?

spirv_msl.cpp
HansKristian-ARM: You'll probably want to wrap this block in a do/while here so you can do `ImgQry(samples);` with a semi-colon to not confuse clang-format.

HansKristian-ARM:
I sorted out the build failure. Recent spirv-tools uses SPIRV 1.2 which failed the internal check in SPIRV-Cross. If you rebase against current master it should pass.

spirv_glsl.cpp
billhollings: Odd. I was getting a const cast error earlier...so that cast was to remove the const.

It is not happening now...so I've removed the messy cast.

spirv_common.hpp
billhollings: `Image` `is_read` and `is_written` track whether the image is read or written to from within the shader. Under Metal, texture types need to be declared as writable in order to be written to. The `is_read` allows the converter to distinguish whether a writable texture should be declared with `write` or `read_write`...in case there is any efficiency to be gained by the Metal compiler by distinguishing between those two states.

spirv_msl.cpp
billhollings: I'm not sure what you are asking.

spirv_msl.cpp
billhollings: Thanks...good catch on D/W.

Done.

spirv_msl.cpp
billhollings: Just hanging around from earlier editing. Removed now.

spirv_common.hpp
HansKristian-ARM: I think you can use the appropriate decorations for this. GLSL has the same deal with image load/store and readonly/writeonly. See DecorationNonWriteable and DecorationNonReadable. I'd like to avoid adding a new way to deal with this.

test_shaders.py
```diff
@@ -75,7 +75,8 @@ def print_msl_compiler_version():
 def validate_shader_msl(shader):
     msl_path = reference_path(shader[0], shader[1])
     try:
-        subprocess.check_call(['xcrun', '--sdk', 'iphoneos', 'metal', '-x', 'metal', '-std=ios-metal1.2', '-Werror', msl_path])
+        subprocess.check_call(['xcrun', '--sdk', 'macosx', 'metal', '-x', 'metal', '-std=osx-metal1.2', '-Werror', msl_path])
+#        subprocess.check_call(['xcrun', '--sdk', 'iphoneos', 'metal', '-x', 'metal', '-std=ios-metal1.2', '-Werror', msl_path])
```
HansKristian-ARM: Any reason to change this? Atomics support?

spirv_msl.cpp
```diff
@@ -1920,11 +2486,29 @@ string CompilerMSL::image_type_glsl(const SPIRType &type)
 
 	// Append the pixel type
 	auto &img_pix_type = get<SPIRType>(img_type.type);
-	img_type_name += "<" + type_to_glsl(img_pix_type) + ">";
+	img_type_name += "<";
+	img_type_name += type_to_glsl(img_pix_type);
+
+	if (img_type.is_written)
```
HansKristian-ARM: Seems a bit odd for read/write access to be tied to the type?

spirv_msl.cpp
```diff
@@ -854,65 +1045,377 @@ void CompilerMSL::emit_instruction(const Instruction &instruction)
...
+		uint32_t id = ops[1];
+		uint32_t arg = ops[2];
+
+		string exp;
+		auto &type = get<SPIRType>(result_type);
+
+		switch (type.vecsize)
+		{
+		case 1:
+			exp = join("float(half(", to_expression(arg), "))");
```
HansKristian-ARM: I assume that casting something to half really turns it into a true FP16 value? Could the compiler optimize this away?

spirv_common.hpp
HansKristian-ARM: Also, is_read/is_written does not belong in the type, multiple images using the same type will get very confusing results.

billhollings: @HansKristian-ARM 

Agreed.

The challenge is that in MSL R/W access is declared as part of the texture type.

See  [SPIRV-Cross/reference/shaders-msl/desktop-only/frag/image-ms.desktop.frag](https://github.com/brenwill/SPIRV-Cross/blob/c1b8154f2ca1c82579cea53a646e37ef79bbea3d/reference/shaders-msl/desktop-only/frag/image-ms.desktop.frag).

Generation of this is currently buried in `image_type_glsl()` which is called from `type_to_glsl()`.

I'm not sure how to handle this without making it too disruptive everywhere.

Perhaps we could pass an optional `var_id` param to those two functions where they could look up decorations if it is non-zero?

Thoughts or suggestions?

test_shaders.py
billhollings: Texture write access is not supported on iOS.

spirv_msl.cpp
billhollings: Yes. See response and question above.

spirv_msl.cpp
billhollings: Yes. Metal's `half` is a 16-bit value.

I assume the compiler will honour it...and not look ahead that far.

However...I do not have test cases for these yet. Let's release it to the wild and modify as necessary base on user testing.

HansKristian-ARM: Ok, good enough to me. QuantizeToF16 is not really used by real shaders.

test_shaders.py
HansKristian-ARM: Really? o_O Oh well.

spirv_common.hpp
HansKristian-ARM: I have a similar situation with samplers in GLSL and HLSL, where I have the distinction between samplerShadow and just plain sampler, but it does not exist in the type system. In that case, I do some hackery to pretend it's part of the type system. In variable_decl I know the variable ID, and I can based on flags modify the type in-place, declare it, and then change it back once the type_to_glsl() has been called. This however, is a bit ugly. The main difference is that at least the shadow information does not leak into other variables.

I think we should try to optionally pass down the variable ID to these functions going forward so that we can deal with this kind of information without hacks and tricks.

Would it be okay for now that all textures are declared as read/write, the is_read/is_write is removed, and then I can have a look at how to refactor the top-level interfaces to deal with this in GLSL/MSL/HLSL?

billhollings: No...I don't want to make all textures writable.

Marking a texture as writable will in most cases cause Metal to optimize it differently, likely negatively relative to a read-only texture.

The MSL produced by CompilerMSL is used in production shaders, where performance is a concern. Writable textures are very rare, and making all textures writable will likely introduce a performance overhead on all basic sampled textures. I would like Metal to be free to optimize for those vast majority read-only cases.

For now...can we leave the mechanism as it is in this PR...and then move quickly to deploy a refactoring of `type_to_glsl()`?

HansKristian-ARM: Hm, so to be clear, does this read/write difference only concern image2D (image load/store), or sampler2D as well? Making all sampler2D writeable would certainly be a really bad idea, but I don't see the big problem if it's image2D only, but there might be issues in Metal I don't know about. For sampler2D, this is known by the type system (sampled == 2), so sampler2D can always be forced to read-only.

I'm fine with keeping things as it is, but please add a comment in the SPIRType structure that this is a temporary workaround for Metal.

billhollings: In MSL, textures and samplers are separate entities.textures can be declared with an `access`:

    enum class access { sample, read, write, read_write };

Default is `sample` for all but multisampled textures. Declaring a texture as `read_write` therefore, removes the `sample` declaration, which presumably does affect any performance optimizations around sampling.

billhollings: I'll add the internal comment you suggest.

billhollings:
By "refactoring the image read/writes"...do you mean looking at passing an optional `var_id` information into `type_to_glsl()`?

That sounds fairly straightforward...and I'm happy to take care of that.If you're thinking of a

HansKristian-ARM:
Refactoring the type_to_glsl() system to also take an optional var_id so backends can deal with per-variable extra type information yes. The default argument can be 0, which would mean "no particular variable".

billhollings:
Okay. I'll take care of implementing that and submit another PR.

And I'll take care of reverting the existing `is_read/is_written` image type flags as part of that.

HansKristian-ARM:
This also means fixing the current GLSL/MSL/HLSL code. We can split up the MSL and GLSL/HLSL if you'd prefer that.

billhollings:
Ah...right. Yes...probably best to split it up...as I don't currently have a strong-enough understanding of the issues with GLSL/HLSL.

What's the priority for the GLSL/HLSL issue? Shall I put together the MSL part...and just leave the GLSL/HLSL use of the `var_id` unused...and then you can go in and make use of it? Or do you feel it is important to get the GLSL/HLSL work done first?

HansKristian-ARM:
Fixing the MSL is higher priority than GLSL/HLSL. I'll make sure the solution is adequate for GLSL/HLSL in the review. GLSL/HLSL will just ignore the var_id for now.

billhollings:
Okay. Sounds good.

billhollings:
@HansKristian-ARM 

Looking into this now.

I'm trying to figure out how to mark the appropriate image object with the readable/writable decorations.

Presumably, an `OpImageWrite` might access an image through several mechanisms, including `OpLoad`, `OpAccessChain`, etc.

Will calling `maybe_get_backing_variable` from the `OpImageWrite` tunnel through all those cases to get to the underlying image to be marked?

In Metal, an input to a function can be a struct that contains textures and buffers. In that case, it would be the struct member (type) that we would have to mark with the read-write I guess.

But is that the case for GLSL? Or will all GLSL-originated images ultimately come from an `OpVariable` and be accessible through `maybe_get_backing_variable` ?

HansKristian-ARM:
Images cannot be inside structs, so any image will always be directly related to a variable.
maybe_get_backing_variable will find the originating variable ID.

billhollings:
@HansKristian-ARM 

In the SPIR-V spec...I have just stumbled across the `OpTypeImage` [Access Qualifier](https://www.khronos.org/registry/spir-v/specs/1.0/SPIRV.html#Access_Qualifier) parameter.

It appears to be performing exactly the same functionality that Metal requires...AND it is attached to the type, not the variable.

It would be interesting to understand the purpose of the access parameters being applied to types instead of variables in SPIR-V (and Metal). Perhaps the intention is that SPIR-V will include different `OpTypeImage` entries for each texture that has different access (as it does for all other image parameters).

The SPIR-V generated by `glslang` for `shaders-msl/desktop-only/frag/image-ms.desktop.frag`, which includes images that require read, write, and read-write access, does NOT include this optional parameter...nor does `glslang` generate different `OpTypeImage` entries, or include access qualifiers if I add two images that differ only in one being read and the other being written.

I wonder if this is just an issue with `glslang`? Perhaps it SHOULD be generating different `OpTypeImage` entries with differing access qualifiers in that case?

HansKristian-ARM:
So this access qualifier is only for "kernel", which is OpenCL. We cannot use it. I'm not sure why there's a difference, but OpenCL might have some other constraints I don't know about.

Access qualfiers for shaders are done through variable decorations NonWriteable and NonReadable. Note that this only matters for image load/store, which are images with sampled parameter == 2. For sampled images, the image is obviously always just read-only.

billhollings:
Thanks for the clarification.

billhollings:
While I am implementing this then...I think I will add support for BOTH variable NW & NR decorations AND kernel access qualifiers.

The decorations I will handle as we've described. And the kernel access qualifiers I can handle by replacing the `is_read` & `is_written` flags in `SPIRVType::Image` with a single `access_qualifer` enum value. It would be populated directly from `OpTypeImage`, and would influence Metal's access qualifiers in the same way the `is_read` and `is_written` flags currently do.

So...Metal would have its variable declarations set through a combination of the NW & NR decorations...AND the `OpTypeImage` `access_qualifer`.

Do you have any objections to that idea?

HansKristian-ARM:
Well, I suppose there is no harm in adding support for the access_qualifier, but I don't really see the value either. Valid SPIR-V code for shaders cannot use those qualifiers anyways.

PR #190, opened 2017-05-30
Emit type declarations tuned for specified SPIR-V objects + MSL builtin enhancements.

billhollings:
Per extended discussion in issue #186, add support for emitting SPIR-V type declarations tuned for specified SPIR-V objects, as well as SPIR-V `OpTypeImage` access qualifiers. CompilerMSL uses this to output appropriate MSL image access qualifiers.

Also includes several MSL builtin enhancements, including eliding unused builtin inputs and outputs, and support for gl_ClipDistance in Metal.

### DIFF
--- a/spirv_glsl.cpp
+++ b/spirv_glsl.cpp
@@ -2893,11 +2893,6 @@ void CompilerGLSL::emit_texture_op(const Instruction &i)
 	auto &type = expression_type(img);
 	auto &imgtype = get<SPIRType>(type.self);
 
-	// Mark that this shader reads from this image
-	auto *p_var = maybe_get_backing_variable(img);
-	if (p_var)
-		unset_decoration(p_var->self, DecorationNonReadable);
-
 	uint32_t coord_components = 0;
 	switch (imgtype.image.dim)
 	{
--- a/spirv_glsl.cpp
+++ b/spirv_glsl.cpp
@@ -2802,7 +2802,9 @@ void CompilerGLSL::emit_texture_op(const Instruction &i)
 	auto &imgtype = get<SPIRType>(type.self);
 
 	// Mark that this shader reads from this image
-	imgtype.image.is_read = true;
+	auto *p_var = maybe_get_backing_variable(img);
+	if (p_var)
+		unset_decoration(p_var->self, DecorationNonReadable);
 
 	uint32_t coord_components = 0;
 	switch (imgtype.image.dim)
@@ -5944,7 +5946,7 @@ string CompilerGLSL::type_to_array_glsl(const SPIRType &type)
 	return res;
 }
 
-string CompilerGLSL::image_type_glsl(const SPIRType &type)
+string CompilerGLSL::image_type_glsl(const SPIRType &type, uint32_t /* id */)
 {
 	auto &imagetype = get<SPIRType>(type.image.type);
 	string res;
@@ -6029,7 +6031,10 @@ string CompilerGLSL::type_to_glsl_constructor(const SPIRType &type)
 	return e;
 }
 
-string CompilerGLSL::type_to_glsl(const SPIRType &type)
+// The optional id parameter indicates the object whose type we are trying
+// to find the description for. It is optional. Most type descriptions do not
+// depend on a specific object's use of that type.
+string CompilerGLSL::type_to_glsl(const SPIRType &type, uint32_t id)
 {
 	// Ignore the pointer type since GLSL doesn't have pointers.
 
@@ -6044,7 +6049,7 @@ string CompilerGLSL::type_to_glsl(const SPIRType &type)
 
 	case SPIRType::Image:
 	case SPIRType::SampledImage:
-		return image_type_glsl(type);
+		return image_type_glsl(type, id);
 
 	case SPIRType::Sampler:
 		// This is a hacky workaround. The sampler type in SPIR-V doesn't actually signal this distinction,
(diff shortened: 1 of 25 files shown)


HansKristian-ARM:
Thanks. I will review tomorrow.

spirv_glsl.cpp
```diff
@@ -2802,7 +2802,9 @@ void CompilerGLSL::emit_texture_op(const Instruction &i)
 	auto &imgtype = get<SPIRType>(type.self);
 
 	// Mark that this shader reads from this image
-	imgtype.image.is_read = true;
+	auto *p_var = maybe_get_backing_variable(img);
+	if (p_var)
+		unset_decoration(p_var->self, DecorationNonReadable);
```
HansKristian-ARM: I don't understand this. DecorationNonReadable is only used for image load/store textures, not normal sampled textures (they are always read-only in SPIR-V no matter what). Also, changing decorations like this is unsafe while emitting code.

If you want to modify the flags of an image load/store texture, please look at OpImageRead/OpImageWrite implementation:

https://github.com/KhronosGroup/SPIRV-Cross/blob/master/spirv_glsl.cpp#L5387
https://github.com/KhronosGroup/SPIRV-Cross/blob/master/spirv_glsl.cpp#L5520

spirv_msl.cpp
```diff
@@ -2491,18 +2484,43 @@ string CompilerMSL::image_type_glsl(const SPIRType &type)
...
-	auto &img_pix_type = get<SPIRType>(img_type.type);
 	img_type_name += "<";
-	img_type_name += type_to_glsl(img_pix_type);
+	img_type_name += type_to_glsl(get<SPIRType>(img_type.type));
 
-	if (img_type.is_written)
+	// For unsampled images, append the sample/read/write access qualifier.
+	// For kernel images, the access qualifier my be supplied directly by SPIR-V.
+	// Otherwise it may be set based on whether the image is read from or written to within the shader.
+	if (type.basetype == SPIRType::Image)
```
HansKristian-ARM: Here you will need to check for `type.image.sampled == 2` as well. Image type can mean both sampled and unsampled.

spirv_glsl.cpp
HansKristian-ARM: Ok, I think I see why ... This is used because OpImageRead is implemented through a normal fetch in MSL.

It is a bit confusing to have this code here, could you just move it to OpImageRead case statement in spirv_msl.cpp instead?

HansKristian-ARM: OpImageFetch is for sampled images (sampler2D), this maps to texelFetch in GLSL.
OpImageRead maps to imageLoad() (on image2D).

OpImageFetch is a bit special since it doesn't use samplers, but it only works on sampled images (so decorations do not matter). The typical sequence to for texelFetch() is
```
OpTypeSampledImage %some_sampled_image
%image_without_sampler = OpImage %some_sampled_image
%result = OpImageFetch %image_without_sampler
```

The main reason for this is that you have samplerBuffers in GLSL which are sampled images, but they don't have a sampler.

spirv_msl.cpp
billhollings: Why are we not considering kernel mode?

Vulkan and SPIR-V support compute shaders, which I presume might get compiled from something like OpenCL. I understand that it might not be a strict requirement for regenerating GLSL...but I need to make sure `spirv_cross` can generate MSL can be used for Vulkan compute requirements in MoltenVK.

HansKristian-ARM: There are two execution models for compute shaders, Kernel and Shader.
Compute shaders in Vulkan still use the "Shader" model. See

```
    ExecutionModelGLCompute = 5, // Vulkan compute
    ExecutionModelKernel = 6, // CL compute
```
Kernel mode is only supported by OpenCL runtime, so SPIR-V created from OpenCL compilers are not compatible with Vulkan. This is explicitly stated in the Vulkan specification.

billhollings: What it means here is that I'm not sure I want to ignore `OpTypeImage` _Sampled_ == 0.

HansKristian-ARM: I would rather add an assert in OpTypeImage if sampled == 0, it should never happen in valid code.

billhollings: I've added the test for `type.image.sampled == 2`...and the `SPIRV_CROSS_THROW` in `OpTypeImage` if `type.image.sampled == 0`.

spirv_glsl.cpp
billhollings: This code test is now moved to `OpImageRead` handling in `CompilerMSL`.

spirv_msl.cpp
```diff
@@ -1176,10 +1179,18 @@ void CompilerMSL::emit_instruction(const Instruction &instruction)
...
 
-	// Reads == fetches in Metal
+	// Reads == Fetches in Metal
 	case OpImageRead:
+	{
+		// Mark that this shader reads from this image
+		uint32_t img_id = ops[2];
+		auto *p_var = maybe_get_backing_variable(img_id);
+		if (p_var)
+			unset_decoration(p_var->self, DecorationNonReadable);
```
HansKristian-ARM: Make sure to do the same check as OpImageWrite with force_recompile.

billhollings: Thanks...but INABIAF. 😏

I deliberately left the `force_recompile` out of `OpImageRead` because in Metal the default (non-attributed) textures are set to `access::sample` which allows both sampling and reading.

In this context, the `force_recompile` is used to enable the output of the access qualifiers... which only needs to be performed if some combination involving write access is required.

HansKristian-ARM: Fine.

PR #191, opened 2017-05-30
Fix null pointer dereference

hesiod:
write_string_to_file would try to write the error message that it failed to open a file stream to the very same file stream it just tried to open.
Fix this by writing to stderr instead.


HansKristian-ARM:
Derp. Thanks.

PR #194, opened 2017-05-31
Fix shader model 2/3 texture sampling

RobDangerous:
This just broke. There are no tests for shader model 2/3 and I don't think we can easily test that on Travis. We could setup an appveyor script.

### DIFF
--- a/spirv_hlsl.cpp
+++ b/spirv_hlsl.cpp
@@ -1398,7 +1398,13 @@ void CompilerHLSL::emit_texture_op(const Instruction &i)
 
 	expr += texop;
 	expr += "(";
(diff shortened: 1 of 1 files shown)


HansKristian-ARM:
I think we can add legacy HLSL output in the test script, like we do for GLSL, Travis should be fine assuming glslangValidator can understand SM 2/3 code.

RobDangerous:
Hm, https://github.com/KhronosGroup/glslang/issues/362 looks like there is some 2/3 support but samplers are explicitly marked as missing.

HansKristian-ARM:
If you know how to set up appveyor, I'd be happy to merge that.

RobDangerous:
Can do, will do.

PR #200, opened 2017-06-17
Geometry shader core since GLSL 150

h3xl3r:
According to https://www.khronos.org/opengl/wiki/History_of_OpenGL#OpenGL_3.2_.282009.29

I am getting errors on OS X when output GLSL version is set to 150 and the unnecessary 
`#extension GL_ARB_geometry_shader4 : require`
string is still generated


HansKristian-ARM:
Mixed up ESSL versioning and GLSL :P

PR #211, opened 2017-06-26
Always make a copy when handling OpCompositeInsert

dsrbecky:
The modified object might not be mutable (e.g. shader input).

Added a test for the case when this happens.


HansKristian-ARM:
Upon further inspection, the old code is indeed incorrect in some other corner cases.

Fortunately, this workaround is clearly not needed anymore since fixing this did not change any other code, so I'll merge this and clean up the remnants of this workaround.

dsrbecky:
Thank you

PR #213, opened 2017-06-27
Ignore debug instruction OpString

dsrbecky:
It may be used with OpLine or OpSource to specify filename.


HansKristian-ARM:
Thanks

PR #217, opened 2017-06-28
WIP - First incomplete attempt at supporting DecorationRelaxedPrecision in MSL.

billhollings:
This is a first attempt at adding support for the `DecorationRelaxedPrecision` in MSL. The purpose is to allow GLSL `medump` declarations to appear as 16-bit precision `short`, `ushort`, and `half` declarations in MSL.

As currently designed...this is turning out to be quite an intrusive enhancement...because there are many, many areas where `type_as_glsl()` is invoked...including textures, assignments, math ops, etc.

It is currently broken on `shaders-msl/frag/constant-array.frag` because the shader contains a number of constant declarations such as `vec4` and `vec4[]` that are NOT decorated with `DecorationRelaxedPrecision` and it then breaks the MSL compiler when it tries to use those constants to assign, or operate on (e.g.- add, subtract) variables that ARE decorated with `DecorationRelaxedPrecision`.

To illustrate...simply run `./test_shaders.py --msl shaders-msl`.

I'm throwing this up here to solicit suggestions for one of...

- Ensuring we can deal with all declarations (variables, constants, spec constants, inputs, outputs, uniforms, texture) correctly and consistently.
- Throwing this approach away in favour of a more elegant and less intrusive solution that is immediately available.
- Postponing this update, in favour of a later refactoring, that might make a different solution much more elegant and feasible.

Regarding the latter two options...I have been thinking of the following possible alternate design...

Replace the `SPIRType` in any object that has relaxed precision with an invented `SPIRType` that looks like the original...but has it's width cut to 16 bits. Then `type_as_glsl()` would just work.

This seems clean and elegant...except for `SPIRType` inheritance and decorations...which might get in the way...hence my thinking of possible refactoring...in this case decoration support or how `SPIRType` inheritance works...which can sometimes get in the way of other things too...like pointers and arrays.

The lack of decoration on the constants described above would still be a problem with this design too.

### DIFF
--- a/reference/shaders-msl/flatten/swizzle.flatten.vert
+++ b/reference/shaders-msl/flatten/swizzle.flatten.vert
@@ -23,12 +23,12 @@ struct UBO
 
 struct main0_out
 {
-    float4 oF [[user(locn5)]];
-    float4 oE [[user(locn4)]];
-    float4 oD [[user(locn3)]];
-    float4 oC [[user(locn2)]];
-    float4 oB [[user(locn1)]];
     float4 oA [[user(locn0)]];
+    float4 oB [[user(locn1)]];
+    float4 oC [[user(locn2)]];
+    float4 oD [[user(locn3)]];
+    float4 oE [[user(locn4)]];
+    float4 oF [[user(locn5)]];
     float4 gl_Position [[position]];
 };
 
--- a/reference/shaders-msl/flatten/types.flatten.frag
+++ b/reference/shaders-msl/flatten/types.flatten.frag
@@ -5,31 +5,31 @@ using namespace metal;
 
 struct UBO1
 {
-    int4 c;
-    int4 d;
+    short4 c;
+    short4 d;
 };
 
 struct UBO2
 {
-    uint4 e;
-    uint4 f;
+    ushort4 e;
+    ushort4 f;
 };
 
 struct UBO0
 {
-    float4 a;
-    float4 b;
+    half4 a;
+    half4 b;
 };
 
 struct main0_out
 {
-    float4 FragColor [[color(0)]];
+    half4 FragColor [[color(0)]];
 };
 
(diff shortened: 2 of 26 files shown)


billhollings:
The CI-Travis failure is expected. It's failing for the reason described above.

HansKristian-ARM:
Thanks, I will have a look tomorrow.

HansKristian-ARM:
Hm ... Alright, so if I understand this correctly:

```
float v = 2.0;
half h = v;
```
is not legal MSL because you cannot get implicit conversion between float and half? That sounds almost like a deal breaker, because the main idea behind RelaxedPrecision is that you get implicit conversions, but the type remains the same. If MSL is very strict about this conversion, things will certainly get quite messy.

billhollings:
The simple case of scalars is not an issue because the constants get compiled by glslang into literals. It's the vectors and arrays that cause problems.

Your custom struct idea is interesting. I have a bunch of custom functions already in `CompilerMSL` that are spit out when needed...so we could follow the same pattern of spitting out custom struct definitions.

How would we handle when it is on the RHS of a math operator?

    float + myhalf

HansKristian-ARM:
Perhaps something like overriding operator+, -, *, / and so on, but yes, I agree this is going to rather extreme lengths to support an optional optimization feature.

I am fine with deferring this until there is a real use case which is proven to show a real uplift by going to half. The case we are dealing with here is related to ESSL fragment shaders indeed.

billhollings:
Overriding operators might work for the LHS math...

    myhalf + float

...but I don't see how it would work for RHS math...

    float + myhalf

...unless we also created an enhanced `float` class as well.

HansKristian-ARM:
You could make a global operator overload perhaps?

```
float operator+(float, half)
```

HansKristian-ARM:
I'll assume this is abandoned.

PR #220, opened 2017-07-07
Basic support for HLSL compute shaders with StructuredBuffers

austinEng:
This adds HLSL support for StructuredBuffers / Unordered Access Views in order to get basic compute shaders working.

StructuredBuffer<T> and RWStructuredBuffer<T> must be accessed as arrays so in HLSL it is impossible to do something like:
```
struct myBuffer {
  int a;
};
RWStructuredBuffer<myBuffer> buffer : register(u0);

void main() {
  buffer.a = 2;
}
```

an array access is required:
`buffer[0].a = 2;`

So essentially all structured buffers are unsized arrays.

To account for this, I've added `HLSLCompiler::to_name` to replace struct-usages of structured buffers with an array-usage that accesses element 0. Would this be the best place to do this (modifying the HLSL interpretation of the SPIRV) or would it be better to modify the SPIRV to always treat these as arrays?

### DIFF
--- a/reference/shaders-hlsl/comp/global-invocation-id-writable-ssbo-in-function.comp
+++ b/reference/shaders-hlsl/comp/global-invocation-id-writable-ssbo-in-function.comp
@@ -1,0 +1,36 @@
+struct myBlock
+{
+    int a;
+    float b[1];
+};
+
+RWStructuredBuffer<myBlock> myStorage : register(u0);
+
+static uint3 gl_GlobalInvocationID;
+struct SPIRV_Cross_Input
+{
+    uint3 gl_GlobalInvocationID : SV_DispatchThreadID;
+};
+
+float mod(float x, float y)
+{
+    return x - y * floor(x / y);
+}
+
+float getB()
+{
+    return myStorage[0].b[gl_GlobalInvocationID.x];
+}
+
+void comp_main()
+{
+    myStorage[0].a = (myStorage[0].a + 1) % 256;
+    myStorage[0].b[gl_GlobalInvocationID.x] = mod(getB() + 0.0199999995529651641845703125f, 1.0f);
+}
+
+[numthreads(1, 1, 1)]
+void main(SPIRV_Cross_Input stage_input)
+{
+    gl_GlobalInvocationID = stage_input.gl_GlobalInvocationID;
+    comp_main();
+}
--- a/reference/shaders-hlsl/comp/ssbo-array.comp
+++ b/reference/shaders-hlsl/comp/ssbo-array.comp
@@ -1,0 +1,25 @@
+struct SSBO
+{
+    float4 data[1];
+};
+
+RWStructuredBuffer<SSBO> ssbos : register(u0);
+
+static uint3 gl_GlobalInvocationID;
+struct SPIRV_Cross_Input
+{
+    uint3 gl_GlobalInvocationID : SV_DispatchThreadID;
+};
+
+void comp_main()
+{
+    uint ident = gl_GlobalInvocationID.x;
+    ssbos[1].data[ident] = ssbos[0].data[ident];
+}
+
+[numthreads(1, 1, 1)]
+void main(SPIRV_Cross_Input stage_input)
+{
+    gl_GlobalInvocationID = stage_input.gl_GlobalInvocationID;
+    comp_main();
+}
--- a/shaders-hlsl/comp/ssbo-array.comp
+++ b/shaders-hlsl/comp/ssbo-array.comp
@@ -1,0 +1,14 @@
+#version 310 es
+layout(local_size_x = 1) in;
+
+layout(std430, binding = 0) buffer SSBO
+{
+    vec4 data[];
+} ssbos[2];
+
+void main()
+{
+    uint ident = gl_GlobalInvocationID.x;
+    ssbos[1].data[ident] = ssbos[0].data[ident];
+}
+
--- a/spirv_cross.cpp
+++ b/spirv_cross.cpp
@@ -3360,6 +3360,8 @@ bool Compiler::ActiveBuiltinHandler::handle(spv::Op opcode, const uint32_t *args
 		if (!var)
 			break;
 
+		add_if_builtin(args[2]);
+
 		auto *type = &compiler.get<SPIRType>(var->basetype);
 
 		// Start traversing type hierarchy at the proper non-pointer types.
--- a/spirv_hlsl.cpp
+++ b/spirv_hlsl.cpp
@@ -942,7 +942,14 @@ void CompilerHLSL::emit_buffer_block(const SPIRVariable &var)
 	if (is_uav)
 	{
 		// type_to_array_glsl is excluded because RWStructuredBuffer<T> must always be accessed as an array but not declared as one
-		statement("RWStructuredBuffer<", struct_name, "> ", to_name(var.self), to_resource_binding(var), ";");
+		if (has_decoration(type.self, DecorationNonWritable))
+		{
+			statement("StructuredBuffer<", struct_name, "> ", to_name(var.self), to_resource_binding(var), ";");
+		}
+		else
+		{
+			statement("RWStructuredBuffer<", struct_name, "> ", to_name(var.self), to_resource_binding(var), ";");
+		}
 	}
 	else if (options.shader_model >= 51) // SM 5.1 uses ConstantBuffer<T> instead of cbuffer.
 	{
@@ -1008,7 +1015,7 @@ string CompilerHLSL::to_name(uint32_t id, bool allow_alias) const
 		auto &type = get<SPIRType>(var->basetype);
 		if (type.basetype == SPIRType::BaseType::Struct && var->storage != StorageClassFunction && type.pointer &&
 		    type.storage == StorageClassUniform && !is_hidden_variable(*var) && type.array.size() == 0 &&
-		    (meta[type.self].decoration.decoration_flags & (1ull << DecorationBufferBlock)))
+		    has_decoration(type.self, DecorationBufferBlock))
 		{
 			name += "[0]";
 		}
--- a/spirv_hlsl.cpp
+++ b/spirv_hlsl.cpp
@@ -369,6 +369,26 @@ void CompilerHLSL::emit_builtin_inputs_in_struct()
 			semantic = "SV_SampleIndex";
 			break;
 
+		case BuiltInGlobalInvocationId:
+			type = "uint3";
+			semantic = "SV_DispatchThreadID";
+			break;
+
+		case BuiltInLocalInvocationId:
+			type = "uint3";
+			semantic = "SV_GroupThreadID";
+			break;
+
+		case BuiltInLocalInvocationIndex:
+			type = "uint";
+			semantic = "SV_GroupIndex";
+			break;
+
+		case BuiltInWorkgroupId:
+			type = "uint3";
+			semantic = "SV_GroupID";
+			break;
+
 		default:
 			SPIRV_CROSS_THROW("Unsupported builtin in HLSL.");
 			break;
@@ -568,6 +588,16 @@ void CompilerHLSL::emit_builtin_variables()
 			}
 			else
 				SPIRV_CROSS_THROW(join("Unsupported builtin in HLSL: ", unsigned(builtin)));
+
+		case BuiltInGlobalInvocationId:
+		case BuiltInLocalInvocationId:
+		case BuiltInWorkgroupId:
+			type = "uint3";
+			break;
+
+		case BuiltInLocalInvocationIndex:
+			type = "uint";
+			break;
 
 		default:
 			SPIRV_CROSS_THROW(join("Unsupported builtin in HLSL: ", unsigned(builtin)));
@@ -884,13 +914,11 @@ void CompilerHLSL::emit_buffer_block(const SPIRVariable &var)
 	auto &type = get<SPIRType>(var.basetype);
 
 	bool is_uav = has_decoration(type.self, DecorationBufferBlock);
-	if (is_uav)
-		SPIRV_CROSS_THROW("Buffer is SSBO (UAV). This is currently unsupported.");
 
 	add_resource_name(type.self);
 
 	string struct_name;
-	if (options.shader_model >= 51)
+	if (options.shader_model >= 51 || is_uav)
 		struct_name = to_name(type.self);
 	else
 		struct_name = join("_", to_name(type.self));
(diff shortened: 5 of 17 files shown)


HansKristian-ARM:
Not sure when I can look at this, but I will as soon as I have time.

Kangz:
Just to point out that CI failed because of a timeout on OSX while fetching glslang.git but I wasn't able to restart the job.

austinEng:
@HansKristian-ARM Do you have time to take a look at this? It's blocking us from supporting compute shaders on our D3D12 backend.

oscarbg:
@austinEng are you aware of https://github.com/google/clspv effort I say that because now seem very easy to translate OpenCL kernels to Spir-V Vulkan compute shaders (before that OpenCL kernels translated to SPIR-V via Khronos tool but where only usable from OpenCL SPIR-V capable implelentations only no Vulkan graphic drivers)..
I say that because you now add HLSL compute to SPIRV-cross and should be interesting to know if you & D3D compute shaders support features needed by the SPIRV code generated by that tool like:
*SPV_KHR_variable_pointers
which seems the most important..
and also:
*SPV_KHR_16bit_storage & shaderInt16
*For image access:
shaderStorageImageReadWithoutFormat
shaderStorageImageWriteWithoutFormat

For example with this tool I get for a simple kernel like:

```
void kernel __attribute__((reqd_work_group_size(1, 1, 1))) foo(sampler_t s, read_only image2d_t i, float2 c, global float4* a)
{
  *a = read_imagef(i, s, c);
}
```

SPIRV output like:

```
; SPIR-V
; Version: 1.0
; Generator: Codeplay; 0
; Bound: 34        
; Schema: 0
	OpCapability Shader
	OpCapability StorageImageReadWithoutFormat
	OpCapability VariablePointers
	OpExtension "SPV_KHR_variable_pointers"
	OpMemoryModel Logical GLSL450
	OpEntryPoint GLCompute %25 "foo"
	OpExecutionMode %25 LocalSize 1 1 1
	OpSource OpenCL_C 120
	OpMemberDecorate %7 0 Offset 0
	OpDecorate %7 BufferBlock
	OpDecorate %12 ArrayStride 16
	OpMemberDecorate %13 0 Offset 0
	OpDecorate %13 BufferBlock
	OpDecorate %21 DescriptorSet 0
	OpDecorate %21 Binding 0
	OpDecorate %22 DescriptorSet 0
	OpDecorate %22 Binding 1
	OpDecorate %22 NonWritable
	OpDecorate %23 DescriptorSet 0
	OpDecorate %23 Binding 2
	OpDecorate %24 DescriptorSet 0
	OpDecorate %24 Binding 3
%1 = 	OpTypeFloat 32
%2 = 	OpTypeSampler
%3 = 	OpTypePointer UniformConstant %2
%4 = 	OpTypeImage %1 2D 0 0 0 1 Unknown
%5 = 	OpTypePointer UniformConstant %4
%6 = 	OpTypeVector %1 2
%7 = 	OpTypeStruct %6
%8 = 	OpTypePointer StorageBuffer %7
%9 = 	OpTypePointer StorageBuffer %6
%10 = 	OpTypeVector %1 4
%11 = 	OpTypePointer StorageBuffer %10
%12 = 	OpTypeRuntimeArray %10
%13 = 	OpTypeStruct %12
%14 = 	OpTypePointer StorageBuffer %13
%15 = 	OpTypeInt 32 0
%16 = 	OpTypeVoid
%17 = 	OpTypeFunction %16
%18 = 	OpTypeSampledImage %4
%19 = 	OpConstant %1 0
%20 = 	OpConstant %15 0
%21 = 	OpVariable %3 UniformConstant
%22 = 	OpVariable %5 UniformConstant
%23 = 	OpVariable %8 StorageBuffer
%24 = 	OpVariable %14 StorageBuffer
%25 = 	OpFunction %16 None %17
%26 = 	OpLabel
%27 = 	OpLoad %2 %21
%28 = 	OpLoad %4 %22
%29 = 	OpAccessChain %9 %23 %20
%30 = 	OpLoad %6 %29
%31 = 	OpAccessChain %11 %24 %20 %20
%32 = 	OpSampledImage %18 %28 %27
%33 = 	OpImageSampleExplicitLod %10 %32 %30 Lod %19
	OpStore %31 %33
	OpReturn
	OpFunctionEnd

```

Kangz:
@oscarbg: we are not HLSL experts, just user of SPIRV-cross that try to add support for our use case :)

That said, it might be difficult to support the variable pointer extension in HLSL because HLSL StructuredBuffers are templated on the type of the element inside the buffer which means that it would require the variable pointer to statically have only one buffer type it can point to. Half floats look like they might be supported, and reading without format too. People that would know more about HLSL are John Kessenich that's writing the HLSL frontend to glslang and Chas Boyd that is leading (?) the new HLSL to DXIL compiler.

reference/shaders-hlsl/comp/global-invocation-id-writable-ssbo-in-function.comp
```diff
@@ -0,0 +1,36 @@
+struct myBlock
+{
+    int a;
+    float b[1];
```
HansKristian-ARM: I assume this is for unsized array? This approach works?

spirv_cross.cpp
```diff
@@ -3356,6 +3356,8 @@ bool Compiler::ActiveBuiltinHandler::handle(spv::Op opcode, const uint32_t *args
 		if (!var)
 			break;
 
+		add_if_builtin(args[2]);
```
HansKristian-ARM: I assume this is to support access chains of InvocationID and friends?

spirv_hlsl.cpp
```diff
@@ -907,7 +935,19 @@ void CompilerHLSL::emit_buffer_block(const SPIRVariable &var)
 	end_scope_decl();
 	statement("");
 
-	if (options.shader_model >= 51) // SM 5.1 uses ConstantBuffer<T> instead of cbuffer.
+	if (is_uav)
+	{
+		// type_to_array_glsl is excluded because RWStructuredBuffer<T> must always be accessed as an array but not declared as one
+		if (has_decoration(type.self, DecorationNonWritable))
+		{
+		  statement("StructuredBuffer<", struct_name, "> ", to_name(var.self), to_resource_binding(var), ";");
```
HansKristian-ARM: Formatting?

spirv_hlsl.cpp
```diff
@@ -958,6 +998,31 @@ string CompilerHLSL::to_func_call_arg(uint32_t id)
 	return arg_str;
 }
 
+string CompilerHLSL::to_name(uint32_t id, bool allow_alias) const
```
HansKristian-ARM: Formatting. Use format_all.sh.

spirv_hlsl.cpp
```diff
@@ -958,6 +998,31 @@ string CompilerHLSL::to_func_call_arg(uint32_t id)
...
+{
+  string name = Compiler::to_name(id, allow_alias);
+
```
HansKristian-ARM: This is a bit crude, but it is probably okay for now.

reference/shaders-hlsl/comp/global-invocation-id-writable-ssbo-in-function.comp
austinEng: This array is just because I am using the same source shader that is in [`shaders-msl/comp/global-invocation-id-writable-ssbo-in-function.comp`](https://github.com/KhronosGroup/SPIRV-Cross/blob/master/shaders-msl/comp/global-invocation-id-writable-ssbo-in-function.comp)

The implicit unsized array is seen in changing `myStorage` to `myStorage[0]` [here](https://github.com/KhronosGroup/SPIRV-Cross/pull/220/files#diff-a268d1a6fec87b22037be5bf40f35bc8R27)

spirv_cross.cpp
austinEng: Yes, it is

spirv_hlsl.cpp
austinEng: Fixed

spirv_hlsl.cpp
austinEng: Fixed

reference/shaders-hlsl/comp/ssbo-array.comp
```diff
@@ -0,0 +1,25 @@
...
+static uint3 gl_GlobalInvocationID;
+struct SPIRV_Cross_Input
+{
+    uint3 gl_GlobalInvocationID : SV_DispatchThreadID;
+};
+
+void comp_main()
+{
+    uint ident = gl_GlobalInvocationID.x;
+    ssbos[1].data[ident] = ssbos[0].data[ident];
```
HansKristian-ARM: This is wrong. ssbos should be an array (array of SSBOs), so you would need to first do the array access, then the [0].

shaders-hlsl/comp/ssbo-array.comp
```diff
@@ -0,0 +1,14 @@
+#version 310 es
+layout(local_size_x = 1) in;
+
+layout(std430, binding = 0) buffer SSBO
+{
+    vec4 data[];
```
HansKristian-ARM: This is the only place I see unsized array in the test shaders, but the reference shader using it is not correct.

reference/shaders-hlsl/comp/ssbo-array.comp
```diff
@@ -0,0 +1,25 @@
+struct SSBO
+{
+    float4 data[1];
```
HansKristian-ARM: This isn't data[0] either, or should it be 1?

reference/shaders-hlsl/comp/ssbo-array.comp
HansKristian-ARM: This would break the to_name() approach though ...

HansKristian-ARM: Array of SSBO is rather niche. We could avoid supporting that for now and solve it cleaner later, considering it's blocking basic compute support.

Could you change the test to just be an unsized array and not array of SSBO?

austinEng: You're right this test case is incorrect. The way this is working right now is that it doesn't support an array of separate SSBOs.
```
layout(std430, binding = 0) buffer SSBO
{
    vec4 data[];
} ssbos[2];
```
is being interpreted as meaning "I have **one** SSBO which **is** and array of `struct { vec4 data[]; }`," instead of "I have an **array of** SSBOs each which contains `vec4 data[];`.

So I think the tough point here is that the contents of the SSBOs are unsized arrays.
Each UAV binding becomes an `RWStructuredBuffer<T>` which is fine if `T` has a fixed size.
This however, doesn't work so well if T has an unsized array. `RWStructuredBuffer<T> buffer` is sort of implicitly the same as `T buffer[]`.

I'm not sure why a `[1]` was added to the generated code, but the generated SPIRV should be:
```
struct SSBO
{
    float4 data[];
};

RWStructuredBuffer<SSBO> ssbos : register(u0);
```
which doesn't really make sense because then `RWStructuredBuffer<SSBO> ssbos` which is implicitly `SSBO ssbos[]` is an unsized array containing unsized float4 arrays.

If, however, `T` is fixed size, then there are no issues. For example:
```
struct Particle {
    vec2 pos;
    vec2 vel;
};

layout(std430, set = 0, binding = 0) buffer ParticlesA {
    Particle particle;
} particlesA[1000];
```
becomes something like:
```
struct Particle {
    vec2 pos;
    vec2 vel;
};

struct _ParticlesA {
    Particle particle;
};

RWStructuredBuffer<_ParticlesA> particlesA: register(u0);
```

HansKristian-ARM: So ... GLSL, and SPIR-V support scenarios like these:

```
buffer SSBO
{
    vec4 a;
    vec4 b;
    vec2 other_data[]; // Unsized array, can only be at the end.
};
```

The way I see it now, we can support the case where we do not have any unsized arrays by using the [0] trick.

A lone unsized array can only be supported directly because this is basically what structured buffers are ...

The combination however is a bit murky. I was under the impression that we could "pretend" that other_data is just an array of size 1, and we just access the array beyond that, but apparently that won't work?

austinEng: I think it would work for this case:
```
layout(std430, binding = 0) buffer SSBO
{
    vec4 unsized_arr[];
} data;
```

But it is problematic if you have an array of unsized arrays like 
```
layout(std430, binding = 0) buffer SSBO
{
    vec4 unsized_arr[];
} data[10];
```
Because both of these will translate to
```
struct SSBO {
 vec4 unsized_arr[1];
};

RWStructuredBuffer<SSBO> data;
```

The first case is fine because you would just do `data[0].unsized_arr[index]` to access any element of `unsized_arr` and just go "out of bounds". But in the second case, the offsets to `data[1].unsized_arr[0]` and `data[2].unsized_arr[0]` are unknown.

HansKristian-ARM: Well, ok. So if we ignore this case for now:
```
layout(std430, binding = 0) buffer SSBO
{
    // Anything
} data[10];
```
The current implementation should work?

austinEng: This will work:
```
layout(std430, binding = 0) buffer SSBO0
{
    // Something**fixed** size
} data[10];

layout(std430, binding = 1) buffer SSBO1
{
    // Something**fixed** size
} data;

layout(std430, binding = 2) buffer SSBO2
{
    // Something**fixed** size
} data[];
```

This might work but will result in accessing out of bounds on an array of size 1:
```
layout(std430, binding = 3) buffer SSBO3
{
    vec2 array[];
} data;
```

This will not work:
```
layout(std430, binding = 4) buffer SSBO4
{
    vec2 array[];
} data[10];

// The contents are too big for a RWStructuredBuffer
layout(std430, binding = 5) buffer SSBO5
{
    float buf[10000000];
} data
```

That being said, after further investigation, if we want more complete support for HLSL compute, this will all probably need to change to use `RWByteAddressBuffer` instead of `RWStructuredBuffer`. `RWByteAddressBuffer` is just a raw view of bytes and has fewer limitations than `RWStructuredBuffer`. (`RWStructuredBuffer<T>` has a limited size for `T` as well. You cannot do `RWStructuredBuffer<ReallyBigStruct>`).

Using `RWStructuredBuffer` would require converting all `OpAccessChains` for UAVs to Load/Store operations with the correct byte offsets, so that may be quite a bit of work.

HansKristian-ARM: Ye, that sounds similar to the UBO flattening support we have for legacy GLSL targets.
Just going for ByteAddressBuffer and doing offset computation manually seems like the only workable solution long-term ...

What do you propose we do then? I can try making the flattening code for GLSL apply to HLSL as well, but might take a week or so to get around to. If RWStructuredBuffer<T> hack doesn't really work for somewhat large SSBOs, it's not something I would like to merge.

austinEng: I am fine with not merging for now

HansKristian-ARM:
I've begun implementation of ByteAddressBuffer in HLSL here: https://github.com/KhronosGroup/SPIRV-Cross/tree/fix-249

HansKristian-ARM:
Ok, it took a while, but I've merged basic ByteAddressBuffer/RWByteAddressBuffer support. Thanks for the contribution, even though it was not merged in the end.

austinEng:
Thank you for taking a look at this!

PR #224, opened 2017-07-17
Add support for building a shared library

hesiod:
Add support for buliding SPIRV-Cross as a shared library,
in addition to building a static library.

Shared libraries are better suited and generally preferred for
distribution via packages.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,5 +13,6 @@
 *.opensdf
 *.shader
 *.a
+*.so
 
 !CMakeLists.txt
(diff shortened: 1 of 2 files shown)


HansKristian-ARM:
I'm not sure how to deal with shared libraries. The API headers of SPIRV-Cross are not particularly stable atm (and it's C++ with lots of inlining, etc). Distributing a pre-built shared library will be very painful at the current time w.r.t. versioning. I do not wish to send the message to potential upstream maintainers that they should ship an .so in a distribution, because it probably won't work.

The only use for shared libraries would be local use.

HansKristian-ARM:
I'll assume this is abandoned.

Issue #225, opened 2017-07-17
Double negation translated incorrectly


turol:
This shader is translated incorrectly:

```
#version 450 core

layout(location = 0) in vec4 input1;
layout(location = 0) out vec4 outColor;

void main() {
    outColor = vec4(input1.x, -(-input1.y), input1.z, -(-input1.w));
}
```

to this:

```
    outColor = vec4(input1.x, --input1.y, input1.z, --input1.w);
```

The double negation in the original is due to preprocessor trickery.

HansKristian-ARM:
Ah, probably some missed case with parenthesis removal optimization. Should be easy to fix.

turol:
Yes, I tracked it down to [CompilerGLSL::enclose_expression](https://github.com/KhronosGroup/SPIRV-Cross/blob/master/spirv_glsl.cpp#L1812) which should check if `expr` begins with minus. I hacked around it locally but didn't make a PR because I'd have to write tests and everything. Also should probably check if the other backends are affected.

HansKristian-ARM:
Fixed on master.

Issue #223, opened 2017-07-13
gl_SampleID missing


HansKristian-ARM:
Trivial issue. gl_SampleID builtin support is missing in GLSL output.

HansKristian-ARM:
Fixed on master.

Issue #219, opened 2017-07-05
HLSL's SampleCmpLevelZero not translated correctly to GLSL when using TextureCube


Loki7979:
Hi,

Some days ago I posted a similar issue because hlsl's SampleCmpLevelZero was not translated correctly to glsl when using Texture2DArray : https://github.com/KhronosGroup/SPIRV-Cross/issues/207

Well, the same issue happens when using hlsl's TextureCube:

HLSL code:

```
TextureCube pointLightShadowMap;
SamplerComparisonState shadowSamplerPCF;

float comparisonResult = pointLightShadowMap.SampleCmpLevelZero(shadowSamplerPCF, lightDir, fragDepth);
```

translated GLSL code:

```
uniform samplerCubeShadow pointLightShadowMap;

vec4 _799 = vec4(lightDir, fragDepth);
float comparisonResult = textureLod(pointLightShadowMap, vec4(_799.xyz, _799.w), 0.0);
```

the glsl shader fails when compiling because textureLod doesn't accept samplerCubeShadow as parameter.

Please, could you perform the same fix in this specific case?

Thank you very much

HansKristian-ARM:
Will do, but might take a while due to holidays.

Loki7979:
No problem, I can wait. ;-)

Thanks for responding and have a nice holidays!

HansKristian-ARM:
Implemented on master.

Loki7979:
Thanks Mr! Works like a charm! :-)

PR #228, opened 2017-07-24
Various bug fixes

HansKristian-ARM:
Fixes:
- https://github.com/KhronosGroup/SPIRV-Cross/issues/223
- https://github.com/KhronosGroup/SPIRV-Cross/issues/219
- https://github.com/KhronosGroup/SPIRV-Cross/issues/225

### DIFF
--- a/spirv_glsl.cpp
+++ b/spirv_glsl.cpp
@@ -1812,6 +1812,16 @@ void CompilerGLSL::strip_enclosed_expression(string &expr)
 string CompilerGLSL::enclose_expression(const string &expr)
 {
 	bool need_parens = false;
+
+	// If the expression starts with a unary we need to enclose to deal with cases where we have back-to-back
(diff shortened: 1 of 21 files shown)


spirv_glsl.cpp
```diff
@@ -1812,6 +1812,16 @@ void CompilerGLSL::strip_enclosed_expression(string &expr)
...
+	// unary expressions.
+	if (!expr.empty())
+	{
+		auto c = expr.front();
+		if (c == '-' || c == '+' || c == '!' || c == '~')
+			need_parens = true;
+	}
+
 	uint32_t paren_count = 0;
 	for (auto c : expr)
```
turol: Should this loop be skipped if we already have `need_parens = true`?

HansKristian-ARM: I suppose it could.

Issue #226, opened 2017-07-21
SPIRVCross hangs on cfg.walk_from


redorav:
I have a relatively complex shader that SPIRVCross seems to be having a lot of difficulty converting to glsl. Specifically this block of code is looping for a very long time.

```
cfg.walk_from(header_block.merge_block, [&](uint32_t walk_block) {
			// We found a block which accesses the variable outside the loop.
			if (blocks.find(walk_block) != end(blocks))
				static_loop_init = false;
		});
```

This is happening after I apply an inlining pass to the original SPIRV which was sourced from hlsl. If I pass unmodified SPIRV to SPIRVCross, it works. The root cause seems to be a big block of if/else statements that gets inlined into the main entry point, as I comment them out it seems to do a faster job of analyzing them. I don't have a minimal example that reproduces this behavior, but I'd like to send you the spv and readable spv to see if we can work through it (github won't let me attach the spv extension and I also don't want to put it here publicly if possible)

redorav:
As an update, disabling the cfg analysis makes it compile properly and works.

HansKristian-ARM:
That's bad. Can you upload the SPIR-V and if possible a way to generate that SPIR-V?

redorav:
Is there a way I can send you this privately? It's company code I'm trying to compile and I wouldn't like to link it here publicly.

HansKristian-ARM:
You can send it to hans-kristian.arntzen@arm.com.

redorav:
Done

HansKristian-ARM:
I'll close this as the failing shader after the fix was likely caused by SPIRV-Tools. Please open a new ticket if you find anything.

PR #229, opened 2017-07-25
Optimize CFG walking

HansKristian-ARM:
Old implementation would explode runtime for long shaders with many blocks.
Fixes https://github.com/KhronosGroup/SPIRV-Cross/issues/226.

### DIFF
--- a/spirv_cfg.hpp
+++ b/spirv_cfg.hpp
@@ -68,11 +68,15 @@ public:
 	}
 
 	template <typename Op>
-	void walk_from(uint32_t block, const Op &op) const
+	void walk_from(std::unordered_set<uint32_t> &seen_blocks, uint32_t block, const Op &op) const
 	{
+		if (seen_blocks.count(block))
+			return;
+		seen_blocks.insert(block);
+
 		op(block);
 		for (auto b : succeeding_edges[block])
-			walk_from(b, op);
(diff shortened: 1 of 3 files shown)


redorav:
It's seems to still be hanging unfortunately, and the glsl it produces also seems to have some sort of error after the cfg analysis. It places a variable within a scope not visible to another variable. I can send you the problematic spv and glsl produced without the cfg by email later during the day.

HansKristian-ARM:
Is this the same shader you sent initially? Decompile was very fast and the shader passed glslang validation for me.

redorav:
I'm not sure it's the same, perhaps it passed the problematic shader and hung on a different one, I only did a quick test to see if all shaders (~130) passed. I can check later and give more details, plus send code.

HansKristian-ARM:
Please do. I will still merge this particular PR though since it's a generally good optimization anyways, but I won't close the ticket.

redorav:
I've been trying to get the shader for you but unfortunately the SPIR-V optimizer is also creating invalid code in some cases so I don't want to send you off on some hunt that ends up being a bug for SPIRVTools. I've been sidetracked a bit but I'm still keeping this in mind.

HansKristian-ARM:
Ok. Yes, wrong SPIR-V inlining is known to cause weird issues like that.

PR #230, opened 2017-07-25
SPIRV GLSL: Add vertex.flip_y option.

Kangz:
This adds a "gl_Position = -gl_Position" at the end of the shader so
that the OpenGL pixel position correspond to Vulkan pixel positions.

### DIFF
--- a/spirv_glsl.hpp
+++ b/spirv_glsl.hpp
@@ -91,6 +91,8 @@ public:
 		{
 			// In vertex shaders, rewrite [0, w] depth (Vulkan/D3D style) to [-w, w] depth (GL style).
 			bool fixup_clipspace = true;
+			// Inverts the Y coordinate so that the [-1, -1] pixel corresponds to texture coordinate [0, 0]
+			bool flip_y = false;
 		} vertex;
 
 		struct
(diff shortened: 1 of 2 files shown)


Kangz:
@HansKristian-ARM PTAL

Technically we should also fixup gl_FragCoord but it isn't done in the HLSL backend when flip_vert_y is used so I skipped it for now.

spirv_glsl.hpp
```diff
@@ -91,6 +91,8 @@ class CompilerGLSL : public Compiler
 		{
 			// In vertex shaders, rewrite [0, w] depth (Vulkan/D3D style) to [-w, w] depth (GL style).
 			bool fixup_clipspace = true;
+			// Inverts the Y coordinate so that the [-1, -1] pixel corresponds to texture coordinate [0, 0]
+			bool flip_y = false;
```
HansKristian-ARM: Probably best to call this flip_vert_y to match HLSL?
Also, the comment seems a bit weird. In GL, [-1, -1] already maps to window space [0, 0], and hence texture coordinate [0, 0].

HansKristian-ARM: Flipping gl_FragCoord is a bad idea. It would need a magic uniform for the Y resolution.
Y-flipping is a dirty hack at best.

Kangz:
I'm still trying to figure out what the correct approach would be for our use-case but I noticed #246 which is a much better version of this patch. Thanks for putting it together!

Abandoning this change.

PR #246, opened 2017-08-03
Unify vertex clip-coord fixups.

### LINKED ISSUE #241: hlsl back end does not respect option "fixup_clipspace"
Hello!
The problem is, CompilerHlsl inherates CompilerGlsl, both of them have a options member, but of different types. And the default value of fixup_clipspace of the two options are different.
In the following call stack, hlsl compiler ends up calling glsl compiler:
CompilerGLSL::emit_fixup() //use options of glsl
CompilerGLSL::emit_block_chain()
CompilerGLSL::emit_function()
CompilerHLSL::compile()

HansKristian-ARM:
The different options were scattered around in different backends and
didn't really work right with CLI at all.

This slightly changes the API, so @billhollings, @RobDangerous, please see if this will break anything. Also unifies the interface so Y-flip / clipspace fixup works as expected for all targets now.

Fixes #241.

### DIFF
--- a/README.md
+++ b/README.md
@@ -241,6 +241,20 @@ but older GLSL relies on symbol names to perform the linking. When emitting shad
 so it is important that the API user ensures that the names of I/O variables are sanitized so that linking will work properly.
 The reflection API can rename variables, struct types and struct members to deal with these scenarios using `Compiler::set_name` and friends.
 
+#### Clip-space conventions
+
+SPIRV-Cross can perform some common clip space conversions on gl_Position/SV_Position by enabling `CompilerGLSL::Options.vertex.fixup_clipspace`.
+While this can be convenient, it is recommended to modify the projection matrices instead as that can achieve the same result.
+
+For GLSL targets, enabling this will convert a shader which assumes `[0, w]` depth range (Vulkan / D3D / Metal) into `[-w, w]` range.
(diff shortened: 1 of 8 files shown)


billhollings:
MSL is fine. Good to go.

HansKristian-ARM:
Ok, thanks for confiming. Don't want to leave this hanging too long, so just merging now.

billhollings:
One last question...

@HansKristian-ARM...your note above indicates "works as expected for all targets now".

The default value for `options.vertex.flip_vert_y` is `false`.

For MSL, standard SPIR-V coming from a Vulkan implementation will only render correctly if `flip_vert_y` is set to `true`...and therefore would only "work as expected" for SPIR-V that already included a flip to MSL coordinates...that is...SPIR-V that was tailored specifically for Metal.

It doesn't matter for MoltenVK...because MoltenVK sets `flip_vert_y` to `true` by default. But would it not make sense for other users of MSL to have `CompilerMSL` set `flip_vert_y` to `true` in its constructor?

Or do other users want to do that coordinate flip in SPIR-V before conversion to MSL?

HansKristian-ARM:
I think this option is something which needs/should be set explicitly. It is certainly possible for CompilerMSL constructor to just force the option to true, but that makes the CLI a bit hairier, because we will need a way to turn it off ...

Standalone applications for example might want to deal with Y-flip by modifying their projection matrices, so I'd prefer to keep the default as the least surprising one.

billhollings:
Okay. That works.

PR #255, opened 2017-08-15
Basic HLSL compute shader support

### LINKED ISSUE #249: Implement SSBO flattening for HLSL
We will need to use ByteAddressBuffer and RWAddressBuffer to implement SSBOs. This will require special handling of SSBO access to handle address flattening not only for loads, but also for stores.

HansKristian-ARM:
Built upon https://github.com/KhronosGroup/SPIRV-Cross/pull/220.
Fixes #249.

Lots of stuff is certainly missing, but it should be usable now for basic compute shaders.

### DIFF
--- a/spirv_hlsl.cpp
+++ b/spirv_hlsl.cpp
@@ -1798,11 +1798,180 @@ void CompilerHLSL::emit_glsl_op(uint32_t result_type, uint32_t id, uint32_t eop,
 	}
 }
 
+void CompilerHLSL::emit_load(const Instruction &instruction)
+{
+	auto ops = stream(instruction);
+
+	auto *chain = maybe_get<SPIRAccessChain>(ops[2]);
+	if (chain)
+	{
+		auto &type = get<SPIRType>(chain->basetype);
+		uint32_t result_type = ops[0];
+		uint32_t id = ops[1];
+		uint32_t ptr = ops[2];
+
+		SPIRType target_type;
+		target_type.basetype = SPIRType::UInt;
+		target_type.vecsize = type.vecsize;
+		target_type.columns = type.columns;
+
+		if (type.columns != 1)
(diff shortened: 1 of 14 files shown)


spirv_hlsl.cpp
```diff
@@ -1798,6 +1850,199 @@ void CompilerHLSL::emit_glsl_op(uint32_t result_type, uint32_t id, uint32_t eop,
...
+		load_op = "Load3";
+		break;
+	case 4:
+		load_op = "Load4";
+		break;
+	default:
+		SPIRV_CROSS_THROW("Unknown vector size.");
+	}
+
+	auto load_expr = join(chain.base, ".", load_op, "(", chain.dynamic_index, chain.static_index, ")");
```
Kangz: Is a + missing between dynamic_index and static_index here? The same happens in the handling of store.

HansKristian-ARM: This is intentional. dynamic_index is something like "16 * foo + ", or blank.

Kangz: Sounds good

Kangz:
Thanks for taking a look at this, approach looks like the proper way to support SSBOs. How will passing SPIRCAccessChains to functions work, will it pass a pair of (buffer, offset)?

HansKristian-ARM:
I'm not entirely sure if you can pass access chains like that to functions, but glslang codegen typically goes via a dummy variable to satisfy the wonky requirements of function calls in GLSL/HLSL. Either we do that as well if needed, or pass down Buffer + Offset. The annoying bit with latter of course is that you probably cannot cast between ByteAddressBuffer and RWByteAddressBuffer, hnnngggggg. <_<

Kangz:
Pulling @dneto0 into the conversation. From what I remember, there is a constraint of SPIRV that makes it so we can know in the function if the parameter is a pointer in a readable or RW SSBO, but I can't find it in the spec. Do you know about it?

PR #272, opened 2017-09-05
Fix assignment optimization for logical and/or.

dsrbecky:
"x = x && y" was converted to "x &= y", which is invalid.


HansKristian-ARM:
Thanks for finding this. *facepalm*, I'll merge and add some regression tests for this.

PR #273, opened 2017-09-06
RMW optimizations tests

HansKristian-ARM:


HansKristian-ARM:
Travis hung, and already passes tests, so merging now.

PR #280, opened 2017-09-11
Fix texture GatherRed/Green/etc. methods for HLSL SM 5.0

rossy:
In GLSL, it seems like the component is selected with an integer constant argument, but in HLSL, it's selected by calling a different method on the texture itself.

EDIT: I noticed the Travis build didn't start again. I've contacted Travis support in case this is something to do with my account. 😕 The tests pass locally (including with fxc.exe.)


HansKristian-ARM:
Weird ... Well, ok. I've ran the test suite locally, and it passes, so I'll just merge it. PR looks good!

PR #284, opened 2017-09-20
Implement texture query APIs in HLSL

### LINKED ISSUE #283: can not convert textureSize to HLSL
layout (binding = 1) uniform sampler samLinear;
layout (binding = 2) uniform texture2D texGlow;
layout (location = 0) out vec4 outFragColor;
void main()
{
    vec2 texSize = textureSize(sampler2D(texGlow, samLinear), 0);
    outFragColor = vec4(0.0, 0.0, 0.0, 1.0);
}

this fragment shader world trigger assertion "Combined image samplers have no default expression representaion"

HansKristian-ARM:
Fixes #283.

### DIFF
--- a/reference/shaders-hlsl/frag/image-query-selective.frag
+++ b/reference/shaders-hlsl/frag/image-query-selective.frag
@@ -1,0 +1,145 @@
+Texture1D<uint4> uSampler1DUint : register(t0);
+SamplerState _uSampler1DUint_sampler : register(s0);
+Texture1D<int4> uSampler1DInt : register(t0);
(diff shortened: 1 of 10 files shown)


amerkoleci:
Why not inline GetDimensions calls to hlsl with some local variable generation?

HansKristian-ARM:
The main reason was that it was the easiest way, and GetDimensions is a very annoying API.
There are 4 opcodes which all would need to hit the same, but slightly different path.
Plus, this method is consistent with how other opcodes have been implemented, i.e. generate convenience functions instead of emitting many temporaries and duplicate code at the callsite.

PR #286, opened 2017-09-23
Fix HLSL atan2() argument order

rossy:
It seems like HLSL uses the same atan2(y, x) argument order as GLSL. Fixing this corrected an error in one of our cross-compiled shaders.

I also used a small test shader to verify that this was correct. Left is a GLSL shader that graphs the range of atan(y, x), cross-compiled to HLSL with glslang and SPIRV-Cross (with this fix.) Right is the same shader running natively in OpenGL.
image

References:
http://docs.gl/sl4/atan
https://msdn.microsoft.com/en-us/library/windows/desktop/bb509575.aspx


HansKristian-ARM:
Yup, that looks like a typo. Thanks for finding it.

PR #289, opened 2017-09-27
Fix compilation with clang

Wotan:
Hi,

Project doesn't compile on my machine (clang 5.0.0, linux), spirv-cross doesn't have the correct compile options set. It was working on gcc since c++11 is enabled by default. 
The commit adds compile options & defines.


HansKristian-ARM:
Thanks.

PR #300, opened 2017-10-06
Require GL_ARB_shader_image_load_store extension.

ben-clayton:
The early_fragment_tests layout qualifier requires either OpenGL 4.2 or use of the GL_ARB_shader_image_load_store extension.

### DIFF
--- a/spirv_glsl.cpp
+++ b/spirv_glsl.cpp
@@ -349,12 +349,18 @@ void CompilerGLSL::emit_header()
 	auto &execution = get_entry_point();
 	statement("#version ", options.version, options.es && options.version > 100 ? " es" : "");
 
-	// Needed for binding = # on UBOs, etc.
-	if (!options.es && options.version < 420 && options.enable_420pack_extension)
+	if (!options.es && options.version < 420)
 	{
-		statement("#ifdef GL_ARB_shading_language_420pack");
-		statement("#extension GL_ARB_shading_language_420pack : require");
-		statement("#endif");
+		// Needed for binding = # on UBOs, etc.
+		if (options.enable_420pack_extension)
+		{
+			statement("#ifdef GL_ARB_shading_language_420pack");
+			statement("#extension GL_ARB_shading_language_420pack : require");
+			statement("#endif");
+		}
+		// Needed for: layout(early_fragment_tests) in;
+		if (execution.flags & (1ull << ExecutionModeEarlyFragmentTests))
+			require_extension("GL_ARB_shader_image_load_store");
 	}
 
 	for (auto &ext : forced_extensions)


spirv_glsl.cpp
```diff
@@ -349,12 +349,24 @@ void CompilerGLSL::emit_header()
...
-	// Needed for binding = # on UBOs, etc.
-	if (!options.es && options.version < 420 && options.enable_420pack_extension)
+	bool can_use_image_load_store = true;
+
+	if (!options.es && options.version < 420)
 	{
-		statement("#ifdef GL_ARB_shading_language_420pack");
-		statement("#extension GL_ARB_shading_language_420pack : require");
-		statement("#endif");
+		// Needed for binding = # on UBOs, etc.
```
HansKristian-ARM: I think it makes more sense to only enable the shader_image_load_store extension if the extension is actually required. You know this because you only need it if ExecutionModeEarlyFragmentTests is actually enabled, and this can be checked here. Also, there can be double declaration, because image_load_store extension will be placed in forced_extensions if image load store itself is used.

Propose change to check if early_fragment_tests is enabled and version < 420, then add image_load_store to forced_extensions.

ben-clayton: Thank you for pointing me towards `forced_extensions`, I hadn't spotted that.

So I'm trying to do two things here:
1) Make sure that the code emitted by SPIRV-Cross actually generates compilable code for versions < 420 and the use of `EarlyFragmentTests`. I believe this is now correctly done with this updated PR.
2) Make shaders that use `early_fragment_tests` usable on machines that do not support the extension, nor support OpenGL 420, by stripping the `early_fragment_tests` qualifier and any explicit requirement of the extension. I'm guessing this should be done as a transform on the SPIRV, and not as an option in SPIRV-Cross?

HansKristian-ARM: You can use the unset_execution_mode() or whatever it is called to unset the early_fragment_tests execution mode. This way, the shader can be compat with older shader versions without introducing more options.

ben-clayton: Perfect - as we're using this as a library, I can call this function directly and I do not require any other changes. Thank you.

ben-clayton: (That said, this PR still fixes a bug, which I'd like to fix)

HansKristian-ARM: I assume the bug you are referring to is that use of EarlyFragmentTests should enable image_load_store extension on < 420 targets? If so, yes, I'm happy to merge that.

ben-clayton: > I assume the bug you are referring to is that use of EarlyFragmentTests should enable image_load_store extension on < 420 targets?

Yes. Please let me know if there are any more changes you'd like from this PR.

HansKristian-ARM: This PR should only do that then, no option for image load store extension should be there.

ben-clayton: I force-pushed the branch, as the roll-back was significant.
Can I confirm you're seeing the latest PR?

HansKristian-ARM: I can see it now.

HansKristian-ARM: Will merge once it passes Travis.

ben-clayton: Great, thank you.

HansKristian-ARM:
Travis is being a bit grumpy with OSX, but it has nothing to do with SPIRV-Cross and Linux is passing, so will merge now.

Issue #295, opened 2017-10-05
Overhaul HLSL row-major/column-major confusion


HansKristian-ARM:
The HLSL backend is in need of some testing and validation that the matrix order is working as intended. We will likely need to actually test with a proper application for this.

amerkoleci:
Note that D3DCompile has also option to specify the matrix order, but I think it can be handled by SPIRV-Cross also

HansKristian-ARM:
SPIR-V can declare matrices in both row-major/column-major and so can HLSL, so SPIRV-Cross will need to handle both cases, also different layouts in the same file.

HansKristian-ARM:
I believe this is more or less fixed, need to do some extra testing to close this.

BartmanAbyss:
It seems CompilerHLSL::layout_for_member should always output a matrix decoration, even if not in block interfaces.
We had the following GLSL->SPIRV code
`
struct ShadowMap {
   matrix lightMtx;
   ...
};

uniform Light {
  matrix modelMtx;
   ...
  ShadowMap shadows[6];
}`
where SPIRV-Cross would only add `row_major` decoration to `Light.modelMtx`, but not to `ShadowMap.lightMtx`

I commented out that check, and it seems to work, however I'm not 100% sure if that's the right thing to do...

sgorsten:
Thank you for your hard work maintaining this library!

I wanted to let you know that this issue tripped me up when writing a Direct3D 11 renderer for a small project I'm working on. In my case, the "matrix multiplication order inversion" behavior that SPIRV-Cross applies when generating HLSL was unhelpful and took me a little while to figure out. My engine stores matrices in column major order, and I memcpy CPU-side objects directly into mapped uniform/constant buffers, and my GLSL shader code uses (matrix * vector) multiplication order. This works for OpenGL and Vulkan, but I was surprised when I was getting garbage from my Direct3D backend, which I eventually tracked down to the `emit_binary_func_op(ops[0], ops[1], ops[3], ops[2], "mul");` lines in spirv_hlsl.cpp. It seems you also have logic to reverse the row_major/column_major annotations if any are present.

[...]

PR #304, opened 2017-10-10
Overhaul buffer packing in HLSL and GLSL

### LINKED ISSUE #302: Support for packoffset in the HLSL backend
Shader Model 4.0 seems to support explicit offsets in constant buffers with the [packoffset][1] keyword. It would be great if SPIRV-Cross could use this to replicate the UBO layout from the input SPIR-V. That way, code that uses the shader could use the same buffer layout for GLSL and HLSL targets.

One issue is that I think this will require removing the intermediate struct that SPIRV-Cross emits for UBOs, since packoffset only works on direct members of the cbuffer definition.

[1]: https://msdn.microsoft.com/en-us/library/windows/desktop/dd607358.aspx

HansKristian-ARM:
This is a fairly large change to support:
- GL_ARB_enhanced_layouts in GLSL to better support HLSL packing patterns.
- HLSL cbuffer code has been overhauled to support packoffset to better support std140 layouts.
- row_major/column_major is now declared in HLSL which should remove most issues with having to magically transpose matrices. (This needs more testing ...)
- Validation that the std140 fallback layout actually works. An error will be thrown if neither std140 nor std430 is possible as a buffer layout.
- To support packoffset, the cbuffer struct is flattened, so no more dummy struct outside the cbuffer declaration.

Closes #302,
Progresses #295.


HansKristian-ARM:
@rossy please give it a spin.

rossy:
Hmm, I get an error: undeclared identifier '_22_m0'

Full mpv output is as follows (includes raw GLSL and HLSL shader source, plus error output from D3DCompile()):
https://gist.github.com/rossy/55f71c4d6e03d0bcd7fd54d53047604a

HansKristian-ARM:
Ah, right ... Some silly cases with stripped SPIR-V modules, will fix.

HansKristian-ARM:
@rossy should work now.

rossy:
Yep, this appears to work for our shaders. Nice work.

PR #311, opened 2017-10-20
HLSL image load/store, atomics, shared memory, barriers

### LINKED ISSUE #303: HLSL: Barriers not supported
Translating compute shaders from GLSL -> HLSL will not produce correct barrier calls.
SPIRV-Cross will insert `barrier()` instead of the HLSL intrinsic.

HansKristian-ARM:
Fills in some gaps in the compute support of HLSL.

Fixes #303.


rossy:
Nice. FWIW, these changes seem to work with mpv's [EWA upscaling][1] compute shader and pass our test suite.

[1]: https://www.imagemagick.org/Usage/distorts/#distort_ewa

PR #312, opened 2017-10-21
Use 'static const' for spec constants in HLSL

rossy:
At least, I think this is correct. I don't know if HLSL has an equivalent for specialization constants in SPIR-V where the work group size can be specified after the shader is compiled. Without this change, the HLSL compiler ignores the initializers for the constants and expects them to be provided by a cbuffer.


HansKristian-ARM:
Wow, that is ... weird. HLSL does not have specialization constants, so they must be fixed to constants at compile-time. static const seems fine to me.

PR #320, opened 2017-11-06
HLSL Vertex Attribute remap

amerkoleci:
Add overload compile method for hlsl with vertex attribute remap.
This adds remap to semantic and semantic_index as HLSL instead of always binding to TEXCOORD + binding_number.

### DIFF
--- a/spirv_hlsl.cpp
+++ b/spirv_hlsl.cpp
@@ -666,6 +666,23 @@ void CompilerHLSL::emit_interface_block_in_struct(const SPIRVariable &var, unord
 		else
 			binding_number = get_vacant_location();
 
+		// Allow semantic remap if specified.
+		string semantic = "TEXCOORD";
+		uint32_t semantic_index = binding_number;
+
+		if (vertex_attributes.size())
+		{
+			for (auto &attribute : vertex_attributes)
+			{
+				if (attribute.name == name)
+				{
+					semantic = attribute.semantic;
+					semantic_index = attribute.semantic_index;
+					break;
+				}
+			}
+		}
+
 		if (need_matrix_unroll && type.columns > 1)
 		{
 			if (!type.array.empty())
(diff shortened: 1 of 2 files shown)


spirv_hlsl.cpp
```diff
@@ -3147,6 +3164,18 @@ void CompilerHLSL::require_texture_query_variant(const SPIRType &type)
 	}
 }
 
+string CompilerHLSL::compile(std::vector<HLSLVertexAttr> *p_vertex_attributes)
```
HansKristian-ARM: Better to take the vector by const<vector> &. There is no need for this to be a pointer.

spirv_hlsl.cpp
```diff
@@ -3147,6 +3164,18 @@ void CompilerHLSL::require_texture_query_variant(const SPIRType &type)
...
+{
+	if (p_vertex_attributes)
```
HansKristian-ARM: Just copy the vector directly.

spirv_hlsl.hpp
```diff
@@ -23,6 +23,13 @@
 
 namespace spirv_cross
 {
+struct HLSLVertexAttr
+{
+	std::string name;
```
HansKristian-ARM: This is a bit problematic, as this maps a *name* to a new semantic. However, you cannot rely on a name existing in the SPIR-V, e.g. if it has been stripped. Name information is purely debug info, and this interface should map location to a custom semantic rather than TEXCOORD#{location}. If your interfaces deal with names, it's trivial to map names to location.

spirv_hlsl.hpp
```diff
@@ -23,6 +23,13 @@
...
+	std::string semantic;
+	uint32_t semantic_index;
```
HansKristian-ARM: Is there a reason we need the semantic index? What does this mean?

spirv_hlsl.cpp
```diff
@@ -666,6 +666,23 @@ void CompilerHLSL::emit_interface_block_in_struct(const SPIRVariable &var, unord
 		else
 			binding_number = get_vacant_location();
 
+		// Allow semantic remap if specified.
+		string semantic = "TEXCOORD";
+		uint32_t semantic_index = binding_number;
+
+		if (vertex_attributes.size())
```
HansKristian-ARM: No need to check for size when you're iterating through vertex_attributes.

spirv_hlsl.cpp
HansKristian-ARM: This is still a mutable pointer, this should be `const std::vector<HLSLVertexAttr> &` type.

spirv_hlsl.hpp
```diff
@@ -23,6 +23,13 @@
 
 namespace spirv_cross
 {
+struct HLSLVertexAttr
+{
+	uint32_t binding;
```
HansKristian-ARM: This should be called location to match the API. Bindings are used for textures and buffers, etc.

spirv_hlsl.hpp
```diff
@@ -23,6 +23,13 @@
...
+	std::string semantic;
+	uint32_t semantic_index;
```
HansKristian-ARM: Is there a need for semantic_index?

spirv_hlsl.hpp
HansKristian-ARM: I do see the code around it being called mostly "binding", but that's a bit misleading. I should fix that sometime.

amerkoleci:
If you want merge then cleanup

HansKristian-ARM:
Ok, will do on Monday then.

HansKristian-ARM:
Merged.

Issue #316, opened 2017-10-29
Map from original to transformed entry point names


grovesNL:
Currently entry points may be transformed during compilation, but it's not obvious how to determine whether a transformation has occurred.

For example, if a default entry point is defined as `main` (which is quite common), the [MSL output will rename it to `main0`](https://github.com/KhronosGroup/SPIRV-Cross/blob/master/spirv_msl.cpp#L63) because `main` is not allowed in MSL. That behavior seems reasonable, but how can we determine that this renaming occurred in order to use the transformed name when configuring Metal shaders?

More generally, for any of the compilers, if a function named has been transformed (i.e. because it is reserved in that SL or some other issue), how can we map from the original entry point names to the transformed entry point names?

Perhaps it's as straightforward as creating and exposing a map prior to the compile step (i.e. during the constructor). Although maybe it's possible that some dynamic renaming would need to occur during compile in the future, in which case we wouldn't be able to do this. Let me know what you think.

HansKristian-ARM:
Setting an entry point name is something I think could already work.
It shouldn't be that much work to add a way to change the name of an entry point during reflection. This name should be used somehow during compilation (except on GLSL of course, because it only understands main()). I don't think this is happening at the moment given that MSL hardcodes the main -> main0 transform, but sounds trivial to make this more generic.

Of course there are issues with "main" as an entry point name as it is reserved, I think the easiest way there is just to say it is reserved and it will be called something else.

From an API perspective something like:
```
Compiler::rename_entry_point(from, to);
```
should be good enough, really.

grovesNL:
@HansKristian-ARM That sounds good to me!

It would be useful to either have 1) a list of reserved entry point names for each compiler so we can explicitly set a new name (i.e. so we know `to` is not reserved), or 2) for the current approach, a way to know whether it was renamed automatically by the compiler.

Being able to know all reserved names is not too important anyway. Applications could use `rename_entry_point` to automatically prefix all entry points with some random characters to avoid reserved words.

billhollings:
I'm not sure what the problem is that is trying to be fixed here. Can you clarify, please?

Currently MSL requirements force a renaming of `main()` to something else. Somewhat arbitrarily, I chose `main0(...)`.

As far as an app knowing how the compiler renames entry points...for MSL, the rules are basically...

1. `main` is converted to `main0`.
2. All other entry point names are left alone.

Do we have a use case for a more general need for more flexible renaming requirements?

[...]

PR #321, opened 2017-11-06
 Enhancements to MSL compute and entry point naming.

billhollings:
This adds significant enhancements to computer shaders for MSL.

All but one of the Sasha-Willems Vulkan compute samples will now run under these enhancements.

It also includes a fix for the entry point renaming discussed in Issue #316.


spirv_common.hpp
```diff
@@ -343,14 +343,16 @@ struct SPIREntryPoint
 {
 	SPIREntryPoint(uint32_t self_, spv::ExecutionModel execution_model, std::string entry_name)
```
HansKristian-ARM: Make this const& since you're not moving from it anymore.

spirv_cross.hpp
```diff
@@ -346,6 +351,10 @@ class Compiler
 	SPIRConstant &get_constant(uint32_t id);
 	const SPIRConstant &get_constant(uint32_t id) const;
 
+	// Recursively marks any constants referenced by the specified constant instruction as being used
```
HansKristian-ARM: Does this need to be in the public interface?

spirv_msl.cpp
```diff
@@ -114,12 +105,13 @@ string CompilerMSL::compile()
 	backend.swizzle_is_function = false;
 	backend.shared_is_implied = false;
 	backend.native_row_major_matrix = false;
+	backend.flexible_member_array_supported = false;
 
 	uint32_t pass_count = 0;
 	do
 	{
 		if (pass_count >= 3)
```
HansKristian-ARM: Do you need to change the branch as well, or?

spirv_msl.hpp
```diff
@@ -76,8 +76,10 @@ class CompilerMSL : public CompilerGLSL
 	// Options for compiling to Metal Shading Language
 	struct Options
 	{
-		bool enable_point_size_builtin = true;
 		std::string entry_point_name;
+		float msl_version = 1.2f;
```
HansKristian-ARM: Having a floating point value for version seems a bit dubious to me. Can we make this an integer like HLSL? We could have a static function which creates this integer version ala CompilerMSL::version(major, minor)

spirv_common.hpp
```diff
@@ -343,14 +343,16 @@ struct SPIREntryPoint
...
+	    , name(entry_name)
+	    , orig_name(entry_name)
 	    , model(execution_model)
 	{
 	}
 	SPIREntryPoint() = default;
 
 	uint32_t self = 0;
 	std::string name;
+	std::string orig_name;
```
HansKristian-ARM: So if I understand it correctly, orig_name will be used for "set_entry_point()" and friends, but name is remapped as necessary?

spirv_cross.cpp
```diff
@@ -2527,6 +2530,14 @@ vector<string> Compiler::get_entry_points() const
 	return entries;
 }
 
+unordered_map<string, string> Compiler::get_entry_point_name_map() const
```
HansKristian-ARM: Can this be called before compile()?

spirv_msl.cpp
```diff
@@ -1434,74 +1493,185 @@ void CompilerMSL::emit_instruction(const Instruction &instruction)
...
+	case MemorySemanticsAcquireReleaseMask:
+	case MemorySemanticsSequentiallyConsistentMask:
+	case MemorySemanticsUniformMemoryMask:
+	case MemorySemanticsMaskNone:
+	default:
+		bar_stmt += "mem_none";
+		break;
+	}
+
+	if (options.msl_version >= 2.0)
```
HansKristian-ARM: Floating point versions is a bit funky.

spirv_msl.hpp
```diff
@@ -76,8 +76,10 @@ class CompilerMSL : public CompilerGLSL
 	// Options for compiling to Metal Shading Language
 	struct Options
 	{
-		bool enable_point_size_builtin = true;
 		std::string entry_point_name;
```
HansKristian-ARM: Is entry_point_name still needed?

HansKristian-ARM:
Thanks for doing this work, we should be close to useable compute shader support for both MSL and HLSL now! :D

HansKristian-ARM:
Is there a way to change the entry point name in reflection? Like change main -> my_main()?

spirv_cross.cpp
HansKristian-ARM: It might be easier to just query the remapped name for a particular entry point (as the common case is single entry point), and have an interface like string `Compiler::get_entry_point_remapped_name(const string &entry)`.

spirv_msl.hpp
```diff
@@ -63,7 +63,7 @@ using MSLStructMemberKey = uint64_t;
 
 // Special constant used in a MSLResourceBinding desc_set
 // element to indicate the bindings for the push constants.
-static const uint32_t kPushConstDescSet = UINT32_MAX;
+static const uint32_t kPushConstDescSet = std::numeric_limits<uint32_t>::max();
```
HansKristian-ARM: Build failure on Linux:
```
spirv_msl.hpp:66:43: error: ‘numeric_limits’ is not a member of ‘std’
 static const uint32_t kPushConstDescSet = std::numeric_limits<uint32_t>::max();
                                           ^
spirv_msl.hpp:66:71: error: expected primary-expression before ‘>’ token
 static const uint32_t kPushConstDescSet = std::numeric_limits<uint32_t>::max();
                                                                       ^
spirv_msl.hpp:66:72: error: ‘::max’ has not been declared
 static const uint32_t kPushConstDescSet = std::numeric_limits<uint32_t>::max();
```

spirv_common.hpp
billhollings: Done.

spirv_cross.hpp
billhollings: Moved to protected section.

spirv_msl.cpp
billhollings: I'm not sure what you are asking here. What "branch" do you mean?

spirv_msl.hpp
billhollings: I've changed this to an int and added setter and validation functions.

spirv_common.hpp
billhollings: No. The entry point renaming is universal.

The only thing `orig_name` is used for is extracting the mapping via the `get_entry_point_name_map()` API function, per issue #316.

spirv_cross.cpp
billhollings: No. The remapping is part of compile().

I implemented the mapping per the request in issue #316. I don't mind either way.

spirv_msl.cpp
billhollings: Changed to integer now.

spirv_msl.hpp
billhollings: Oh-oh!

AFAIK...this is standard C++11 (actually C++98) [functionality](http://www.cplusplus.com/reference/limits/numeric_limits/).

Can someone with Linux C++ experience weigh in here, please?

billhollings:
I've resubmitted based on feedback...and responded to comments above.

Not sure what to do with the `std::numeric_limits` issue on Linux. Anyone have any idea why Linux would not support standard C++11 behaviour?

spirv_msl.cpp
HansKristian-ARM: if pass_count >= 3

spirv_cross.cpp
HansKristian-ARM: Ok, it's probably okay, but this behavior needs to be documented I think. We can enforce this in runtime by throwing an exception if this is called before compile().

I still prefer returning a single string here though. Returning an unordered_map for this seems a bit too much.

spirv_common.hpp
HansKristian-ARM: So orig_name is recorded during parse, but name is the remapped name which happens during compile() in case the entry point name is incompatible?

spirv_msl.hpp
HansKristian-ARM: You need <limits>, yes, that should do it.

spirv_msl.cpp
billhollings: Oh...I see what you are commenting on...the change to the error text.

Hmmm...looks like a brain-farted and made that change incorrectly. I must have misread the >= as >. I've turned it back.

spirv_common.hpp
billhollings: Yes...illegal function names...both in the entry point and the function alias are renamed early in the compile stage.

The `orig_name` is retained strictly to allow a post-compile query to be made about which names were changed and from what to what.

spirv_cross.cpp
grovesNL: I don't really have a preference for #316, either `get_entry_point_name_map` or `get_entry_point_remapped_name` are fine IMO

billhollings: Once renaming happens...the existing `get_entry_points()` function will return a list of the **_new_** names (i.e.. in the most common case it will contain `main0`). So the app has no way of getting a list of the **_old_** names...which are the ones the app is likely expecting to be familiar with already.

So...one potential issue with a single-string query-answer function is the app knowing what to use as the look-up key?

1. Would we use the original name as the key...effectively expecting the app to somehow know what the SPIR-V entry points were before renaming...without being able to ask the compiler for that info?

2. Or would the app use `get_entry_points()` to get the new names...and then  cycle through them to find which one mapped to an old name? That seems a bit backwards to me.

billhollings: Regarding the question of timing...

We wouldn't want the renaming to occur during parsing...because that is backend-dependent...and would break the design goal of being able to reuse parsed results to output multiple backend languages.

After parsing but before compiling, `get_entry_point_name_map()` will return a mapping of:

    original names -> original names

After compiling, it will return a mapping of:

    original names -> new names

I have documented this now.

billhollings:
I've pushed fixes from the second review now.

Discussion on entry-point renaming is still outstanding. See my clarification comments there.

spirv_cross.cpp
HansKristian-ARM: I see the API flow as something like this:

- get_entry_points() is generally something you do before compile(), because after all, you need to figure out which entry points you want to use.

- Then, you call compile(), and the entry point names are potentially changed in MSL.

- Before you pass this MSL to the compiler, you can remap the entry point you compiled using get_entry_point_name_mapping(), to see what your actual entry point is.


It should be possible for get_entry_points() to always return the original names I think, it could just return "orig_name". The entry points are a property of the SPIR-V module itself, and I don't think it needs to change like this. The API should always accept original names for entry points, except that the "remapped name" is something which happens in the MSL output, but only there.

spirv_msl.hpp
```diff
@@ -73,11 +74,24 @@ static const uint32_t kPushConstBinding = 0;
 class CompilerMSL : public CompilerGLSL
 {
 public:
+#define MAKE_MSL_VERSION(major, minor, patch) (((major)*10000) + ((minor)*100) + (patch))
```
HansKristian-ARM: Please make this a public static function. This macro does not have a namespace which could conflict with other modules.

spirv_cross.cpp
HansKristian-ARM: It might be better to throw an exception if get_entry_point_name_mapping is called before compile() imo. API users will call it, get some sensible result, and be surprised when they get the wrong result. Wrong API use should have an unignorable error.

spirv_msl.hpp
billhollings: I've made it a static function now.

spirv_cross.cpp
billhollings: I have changed `get_entry_points()` and `get_entry_point(name)` functions to use the `orig_name` of the entry point instead of `name`.

I have removed the `get_entry_point_name_mapping()` function and introduced a `get_cleansed_entry_point_name(name)` function.

I could not determine an consistent way of detecting whether the `compile()` function had been run across all backends...so I left the idea of throwing an exception unimplemented...and simply documented the `get_cleansed_entry_point_name(name)` function to indicate what the API user should expect.

billhollings:
I've pushed fixes from the third review now.

HansKristian-ARM:
Thanks for fixing stuff.

grovesNL:
Thanks @billhollings and @HansKristian-ARM. This looks great.

billhollings:
Thanks. Glad it's all good...and useful!

PR #325, opened 2017-11-15
Support new hlsl 5.1 space for registers.

amerkoleci:
As described here, support the new space{set} under hlsl 5.1
https://msdn.microsoft.com/en-us/library/windows/desktop/dn899207(v=vs.85).aspx


HansKristian-ARM:
Tests fail, you'll need to `./test_shaders.py shaders-hlsl --hlsl --update` to update the reference files. Let me know if you have problems doing this.

HansKristian-ARM:
Fixed up and merged.

PR #326, opened 2017-11-16
CompilerMSL support and tests for OpUndef.

billhollings:
Changes outside `spirv_msl.hpp/cpp` are result of running `format_all.sh`.

### DIFF
--- a/reference/shaders-msl/asm/frag/default-member-names.asm.frag
+++ b/reference/shaders-msl/asm/frag/default-member-names.asm.frag
@@ -1,0 +1,40 @@
+#include <metal_stdlib>
+#include <simd/simd.h>
+
+using namespace metal;
+
+struct _9
+{
+    float _m0;
+};
+
+struct _10
+{
+    float _m0;
+    float _m1;
+    float _m2;
+    float _m3;
+    float _m4;
+    float _m5;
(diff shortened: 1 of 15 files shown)


HansKristian-ARM:
It seems like we have mismatching clang-format versions, mine is on 5.0.0 at least.

billhollings:
Thanks.

I've upgraded to `clang-format 5.0.0` now.

I wonder if there is a way of validating or enforcing the minimum `clang-format` version within `format_all.sh`...to avoid successive submissions flipping the formatting back and forth as different submitters drift versions of `clang-format` over time.

HansKristian-ARM:
Right, I'm a bit concerned that putting some mandate on the clang-format would drive potential contributors away. I'm on Arch, so I always have the bleeding edge stuff, but not everyone will have that. I suppose I could look into adding some kind of warning at least when running ./format_all.sh and try to keep that up-to-date.

billhollings:
Hmmm...is it possible to run `format_all.sh` on the server during a pull request submission (like Travis, etc)...and avoid devs having to think about it at all?

HansKristian-ARM:
This implies that TravisCI is able to magically modify the commit in question ... I doubt that is possible or practical.

PR #337, opened 2017-11-21
HLSL: Support early_fragment_tests attribute

msiglreith:
Other backends seem to support it already, but HLSL is missing it.


HansKristian-ARM:
Merging when it passes travis.

PR #353, opened 2017-11-30
For MSL, use {} instead of constructors to init OpUndef values.

billhollings:
Fixes an MSL compiler complaint about global constructors.


HansKristian-ARM:
Tests fail, because you'll need to run --opt now as well to update tests for optimized SPIR-V.

spirv_msl.cpp
```diff
@@ -1851,10 +1837,10 @@ void CompilerMSL::emit_glsl_op(uint32_t result_type, uint32_t id, uint32_t eop,
 		break;
 	}
 
-		// TODO:
-		//        GLSLstd450InterpolateAtCentroid (centroid_no_perspective qualifier)
-		//        GLSLstd450InterpolateAtSample (sample_no_perspective qualifier)
-		//        GLSLstd450InterpolateAtOffset
+	// TODO:
```
HansKristian-ARM: Is this with clang-format 5.0?

billhollings: Here's what I'm using on macOS 10.13.1...

    SPIRV-Cross $ clang-format -version
    clang-format version 5.0.0 (tags/google/stable/2017-06-22)

HansKristian-ARM: :\ Oh well, guess the number of conflicts is much lower now at least.

billhollings: The formatting effect is to bounce some // comments in all files back and forth by one indentation level.

Surely this can be fixed in `clang-format` configuration or the `format_all.sh` script, can it not?

Otherwise, this is going to be distracting, as it usually affects a number of files...even those that have not been edited. In this PR, I deliberately removed the other files that were ONLY changed by this reformatting, but we can't do that every time.

HansKristian-ARM: You can try // clang-format off, and // clang-format on, that should block clang-format from doing anything with the section.

billhollings: I can do that for this one. But these kind of comments are peppered around all the files. If I run `format_all.sh`...it makes similar changes to almost all the files.

I think we need to find a configurational way of making this change.

billhollings:
I've updated reference shaders in `reference/opt/shaders-msl`.

HansKristian-ARM:
Hum ... Yes, SPIRV-Tools doesn't appear to be as stable as I'd like. If different versions is a problem, I think we can try an approach where we build glslang and SPIRV-Tools locally with a fixed version so that contributors can build a known SPIRV-Tools version and glslang, then a second script to update all test files using that build. Does that sound okay?

I can fixup this PR myself for now. We'll also have to figure out the clang-format issues ...

billhollings:
@HansKristian-ARM 

Thanks for handling the SPIRV-Tools optimizer issues!

PR #358, opened 2017-12-05
msl generator enhancements/fixes

fatkas:
- pack/unpack nested constant buffer structs
- support for write-only textures (only global ones for now)
- better rt index support for msl generator

### DIFF
--- a/spirv_common.hpp
+++ b/spirv_common.hpp
@@ -657,10 +657,11 @@ struct SPIRVariable : IVariant
 	};
 
 	SPIRVariable() = default;
-	SPIRVariable(uint32_t basetype_, spv::StorageClass storage_, uint32_t initializer_ = 0)
+	SPIRVariable(uint32_t basetype_, spv::StorageClass storage_, uint32_t initializer_ = 0, uint32_t basevariable_ = 0)
 	    : basetype(basetype_)
 	    , storage(storage_)
 	    , initializer(initializer_)
+	    , basevariable(basevariable_)
 	{
 	}
 
@@ -668,6 +669,7 @@ struct SPIRVariable : IVariant
 	spv::StorageClass storage = spv::StorageClassGeneric;
 	uint32_t decoration = 0;
 	uint32_t initializer = 0;
+	uint32_t basevariable = 0;
 
 	std::vector<uint32_t> dereference_chain;
 	bool compat_builtin = false;
--- a/spirv_cross.cpp
+++ b/spirv_cross.cpp
@@ -1706,6 +1706,7 @@ void Compiler::parse(const Instruction &instruction)
 
 		auto &var = set<SPIRVariable>(id, type, storage, initializer);
 
+		// hlsl based shaders don't have those decorations. force them and then reset when reading/writing images
 		auto &ttype = get<SPIRType>(type);
 		if (ttype.basetype == SPIRType::BaseType::Image)
 		{
--- a/spirv_cross.cpp
+++ b/spirv_cross.cpp
@@ -1706,6 +1706,13 @@ void Compiler::parse(const Instruction &instruction)
 
 		auto &var = set<SPIRVariable>(id, type, storage, initializer);
 
+		auto &ttype = get<SPIRType>(type);
+		if (ttype.basetype == SPIRType::BaseType::Image)
+		{
+			set_decoration(id, DecorationNonWritable);
+			set_decoration(id, DecorationNonReadable);
+		}
+
 		if (variable_storage_is_aliased(var))
 			aliased_variables.push_back(var.self);
 
--- a/spirv_msl.cpp
+++ b/spirv_msl.cpp
@@ -301,9 +301,10 @@ void CompilerMSL::extract_global_variables_from_function(uint32_t func_id, std::
 		uint32_t next_id = increase_bound_by(uint32_t(added_arg_ids.size()));
 		for (uint32_t arg_id : added_arg_ids)
 		{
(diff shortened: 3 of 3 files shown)


spirv_cross.cpp
```diff
@@ -417,7 +417,7 @@ bool Compiler::is_immutable(uint32_t id) const
 		auto &var = get<SPIRVariable>(id);
 
 		// Anything we load from the UniformConstant address space is guaranteed to be immutable.
-		bool pointer_to_const = var.storage == StorageClassUniformConstant;
+		bool pointer_to_const = var.storage == StorageClassUniformConstant || var.storage == StorageClassUniform;
```
HansKristian-ARM: This doesn't seem right. Why is StorageClassUniform added here?

spirv_cross.cpp
```diff
@@ -1703,6 +1703,13 @@ void Compiler::parse(const Instruction &instruction)
 		}
 
 		auto &var = set<SPIRVariable>(id, type, storage, initializer);
+        
+        auto &ttype = get<SPIRType>(type);
```
HansKristian-ARM: Does the input SPIR-V not have these decorations?

spirv_msl.cpp
```diff
@@ -1365,8 +1371,11 @@ void CompilerMSL::emit_instruction(const Instruction &instruction)
 		// Mark that this shader reads from this image
 		uint32_t img_id = ops[2];
 		auto *p_var = maybe_get_backing_variable(img_id);
-		if (p_var)
+		if (p_var && has_decoration(p_var->self, DecorationNonReadable))
```
HansKristian-ARM: Is there something similar for NonWriteable?

spirv_common.hpp
```diff
@@ -657,17 +657,19 @@ struct SPIRVariable : IVariant
...
 	    , initializer(initializer_)
+        , base_variable(basevar)
 	{
 	}
 
 	uint32_t basetype = 0;
 	spv::StorageClass storage = spv::StorageClassGeneric;
 	uint32_t decoration = 0;
 	uint32_t initializer = 0;
+    uint32_t base_variable = 0;
```
HansKristian-ARM: Is this used to figure out which global variable maps to a shadow parameter?

spirv_msl.cpp
fatkas: yes, the code was already there

spirv_common.hpp
fatkas: yes

spirv_cross.cpp
fatkas: not for NonReadable at least. since the texture reading/writing code resets those decorations it looks safe to set both at the beginning

spirv_cross.cpp
fatkas: I think it missed nested struct unpacking but I didn't try running without that code after merging latest head. will double check

HansKristian-ARM:
Pinging @billhollings so he can have a look.

spirv_cross.cpp
HansKristian-ARM: Hm ... Is this coming from HLSL by any chance? In GLSL there is writeonly/readonly which should cover these things automatically.

fatkas: yeah, all of our sources are hlsl

spirv_msl.cpp
```diff
@@ -2996,12 +3007,13 @@ string CompilerMSL::builtin_to_glsl(BuiltIn builtin, StorageClass storage)
 		return "gl_VertexIndex";
 	case BuiltInInstanceIndex:
 		return "gl_InstanceIndex";
+	case BuiltInLayer:
+        return current_function && (current_function->self == entry_point) ? stage_out_var_name + ".gl_Layer": "gl_Layer";
```
HansKristian-ARM: Formatting

spirv_msl.cpp
```diff
@@ -3096,6 +3108,8 @@ string CompilerMSL::builtin_type_decl(BuiltIn builtin)
 		return "uint";
 	case BuiltInInstanceIndex:
 		return "uint";
+	case BuiltInLayer:
+        return "uint";
```
HansKristian-ARM: Formatting.

spirv_cross.cpp
```diff
@@ -1706,6 +1706,13 @@ void Compiler::parse(const Instruction &instruction)
 
 		auto &var = set<SPIRVariable>(id, type, storage, initializer);
 
+		auto &ttype = get<SPIRType>(type);
```
HansKristian-ARM: I would like a comment here stating why this is needed (HLSL does not set stuff).

billhollings:
Please make organizational mods I've suggested to use of `BuiltInLayer`. Then it's good to go.

fatkas:
> Please make organizational mods I've suggested to use of BuiltInLayer. Then it's good to go.

could you please elaborate what mods are you referring to?

billhollings:
>> Please make organizational mods I've suggested to use of BuiltInLayer. Then it's good to go.

> could you please elaborate what mods are you referring to?

See the two (now outdated) change requests I asked for above...which I've also summarized in the new change request I've made.

billhollings:
Good to go.

HansKristian-ARM:
Thanks for review and contribution.
@billhollings FYI, I'm on vacation starting tomorrow.

billhollings:
Have a great vacation.

When are you back...just so we can plan for future PR's?

HansKristian-ARM:
Thanks. I'm back 4th Jan.

Issue #372, opened 2017-12-24
Select/Deselect reference shaders based on platform/version being tested


billhollings:
For MSL, I've encountered a couple of use cases where being able to select and deselect reference shaders based on characteristics of the build...such as a platform type (iOS vs macOS)...or even a shader language version (MSL 2.0 vs 1.2).

Essentially...some MSL functions and parameters are not available in all MSL versions and on all platforms. Apple distinguishes between iOS and macOS...and MSL versions.

Another example is the use of SPIR-V optimizations (`--opt`). Optimizations of certain complex shaders sometimes create SPIR-V that is not straightforward to convert to MSL. Since we have no control over changes to SPIR-V optimizations...it would be useful to be able to selectively deselect those shaders from the optimization run.

What are our options for an approach here? Could we start to differentiate shader directories and selectively run them for platforms, versions, and optimizations from `test_shaders.sh`? Or could we think in terms of selecting shaders based on characteristics (eg- maybe a platform & version shader naming convention)?

billhollings:
PR #373 partially addresses this.

The MSL shaders that cannot be optimized have been moved to the new `shaders-msl-no-opts` folder.

MSL shaders in the `shaders-msl` are run both with and without optimizations. MSL shaders in the `shaders-msl-no-opts` folder are only run WITHOUT SPIR-V optimizations.

`CompilerMSL` now supports a platform (iOS or macOS) indicator to selectively output MSL according to platform requirements. This is in addition to the existing MSL version indication.

The ability to select this via the `spirv-cross` command line has NOT been implemented. Nor have any separate tests and reference shaders been developed for each separate platform...pending further discussion.

HansKristian-ARM:
So, the original implementation also had certain shaders which refused to work properly with optimizations (glslang issues for the most part), for that case I tacked on .noopt. to the file name, but in hindsight with test_shaders.sh/update_test_shaders.sh now, it probably makes more sense to move those shaders to special folders as well. I don't mind that approach.

For dealing with OSX vs iOS specific shaders, I think it makes sense to use file names. I assume it will only be a handful shaders which should be tested specifically for iOS vs. OSX. Another route is to run iOS and OSX testing for all the MSL shaders, and opt out as necessary, either is fine with me.

HansKristian-Work:
FWIW, I think this is resolved long ago.

PR #373, opened 2017-12-26
CompilerMSL enhancements PLUS refactoring of MSL tests to avoid optimization crashes

billhollings:
This PR includes several MSL enhancements to support platforms and nested function arguments.

It also includes a refactoring of some of the MSL test shaders to run some of them without optimization. This is to partly address the requests in Issue #372.

SPIR-V optimizations is problematic for the reference shaders:

1. Optimization results change frequently...changing the reference shader output.

2. Many shaders cause the SPIR-V optimizer and validator to crash.

3. Some optimizations are not correct (eg. - `while(true)` instead of `for` loops).

As a temporary solution, to avoid slowing down SPIRV-Cross development, the problematic MSL shaders have been moved to the new `shaders-msl-no-opts` folder.

MSL shaders in the `shaders-msl` are run both with and without optimizations. MSL shaders in the `shaders-msl-no-opts` folder are only run WITHOUT SPIR-V optimizations.

There are similar problems with GLSL and HLSL shader optimizations...but I have not attempted to fix this. I expect the GitHub CI tests to fail on this PR as a result.

These optimization issues arose by via the latest SPIRV-Tools version.

I think a partial longer-term solution might include establishing SPIRV-Tools as a submodule of SPIRV-Cross...so that a given SPIRV-Cross commit would use a consistent SPIRV-Tools commit for optimization...ensuring that once working, the optimizations would remain unchanged.


HansKristian-ARM:
For glslang/SPIRV-Tools, check checkout_glslang_spirv_tools.sh, this checks out a fixed version and builds it locally.

spirv_glsl.cpp
```diff
@@ -5651,9 +5654,10 @@ void CompilerGLSL::emit_instruction(const Instruction &instruction)
 	case OpVectorTimesMatrix:
 	case OpMatrixTimesVector:
 	{
-		// If the matrix needs transpose, just flip the multiply order.
+		// If the matrix needs transpose and it is square, just flip the multiply order.
+		SPIRType *t;
```
HansKristian-ARM: Should probably just assign t here, I feel it's a bit sketchky to assign in the branch.

spirv_glsl.hpp
```diff
@@ -90,6 +90,13 @@ class CompilerGLSL : public Compiler
 		// If disabled on older targets, binding decorations will be stripped.
 		bool enable_420pack_extension = true;
 
+		// If true, all two-vector OpVectorShuffles will be forced to use a temporary variable.
```
HansKristian-ARM: This is a bit awkward to make into an option, this should just be forced.

spirv_msl.cpp
```diff
@@ -2229,6 +2278,82 @@ string CompilerMSL::to_sampler_expression(uint32_t id)
 	return samp_id ? to_expression(samp_id) : to_expression(id) + sampler_name_suffix;
 }
 
+// Checks whether the ID is a row_major matrix that requires conversion before use
+bool CompilerMSL::is_non_native_row_major_matrix(uint32_t id)
+{
+	// Natively supported row-major matrices do not need to be converted.
+	// Legacy targets do not support row major.
+	if (backend.native_row_major_matrix && !is_legacy())
```
HansKristian-ARM: is_legacy doesn't really make sense on MSL? it's more of a GLSL/HLSL thing.

spirv_msl.cpp
```diff
@@ -2229,6 +2278,82 @@ string CompilerMSL::to_sampler_expression(uint32_t id)
...
+	const auto type = expression_type(id);
+	add_convert_row_major_matrix_function(type.columns, type.vecsize);
+	return true;
+}
+
+// Checks whether the member is a row_major matrix that requires conversion before use
+bool CompilerMSL::member_is_non_native_row_major_matrix(const SPIRType &type, uint32_t index)
+{
+	// Natively supported row-major matrices do not need to be converted.
+	if (backend.native_row_major_matrix && !is_legacy())
```
HansKristian-ARM: is_legacy.

spirv_msl.cpp
```diff
@@ -2599,7 +2722,7 @@ string CompilerMSL::entry_point_args(bool append_comma)
 			{
 				if (!ep_args.empty())
 					ep_args += ", ";
-				BuiltIn bi_type = meta[var_id].decoration.builtin_type;
+				BuiltIn bi_type = (BuiltIn)get_decoration(var_id, DecorationBuiltIn);
```
HansKristian-ARM: This is a bit awkward with the casting back to an enum which meta[] does not need.

spirv_msl.cpp
```diff
@@ -3298,9 +3424,44 @@ bool CompilerMSL::OpCodePreprocessor::handle(Op opcode, const uint32_t *args, ui
...
-	if (length > 1)
-		result_types[args[1]] = args[0];
+	// If it has one, keep track of the instruction's result type, mapped by ID
+	switch (opcode)
+	{
+	case OpStore:
+	case OpCopyMemory:
+	case OpCopyMemorySized:
+	case OpImageWrite:
+	case OpLoopMerge:
```
HansKristian-ARM: Control flow related op-codes are unnecessary, like LoopMerge, SelectionMerge, Label, Branch*, Switch

spirv_msl.hpp
```diff
@@ -77,10 +77,25 @@ class CompilerMSL : public CompilerGLSL
...
+		Platform platform = macOS;
 		uint32_t msl_version = make_msl_version(1, 2);
 		bool enable_point_size_builtin = true;
 		bool resolve_specialized_array_lengths = true;
 
+		bool is_ios()
+		{
+			return platform == iOS;
+		}
+		bool is_macos()
```
HansKristian-ARM: Empty line between is_ios(), is_macos().

spirv_glsl.cpp
billhollings: Done.

spirv_glsl.hpp
billhollings: Always forcing makes sense for MSL. But maybe not GLSL and HLSL.

I've moved it to a backend config selection.

spirv_msl.cpp
billhollings: Copy and pasted code. Good catch. Removed now.

spirv_msl.cpp
billhollings: Copy and pasted code. Good catch. Removed now.

spirv_msl.cpp
billhollings: Reverted.

spirv_msl.hpp
billhollings: Done.

spirv_msl.cpp
billhollings: Removed them.

spirv_msl.cpp
billhollings: Removed them.

billhollings:
I've update code as per the review. Thanks.

I'm now using the `checkout_glslang_spirv_tools.sh` script to lock in the correct `SPIRV-Tools` and `glslang` versions. That has fixed all but one of the shaders that were breaking under heavy `SPRIV-Tools` optimization.

I've left that one shader in `shaders-msl-no-opt`folder so that it is not run with optimizations. I will address why it is broken at a later date. It is a complicated nested function shader.

PR #376, opened 2017-12-31
better? type handling when initial parsing OpStore by MSL generator

fatkas:
We're trying to add SPIRV-Tools optimizer to our shader pipeline and MSL generator is failing to parse optimized SPIRV which seems pretty legit (https://gist.github.com/fatkas/2337cb50ed713c2586d949e6dec651c7) . The following diff seems to workaround the issue but i'm not sure if it makes sense

### DIFF
--- a/spirv_msl.cpp
+++ b/spirv_msl.cpp
@@ -3319,9 +3319,12 @@ CompilerMSL::SPVFuncImpl CompilerMSL::OpCodePreprocessor::get_spv_func_impl(Op o
 		// we must extract the result type directly from the Instruction, rather than the ID.
 		uint32_t id_rhs = args[1];
 		uint32_t type_id_rhs = result_types[id_rhs];
-		if ((compiler.ids[id_rhs].get_type() != TypeConstant) && type_id_rhs &&
-		    compiler.is_array(compiler.get<SPIRType>(type_id_rhs)))
-			return SPVFuncImplArrayCopy;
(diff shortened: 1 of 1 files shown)


HansKristian-ARM:
Looking at the code, I think the really brittle part is:
```
	// Keep track of the instruction return types, mapped by ID
	if (length > 1)
		result_types[args[1]] = args[0];
```
I wouldn't be surprised if this is incorrect in some cases, we should rather try to fix that.
```
diff --git a/spirv_msl.cpp b/spirv_msl.cpp
index 087c415..ecd0e41 100644
--- a/spirv_msl.cpp
+++ b/spirv_msl.cpp
@@ -3299,7 +3299,7 @@ bool CompilerMSL::OpCodePreprocessor::handle(Op opcode, const uint32_t *args, ui
        }
 
        // Keep track of the instruction return types, mapped by ID
-       if (length > 1)
+       if (length > 1 && opcode != OpStore)
                result_types[args[1]] = args[0];
 
        return true;
```
is an alternative patch which seems to fix it. I think a proper fix would be to just also track proper AccessChains to be able to use the correct type in OpStore.

billhollings:
PR #373 now includes an fix similar to the suggestion from @HansKristian-ARM, but also adding other Op codes that do not contain result type.

HansKristian-ARM:
373 was merged, so this is fixed as well.

PR #386, opened 2018-01-09
Fix naive barrier handling in GLSL and HLSL and various other fixes

HansKristian-ARM:
glslang recently changed its barrier behavior which exposed faulty implementation in SPIRV-Cross. Overhaul the implementation.

Struct flattening is also changed to be done on instance names rather than block names, spirv-opt now tends to merge blocks together, so relying on block names will be more brittle going forward.

Also some slight cleanups to the clumsy gl_in/gl_out redirection.


reference/shaders-msl/comp/barriers.comp
```diff
@@ -0,0 +1,79 @@
+#pragma clang diagnostic ignored "-Wmissing-prototypes"
```
HansKristian-ARM: @billhollings I included this, but this output seems wrong to me. memoryBarrierBuffer() is wrong here. Looking at the implementation in spirv_msl.cpp it doesn't seem to take into account that OpControlBarrier has its own memory barriers as well (typically Workgroup).

HansKristian-ARM:
@rossy Rewrote the barrier handling for GLSL/HLSL, I assume this PR works for you now?

PR #389, opened 2018-01-12
Analyze the CFG for temporaries as well.

### LINKED ISSUE #384: Invalid MSL generated for an optimized SPIRV shader
Several of our more complex shaders convert to invalid MSL that can't be consumed by the MSL frontend. The shaders go through HLSL->glslang compilation step w/optimization; the output SPIRV is valid or at least the validator doesn't complain about it :)

The specific error is that the identifier is declared in one `do {} while (false)` block, and is used in another `do {} while (false)` block where it's not declared so it can't be found. This might be a recent regression - I haven't seen this before and we did test the optimized SPIRV output for the offending shader at one point, IIRC.

Offending MSL fragment; note that `_1124` is declared in one `do` block and used in another one.

```
    do
    {
        bool _1124 = 0.0 == 0.0;
        if (_1124)
        {
            _1132 = DiffuseMapTexture.sample(DiffuseMapSampler, _1024);
            _1517 = _1132;
            break;
        }
        else
        {
            float _1135 = 1.0 / (1.0 - 0.0);
            _1156 = mix(DiffuseMapTexture.sample(DiffuseMapSampler, (_1024 * 0.25)), DiffuseMapTexture.sample(DiffuseMapSampler, _1024), float4(clamp((clamp(1.0 - (_1396.View_Depth.w * 0.0033333334140479564666748046875), 0.0, 1.0) * _1135) - (0.0 * _1135), 0.0, 1.0)));
            _1517 = _1156;
            break;
        }
        _1517 = _1157;
        break;
    } while (false);
    float4 _1523;
    do
    {
        if (_1124)
        {
            _1178 = NormalMapTexture.sample(NormalMapSampler, _1024);
            _1523 = _1178;
            break;
        }
        else
        {
            float _1181 = 1.0 / (1.0 - 0.0);
            _1202 = mix(NormalMapTexture.sample(NormalMapSampler, (_1024 * 0.25)), NormalMapTexture.sample(NormalMapSampler, _1024), float4(clamp((_990 * _1181) - (0.0 * _1181), 0.0, 1.0)));
            _1523 = _1202;
            break;
        }
        _1523 = _1203;
        break;
    } while (false);
```

Also attaching complete SPIRV input and MSL output.

[shaders.zip](https://github.com/KhronosGroup/SPIRV-Cross/files/1612406/shaders.zip)

HansKristian-ARM:
Normally, temporary declaration must dominate any use of it,
so we generally did not need to analyze the CFG for these variables,
but there is an edge case where you have an inliner doing:

```
do {
	create_temporary;
	break;
} while(0);

use_temporary;
```

The inside of the loop dominates the outer scope, but we cannot emit
code like this, so make sure we hoist these temporaries outside
the "loop".

Fixes #384.

### DIFF
--- a/reference/opt/shaders-msl/asm/frag/inliner-dominator-inside-loop.asm.frag
+++ b/reference/opt/shaders-msl/asm/frag/inliner-dominator-inside-loop.asm.frag
@@ -1,0 +1,249 @@
+#include <metal_stdlib>
+#include <simd/simd.h>
+
+using namespace metal;
+
+struct VertexOutput
+{
+    float4 HPosition;
+    float4 Uv_EdgeDistance1;
+    float4 UvStuds_EdgeDistance2;
+    float4 Color;
+    float4 LightPosition_Fog;
+    float4 View_Depth;
+    float4 Normal_SpecPower;
+    float3 Tangent;
+    float4 PosLightSpace_Reflectance;
+    float studIndex;
+};
+
+struct Surface
+{
+    float3 albedo;
+    float3 normal;
+    float specular;
+    float gloss;
+    float reflectance;
+    float opacity;
+};
+
+struct SurfaceInput
+{
+    float4 Color;
+    float2 Uv;
+    float2 UvStuds;
+};
+
+struct Globals
+{
+    float4x4 ViewProjection;
+    float4 ViewRight;
+    float4 ViewUp;
+    float4 ViewDir;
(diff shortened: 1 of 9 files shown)


HansKristian-ARM:
@zeux @billhollings I assume this fixes your issues, I added the shader in assembly form into the test suite.

PR #398, opened 2018-01-16
Do not use inline for-loop initializers with different types.

### LINKED ISSUE #396: Implicit conversion compile errors
I have another error now. Spirv-cross is generating shaders that declare variables that do implicit conversions. Glsl compilers seem less than happy to handle these automatically. I need to upload the actual SPIR-V as I can't find a simple repro after optimization (this is using -O and latest spirv-tools and spirv-cross)

[ImplicitConversions_SPV.TXT](https://github.com/KhronosGroup/SPIRV-Cross/files/1631716/ImplicitConversions_SPV.TXT)
[ImplicitConversions_Readable.TXT](https://github.com/KhronosGroup/SPIRV-Cross/files/1631717/ImplicitConversions_Readable.TXT)
[ImplicitConversions_GLSL.TXT](https://github.com/KhronosGroup/SPIRV-Cross/files/1631749/ImplicitConversions_GLSL.TXT)

The result from this is lines of code in glsl like the following:

```glsl
// bool to float
// line 266
float _5226 = 0.0, _5227 = false
```

where 5227 is a float.

or

```glsl
// float to bool
// line 266
_5229 = _5227;
```

where 5229 is a bool and 5227 is the previous float. I'm also getting a float to int here

```glsl
// float to int
// line 323
_3346 = _5241 + 1
```

presumably because _5241 is a float, the addition promotes it to float, and then an implicit conversion happens.

HansKristian-ARM:
Fixes #396.

### DIFF
--- a/reference/opt/shaders/asm/frag/multi-for-loop-init.asm.frag
+++ b/reference/opt/shaders/asm/frag/multi-for-loop-init.asm.frag
@@ -1,0 +1,23 @@
+#version 310 es
+precision mediump float;
+precision highp int;
+
+layout(location = 0) out vec4 FragColor;
+layout(location = 0) flat in mediump int counter;
+
+void main()
+{
+    int _37;
+    mediump int _22;
+    uint _38;
+    mediump uint _25;
+    FragColor = vec4(0.0);
+    int _53 = 0;
+    uint _54 = 1u;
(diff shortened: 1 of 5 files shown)


HansKristian-ARM:
Linux passed, but waiting time for OSX is too long.

PR #399, opened 2018-01-16
Specialize CFG traversal for some opcodes which use literals.

HansKristian-ARM:


HansKristian-ARM:
Linux passed, but OSX is really taking its time, merging.

PR #406, opened 2018-01-23
Fix HLSL regression with struct declaration.

### LINKED ISSUE #405: [HLSL] problem with struct initialization
Hi, after having updated SPIRV-Cross with the latest version, my generated HLSL outputs has been changed from this syntax:

```
struct aStruct
{
    float f;
    int i;
};

aStruct aFunction()
{
    aStruct res = { 0.0, 0 };
    return res;
}

```
To this one:
```
struct aStruct
{
    float f;
    int i;
};

aStruct aFunction()
{
    aStruct res = aStruct{ 0.0, 0 };
    return res;
}
```

Problem is that the HLSL compiler now returns an error with the new syntax:
error X3000: syntax error: unexpected token '{'

HansKristian-ARM:
It actually worked surprisingly. Fix it properly.

Fix #405.

### DIFF
--- a/reference/opt/shaders-hlsl/vert/struct-composite-decl.vert
+++ b/reference/opt/shaders-hlsl/vert/struct-composite-decl.vert
@@ -1,0 +1,44 @@
+struct VOut
+{
+    float4 a;
+    float4 b;
+    float4 c;
+    float4 d;
+};
+
+static VOut vout;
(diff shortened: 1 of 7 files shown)


HansKristian-ARM:
Expedite. Travis OSX is very slow and no significant OSX changes.

Issue #382, opened 2018-01-04
Metal shared memory emit before OpFunctionCall


JJoosten:
We found that when using shared memory in a function, the msl compiler forgets to emit the threadgroup variable(s) before the function call.

This is the repro case that we have used to validate the issue:

**GLSL Input**
```
#version 450
shared int foo[1337];

void myfunc()
{
	foo[0]=13;
}

void main()
{
	myfunc();
}
```

**Metal Output**
```#pragma clang diagnostic ignored "-Wmissing-prototypes"

#include <metal_stdlib>
#include <simd/simd.h>

using namespace metal;

void myfunc(thread int* foo)
{
    foo[0] = 13;
}

kernel void main0()
{
    myfunc(foo);
}
```

HansKristian-ARM:
Ping @billhollings.

billhollings:
Fixed in PR #411.

JJoosten:
Hey everyone,
Sorry for the late reply. Thanks for the fix; have not yet been able to verify that it works since we are running into some other stuff. But hopefully will be able to, later this week!
Cheers and keep up the awesome work!

PR #411, opened 2018-01-24
Fix for #382

billhollings:
CompilerMSL declare threadgroup variables accessed in called functions.

Passes all tests in `test_shaders.sh`


HansKristian-ARM:
Passed Linux, and I assume you've tested on OSX, so merging.

PR #413, opened 2018-01-26
MSL: Order resources by type and binding index in the output

zeux:
We've hit a bizarre bug on NVidia / macOS 10.13 where if two subsequent draw
calls use two different shaders that both have VS use buffers 0 & 1, but one
declares them in the increasing binding order and another one declares them
in the decreasing binding order, then the second draw call (with the decreasing
order) doesn't get correct data in some cases.

This has been reported to Apple and they will probably fix it at some point;
to work around that it's sufficient to sort resources by their binding index.

For output consistency we also sort by type to get a stable order, and output builtins
after that.


zeux:
Interesting - results of torture-loop.comp seem to disagree between Mac & Linux (I initially thought that it's an issue with my local build which was done on Mac, but it looks like an issue that was introduced before this PR - might indicate non-deterministic traversal/insertion order in some recent change?

zeux:
I'm not super happy about texture_buffer test changing behavior; but also it seems like SampledImage handling isn't very nice atm - in case auto-assignment is used, it may distribute texture & sampler to different registers, which feels counter-intuitive for higher level code to handle. I will update the change to assign *one* register to SampledImage which should also fix the behavior change...

zeux:
Okay, this should be ready for review (torture-loop.comp is still an issue though).

HansKristian-ARM:
As a workaround, this approach seems reasonable to me.
I think I know why the OSX output has a mismatch, I'll try to fix that so you can rebase against that. @billhollings should also have a look.

HansKristian-ARM:
@zeux, torture loop passes on OSX now.

zeux:
Thanks - rebased and formatted.

billhollings:
@HansKristian-ARM 

Yes...looks good to me...and fixes an issue I was encountering on 10.13 Nvidia but hadn't looked into yet.

It's good to go.

danginsburg:
@billhollings pointed me to this change because of a problem I'd been debugging on NVIDIA GT 650M macOS 10.13.2 where a constant buffer had incorrect values.  This patch fixed it.  Thank you @zeux!

PR #417, opened 2018-01-29
hlsl: Support custom root constant layout

msiglreith:
That's a feature we use for our dx12 backend in gfx-rs. Due to the definition of push constants in Vulkan, we need to split push constant blocks if they are (partially) used in multiple shader stages. We divide them into multiple constant buffers and define the root signature layout accordingly. Therefore, we need to be able to mirror this splitting on the shader side.

It's purely optional and shouldn't affect other applications.

### DIFF
--- a/spirv_glsl.cpp
+++ b/spirv_glsl.cpp
@@ -1040,7 +1040,8 @@ uint32_t CompilerGLSL::type_to_packed_size(const SPIRType &type, uint64_t flags,
 	return size;
 }
 
-bool CompilerGLSL::buffer_is_packing_standard(const SPIRType &type, BufferPackingStandard packing)
+bool CompilerGLSL::buffer_is_packing_standard(const SPIRType &type, BufferPackingStandard packing,
+                                              uint32_t start_offset, uint32_t end_offset)
 {
 	// This is very tricky and error prone, but try to be exhaustive and correct here.
 	// SPIR-V doesn't directly say if we're using std430 or std140.
@@ -1079,6 +1080,10 @@ bool CompilerGLSL::buffer_is_packing_standard(const SPIRType &type, BufferPackin
 		uint32_t alignment = max(packed_alignment, pad_alignment);
 		offset = (offset + alignment - 1) & ~(alignment - 1);
 
+		// Field is not in the specified range anymore and we can ignore any further fields.
+		if (offset >= end_offset)
+			break;
+
 		// The next member following a struct member is aligned to the base alignment of the struct that came before.
 		// GL 4.5 spec, 7.6.2.2.
 		if (memb_type.basetype == SPIRType::Struct)
@@ -1086,26 +1091,30 @@ bool CompilerGLSL::buffer_is_packing_standard(const SPIRType &type, BufferPackin
 		else
 			pad_alignment = 1;
 
-		// We only care about offsets in std140, std430, etc ...
-		// For EnhancedLayout variants, we have the flexibility to choose our own offsets.
-		if (!packing_has_flexible_offset(packing))
+		// Only care about packing if we are in the given range
+		if (offset >= start_offset)
 		{
-			uint32_t actual_offset = type_struct_member_offset(type, i);
-			if (actual_offset != offset) // This cannot be the packing we're looking for.
+			// We only care about offsets in std140, std430, etc ...
+			// For EnhancedLayout variants, we have the flexibility to choose our own offsets.
+			if (!packing_has_flexible_offset(packing))
+			{
+				uint32_t actual_offset = type_struct_member_offset(type, i);
+				if (actual_offset != offset) // This cannot be the packing we're looking for.
+					return false;
+			}
+
+			// Verify array stride rules.
+			if (!memb_type.array.empty() && type_to_packed_array_stride(memb_type, member_flags, packing) !=
+			                                    type_struct_member_array_stride(type, i))
+				return false;
+
+			// Verify that sub-structs also follow packing rules.
+			// We cannot use enhanced layouts on substructs, so they better be up to spec.
+			auto substruct_packing = packing_to_substruct_packing(packing);
+
+			if (!memb_type.member_types.empty() && !buffer_is_packing_standard(memb_type, substruct_packing))
 				return false;
 		}
-
-		// Verify array stride rules.
-		if (!memb_type.array.empty() &&
-		    type_to_packed_array_stride(memb_type, member_flags, packing) != type_struct_member_array_stride(type, i))
-			return false;
-
-		// Verify that sub-structs also follow packing rules.
-		// We cannot use enhanced layouts on substructs, so they better be up to spec.
-		auto substruct_packing = packing_to_substruct_packing(packing);
-
-		if (!memb_type.member_types.empty() && !buffer_is_packing_standard(memb_type, substruct_packing))
-			return false;
 
 		// Bump size.
 		offset += packed_size;
@@ -7066,7 +7075,7 @@ string CompilerGLSL::variable_decl(const SPIRType &type, const string &name, uin
 // Emit a structure member. Subclasses may override to modify output,
 // or to dynamically add a padding member if needed.
 void CompilerGLSL::emit_struct_member(const SPIRType &type, uint32_t member_type_id, uint32_t index,
-                                      const string &qualifier)
+                                      const string &qualifier, uint32_t)
 {
 	auto &membertype = get<SPIRType>(member_type_id);
 
--- a/spirv_glsl.hpp
+++ b/spirv_glsl.hpp
@@ -202,7 +202,7 @@ protected:
 	virtual std::string type_to_glsl(const SPIRType &type, uint32_t id = 0);
 	virtual std::string builtin_to_glsl(spv::BuiltIn builtin, spv::StorageClass storage);
 	virtual void emit_struct_member(const SPIRType &type, uint32_t member_type_id, uint32_t index,
-	                                const std::string &qualifier = "");
+	                                const std::string &qualifier = "", uint32_t base_offset = 0);
 	virtual std::string image_type_glsl(const SPIRType &type, uint32_t id = 0);
 	virtual std::string constant_expression(const SPIRConstant &c);
 	std::string constant_op_expression(const SPIRConstantOp &cop);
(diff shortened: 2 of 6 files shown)


spirv_hlsl.cpp
```diff
@@ -19,6 +19,8 @@
 #include <algorithm>
 #include <assert.h>
 
+#include <iostream>
```
HansKristian-ARM: ?

spirv_msl.cpp
```diff
@@ -2394,7 +2394,7 @@ void CompilerMSL::emit_fixup()
 
 // Emit a structure member, padding and packing to maintain the correct memeber alignments.
 void CompilerMSL::emit_struct_member(const SPIRType &type, uint32_t member_type_id, uint32_t index,
-                                     const string &qualifier)
+                                     const string &qualifier, uint32_t base_offset)
```
HansKristian-ARM: Unused variable, remove base_offset.

spirv_hlsl.hpp
```diff
@@ -30,6 +30,11 @@ struct HLSLVertexAttributeRemap
 	std::string semantic;
 };
 
+struct RootConstants {
```
HansKristian-ARM: Does this map to a VkPushConstantRange?

spirv_hlsl.cpp
```diff
@@ -1508,7 +1512,45 @@ void CompilerHLSL::emit_buffer_block(const SPIRVariable &var)
...
+	}
+	else
+	{
+		for(auto const& layout : options.root_constants_layout)
+		{
+			auto &type = get<SPIRType>(var.basetype);
+
+			flattened_structs.insert(var.self);
+			type.member_name_cache.clear();
+			add_resource_name(var.self);
```
HansKristian-ARM: If multiple root constant layout ranges are used, the name of var will get out of sync, probably best to call the cbuffer something fixed like SPIRV_Cross_RootConstants(Number) or something.

spirv_hlsl.cpp
```diff
@@ -1508,7 +1512,45 @@ void CompilerHLSL::emit_buffer_block(const SPIRVariable &var)
...
+			auto &memb = meta[type.self].members;
+
+			statement("cbuffer ", to_name(var.self), ": register(b", layout.start / 4, ", space0)");
```
HansKristian-ARM: Is this binding assignment arbitrary? It probably makes sense to add explicit namespace/binding for the RootConstant struct?

spirv_hlsl.cpp
```diff
@@ -1508,7 +1512,45 @@ void CompilerHLSL::emit_buffer_block(const SPIRVariable &var)
...
+				uint32_t i = 0;
+				for (auto &member : type.member_types)
+				{
+					if (memb[i].offset == offset)
+					{
+						add_member_name(type, i);
+						auto backup_name = get_member_name(type.self, i);
+						auto member_name = to_member_name(type, i);
+						set_member_name(type.self, i, sanitize_underscores(join(to_name(type.self), "_", member_name)));
+						emit_struct_member(type, member, i, "", layout.start);
```
HansKristian-ARM: How does this deal with buffer blocks with "odd" layouts where manual packoffset() must be used?

spirv_hlsl.hpp
```diff
@@ -39,6 +44,8 @@ class CompilerHLSL : public CompilerGLSL
 
 		// Allows the PointSize builtin, and ignores it, as PointSize is not supported in HLSL.
 		bool point_size_compat = false;
+
+		std::vector<RootConstants> root_constants_layout = {};
```
HansKristian-ARM: It's probably best to pass this as a separate function call rather than through options.

spirv_glsl.cpp
```diff
@@ -7059,7 +7059,7 @@ string CompilerGLSL::variable_decl(const SPIRType &type, const string &name, uin
 // Emit a structure member. Subclasses may override to modify output,
 // or to dynamically add a padding member if needed.
 void CompilerGLSL::emit_struct_member(const SPIRType &type, uint32_t member_type_id, uint32_t index,
-                                      const string &qualifier)
+                                      const string &qualifier, uint32_t base_offset)
```
HansKristian-ARM: Unused variable.

spirv_hlsl.cpp
```diff
@@ -1508,7 +1512,45 @@ void CompilerHLSL::emit_buffer_block(const SPIRVariable &var)
 
 void CompilerHLSL::emit_push_constant_block(const SPIRVariable &var)
 {
-	emit_buffer_block(var);
+	std::cout << options.root_constants_layout.size() << std::endl;
+	if (options.root_constants_layout.empty())
+	{
+		emit_buffer_block(var);
+		return;
```
HansKristian-ARM: return is redundant.

spirv_hlsl.cpp
```diff
@@ -1508,7 +1512,45 @@ void CompilerHLSL::emit_buffer_block(const SPIRVariable &var)
 
 void CompilerHLSL::emit_push_constant_block(const SPIRVariable &var)
 {
-	emit_buffer_block(var);
+	std::cout << options.root_constants_layout.size() << std::endl;
```
HansKristian-ARM: Old debug code?

spirv_hlsl.cpp
HansKristian-ARM: It might be better to add some kind of begin/end range support to the already implemented emit_buffer_block() functions to reduce maintenance.

msiglreith: > Is this binding assignment arbitrary? It probably makes sense to add explicit namespace/binding for the RootConstant struct?

right, that's quite tailored towards our specific root signature layout (space0, binding equals to start of range). I'm going to make it a bit more generic.

spirv_hlsl.cpp
msiglreith: Could you give an example for these odd layouts? It might not apply for push constants but I'm not sure atm.

spirv_hlsl.hpp
msiglreith: On constructor call, overloading `compile` or do you have something else in mind?

spirv_hlsl.cpp
HansKristian-ARM: A classic example here is:
```
push_constant {
   float foo; // Offset 0
   vec3 bar; // Offset 16 in GLSL, 4 in HLSL
}
```
The main implementation of buffer blocks validates the buffer blocks for offsets and uses manual layout(offset) or packoffset() as necessary.

There are also cases where packoffset is impossible, for example substructs within a cbuffer. If there is a mismatch between the SPIR-V and equivalent HLSL packing for that substruct, an exception is thrown.

spirv_hlsl.hpp
HansKristian-ARM: A separate `void set_root_constant_layouts()` is probably fine, which can be called before compile().

spirv_hlsl.hpp
```diff
@@ -72,6 +70,15 @@ class CompilerHLSL : public CompilerGLSL
 		options = opts;
 	}
 
+	// Optionally specify a custom root constant layout.
+	//
+	// Push constants ranges will be split up according to the
+	// layout specified.
+	void set_root_constant_layouts(std::vector<RootConstants> layout)
+	{
+		root_constants_layout = layout;
```
HansKristian-ARM: std::move()

spirv_hlsl.cpp
msiglreith: Thanks for bringing this point up! Played a bit around with it. I works depending on the specified root constant ranges. There are certain cases where I see no way to work around it, but I need to double-check it.
For example: 
```
push_constant {
   float foo; // VERTEX shader only, other fields VERTEX and FRAGMENT shader

   float foo1;
   float foo2;
   float foo3;
   vec3 bar; // Misalignment here
}
```
Splitting after the first float will result in misaligment of `bar` in the fragment shader.
Splitting at offset 16 in the example you mentioned would be fine on the other hand.

spirv_hlsl.cpp
msiglreith: I used a numbering scheme similar to the one proposed, just wondering if just renaming the cbuffer is fine as currently done by me?
It will emit something like
```
cbuffer SPIRV_CROSS_RootConstant_u_foo : register(b0, space0)
{
    float foo_a : packoffset(c0);
};
cbuffer SPIRV_CROSS_RootConstant_u_foo_1 : register(b1, space0)
{
    float3 foo_b : packoffset(c0.w); // <- TODO: offset
};
```

HansKristian-ARM:
Some of the reference files are failing, you might want to try rebasing: https://travis-ci.org/KhronosGroup/SPIRV-Cross/jobs/336169533

spirv_glsl.hpp
```diff
@@ -424,7 +424,8 @@ class CompilerGLSL : public Compiler
 	std::string to_combined_image_sampler(uint32_t image_id, uint32_t samp_id);
 	virtual bool skip_argument(uint32_t id) const;
 
-	bool buffer_is_packing_standard(const SPIRType &type, BufferPackingStandard packing);
+	bool buffer_is_packing_standard(const SPIRType &type, BufferPackingStandard packing, uint32_t start_offset = 0,
+	                                uint32_t end_offset = -1);
```
HansKristian-ARM: -1u should be ~(0u) or something to avoid implicit cast warnings.

spirv_hlsl.cpp
```diff
@@ -1557,7 +1561,50 @@ void CompilerHLSL::emit_buffer_block(const SPIRVariable &var)
...
 void CompilerHLSL::emit_push_constant_block(const SPIRVariable &var)
 {
-	emit_buffer_block(var);
+	if (root_constants_layout.empty())
+	{
+		emit_buffer_block(var);
+	}
+	else
+	{
+		for (auto const &layout : root_constants_layout)
```
HansKristian-ARM: `const auto &`

spirv_hlsl.cpp
```diff
@@ -2317,23 +2364,23 @@ string CompilerHLSL::to_resource_binding(const SPIRVariable &var)
 		return "";
 
 	auto &type = get<SPIRType>(var.basetype);
-	const char *space = nullptr;
+	char space = 0;
```
HansKristian-ARM: `\0`

spirv_hlsl.cpp
```diff
@@ -1557,7 +1561,50 @@ void CompilerHLSL::emit_buffer_block(const SPIRVariable &var)
...
+
+			flattened_structs.insert(var.self);
+			type.member_name_cache.clear();
+			add_resource_name(var.self);
+			auto &memb = meta[type.self].members;
+
+			statement("cbuffer SPIRV_CROSS_RootConstant_", to_name(var.self),
+			          to_resource_register('b', layout.binding, layout.space));
+			begin_scope();
+			for (auto offset = layout.start; offset < layout.end; offset += 4)
```
HansKristian-ARM: This seems like an unnecessary O(n^2) way of emitting things. The offsets are in order, so I think it might make more sense to find first/last member and loop over that instead.

msiglreith:
Thanks for the review!
Rebased and addressed the mentioned points. I can squash the commits if you are ok with the changes.

HansKristian-ARM:
Ok, please squash then.

HansKristian-ARM:
yup, you'll need <limits>, build is failing on Linux.

PR #427, opened 2018-02-02
add `matrix` to the list of illegal names.

aliaspider:


pmuetschard:
May I ask why? "matrix" is not a reserved keyword in glsl...

aliaspider:
`CompilerHLSL` inherits `replace_illegal_names` from `CompilerGLSL`

billhollings:
Should we think about separating the lists per backend via a virtual population function?

`CompilerGLSL` could create the base list...and then each subclass could add or subtract specific platform requirements.

HansKristian-ARM:
We certainly could if people feel it's important. PR welcome.

pmuetschard:
#565 PR delivered

PR #433, opened 2018-02-05
Support returning arrays from functions in GLSL/MSL, refactor array copies in MSL.

### LINKED ISSUE #428: Incorrect HLSL generated for array return type
I have some GLSL with an array return type (which I believe is valid):
```glsl
#version 150

vec4[2] test() {
  return vec4[2](vec4(1.0, 2.0, 3.0, 4.0), vec4(5.0, 6.0, 7.0, 8.0));
}

void main() {
  gl_Position = test()[0];
}
```
When running `glslangValidator.exe test.vert -V && SPIRV-Cross vert.spv --hlsl`, I receive:
```hlsl
uniform float4 gl_HalfPixel;

static float4 gl_Position;
struct SPIRV_Cross_Output
{
    float4 gl_Position : POSITION;
};

float4 test()
{
    return { float4(1.0f, 2.0f, 3.0f, 4.0f), float4(5.0f, 6.0f, 7.0f, 8.0f) };
}

void vert_main()
{
    gl_Position = test()[0];
    gl_Position.x = gl_Position.x - gl_HalfPixel.x * gl_Position.w;
    gl_Position.y = gl_Position.y + gl_HalfPixel.y * gl_Position.w;
}

SPIRV_Cross_Output main()
{
    vert_main();
    SPIRV_Cross_Output stage_output;
    stage_output.gl_Position = gl_Position;
    return stage_output;
}
```
This fails to compile because of the `float4` return type.

HansKristian-ARM:
Not possible in HLSL apparently, need workaround ...
... And neither can MSL.

The workaround for returned arrays is to inject a magic "out" argument to functions which have this return type.

For HLSL, we can directly assign the array, while in MSL, we need to inject a spvArrayCopy, because arrays are not value types (sigh).

While testing this I ran into a different issue with MSL where spvArrayCopy would be ignored if the right hand side of an assignment is constant. This however was buggy except in the cases where a variable is declared.

```
float4 array[4] = { ... };
```
is fine, but
```
float4 array[4];
array = { ... };
```
is not.

The workaround here is similar to HLSL, declare arrays as constant globals, then assign using that variable.
MSL arrays cannot be assigned so, we end up with:

```
constant float4 array[4] = { ... };
float4 t[4];
spvArrayCopy(t, array);
```
This will likely break performance of LUTs, so I added a rather ugly special case which is such that if a variable is declared and we assign a constant expression, we can declare the array inline. This leads to some redundancies (same constant is declared both as a constant and declared inline on assignment), but I don't see a better solution which is also robust ... :\
If the MSL compiler is able to optimize the spvArrayCopy away, we can remove this hack.

Fix #428.


HansKristian-ARM:
@billhollings This modifies MSL significantly, so please have a look if this is acceptable.

billhollings:
@HansKristian-ARM  Looks good to me. Thanks for all the effort!

grovesNL:
Thanks very much @HansKristian-ARM!

Issue #436, opened 2018-02-06
RowMajor float4x3 matrix in uniform buffer generates incorrect MSL


danginsburg:
I have a shader that has two RowMajor float4x3 matrices in a uniform buffer:

```
               OpMemberDecorate %_struct_1365 0 RowMajor
               OpMemberDecorate %_struct_1365 0 Offset 0
               OpMemberDecorate %_struct_1365 0 MatrixStride 16
               OpMemberDecorate %_struct_1365 1 RowMajor
               OpMemberDecorate %_struct_1365 1 Offset 48
               OpMemberDecorate %_struct_1365 1 MatrixStride 16
               OpDecorate %_struct_1365 Block
```

The MSL output for this cbuffer is the following (./spirv-cross --output sky_vs.msl sky_vs.spv --msl):

```
struct _1365
{
    float4x3 _m0;
    float4x3 _m1;
};
```

This results in rendering corruption because according to the Metal Shading Language spec Table 5 a float4x3 is 64 bytes whereas these are supposed to be 48 bytes (a float3x4 in MSL).  It seems like spirv-cross would need to generate these as flaot3x4 in the struct definition and then transform them at point of use?  (unless there is some other way to specify row major that I'm not aware of in MSL).

I notice that the GLSL looks like it's probably correct:

```
layout(binding = 13, std140) uniform _1365_18812
{
    layout(row_major) mat4x3 _m0;
    layout(row_major) mat4x3 _m1;
} _18812;
```

The offending shader is attached to the bug.

Seems possibly related to #295.

[sky_vs.zip](https://github.com/KhronosGroup/SPIRV-Cross/files/1699470/sky_vs.zip)

billhollings:
@danginsburg 

As far as I can tell...in SPIR-V, the matrices are 4 columns and 3 rows (4 columns of vec3)...so float4x3 is the correct declaration...but is the wrong memory footprint.

While float3x4 would have the right memory footprint...it wouldn't be the correct shape for use in the shader. We'd have to convert it to float4x3 dynamically in code.

Current conversions between R-Maj & C-Maj are working...but this issue would also require us to declare the wrong matrix shape and then transform it on the fly whenever we used it. We'd have to track something other than whether it was R-Maj or C-Maj.

Still thinking about this.

danginsburg:
[...]

PR #448, opened 2018-02-11
MSL support for smaller offsets for 3-row row-major matrices (fix issue #436).

billhollings:
MSL support for smaller offsets for 3-row row-major matrices (fix issue #436).

### DIFF
--- a/spirv_msl.cpp
+++ b/spirv_msl.cpp
@@ -744,29 +744,9 @@ void CompilerMSL::align_struct(SPIRType &ib_type)
 	curr_offset = 0;
 	for (uint32_t mbr_idx = 0; mbr_idx < mbr_cnt; mbr_idx++)
 	{
-		// Align current offset to the current member's default alignment.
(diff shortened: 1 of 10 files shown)


spirv_msl.cpp
```diff
@@ -858,14 +838,48 @@ void CompilerMSL::align_struct(SPIRType &ib_type)
...
-	if (mbr_type.vecsize == 3 && mbr_type.columns == 1)
+	// We've already marked it as packable
+	if (has_member_decoration(ib_type.self, index, DecorationCPacked))
 		return true;
 
-	return false;
+	auto &mbr_type = get<SPIRType>(ib_type.member_types[index]);
+
+	// Only 3-element vectors or 3-row matrices need to be packed.
+	if (mbr_type.vecsize != 3)
```
HansKristian-ARM: I have a feeling there will be more cases. This does not take into account MatrixStride for example, so matrices with vec2-columns could get interesting ... But for now, I assume this is a good enough solution.

Issue #443, opened 2018-02-09
MSL that writes to SampleMask fails to compile


danginsburg:
See attached SPIR-V.  Repro:

```
./spirv-cross sample_mask.spv --msl --output sample_mask.msl
xcrun --sdk macosx10.13 metal -x metal -std=osx-metal1.2 -Werror -Wno-unused-variable sample_mask.msl
```

Generates the following errors:
```
sample_mask.msl:165:26: error: type 'uint [1]' is not valid for attribute 'sample_mask'
    uint gl_SampleMask [[sample_mask]][1];
                         ^~~~~~~~~~~
sample_mask.msl:665:5: error: use of undeclared identifier 'gl_SampleMask'
    gl_SampleMask[0] = _265;
    ^
2 errors generated.
```
[sample_mask.zip](https://github.com/KhronosGroup/SPIRV-Cross/files/1711844/sample_mask.zip)

billhollings:
I'm looking into this.

billhollings:
@danginsburg 

`sample_mask.zip` seems to only include the MSL. Do you have the SPIR-V?

danginsburg:
My mistake, SPIR-V in the attached file.

[sample_mask_spv.zip](https://github.com/KhronosGroup/SPIRV-Cross/files/1716585/sample_mask_spv.zip)

HansKristian-ARM:
Overall I think it's going to be near impossible to make sure -Wsometimes-uninitialized never triggers with optimized SPIR-V. Is it possible to disable that warning from within the source?

billhollings:
I've confirmed that the warning/error is successfully removed by adding the following pragma line to the MSL...

		#pragma clang diagnostic ignored "-Wsometimes-uninitialized"

However...turning this off for all shaders, regardless of how well-formed they are, feels intrusive and ugly (not to mention will cause every single MSL test reference shader to change).

Could we use one of the following two approaches?

1. Couldn't this be fixed by ensuring that that all MSL declarations are initialized to their default values if not set to something else explicitly? If so...where would be the most appropriate place to handle that?

2. If not...is there a reasonable way to detect when to add the pragma (ie- only when the warning will be triggered by an uninitialized variable), rather than adding it to all MSL shaders?

HansKristian-ARM:
[...]

PR #449, opened 2018-02-13
CompilerMSL support gl_SampleMask and convert it to scalar uint from array.

billhollings:
This is a partial fix for Issue #443. It fixes the original issue reported in the use of `gl_SampleMask.

Issue #443 still requires work that is independent of this PR...and this PR can be added now.


HansKristian-ARM:
Cute solution. :P

PR #450, opened 2018-02-14
Null crash fix in OpCopyMemory.

jodavis42:
I found a crash in OpCopyMemory where it assumed that var wasn't null. This looked like a small oversight as the lines directly above and below do the null check.

I attached a dissassembly file that caused this crash.
[TestDissassembly.txt](https://github.com/KhronosGroup/SPIRV-Cross/files/1725213/TestDissassembly.txt)


HansKristian-ARM:
Indeed, an oversight. Test coverage on OpCopy* isn't that great. Thanks for the fix.

PR #468, opened 2018-02-23
Support special float constants (NaN/Inf).

### LINKED ISSUE #461: special float constants produce invalid GLSL
If a floating point constant is nan, the generated code is invalid. Here's some example output:
```
float _392(vec2 _391)
{
    return _398(_391 * vec2(-nan.0));
}
```

HansKristian-ARM:
Fix #461.

### DIFF
--- a/spirv_glsl.cpp
+++ b/spirv_glsl.cpp
@@ -19,6 +19,7 @@
 #include "spirv_common.hpp"
 #include <algorithm>
 #include <assert.h>
+#include <cmath>
 #include <utility>
 
 using namespace spv;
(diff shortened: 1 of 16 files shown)


spirv_glsl.cpp
```diff
@@ -2529,6 +2531,149 @@ string CompilerGLSL::constant_expression(const SPIRConstant &c)
...
+			res += "f";
+	}
+
+	return res;
+}
+
+std::string CompilerGLSL::convert_double_to_string(const SPIRConstant &c, uint32_t col, uint32_t row)
+{
+	string res;
+	float double_value = c.scalar_f64(col, row);
```
grovesNL: Was the `float` here intentional? It gives a conversion warning on MSVC.

HansKristian-ARM: Not intentional, no, sigh.

PR #210, opened 2017-06-23
added --rename-interface-variable <in|out> <location> <new_variable_name>

Jasonchan35:
according to issue #196 HLSL source to GLSL Vertex/Fragment interface linking error on MacOSX

PR #473, opened 2018-02-24
Move `rename_interface_variable` to util

### LINKED ISSUE #467: Move `rename_interface_variable` into `Compiler`?
Would it be possible to move [`rename_interface_variable` into the reflection API](https://github.com/KhronosGroup/SPIRV-Cross/blob/master/main.cpp#L587-L612) (probably as a method on `Compiler`)?

grovesNL:
Fixes #467.

I noticed that in #210 the choice was made to explicitly move this method into `main.cpp`, but it would be useful to expose as part of the reflection API if possible (I'd prefer not to re-implement it myself).

Let me know if it's still too specific or belongs on another class. Thanks!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -101,12 +101,17 @@ spirv_cross_add_library(spirv-cross-hlsl spirv_cross_hlsl STATIC
     ${CMAKE_CURRENT_SOURCE_DIR}/spirv_hlsl.hpp
     ${CMAKE_CURRENT_SOURCE_DIR}/spirv_hlsl.cpp)
 
+spirv_cross_add_library(spirv-cross-util spirv_cross_util STATIC
+    ${CMAKE_CURRENT_SOURCE_DIR}/spirv_cross_util.hpp
+    ${CMAKE_CURRENT_SOURCE_DIR}/spirv_cross_util.cpp)
+
 add_executable(spirv-cross main.cpp)
 target_compile_options(spirv-cross PRIVATE ${spirv-compiler-options})
 target_compile_definitions(spirv-cross PRIVATE ${spirv-compiler-defines})
 
 install(TARGETS spirv-cross RUNTIME DESTINATION bin)
-target_link_libraries(spirv-cross spirv-cross-glsl spirv-cross-hlsl spirv-cross-cpp spirv-cross-msl spirv-cross-core)
+target_link_libraries(spirv-cross spirv-cross-glsl spirv-cross-hlsl spirv-cross-cpp spirv-cross-msl spirv-cross-util spirv-cross-core)
+target_link_libraries(spirv-cross-util spirv-cross-core)
 target_link_libraries(spirv-cross-glsl spirv-cross-core)
 target_link_libraries(spirv-cross-msl spirv-cross-glsl)
 target_link_libraries(spirv-cross-hlsl spirv-cross-glsl)
--- a/spirv_cross_util.cpp
+++ b/spirv_cross_util.cpp
@@ -1,0 +1,51 @@
+/*
+ * Copyright 2015-2018 ARM Limited
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "spirv_cross_util.hpp"
+#include "spirv_common.hpp"
+
+using namespace spv;
+using namespace spirv_cross;
+
+namespace spirv_cross_util
+{
+void rename_interface_variable(spirv_cross::Compiler &compiler, const std::vector<spirv_cross::Resource> &resources,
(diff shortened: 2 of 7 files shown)


spirv_cross.cpp
```diff
@@ -4149,3 +4149,30 @@ bool Compiler::instruction_to_result_type(uint32_t &result_type, uint32_t &resul
 			return false;
 	}
 }
+
+void Compiler::rename_interface_variable(const std::vector<Resource> &resources,
```
HansKristian-ARM: I'm almost inclined to think this should live in a separate file, e.g. spirv_cross_util.cpp. There is no particular reason why this needs to live in spirv_cross.cpp. It's also loaded a bit too much with "policy", not mechanics.

The arguments are not well suited for a public API. InterfaceVariableRename is an internal structure which species the storage class as well, but that's redundant, because it's implied by the `resources` argument. Something like `const std::vector<Resource> &resources, uint32_t location, const std::string &name)` would work much better.

grovesNL:
Thanks for the feedback! Please take a look at my updated commit and let me know if this is better.

CMakeLists.txt
```diff
@@ -83,7 +83,9 @@ spirv_cross_add_library(spirv-cross-core spirv_cross_core STATIC
     ${CMAKE_CURRENT_SOURCE_DIR}/spirv_cross.hpp
     ${CMAKE_CURRENT_SOURCE_DIR}/spirv_cross.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/spirv_cfg.hpp
-    ${CMAKE_CURRENT_SOURCE_DIR}/spirv_cfg.cpp)
+    ${CMAKE_CURRENT_SOURCE_DIR}/spirv_cfg.cpp
+    ${CMAKE_CURRENT_SOURCE_DIR}/spirv_cross_util.hpp
```
HansKristian-ARM: This should be a separate CMake target. spirv-cross-util or something and the "main" executable target should link to it.

spirv_cross_util.cpp
```diff
@@ -0,0 +1,47 @@
+/*
+* Copyright 2015-2018 ARM Limited
```
HansKristian-ARM: Space before the asterisk.

spirv_cross_util.cpp
```diff
@@ -0,0 +1,47 @@
...
+* limitations under the License.
+*/
+
+#include "spirv_common.hpp"
+#include "spirv_cross_util.hpp"
+
+using namespace spv;
+using namespace spirv_cross;
+
+void spirv_cross_util::rename_interface_variable(spirv_cross::Compiler &compiler, const std::vector<spirv_cross::Resource>& resources, uint32_t location, const std::string &name)
```
HansKristian-ARM:
```
namespace
{
void rename_interface_variable();
}
```

spirv_cross_util.hpp
```diff
@@ -0,0 +1,28 @@
...
+
+#ifndef SPIRV_CROSS_UTIL_HPP
+#define SPIRV_CROSS_UTIL_HPP
+
+//#include "spirv_common.hpp"
+#include "spirv_cross.hpp"
+
+namespace spirv_cross_util
+{
+	void rename_interface_variable(spirv_cross::Compiler &compiler, const std::vector<spirv_cross::Resource>& resources, uint32_t location, const std::string &name);
```
HansKristian-ARM: No indent here. Please use `./format_all.sh` to format this automatically.

spirv_cross_util.hpp
```diff
@@ -0,0 +1,28 @@
...
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+#ifndef SPIRV_CROSS_UTIL_HPP
+#define SPIRV_CROSS_UTIL_HPP
+
+//#include "spirv_common.hpp"
```
HansKristian-ARM: Remove commented include if it's not needed.

spirv_cross_util.hpp
```diff
@@ -0,0 +1,28 @@
+/*
+* Copyright 2015-2018 ARM Limited
```
HansKristian-ARM: Space before asterisk.

CMakeLists.txt
HansKristian-ARM: I'm not seeing an update Makefile?

grovesNL: Sure, I can add the separate target. I'm not sure what I need to do for your second comment though, it seems like the Makefile should be ok?

spirv_cross_util.hpp
grovesNL: I'll fix the indent manually, VS keeps automatically aligning those.

Side note: `./format_all.sh` modifies quite a few files if I use the clang-format included with the latest LLVM 7 Windows binaries. It seems like `FixNamespaceCommentsL false`  is needed, and it realigns curly braces in a couple areas (I think they followed `typedef struct` and `extern "C"`). Is there a certain version I need to use?

grovesNL:
Updated commit to address comments.

Everything seems to work fine when I build through MSVC, `make`, or `cmake .. && make`. I don't use `cmake` or `Makefile` often so let me know if I'm still doing something wrong.

grovesNL:
Not sure what's happening on the Travis OS X build, seems to be unrelated to my changes?

CMakeLists.txt
HansKristian-ARM: Actually, the Makefile is fine. It uses globbing.

HansKristian-ARM:
Travis breakage is unrelated yes, looking into it.

PR #480, opened 2018-03-01
Take execution model into account for entry point methods.

### LINKED ISSUE #476: Ambiguous entry point names in `get_entry_point`
From the SPIR-V specification:

> Name is a name string for the entry point. A module cannot have two OpEntryPoint instructions with the same Execution Model and the same Name string.

In `get_entry_point` we don't specify which execution model to search. So if there is an `OpEntryPoint` with a duplicate name, we may return the incorrect entry point.

For example, when we use `get_cleansed_entry_point_name` we only pass a string for the name. But if the SPIR-V contains two entry points with the same name but different execution models (i.e. both named `main` for vertex and fragment), we have no way to distinguish between which `main` is the entry point we're searching for. I also expect this probably breaks `rename_entry_point` and some other methods.

I'll try to create a sample shader later.

HansKristian-ARM:
SPIR-V allows names to alias if they implement different stages.
Deprecate the old interface and replace it with a new one which takes
execution modes into account.

Fix #476.

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -286,9 +286,9 @@ static void print_resources(const Compiler &compiler, const ShaderResources &res
 	uint64_t modes = compiler.get_execution_mode_mask();
 
 	fprintf(stderr, "Entry points:\n");
-	auto entry_points = compiler.get_entry_points();
+	auto entry_points = compiler.get_entry_points_and_stages();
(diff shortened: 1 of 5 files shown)


grovesNL:
This should work great, thanks!

PR #483, opened 2018-03-05
Overhaul type_alias for MSL.

### LINKED ISSUE #466: MSL : nested cbuffer structs have incorrectly aligned members
basically fix for https://github.com/KhronosGroup/SPIRV-Cross/issues/457 broke packing of members of nested structs. thats a huge regression for us

HansKristian-ARM:
Fix #466.

The current implementation will not prioritize declaring packed structs over their non-packed counter-part. This fix makes it so the block type is always the type master, and multiple structs are declared as necessary if the alias master is marked as packed.

This makes the MSL potentially a bit more verbose, but there's no reasonable way to support local variable variants of block type structs without declaring multiple variants.

Currently MSL will mark every interface struct as packed, which might be a bit overzealous, but that's a cleanliness optimization which can be done later. It only kicks in if you're loading the structs as a struct.

@fatkas @billhollings @JJoosten, please try this out and see if it solves your problems.


JJoosten:
@HansKristian-ARM thanks! I will give it a shot tomorrow! :)

JJoosten:
@HansKristian-ARM I just ran the patch and it solved our issues, thanks! :)

fatkas:
seems to be fine, thanks!

HansKristian-ARM:
Cool, I'll just merge then, essentially all changes were outside the MSL backend.

PR #487, opened 2018-03-06
Add builtin GLSL functions to blacklist.

### LINKED ISSUE #486: SPIRV-Cross should handle naming conflicts on Metal output backend..
Hi,
I'm posting this as trying to run vkquake under MoltenVK I found a shader issue that should be fixed in SPIRV-Cross according to @billhollings ..
(please see full issue: https://github.com/KhronosGroup/MoltenVK/issues/81#issuecomment-370550989)
running vkquake got into:
```
Compilation failed:
program_source:46:20: error: called object type 'float' is not a function or function pointer
float dot = dot(vertexnormal, ubo.shade_vector);
~~~^
```
and answers were:

> The shader error is because the variable is named “dot” which is also the name of a function. This is valid in SPIRV but not metal shaders. The fix is to simply rename the variable.

I said:
> for example: knowing that a variable and function in Metal can't have the same name shouldn't have to be known by a user..
> perhaps has learned Vulkan but has no interest on Metal.. I mean either SPIRV-Cross or MoltenVK should check these cases (in that case seems a request for SPIRV-cross) and rename the variable or function accordingly transparent to the user..

and:
@billhollings

> Yes...SPIRV-Cross should handle naming conflicts. If you haven't already done so...can you post that to SPIRV-Cross to be fixed, please? If you have a sample shader, please include it.

I have not access to exact shader but created a simple repro case to exercise the bug:
compiled this shader below with:

> glslangValidator.exe metalbug.vert -o metalbug.spv -V

```
#version 450

void main()
{
 vec3   vertexnormal=vec3(1,2,3);
 vec3 shade_vector=vec3(4,5,6);

float dot = dot(vertexnormal, shade_vector);
    gl_Position = vec4(dot);
}
```
and then run:

> spirv-cross metalbug.spv --msl
which shows same bug (there is dot function and variable which will not compile with Metal shader compiler) :

```
#include <metal_stdlib>
#include <simd/simd.h>

using namespace metal;

struct main0_out
{
};

vertex main0_out main0()
{
    main0_out out = {};
    float3 vertexnormal = float3(1.0, 2.0, 3.0);
    float3 shade_vector = float3(4.0, 5.0, 6.0);
    float dot = dot(vertexnormal, shade_vector);
    out.gl_Position = float4(dot);
    return out;
}
```
repro files DOWNLOAD:
[metalbug.zip](https://github.com/KhronosGroup/SPIRV-Cross/files/1785691/metalbug.zip)

HansKristian-ARM:
Fix #486.


billhollings:
@HansKristian-ARM 

Thanks for taking care of this so quickly.

How did you generate the list of MSL functions?

HansKristian-ARM:
I haven't looked at MSL functions specifically, but the common ones should overlap with GLSL. (dot, cross, etc). I just scanned through a GL reference page.

billhollings:
Fair enough. A good start. Thanks.

PR #495, opened 2018-03-08
Fix cases where loops and branches merge into continue blocks.

### LINKED ISSUE #493: Complex loop fallback (when enabled) turns a loop into an infinite loop
The attached shader is converted without issues to MSL by SPIRV-Cross by default.

However, changing `block_is_loop_candidate` to always return false (to trigger complex loop emission) converts this shader to a shader with an infinite loop. This shader has three nested loops, and complex loop continues in one of them in the branch that should break instead.

This happens because in CompilerGLSL::branch, one of the blocks returns true for both `is_continue` and `is_break`; `is_continue` is checked first...

[shader-opt.spv.zip](https://github.com/KhronosGroup/SPIRV-Cross/files/1790469/shader-opt.spv.zip)

HansKristian-ARM:
Fix #493.


HansKristian-ARM:
@zeux I believe this should fix both your issues.

zeux:
@Themaister Thanks! Got a chance to try this out, this seems to resolve all issues we had.

PR #520, opened 2018-04-03
Add more illegal name replacement in MSL.

### LINKED ISSUE #517: Additional MSL keyword screening
MSL-specifid keywords like `vertex`, `fragment`, `kernel`, etc...need to be added to the list of illegal words for MSL.  

While we are at it...`CompilerMSL` and `CompilerGLSL` have slightly different approaches to renaming. Perhaps we can look to consolidate. I prefer name for functions to not simply have a  `_` prepended.

Edit: See [this MoltenVK issue](https://github.com/KhronosGroup/MoltenVK/issues/112) for example of this issue.

HansKristian-ARM:
Fix #517.

### DIFF
--- a/reference/opt/shaders-msl/frag/constant-composites.frag
+++ b/reference/opt/shaders-msl/frag/constant-composites.frag
@@ -16,7 +16,7 @@ constant Foo _28[2] = {{10.0, 20.0}, {30.0, 40.0}};
 
 struct main0_in
 {
-    int line [[user(locn0)]];
+    int _line [[user(locn0)]];
 };
 
 struct main0_out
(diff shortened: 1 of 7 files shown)


HansKristian-ARM:
@billhollings Please make a PR if you want to refactor this. Ideally, there should be no overrides for replace_illegal_names.

billhollings:
Having a single list means all languages will either modify or not modify the same set of illegals.

I didn't want to introduce names that break MSL...but might be fine in GLSL (eg. vertex) into the main GLSL code...and have `_vertex` show up in GLSL shaders.

PR #539, opened 2018-04-17
Add support for constexpr samplers in MSL.

### LINKED ISSUE #533: iOS metal seems to require compile-time constant sampler for sample_compare
Hi, I've raised an issue #529 several days ago. You said "no constexpr samplers are needed" because you tested on OSX, it seems that only iOS metal requires this. I upgraded my iOS system and Xcode to the latest version, but the issue still remains. Is there any way to deal with this problem? Otherwise, maybe I have to use sample instead of sample_compare on iOS.

HansKristian-ARM:
Fix #533.


billhollings:
@HansKristian-Work 

I'm implementing the use of this in MoltenVK...by using immutable samplers to cover samplers that need a compare function on older iOS devices.

https://github.com/KhronosGroup/Vulkan-Portability-PoC/issues/53#issuecomment-495742756
https://github.com/KhronosGroup/MoltenVK/issues/556#issuecomment-491799420

How is the `id` used in

    void remap_constexpr_sampler(uint32_t id, const MSLConstexprSampler &sampler);

meant to be resolved through the API? 

The app would generally map the sampler via a desc set & binding...but I couldn't find any functions in SPIRV-Cross that would retrieve the `id` of a `SPIRVariable` based on desc set & binding.

Am I missing something...or should I add such a function? 

And if I do...since it would be an API function...should I roll it into the C interface too...or would you prefer to have control of that since it is meant to be a constant ABI?

Alternately...I could add an API wrapper function around `remap_constexpr_sampler` that would take the desc set and binding as input...similar to:

    bool is_msl_resource_binding_used(spv::ExecutionModel model, uint32_t set, uint32_t binding);

HansKristian-Work:
@billhollings The intended use is to use the get_shader_resources() reflection information to get the IDs of all samplers, and then you can match id with set/binding. However, this doesn't match up with the special resource remapping interface in MSL which specified set/binding up front, and doesn't care about IDs at all. I'll add a variant which deals with set/binding, rather than ID.

PR #565, opened 2018-05-04
Don't limit GLSL identifiers with HLSL keywords.

pmuetschard:
- The HLSL compiler now has its own list of keywords in addition to the ones from GLSL.
 - Added `buffer`, `precise`, and `shared` to the GLSL keywords.


spirv_hlsl.cpp
```diff
@@ -1047,6 +1047,30 @@ void CompilerHLSL::emit_specialization_constants()
 		statement("");
 }
 
+void CompilerHLSL::replace_illegal_names() {
+	// clang-format off
```
HansKristian-ARM: Remove this line.

pmuetschard: Done. Thanks.

HansKristian-ARM:
Thanks.

PR #583, opened 2018-05-22
no expression type for OpSRem

jfactory07:
fix issue#582

### DIFF
--- a/spirv_glsl.cpp
+++ b/spirv_glsl.cpp
@@ -6664,6 +6664,10 @@ void CompilerGLSL::emit_instruction(const Instruction &instruction)
 		UFOP(transpose);
 		break;
 
+	case OpSRem:
+		BOP_CAST(%, SPIRType::Int);
+		break;
+
 	case OpSDiv:
 		BOP_CAST(/, SPIRType::Int);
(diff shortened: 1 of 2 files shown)


spirv_glsl.cpp
```diff
@@ -6664,6 +6664,10 @@ void CompilerGLSL::emit_instruction(const Instruction &instruction)
 		UFOP(transpose);
 		break;
 
+	case OpSRem:
```
HansKristian-ARM: Are you sure about that? IIRC, SMod has those semantics, but SRem behaves slightly differently I think with negative numbers, so I suspect a workaround function is required, similar to FRem.

jfactory07: Sorry, I don't have too much tests for this change.  For my case, SRem can have the same behavior as SMod. 
You are right. We need handle SRem as FRem. Do you think add SRem to FRem is right ? i.e.:
case OpFRem:
case OpSRem:
{
  .......
  break; 
}

HansKristian-ARM:
Merged. I committed the test results on top since the SPIR-V was slightly wrong. Thanks!

PR #587, opened 2018-05-23
Fix textureGrad compilation for legacy targets

zeux:
textureGrad isn't part of core GL2 or GLES2. In desktop GL, it's
provided by GL_ARB_shader_texture_lod and is called texture2DGradARB; in
ES, it's provided by GL_EXT_shader_texture_lod and is called
texture2DGradEXT.

This change rewrites textureGrad similarly to textureProj as per above.


HansKristian-ARM:
Thanks!

Issue #602, opened 2018-06-08
Discuss approach to supporting in/out matrices in MSL


billhollings:
@HansKristian-ARM, et al...

The GLSL `in` and `out` variable declarations get translated to vertex output structure and fragment `stage_in` structure elements.

In GLSL it is posible to define a matrix as an `in` or `out` variable.

However...In MSL, matrices (and scalar & vector arrays) cannot be passed as between vertex and fragment shaders in the vertex shader output struct and the fragment `stage_in` struct.

One way around this might be to convert the matrices back and forth to 4 (or less depending on matrix size) vectors, each with its own element in the `in/out` struct.

Assuming this makes sense...I'd like to discuss how best to do this.

My thoughts so far revolve around having `CompilerMSL` special-case matrices in the `OpLoad` and `OpStore` operations to inject code that will convert back and forth to vectors appropriately when it knows it is dealing with a matrix that is part of `StorageClassInput` or `StorageClassOutput` struct.

I'm throwing this issue out here for discussion...because I'm not sure this will cover all use-cases...and I'm not sure that there are not better approaches that could be explored.

For example...could the existing flattening functions within `CompilerGLSL` be of any help here? I am not familiar with their use.

Or would it be better to inject the appropriate conversion code into the SPIR-V itself...maybe by `glslang`?

Thoughts?

Thanks.

HansKristian-ARM:
For considering just input/output, I think we can go with a lowering scheme. E.g. the internal variable is a matrix, but at the end of a vertex shader, or start of a fragment shader, we can copy over the matrix.

I did this recently with MSL MRT for example, where if you have an array of render targets, I had to flatten that separately. Internally, I have an actual array, but and the end of main(), the array elements are written out to the MSL render targets.

The problem with flattening a matrix is that you can dynamically index into one, so we need at least to express it as an array-of-vectors if we are going to directly flatten things.

billhollings:
@HansKristian-ARM  I've started to explore this. Let me know if you are too.

HansKristian-ARM:
No, I haven't looked into this yet, go ahead.

billhollings:
This enhancement was added in PR #607.

PR #607, opened 2018-06-12
 CompilerMSL support matrices & arrays in stage-in & stage-out.

billhollings:
Support matrices and arrays as both input and output variables for both fragment and vertex shaders.

This extends the previous support for MRT array output in fragment shaders to the general case for arrays and matrices, both input and output, in both fragment and vertex shaders.

Adds the enhancement requested and discussed in issue #602.

### DIFF
--- a/reference/opt/shaders-msl/asm/frag/combined-sampler-reuse.asm.frag
+++ b/reference/opt/shaders-msl/asm/frag/combined-sampler-reuse.asm.frag
@@ -3,14 +3,14 @@
 
 using namespace metal;
 
-struct main0_in
-{
-    float2 vUV [[user(locn0)]];
-};
-
 struct main0_out
 {
     float4 FragColor [[color(0)]];
+};
+
+struct main0_in
+{
+    float2 vUV [[user(locn0)]];
 };
 
 fragment main0_out main0(main0_in in [[stage_in]], texture2d<float> uTex [[texture(1)]], sampler uSampler [[sampler(0)]])
--- a/reference/opt/shaders-msl/flatten/push-constant.flatten.vert
+++ b/reference/opt/shaders-msl/flatten/push-constant.flatten.vert
@@ -10,16 +10,16 @@ struct PushMe
     float Arr[4];
 };
 
-struct main0_in
-{
-    float4 Pos [[attribute(1)]];
-    float2 Rot [[attribute(0)]];
-};
-
 struct main0_out
 {
     float2 vRot [[user(locn0)]];
     float4 gl_Position [[position]];
+};
+
+struct main0_in
+{
+    float2 Rot [[attribute(0)]];
+    float4 Pos [[attribute(1)]];
 };
 
 vertex main0_out main0(main0_in in [[stage_in]], constant PushMe& registers [[buffer(0)]])
--- a/reference/opt/shaders-msl/frag/mrt-array.frag
+++ b/reference/opt/shaders-msl/frag/mrt-array.frag
@@ -5,18 +5,18 @@
 
 using namespace metal;
 
-struct main0_in
-{
-    float4 vB [[user(locn1)]];
-    float4 vA [[user(locn0)]];
-};
-
 struct main0_out
 {
     float4 FragColor_0 [[color(0)]];
     float4 FragColor_1 [[color(1)]];
     float4 FragColor_2 [[color(2)]];
     float4 FragColor_3 [[color(3)]];
+};
+
+struct main0_in
+{
+    float4 vA [[user(locn0)]];
+    float4 vB [[user(locn1)]];
 };
 
 // Implementation of the GLSL mod() function, which is slightly different than Metal fmod()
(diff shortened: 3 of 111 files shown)


HansKristian-ARM:
Cool, I'll review when I get to the office tomorrow.

reference/opt/shaders-msl/asm/frag/combined-sampler-reuse.asm.frag
```diff
@@ -3,14 +3,14 @@
 
 using namespace metal;
 
-struct main0_in
+struct main0_out
```
HansKristian-ARM: A lot of changes to these reference files, any reason for that?

reference/opt/shaders-msl/flatten/push-constant.flatten.vert
```diff
@@ -10,18 +10,18 @@ struct PushMe
...
-
 struct main0_out
 {
     float2 vRot [[user(locn0)]];
     float4 gl_Position [[position]];
 };
 
+struct main0_in
+{
+    float2 Rot [[attribute(0)]];
```
HansKristian-ARM: The ordering here changed, is that intentional?

reference/opt/shaders-msl/frag/mrt-array.frag
```diff
@@ -29,7 +29,7 @@ Tx mod(Tx x, Ty y)
 fragment main0_out main0(main0_in in [[stage_in]])
 {
     main0_out out = {};
-    float4 FragColor[4];
+    float4 FragColor[4] = {};
```
HansKristian-ARM: Any reason for this initializer?

reference/opt/shaders-msl/asm/frag/combined-sampler-reuse.asm.frag
billhollings: Yes. It's deliberate. No technical requirement...but it reads better.

Previous order of emitted structs was:

     in / out / uniforms

MSL functions read as:

    out funcName(in, uniforms)

and this new ordering of emitted matches that:

     out / in / uniforms

First the output then the inputs (in & uniforms).

reference/opt/shaders-msl/flatten/push-constant.flatten.vert
billhollings: Yes. It's deliberate.

Previously input structs were ordered by _reverse_ location order...and output structs were ordered by location order. This brings them both inline.

The original requirement was because early versions of Metal _seemed_ to like having that reverse order on the inputs. But I can't find any real requirement for that anymore. It's either been fixed in Metal...or it never existed.

So I have made both in and out structs sort by location so they visually match between vertex and fragment shaders.

spirv_msl.cpp
```diff
@@ -855,158 +875,12 @@ uint32_t CompilerMSL::add_interface_block(StorageClass storage)
...
-	MemberSorter::SortAspect sort_aspect =
-	    (storage == StorageClassInput) ? MemberSorter::LocationReverse : MemberSorter::Location;
-	MemberSorter member_sorter(ib_type, meta[ib_type_id], sort_aspect);
+	MemberSorter member_sorter(ib_type, meta[ib_type_id], MemberSorter::Location);
 	member_sorter.sort();
 
 	return ib_var_id;
 }
 
-// Returns whether a variable of type and storage class should be moved from an interface
```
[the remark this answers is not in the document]
billhollings: Thanks. I was certainly happy to see that go. The transfer to a different buffer was a kludge workaround to deal with incoming matrices as inputs. This is a much better solution.

reference/opt/shaders-msl/frag/mrt-array.frag
billhollings: Adding an initializer was a convenient way of getting the variable to be declared early in `CompilerGLSL::emit_function()`.

It's not so important for output variables...but critical for input variables whose content will be filled by statements in `fixup_statements_in`, and need to be declared before those statements are emitted.

I could modify so that only input variables are initialized.

And I'm open to other mechanisms. I toyed with adding a new marker on variables to indicate that they needed to be declared early and then modify `CompilerGLSL::emit_function()` to check it...but in the end I figured adding an initializer could make use of the existing `CompilerGLSL::emit_function()` mechanism.

reference/opt/shaders-msl/asm/frag/combined-sampler-reuse.asm.frag
HansKristian-ARM: Ok, makes sense to me.

reference/opt/shaders-msl/flatten/push-constant.flatten.vert
HansKristian-ARM: Ok, the new output makes more sense to me.

reference/opt/shaders-msl/frag/mrt-array.frag
HansKristian-ARM: I see. That's what I assumed. I was going to check if this was the case, but this saved me some time. It's fine as is.

HansKristian-ARM:
Thanks for the fix, good refactor!

Issue #544, opened 2018-04-23
Uber-reflection format


HansKristian-ARM:
#542 expressed a desire for uber-reflection formats, e.g., output a JSON file with all relevant reflection data in it.

Please ping this thread if this is important to you.

Help wanted is tagged here, because someone should PR this functionality if they want it. A separate class derived from Compiler would work well here. `::compile()` could return a full document.

fredrikzetterman:
I want it, of course :)

I've taken the liberty to promote it on reddit as well, to see if someone wants to champion this (my time is limited, otherwise I'd love to help out)

volcanoauthors:
I wrote a shader reflection implementation that uses the SPIRV-Cross library to write out a C (with C++) header for each .spv file input. I call it "copy_header".

Feel free to take inspiration from it: https://github.com/ndsol/volcano/tree/master/src/tools

prideout:
See also:

https://github.com/chaoticbob/SPIRV-Reflect

jherico:
I'm interested in this as well, although I'm interested in the output being a C/C++ header declaring the structs and enums required for the header interface.  However, if JSON output would also be sufficient if there was enough information enced in JSON for me to produce the C++ header with a simple parser on my own.

jherico:
I'm going to work on tackling this in my fork.  I'm likely going to rely on https://github.com/nlohmann/json for formatting JSON output.  There's no existing pattern for inclusion of external libraries.  I could include it as a submodule, or just add the json.hpp header to 3rdparty folder.  Any suggestions?

HansKristian-ARM:
Hm, that is a good question. I'm not sure if there is some convention within Khronos for hosting 3rdparty code, but it's MIT so I doubt that'd be an issue ...

I'd normally lean towards submodule, but since it's a single-header library, it might be more friendly to just add it directly into the repo.

jherico:
My initial plan is to make reflection emission based on a command line option in the style `--reflect [format]` where format would default to `json`, and then potentially support alternative formats like `xml` and `c++`, but only after I've nailed down the JSON reflection.  I'm currently examining the code triggered by `--dump-resources` as a starting point for what I need to iterate over to produce my output.     I had earlier started work on a derived compiler that would output a C++ header with enums and structs,  but I think building a hierarchical structure like JSON first and then transforming that to other formats is likely going to be easier.

HansKristian-ARM:
Another consideration is the compiler support.

[...]

PR #608, opened 2018-06-15
Output JSON formatted reflection information for a SPIRV shader

jherico:
Feature implementation for #544


HansKristian-ARM:
Cool! Don't think I'll find time to review today, but I'll get to it as soon as possible.

main.cpp
```diff
@@ -768,6 +791,13 @@ static int main_inner(int argc, char *argv[])
 		if (args.cpp_interface_name)
 			static_cast<CompilerCPP *>(compiler.get())->set_interface_name(args.cpp_interface_name);
 	}
+	else if (!args.reflect.empty())
+	{
+		combined_image_samplers = !args.vulkan_semantics;
```
HansKristian-ARM: Both of these should be false.

CMakeLists.txt
```diff
@@ -93,6 +93,10 @@ spirv_cross_add_library(spirv-cross-cpp spirv_cross_cpp STATIC
     ${CMAKE_CURRENT_SOURCE_DIR}/spirv_cpp.hpp
     ${CMAKE_CURRENT_SOURCE_DIR}/spirv_cpp.cpp)
 
+spirv_cross_add_library(spirv-cross-reflect spirv_cross_reflect STATIC
```
HansKristian-ARM: I'm missing an update to the MSVC project (normally it's built through cmake, just add the few lines required by hand.

spirv_glsl.hpp
```diff
@@ -259,6 +259,11 @@ class CompilerGLSL : public Compiler
 		statement_inner(std::forward<Ts>(ts)...);
 	}
 
+	inline void statement_indent() {
```
HansKristian-ARM: This doesn't look formatted, did you run clang-format on it?

spirv_reflect.cpp
```diff
@@ -0,0 +1,514 @@
+/*
+ * Copyright 2015-2018 ARM Limited
```
HansKristian-ARM: This should be your copyright.

spirv_reflect.cpp
```diff
@@ -0,0 +1,514 @@
...
+
+#include "spirv_reflect.hpp"
+#include <cassert>
+
+using namespace spv;
+using namespace spirv_cross;
+using namespace std;
+
+#ifdef SPIRV_CROSS_EXCEPTIONS_TO_ASSERTIONS
+static inline void THROW(const char *str)
```
HansKristian-ARM: Use SPIRV_CROSS_THROW.

spirv_reflect.cpp
```diff
@@ -0,0 +1,514 @@
...
+
+void CompilerReflection::set_format(const std::string &format)
+{
+	if (format != "json")
+	{
+		THROW("Unsupported format");
+	}
+}
+
+// Hackery to emit JSON without using
```
HansKristian-ARM: without using what?

spirv_reflect.cpp
```diff
@@ -0,0 +1,514 @@
...
+	++indent;
+	json_state.emplace(JsonType::Array, false);
+}
+
+string CompilerReflection::compile()
+{
+	// force a GLSL compilation in the parent class in order to set all our internal state
+	CompilerGLSL::compile();
+
+    buffer.reset();
```
HansKristian-ARM: Formatting?

spirv_reflect.cpp
```diff
@@ -0,0 +1,514 @@
...
+
+	//add_resource_name(type.self);
+	//type.member_name_cache.clear();
+
+	if (!emitted_open_tag)
+	{
+		emit_json_key_object("types");
+		emitted_open_tag = true;
+	}
+	emit_json_key_array(name);
```
HansKristian-ARM: Emitting the types based on name is dangerous. SPIR-V modules can have identical names for multiple unrelated types. It's very common to have two types with same name, one living as part of an SSBO, one as a value type. While we need to deal with this in GLSL/HLSL/MSL/etc, this output should just be a "raw" reflection of the SPIR-V module.

I think it makes more sense to have "types" as a flat array, and then objects will refer to types by index into that array (or some other way to index the objects). The name would be part of the inner object in the array. This way there is no ambiguity.

Consider the problem:
```
"types" : [
    "foo" : { "do something" },
    "foo" : { "do something else" }
]
```

spirv_reflect.cpp
```diff
@@ -0,0 +1,514 @@
...
+
+	auto &membertype = get<SPIRType>(type.member_types[index]);
+	emit_type_array(membertype);
+	auto &memb = meta[type.self].members;
+	if (index < memb.size())
+	{
+		auto &dec = memb[index];
+		if (dec.decoration_flags.get(DecorationLocation) && can_use_io_location(type.storage, true))
+			emit_json_key_value("location", dec.location);
+		if (has_decoration(type.self, DecorationCPacked) && dec.decoration_flags.get(DecorationOffset))
```
HansKristian-ARM: CPacked is not relevant here, just check for DecorationOffset.

spirv_reflect.cpp
```diff
@@ -0,0 +1,514 @@
...
+		{
+			begin_json_object();
+			emit_json_key_value("name", e.name);
+			emit_json_key_value("mode", execution_model_to_str(e.execution_model));
+			end_json_object();
+		}
+		end_json_array();
+	}
+
+#if 0
```
HansKristian-ARM: Random debug code should be left out if possible.

spirv_reflect.cpp
```diff
@@ -0,0 +1,514 @@
...
+}
+
+void CompilerReflection::emit_resources(const char *tag, const vector<Resource> &resources)
+{
+	if (resources.empty())
+	{
+		return;
+	}
+
+	bool print_ssbo = !strcmp(tag, "ssbos");
```
HansKristian-ARM: Is this used, or just copy-pasted from main.cpp?

spirv_reflect.cpp
```diff
@@ -0,0 +1,514 @@
...
+					emit_json_key_value("std140", true);
+				else if (buffer_is_packing_standard(type, BufferPackingStd140EnhancedLayout))
+					emit_json_key_value("std140", true);
+				else if (buffer_is_packing_standard(type, BufferPackingStd430EnhancedLayout))
+					emit_json_key_value("std430", true);
+			}
+
+			if (ssbo_block)
+			{
+				auto bufferFlags = get_buffer_block_flags(res.id);
```
HansKristian-ARM: Casing.

spirv_reflect.cpp
```diff
@@ -0,0 +1,514 @@
...
+			if (type.storage == StorageClassUniform && typeflags.get(DecorationBlock))
+			{
+				if (buffer_is_packing_standard(type, BufferPackingStd140))
+					emit_json_key_value("std140", true);
+				else if (buffer_is_packing_standard(type, BufferPackingStd140EnhancedLayout))
+					emit_json_key_value("std140", true);
+			}
+			else if (push_constant_block || ssbo_block)
+			{
+				if (buffer_is_packing_standard(type, BufferPackingStd430))
```
HansKristian-ARM: I don't think this is relevant for SPIR-V reflection.
std430/std140 only make sense for GLSL output/input.

spirv_reflect.hpp
```diff
@@ -0,0 +1,102 @@
...
+#define SPIRV_CROSS_REFLECT_HPP
+
+#include "spirv_cross.hpp"
+#include "spirv_glsl.hpp"
+#include <utility>
+#include <vector>
+
+namespace spirv_cross
+{
+class CompilerReflection : public CompilerGLSL
```
HansKristian-ARM: CompilerReflection should depend on Compiler, not CompilerGLSL.

spirv_reflect.hpp
```diff
@@ -0,0 +1,102 @@
...
+    void emit_json_key_value(const std::string& key, uint32_t value);
+    void emit_json_key_object(const std::string& key);
+    void emit_json_key_array(const std::string& key);
+    void emit_json_array_value(const std::string& value);
+    void emit_json_array_value(uint32_t value);
+    void begin_json_array();
+    void end_json_array();
+
+    template <typename... Ts>
+    inline void statement_no_return(Ts &&... ts) {
```
HansKristian-ARM: Formatting.

spirv_reflect.hpp
```diff
@@ -0,0 +1,102 @@
...
+    void emit_type(const SPIRType &type, bool& emitted_open_tag);
+    void emit_type_member(const SPIRType &type, uint32_t index);
+    void emit_type_member_qualifiers(const SPIRType &type, uint32_t index);
+    void emit_type_array(const SPIRType &type);
+    void emit_resources(const char *tag, const std::vector<Resource> &resources);
+
+    void begin_json_object();
+    void end_json_object();
+    void emit_json_key(const std::string& key);
+    // FIXME switch to templated types with specialization for string?
```
HansKristian-ARM: This is fine.

HansKristian-ARM:
I think having this many test files is overkill. At best, we can probably get away with just a few shaders which exercise all the different resource types.

jherico:
Thanks for the comprehensive review, will start working on the updates.

spirv_reflect.cpp
```diff
@@ -0,0 +1,514 @@
...
+	{
+		end_json_object();
+	}
+}
+
+void CompilerReflection::emit_type(const SPIRType &type, bool &emitted_open_tag)
+{
+	auto name = type_to_glsl(type);
+
+	// Struct types can be stamped out multiple times
```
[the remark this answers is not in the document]
jherico: are there any examples of this kind of thing in the test shaders that I can use to validate my solution?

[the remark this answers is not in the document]
HansKristian-ARM: What you can try is a GLSL shader doing:

```
struct Foo { float b[16]; };
layout(std140, binding = 0) buffer SSBO140 { Foo foo; };
layout(std430, binding = 1) buffer SSBO430 { Foo bar; };
```
This should give you some interesting output.

spirv_reflect.cpp
jherico: using indexing would require maintaining additional internal state between different sections of the JSON emission code.  Would it be reasonable to use the type id (rather than the name) as the key into a map of types?  Also, again, are there any examples in the test data of this kind of name duplication, so I can validate my fix?

HansKristian-ARM: Type ID lookup is also fine in my book. See shaders/comp/struct-packing.comp for some difficult cases.

spirv_reflect.cpp
jherico: If someone intends to use the reflection in combination with GLSL output from SPIRV-Cross, then it's critical, is it not?  That's actually my use case.  This is part of a first step towards a full Vulkan backend, by making all the shaders SPIRV-Compatibile and using SPIRV as our transmission format, even though we _might_ have to use GLSL for the shader upload to the GPU on systems that aren't OpenGL 4.6 or higher.

spirv_reflect.hpp
jherico: A lot of the code I depend on is in `CompilerGLSL`.  All the buffering code like `statement` are in that class.  I could probably extract that to either `Compiler` or some new intermediate class between `Compiler` and `CompilerGLSL` / `CompilerReflection`, but that would still leave a lot of other stuff out.  Currently, in order to ensure that all the internal structures like type member names have been fully populated I actually run the parent class method `CompilerGLSL::compile` before I start introspection.

spirv_reflect.cpp
HansKristian-ARM: Well, the question then is, the API-side of GL should not care about packing used in the shader (unless it's packed/shared layout). What information does your backend use if a buffer is std140 or std430? I don't think that can even be queried. The shader finds whatever layout will conform to the SPIR-V, not the other way around.

spirv_reflect.hpp
jherico: I'm looking into how much breaks if I change the base class.  The emitter stuff obviously, but I've actually been thinking of breaking out the simple JSON streaming stuff into it's own class so that if and when it can be replaced with something like `nlohmann::json` the work will be trivial.  The remaining things that break don't look as daunting to me as they initially did... some of them like `format_to_glsl` and `type_to_glsl` might be fixed by adding some static helpers that are used both by `CompilerGLSL` and `CompilerReflect`.  Others, like `buffer_is_packing_standard` are only used in limited scope where you suggest they might not be needed at all.  

I'll see what I can do.

jherico:
> I think having this many test files is overkill

My goal there was to avoid trying to cherry pick use cases and miss something important.  However, I'll opt to try to scour the existing test shaders data set for a set of features I can use to create stuff specifically to test the reflection and see if I can't use some code-coverage tools to make sure I'm hitting all the cases.

jherico:
Sorry about the many formatting errors.  Unfortunately I can't simply turn on _Format on Save_ in my IDE because my day to day work is on a code base that _has_ a .clang-format file, but hasn't had all it's code migrated to the correct format yet.

spirv_reflect.hpp
jherico: I forgot the other issue that popped up when I initially attempted to derive directly from `Compiler`.  The code in main.cpp uses `unique_ptr<CompilerGLSL> compiler;` to contain the instantiated compiler object, and then uses a significant number of methods (like `get_common_options`, `set_common_options`, `flatten_buffer_block`, etc) which are only visible on the `CompilerGLSL` class, not on the base class.  

I may be able to work around this by putting all the reflection work into the `else if (!args.reflect.empty())` block and doing an early return, but it might be ugly.  The alternative would seem to be a lot of new conditional blocks and static casting in the bulk of the main.cpp code.

HansKristian-ARM: Keeping it as a subclass of GLSL is okay for now, don't spend a lot of time trying to rework this.

spirv_reflect.cpp
jherico: That makes sense.

CMakeLists.txt
jherico: fixed

jherico: fixed

spirv_reflect.hpp
jherico: Oops... too late... already did the refactoring.

HansKristian-ARM:
Thanks for the updates. I'll have time to continue review tomorrow.

spirv_glsl.cpp
```diff
@@ -689,22 +689,6 @@ void CompilerGLSL::emit_struct(SPIRType &type)
 		statement("");
 }
 
-Bitset CompilerGLSL::combined_decoration_for_member(const SPIRType &type, uint32_t index)
```
jherico: moved to parent class unchanged, except declared const

spirv_glsl.cpp
```diff
@@ -786,122 +770,36 @@ string CompilerGLSL::layout_for_member(const SPIRType &type, uint32_t index)
 
 const char *CompilerGLSL::format_to_glsl(spv::ImageFormat format)
 {
-	auto check_desktop = [this] {
```
jherico: enum to string functionality moved to parent class `format_to_string`, retaining only the GLES valid format check

spirv_cross.cpp
```diff
@@ -4567,3 +4567,292 @@ bool Compiler::instruction_to_result_type(uint32_t &result_type, uint32_t &resul
...
+	for (uint32_t i = 0; i < type.member_types.size(); i++)
+		flags.merge_or(combined_decoration_for_member(get<SPIRType>(type.member_types[i]), i));
+
+	return flags;
+}
+
+// The optional id parameter indicates the object whose type we are trying
+// to find the description for. It is optional. Most type descriptions do not
+// depend on a specific object's use of that type.
+string Compiler::type_to_string(const SPIRType &type)
```
jherico: need to refactor the `CompilerGLSL::type_to_glsl` to work in terms of this function, checking only the small number of cases where the glsl will be different than the output here.

jherico:
Another commit incoming shortly to reduce some code duplication between `Compiler` and `CompilerGLSL`, as well as some additional minor cleanup and support for running the regression tests in parallel.

main.cpp
```diff
@@ -149,6 +150,22 @@ struct CLIParser
 		return val;
 	}
 
+	// Return a string only if it's not prefixed with `--`, otherwise return the default value
+	const char *next_value_string(const char *defaultValue)
```
HansKristian-ARM: Casing of defaultValue.

spirv_cross.hpp
```diff
@@ -856,6 +856,13 @@ class Compiler
 	bool instruction_to_result_type(uint32_t &result_type, uint32_t &result_id, spv::Op op, const uint32_t *args,
 	                                uint32_t length);
 
+	Bitset combined_decoration_for_member(const SPIRType &type, uint32_t index) const;
+	std::string type_to_string(const SPIRType &type) const;
```
HansKristian-ARM: Isn't this basically type_to_glsl? I think it would make more sense if this is virtual and the other backends start using type_to_string instead. The old naming of type_to_glsl is a bit awkward to begin with since HLSL/MSL are also using it.

Right now it seems like a helper function specific to GLSL/Reflection, which is a bit weird.

spirv_glsl.cpp
HansKristian-ARM: Makes sense.

spirv_reflect.hpp
```diff
@@ -0,0 +1,70 @@
...
+
+	void emit_type(const SPIRType &type, bool &emitted_open_tag);
+	void emit_type_member(const SPIRType &type, uint32_t index);
+	void emit_type_member_qualifiers(const SPIRType &type, uint32_t index);
+	void emit_type_array(const SPIRType &type);
+	void emit_resources(const char *tag, const std::vector<Resource> &resources);
+
+	std::string to_member_name(const SPIRType &type, uint32_t index) const;
+
+	std::shared_ptr<simple_json::Stream> json_stream;
```
HansKristian-ARM: shared_ptr? Why not unique_ptr?

test_shaders.py
```diff
@@ -459,19 +557,35 @@ def main():
...
             action = 'store_true',
             help = 'Run SPIRV-Tools optimization passes as well.')
+    parser.add_argument('--reflect',
+            action = 'store_true',
+            help = 'Test reflection backend.')
+    parser.add_argument('--parallel',
+            action = 'store_true',
+            help = 'Execute tests in parallel.  Useful for doing regression quickly, but bad for debugging and stat output.')
+    
+    global args;
```
HansKristian-ARM: semicolon?

test_shaders.py
```diff
@@ -410,26 +484,50 @@ def test_shader_hlsl(stats, shader, update, keep, opt):
...
-                test_shader_hlsl(stats, (shader_dir, relpath), update, keep, opt)
-            else:
-                test_shader(stats, (shader_dir, relpath), update, keep, opt)
+            relpath = os.path.relpath(path, args.folder)
+            all_files.append(relpath)
+
+    # The child processes in parallel execution mode don't have the proper state for the global args variable, so 
+    # at this point we need to switch to explicit arguments
+    if args.parallel:
+        pool = multiprocessing.Pool(multiprocessing.cpu_count())
```
HansKristian-ARM: Cool.

spirv_cross.hpp
jherico: I concur with avoiding leaking, but I was unable to convert this to a static method on my first pass because some of the code paths use the internal `meta` array to do lookups.  I'll take a look again and see if there's not a way to push this into some kind GLSL helper library of static functions.

test_shaders.py
jherico: curse my 20 years of C++ & Java reflexes.

spirv_reflect.cpp
```diff
@@ -0,0 +1,533 @@
...
+
+	if (!emitted_open_tag)
+	{
+		json_stream->emit_json_key_object("types");
+		emitted_open_tag = true;
+	}
+	json_stream->emit_json_key_object(std::to_string(type.self));
+	json_stream->emit_json_key_value("name", name);
+
+	// FIXME I'd like to emit the size here, but it triggers a crash in some shaders
```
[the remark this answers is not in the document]
jherico: thanks, I'll investigate

test_shaders.py
```diff
@@ -459,19 +557,35 @@ def main():
...
+            help = 'Execute tests in parallel.  Useful for doing regression quickly, but bad for debugging and stat output.')
+    
+    global args;
     args = parser.parse_args()
-
     if not args.folder:
         sys.stderr.write('Need shader folder.\n')
         sys.exit(1)
 
+    if (args.parallel and (args.malisc or args.force_no_external_validation)):
```
HansKristian-ARM: Should this be blocked for --update paths? Can the tempfile creation create race conditions?

test_shaders.py
jherico: Thanks.  I was running into failures of the regression tests when I was refactoring the type_to_glsl and image_type_to_glsl functions, and it was annoying me that each test would take so long.  Had to learn the python mechanism for parallelism and for partial function binding, but it ended up being not too complicated.  

I think the multiprocessing approach here may be overkill though.  Most stuff I saw about python threading led rapidly to dire warnings about the global interpreter lock killing multi-threading performance gains, hence the need to essentially use a variant of `fork` to go multi-process instead of multi-thread.  However, since the vast majority of the time in this script is actually already spent launching other processes, I suspect it would not have been a problem to just use threads.

spirv_cross.hpp
HansKristian-ARM: I'm also considering that maybe trying to reduce code duplication here at all costs might be counterproductive. Clearly something is off if we need to contort the structure like this ... What do you think?

test_shaders.py
jherico: there's one process per input file, and each output file maps to exactly one input file, so there shouldn't be any chance of a race condition where two processes are trying to write to the same output.  However I can err on the side of caution.

test_shaders.py
HansKristian-ARM: multiprocess makes sense. We just want to invoke completely separate things in parallel.

spirv_reflect.hpp
HansKristian-ARM: You'll need to declare constructor and destructor as non-default/inline, then it should work.

spirv_cross.hpp
HansKristian-ARM: The naming scheme is basically GLSL, but without any validation for ES or extensions. I think it is okay here to just implement the Reflection backends naming convention directly into CompilerReflection. Finding a clean way with standalone functions is fine too if you figure it out.

jherico: If the desire is to keep `Compiler` clean of GLSL conventions, but that the JSON reflection _should_ use GLSL naming conventions (for simplicity and for lack of a better alternative) then perhaps I should revert the `CompilerReflection` class to deriving from `CompilerGLSL`.  That doesn't feel ideal either, though.  It seems more like what's needed is composition, rather than inheritance, where the internal representation of the SPIRV is in one class, and the conversion of that to the syntax of a given shader language is in a separate class hierarchy, of which a single concrete instance is owned by the `Compiler`.  This might simplify the Compiler hierarchy _and_ allow the JSON reflection to be customized so that the core types were expressed in terms of a specific target language (i.e. have a type be `float4` instead of `vec4` etc...).

However, this seems like a _big_ architectural change, at least for someone like me who isn't as familiar with the whole code base.  I think for the time being it would probably be simpler if I just reverted to having `CompilerReflection` derive from `CompilerGLSL`.

HansKristian-ARM: Ye, that works too, CompilerReflection can inherit CompilerGLSL, that is the easiest, least invasive way to get this through. We can revisit later if we need non-GLSL centric reflection output.

Sorry for the back-and-forth. New backends are always a "big deal".

spirv_reflect.hpp
jherico: The code that fails to compile is in main.cpp (so no change to the `Stream` implementation in `spriv_reflect.cpp` is going to help), because that's where the `CompilerReflection` class is instantiated and destroyed.  Because it doesn't have visibility on the full `simple_json::Stream` type, it can't properly generate the destructor.  Explicitly adding a `~CompilerReflection` doesn't seem to help either.  

I can either move the `simple_json::Stream` full declaration into the spirv_reflect.hpp file (or into a new .cpp/.hpp pair in spirv-cross-util), making the full class declaration visible to main.cpp, or I can leave it as a `shared_ptr`.  Let me know.

HansKristian-ARM: Ok, if that doesn't work, just keep shared_ptr.

spirv_cross.hpp
jherico: It's no problem.  I find it pretty exciting to be able to contribute something to the project.

test_shaders.sh
```diff
@@ -16,4 +16,5 @@ echo "Using spirv-opt in: $(which spirv-opt)."
 ./test_shaders.py shaders-hlsl --hlsl || exit 1
 ./test_shaders.py shaders-hlsl --hlsl --opt || exit 1
 ./test_shaders.py shaders-hlsl-no-opt --hlsl || exit 1
+./test_shaders.py shaders-reflection --reflect || exit 1
```
HansKristian-ARM: Probably have to change update_shaders.sh as well.

main.cpp
```diff
@@ -663,6 +681,11 @@ static int main_inner(int argc, char *argv[])
 	cbs.add("--flip-vert-y", [&args](CLIParser &) { args.yflip = true; });
 	cbs.add("--iterations", [&args](CLIParser &parser) { args.iterations = parser.next_uint(); });
 	cbs.add("--cpp", [&args](CLIParser &) { args.cpp = true; });
+	cbs.add("--reflect", [&args](CLIParser &parser) {
+		// force vulkan semantics for reflection
+		args.vulkan_semantics = true;
```
HansKristian-ARM: Is this necessary anymore?

reference/shaders-reflection/asm/aliased-entry-point-names.asm.multi.json
```diff
@@ -0,0 +1,49 @@
...
+            "name" : "main",
+            "mode" : "frag"
+        },
+        {
+            "name" : "main2",
+            "mode" : "frag"
+        }
+    ],
+    "types" : {
+        "8" : {
```
HansKristian-ARM: The member is a string, could it be made an actual integer?

jherico: The problem is that the members of a struct each have a type field which is either a built-in type, like 'vec4' or a key back into this map.  I wanted to avoid making that field switch between string and integer types.  

I _could_ prefix all these identifiers with underscores to make it more clear they're just opaque IDs.

test_shaders.sh
jherico: done

HansKristian-ARM:
Great. Thanks for the contribution, it's now merged to master.

PR #615, opened 2018-06-22
Added support for shadowXY() sample instructions in legacy GLSL

JustSid:
Basically what the title says. If you have a shadow sampler up until now SPIRV-Cross would emit a `texture[XY]()` instruction, which works fine on modern GLSL backends but not on legacy backends. 

This adds the emission of  `shadowXY()` instructions. It's not 100% complete but should cover the most used cases, at least to implement CSM. And it's a starting point for others to pile on more variants if necessary.

### DIFF
--- a/spirv_glsl.cpp
+++ b/spirv_glsl.cpp
@@ -3947,7 +3947,7 @@ void CompilerGLSL::emit_texture_op(const Instruction &i)
 	expr += ")";
 
 	// texture(samplerXShadow) returns float. shadowX() returns vec4. Swizzle here.
-	if (is_legacy() && ((imgtype.basetype == SPIRType::SampledImage) || (imgtype.basetype == SPIRType::Sampler)) && imgtype.image.depth)
+	if (is_legacy() && imgtype.image.depth)
 		expr += ".r";
 
 	emit_op(result_type, id, expr, forward);
--- a/spirv_glsl.cpp
+++ b/spirv_glsl.cpp
@@ -3542,47 +3542,43 @@ string CompilerGLSL::legacy_tex_op(const std::string &op, const SPIRType &imgtyp
 			require_extension_internal("GL_ARB_shader_texture_lod");
 	}
 
-	// texture
+	std::string type_prefix = imgtype.image.depth ? "shadow" : "texture";
+
 	if (op == "texture")
-		return join("texture", type);
+		return join(type_prefix, type);
 	else if (op == "textureLod")
 	{
 		if (use_explicit_lod)
-			return join("texture", type, is_legacy_es() ? "LodEXT" : "Lod");
+			return join(type_prefix, type, is_legacy_es() ? "LodEXT" : "Lod");
 		else
-			return join("texture", type);
+			return join(type_prefix, type);
 	}
 	else if (op == "textureProj")
-		return join("texture", type, "Proj");
+		return join(type_prefix, type, "Proj");
 	else if (op == "textureGrad")
-		return join("texture", type, is_legacy_es() ? "GradEXT" : is_legacy_desktop() ? "GradARB" : "Grad");
+		return join(type_prefix, type, is_legacy_es() ? "GradEXT" : is_legacy_desktop() ? "GradARB" : "Grad");
 	else if (op == "textureProjLod")
 	{
 		if (use_explicit_lod)
-			return join("texture", type, is_legacy_es() ? "ProjLodEXT" : "ProjLod");
+			return join(type_prefix, type, is_legacy_es() ? "ProjLodEXT" : "ProjLod");
 		else
-			return join("texture", type);
+			return join(type_prefix, type);
 	}
-	// shadow
-	else if (op == "shadow")
-		return join("shadow", type);
-	else if (op == "shadowLodOffset")
+	else if (op == "textureLodOffset")
 	{
 		if (use_explicit_lod)
-			return join("shadow", type, "LodOffset");
+			return join(type_prefix, type, "LodOffset");
 		else
-			return join("shadow", type);
+			return join(type_prefix, type);
 	}
-	else if (op == "shadowProjGrad")
-		return join("shadow", type, "ProjGrad");
-	else if (op == "shadowGrad")
-		return join("shadow", type, "Grad");
-	else if (op == "shadowProjLodOffset")
+	else if (op == "textureProjGrad")
+		return join(type_prefix, type, "ProjGrad");
+	else if (op == "textureProjLodOffset")
 	{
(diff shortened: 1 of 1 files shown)


spirv_glsl.cpp
```diff
@@ -3924,6 +3946,10 @@ void CompilerGLSL::emit_texture_op(const Instruction &i)
 	                         coffset, offset, bias, comp, sample, &forward);
 	expr += ")";
 
+	// texture(samplerXShadow) returns float. shadowX() returns vec4. Swizzle here.
+	if (is_legacy() && ((imgtype.basetype == SPIRType::SampledImage) || (imgtype.basetype == SPIRType::Sampler)) && imgtype.image.depth)
```
HansKristian-ARM: Why check for SPIRType::Sampler here?

spirv_glsl.cpp
```diff
@@ -3968,7 +3994,10 @@ string CompilerGLSL::to_function_name(uint32_t, const SPIRType &imgtype, bool is
 		fname += "texelFetch";
 	else
 	{
-		fname += "texture";
+		if (is_legacy() && ((imgtype.basetype == SPIRType::SampledImage) || (imgtype.basetype == SPIRType::Sampler)) && imgtype.image.depth)
```
HansKristian-ARM: Why SPIRType::Sampler?

spirv_glsl.cpp
```diff
@@ -3542,6 +3542,7 @@ string CompilerGLSL::legacy_tex_op(const std::string &op, const SPIRType &imgtyp
 			require_extension_internal("GL_ARB_shader_texture_lod");
 	}
 
+	// texture
 	if (op == "texture")
```
HansKristian-ARM: Isn't it easier to move the check for "texture" vs "shadow" here? You know the imgtype here, so you should be able to isolate the check to think function.

spirv_glsl.cpp
JustSid: @HansKristian-ARM I copy and pasted the if more or less verbatim from line 8480. Same goes for the other instance.  Based on the comment I guess the check is unnecessary? Thinking about this, just checking for depth probably ought to be enough here?

HansKristian-ARM: I think just checking for depth is enough, either that, or check if the op-code has "Dref" in it.

HansKristian-ARM:
I think you screwed up your rebase. :)

JustSid:
Yeah, I was just going to say... That turned a bit ugly... Whoops

JustSid:
Alright, I'll out myself as an absolute github noob. If I force push a change in history (without the screwed up rebase) on my fork, will the pull request here survive? Otherwise I can also open a new pull request and close this one.

HansKristian-ARM:
You can force-push and the PR will still survive fine.

JustSid:
Thanks! :)  
Time to rewrite history!

spirv_glsl.cpp
```diff
@@ -3542,25 +3542,43 @@ string CompilerGLSL::legacy_tex_op(const std::string &op, const SPIRType &imgtyp
...
 	else if (op == "textureGrad")
-		return join("texture", type, is_legacy_es() ? "GradEXT" : is_legacy_desktop() ? "GradARB" : "Grad");
+		return join(type_prefix, type, is_legacy_es() ? "GradEXT" : is_legacy_desktop() ? "GradARB" : "Grad");
 	else if (op == "textureProjLod")
 	{
 		if (use_explicit_lod)
-			return join("texture", type, is_legacy_es() ? "ProjLodEXT" : "ProjLod");
+			return join(type_prefix, type, is_legacy_es() ? "ProjLodEXT" : "ProjLod");
+		else
+			return join(type_prefix, type);
```
HansKristian-ARM: Shouldn't this be Proj?

spirv_glsl.cpp
```diff
@@ -3542,25 +3542,43 @@ string CompilerGLSL::legacy_tex_op(const std::string &op, const SPIRType &imgtyp
...
+	}
+	else if (op == "textureLodOffset")
```
HansKristian-ARM: Are you sure this exists in legacy GLSL?

spirv_glsl.cpp
```diff
@@ -3542,25 +3542,43 @@ string CompilerGLSL::legacy_tex_op(const std::string &op, const SPIRType &imgtyp
...
+	{
+		if (use_explicit_lod)
+			return join(type_prefix, type, "LodOffset");
 		else
-			return join("texture", type);
+			return join(type_prefix, type);
+	}
+	else if (op == "textureProjGrad")
+		return join(type_prefix, type, "ProjGrad");
```
HansKristian-ARM: Does this exist in legacy GLSL? Wouldn't it need EXT/ARB like "GradARB" at the very least?

spirv_glsl.cpp
```diff
@@ -3542,25 +3542,43 @@ string CompilerGLSL::legacy_tex_op(const std::string &op, const SPIRType &imgtyp
...
+	else if (op == "textureProjLodOffset")
```
HansKristian-ARM: Not sure if this exists either.

spirv_glsl.cpp
JustSid: This one and the other two exist through the `EXT_gpu_shader4` extension. I'll add requiring of that extension when encountering these commands.

spirv_glsl.cpp
JustSid: This exists either as ARB suffixed version through `ARB_shader_texture_lod`, or through `EXT_gpu_shader4` without suffix. I'm not really sure how this should be handled. Is there any precedence on how to handle functionality that is provided by two extensions?

HansKristian-ARM: I think ARB_shader_texture_lod is already used, so you can reuse that.

spirv_glsl.cpp
HansKristian-ARM: EXT_gpu_shader4 for these commands is fine.
Do you know anything about GLES though? If it doesn't exist there as an extension to GLES2, please just add a SPIRV_CROSS_THROW for that case.

JustSid: GLES has `EXT_shadow_samplers` which supports at least `shadow2DEXT` and `shadow2DProjEXT`, so these two can be supported. I'll add the throw case for everything else. I'm not aware of any other extension that adds further support here and doing a quick glance through the registered extensions didn't reveal anything either.

HansKristian-ARM: Sounds good.

HansKristian-ARM:
Thanks!

PR #625, opened 2018-06-26
Uniquely name macros across both files (spirv_glsl.cpp/spirv_hlsl.cpp)

twinaphex:
@HansKristian-ARM  The reason for this PR is that we get a lot of warnings when using Griffin builds (for RetroArch) due to these similar macros being named the same across both spirv_glsl.cpp and spirv_hlsl.cpp. As you might recall, with a Griffin build, all source files get compiled into one bigger file.

Merging this PR would mean I wouldn't have to do any major changes on top of the SPIRV-Cross code and it could stay as a shallow fork.

### DIFF
--- a/spirv_glsl.cpp
+++ b/spirv_glsl.cpp
@@ -2620,8 +2620,6 @@ string CompilerGLSL::constant_op_expression(const SPIRConstantOp &cop)
 		break;
 	}
 
-#undef BOP
-#undef UOP
 	if (binary)
 	{
 		if (cop.arguments.size() < 2)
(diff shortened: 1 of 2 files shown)


HansKristian-ARM:
Standalone build fails on Travis: https://travis-ci.org/KhronosGroup/SPIRV-Cross/jobs/396839635. Is it possible to just make sure the macros from one file is properly `#undef`-ed so that it doesn't pollute any other files in a single-file build?

HansKristian-ARM:
Fixed separately on master.

PR #626, opened 2018-06-26
CompilerMSL support larger texel buffers by using 2D Metal textures.

billhollings:
Add CompilerMSL::Options::texture_width_max.
Emit and use spvTexelBufferCoord() function to convert 1D
texel buffer coordinates to 2D Metal texture coordinates.

### DIFF
--- a/spirv_msl.hpp
+++ b/spirv_msl.hpp
@@ -152,7 +152,7 @@ public:
 
 		Platform platform = macOS;
 		uint32_t msl_version = make_msl_version(1, 2);
-		uint32_t texture_width_max = 4096;
+		uint32_t texel_buffer_texture_width = 4096;		// Width of 2D Metal textures used as 1D texel buffers
 		bool enable_point_size_builtin = true;
 		bool resolve_specialized_array_lengths = true;
 
--- a/spirv_msl.hpp
+++ b/spirv_msl.hpp
@@ -152,6 +152,7 @@ public:
 
 		Platform platform = macOS;
 		uint32_t msl_version = make_msl_version(1, 2);
+		uint32_t texture_width_max = 4096;
 		bool enable_point_size_builtin = true;
 		bool resolve_specialized_array_lengths = true;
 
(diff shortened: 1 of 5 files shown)


spirv_msl.hpp
```diff
@@ -152,6 +152,7 @@ class CompilerMSL : public CompilerGLSL
 
 		Platform platform = macOS;
 		uint32_t msl_version = make_msl_version(1, 2);
+		uint32_t texture_width_max = 4096;
```
HansKristian-ARM: There should probably be an option to opt-in to this kind of path.
It seems a bit specific to MoltenVK.
I.e. if true, wrap by modulo here, otherwise, keep the old path.

HansKristian-ARM: The name of the variable should more clearly specify that it's for texel buffer emulation in particular.

billhollings: @HansKristian-ARM 

No...it's not specific to MoltenVK. Metal telex buffers must be 2D with or without MoltenVK. The old path essentially hardcoded the modulo to 1:

    uint2(idx, 0)

and in doing so limited the size of the texel buffer to the max width of a 2D texture (eg- 4096/8192/16384 texels) depending on OS version.

The new implementation allows that limitation to be squared (16M/32M/64M texels)...again with or without MoltenVK.

HansKristian-ARM: Ok, fair enough. But at least the option name could be changed (with a comment) to make it more clear why it's there?

HansKristian-ARM: Out of curiosity, how is this implemented? Can you alias a texture on top of a linear buffer in Metal?

billhollings: It is important that the Metal app creates the texture with the same width as the MSL limit...which is why I just took the max width as the easiest measure of a consistent value. And I figured that limit might have other uses within SPIRV-Cross down the road at some point.

But I guess you're thinking we should explicitly dedicate this setting to texel texture widths? So perhaps a name like `texel_buffer_texture_width`?

HansKristian-ARM: Yup, I think a name like texel_buffer_texture_width makes more sense. If we need actual texture sizes in a shader later, we could add that separately.

billhollings: Done.

HansKristian-ARM:
Will merge when travis passes.

PR #627, opened 2018-06-28
Add basic namespace to internal macros

HansKristian-ARM:
Some projects build SPIRV-Cross as a single translation unit
and this causes a lot of warnings because the same macro is redeclared
multiple times in the different backends. This make sure that each
backend has its own namespace for internal macros.


HansKristian-ARM:
Non-functional change and Linux passes, not going to bother waiting for OSX to start.

PR #629, opened 2018-06-30
Changed OpTypeImage to only flag depth if the op is 1

cedega:
Based on the SPIR-V spec, `Depth` is only guaranteed if the op is 1. An op of 2 may or may not be depth, which is what DXC outputs for all texture types.

https://www.khronos.org/registry/spir-v/specs/1.0/SPIRV.html#OpTypeImage

https://github.com/Microsoft/DirectXShaderCompiler/blob/master/docs/SPIR-V.rst#textures

### DIFF
--- a/spirv_cross.cpp
+++ b/spirv_cross.cpp
@@ -1842,7 +1842,7 @@ void Compiler::parse(const Instruction &instruction)
 		type.basetype = SPIRType::Image;
 		type.image.type = ops[1];
 		type.image.dim = static_cast<Dim>(ops[2]);
-		type.image.depth = ops[3] != 0;
+		type.image.depth = ops[3] == 1;
 		type.image.arrayed = ops[4] != 0;
 		type.image.ms = ops[5] != 0;
 		type.image.sampled = ops[6];


HansKristian-ARM:
Thanks. Does shadow sampling work though? I.e. is the image type for the OpSampledImage at least 1?

cedega:
From what I can tell the shadow sampling works when compiling SPIRV from glslang HLSL.

For DXC however, the `OpTypeImage`'s depth is still 2, so it doesn't cross compile correctly.

It looks like they handled this issue somewhat recently.
https://github.com/Microsoft/DirectXShaderCompiler/issues/1107

Looking at the docs, it seems that the depth value of the `OpTypeImage` doesn't really matter, as whether or not it's a depth compare ultimately depends on the sample. In their case, they rely on `OpImageSampleDrefImplicitLod` deciding whether or not it's a depth sample, not the depth from `OpTypeImage`.

This should probably be how SPIRV-Cross handles it as well, but for now at least the majority of SPIR-V generated from DXC should cross compile correctly.

HansKristian-ARM:
Ew ... :(

PR #644, opened 2018-07-26
CompilerMSL vertex entry point return void when rasterization disabled.

billhollings:
Add CompilerMSL::Options::disable_rasterization input/output API flag.
Disable rasterization via API flag or when writing to textures.
Disable rasterization when shader declares no output.
Add test shaders for vertex no output and write texture forcing void output.

Metal has an IFF relationship between pipeline rasterization and vertex shader void return.

Metal requires that vertex shaders have a void return on any of:
- rasterization is disabled on the pipeline
- writing to textures
- writing to buffers
- no output variables are declared

Furthermore...when a void return is declared by the shader for any of the above reasons, the runtime pipeline must disable rasterization...requiring feedback through the API of the void return.

### DIFF
--- a/reference/opt/shaders-msl/vert/no_stage_out.vert
+++ b/reference/opt/shaders-msl/vert/no_stage_out.vert
@@ -1,0 +1,20 @@
+#include <metal_stdlib>
+#include <simd/simd.h>
+
+using namespace metal;
+
+struct _10
+{
+    uint4 _m0[1024];
+};
+
+struct main0_in
+{
+    uint4 m_19 [[attribute(0)]];
+};
+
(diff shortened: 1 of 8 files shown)


billhollings:
Thanks.

Actually...I think I'll tweak the design to separate the feedback API of disabling rasterization from within the shader conversion...from the initial `CompilerMSL::Options::disable_rasterization`.

In thinking it through...it is kind of a confusing outlier to have one of the options change from within the shader conversion. It would probably be a better design to leave the option itself alone...and provide a separate `CompilerMSL::is_rasterization_disabled()` function.

I'll submit a separate PR for that.

billhollings:
BTW...enjoy your holiday! When are you returning?

HansKristian-ARM:
Yes, the new PR makes sense. I'll be back August 7th.

PR #648, opened 2018-07-27
CompilerMSL disable rasterization on buffer writes in vertex shader.

billhollings:
Continues from PRs #644 & #646 to include writes and atomic writes to buffers from vertex shaders.


HansKristian-ARM:
No problem.

PR #672, opened 2018-08-31
MSL: Support unordered relational operators.

cdavis5e:
The SPIR-V spec says that these check if the operands either are
unordered or satisfy the given condition. So that's just what we'll do,
using Metal's `isunordered()` stdlib function. Apple's optimizers ought
to be able to collapse that to a single unordered compare.


HansKristian-ARM:
Thanks!

PR #673, opened 2018-09-02
MSL: Emit F{Min,Max,Clamp} as fast:: and N{Min,Max,Clamp} as precise::.

cdavis5e:
This roughly matches their semantics in SPIR-V and MSL. For `FMin`,
`FMax`, and `FClamp`, and the Metal functions `fast::min()`,
`fast::max()`, and `fast::clamp()`, the result is undefined if any
operand is NaN. For the 'N' operations and their corresponding MSL
`precise::` functions, the result is consistent with IEEE 754 (first
non-NaN wins; result is NaN if all operands are NaN).

We can only do this with 32-bit floats, though, because Metal only
provides these variants for `float`. `half` only has one variant of
these functions that is presumably consistent with IEEE 754. I guess
that's OK; the SPIR-V spec only says that `F{Min,Max,Clamp}` are
undefined for NaNs. Performance might suffer, though.

### DIFF
--- a/reference/opt/shaders-msl/asm/frag/vector-shuffle-oom.asm.frag
+++ b/reference/opt/shaders-msl/asm/frag/vector-shuffle-oom.asm.frag
@@ -102,14 +102,14 @@ fragment main0_out main0(constant _6& _7 [[buffer(0)]], constant _18& _19 [[buff
     float4 _88 = _7._m2 * _7._m0.xyxy;
     float2 _95 = _88.xy;
     float2 _96 = _88.zw;
-    float2 _97 = clamp(_82 + (float2(0.0, -2.0) * _7._m0.xy), _95, _96);
-    float3 _109 = float3(_11._m5) * clamp(_8.sample(_9, _97, level(0.0)).w * _7._m1, 0.0, 1.0);
+    float2 _97 = fast::clamp(_82 + (float2(0.0, -2.0) * _7._m0.xy), _95, _96);
+    float3 _109 = float3(_11._m5) * fast::clamp(_8.sample(_9, _97, level(0.0)).w * _7._m1, 0.0, 1.0);
     float4 _113 = _12.sample(_13, _97, level(0.0));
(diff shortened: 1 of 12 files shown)


HansKristian-ARM:
According to travis, some of the SPIR-V modules are invalid.

cdavis5e:
How the heck did *that* happen? All I did was change the lowering of `{F,N}{Min,Max,Clamp}`, and now the variables in some of the generated shaders are missing names? And I can't reproduce this locally.

HansKristian-ARM:
Did you run validation locally first? I'll try to run it locally tomorrow and see what's going on.

cdavis5e:
Well, at least we know now it's reproducible on CI.

cdavis5e:
All right, I am able to reproduce this locally, using the exact version that CI uses. If I use a later version of glslang, though, the problem goes away.

cdavis5e:
For the record, I *don't* recommend using a later version—particularly of SPIRV-Tools. It's stricter about validation, which makes some of the other tests fail—especially those that use `half`, since the `VK_AMD_gpu_shader_half_float` extension is required to support that.

cdavis5e:
Actually, I misspoke. It's not using a later glslang that makes the problem go away. Rather, it's when I run spirv-cross manually. I don't know why it's different when the test script runs it.

cdavis5e:
I figured out where I went wrong. The failing test cases were for the optimized output of the shader compiler. I had simply not updated them!

That shouldn't be a problem in the future, now that I've got it set up so I can test this locally.

HansKristian-ARM:
What is the default here for clamp? If it's fast:: would it be possible to avoid clutter and only emit fast:: for the "precise" variants?

cdavis5e:
It depends. If fast math is turned on, it's `fast::`. Otherwise, it's `precise::`.

HansKristian-ARM:
Ok, guess it's okay as-is then. Thanks.

PR #676, opened 2018-09-05
MSL: Correct number of words to skip in OpImageWrite.

cdavis5e:
The length field in `Instruction` doesn't include the initial
opcode/length word. We only need to skip three words instead of four.


HansKristian-ARM:
Thanks. Will merge when Travis goes through.

PR #680, opened 2018-09-06
MSL: Account for components when assigning locations to varyings.

cdavis5e:
Two varyings (vertex outputs/fragment inputs) might have the same
location but be in different components--e.g. the compiler may have
packed what were two different varyings into a single varying vector.
Giving both varyings the same `[[user]]` attribute won't work--it may
yield unexpected results, or flat out fail to link. We could eventually
pack such varyings into a single vector, but that would require us to
handle the case where the varyings are different types--e.g. a `float`
and a `uint` packed into the same vector. For now, it seems most
prudent to give them unique `[[user]]` locations and let Apple's
compiler work out the best way to pack them.

### DIFF
--- a/reference/opt/shaders-msl/asm/frag/locations-components.asm.frag
+++ b/reference/opt/shaders-msl/asm/frag/locations-components.asm.frag
@@ -11,10 +11,10 @@ struct main0_out
 struct main0_in
 {
     float2 m_2 [[user(locn1)]];
-    float m_3 [[user(locn1)]];
+    float m_3 [[user(locn1_2)]];
     float m_4 [[user(locn2), flat]];
-    uint m_5 [[user(locn2)]];
-    uint m_6 [[user(locn2)]];
+    uint m_5 [[user(locn2_1)]];
+    uint m_6 [[user(locn2_2)]];
 };
 
 fragment main0_out main0(main0_in in [[stage_in]])
--- a/reference/shaders-msl/asm/frag/locations-components.asm.frag
+++ b/reference/shaders-msl/asm/frag/locations-components.asm.frag
@@ -11,10 +11,10 @@ struct main0_out
 struct main0_in
 {
     float2 m_2 [[user(locn1)]];
-    float m_3 [[user(locn1)]];
+    float m_3 [[user(locn1_2)]];
     float m_4 [[user(locn2), flat]];
-    uint m_5 [[user(locn2)]];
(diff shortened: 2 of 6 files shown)


reference/opt/shaders-msl/asm/frag/locations-components.asm.frag
```diff
@@ -11,10 +11,10 @@ struct main0_out
 struct main0_in
 {
     float2 m_2 [[user(locn1)]];
-    float m_3 [[user(locn1)]];
+    float m_3 [[user(locn1_2)]];
```
HansKristian-ARM: Does "locn" mean anything in MSL? Or does MSL link user attributes by name?

spirv_cross.cpp
```diff
@@ -1126,6 +1126,10 @@ void Compiler::set_member_decoration(uint32_t id, uint32_t index, Decoration dec
 		dec.location = argument;
 		break;
 
+	case DecorationComponent:
```
HansKristian-ARM: Hm, this isn't implemented in GLSL or HLSL either. Guess I'll implement GLSL after this PR goes through. Not sure if HLSL supports this?

reference/shaders-msl/asm/frag/locations-components.asm.frag
```diff
@@ -11,10 +11,10 @@ struct main0_out
 struct main0_in
```
HansKristian-ARM: I guess you forgot to push the original test?

cdavis5e: That was actually in the last patch.

reference/opt/shaders-msl/asm/frag/locations-components.asm.frag
cdavis5e: Like HLSL with its semantics, MSL links user attributes by name.

reference/shaders-msl/asm/frag/locations-components.asm.frag
HansKristian-ARM: Oh ... :| you're right.

cdavis5e:
I don't know if `spirv-opt` does this, but I know that Microsoft's HLSL->DXBC compiler *does* do this, which leads us (when we're generating SPIR-V) to do it, too.

HansKristian-ARM:
Ah, right. That makes sense ...

PR #681, opened 2018-09-07
MSL: Do not emit function constants for version < 1.2.

### LINKED ISSUE #679: MSL: `function_constant` not available in OS X 10.11
The MSL compiler appears to emit [`function_constant` unconditionally for specialization constants](https://github.com/KhronosGroup/SPIRV-Cross/blob/master/spirv_msl.cpp#L1530), but this is only available in macOS 10.12+. Is there already a recommended approach for OS X 10.11, or should we try to replace these specialization constants outside of SPIRV-Cross?

cc https://github.com/gfx-rs/gfx/issues/2342#issuecomment-417068547

HansKristian-ARM:
Fix #679.


grovesNL:
Looks great, thanks!

billhollings:
@HansKristian-ARM  Nice work!

PR #684, opened 2018-09-09
MSL: Cast uses of builtin vectors to their declared SPIR-V type.

cdavis5e:
In SPIR-V, builtin integral vectors can be either signed or unsigned,
but in MSL they're always unsigned. Unfortunately, the MSL spec forbids
implicit conversions between vector types--even if the corresponding
scalar types would implicitly convert. If you try, the result is a
cryptic error message such as:

```
program_source:37:60: error: cannot convert between vector values of different size ('int4' (aka 'vector_int4') and 'vector_uint4' (vector of 4 'unsigned int' values))
            float4 r3 = as_type<float4>((as_type<int4>(r0) * gl_LocalInvocationID.xyyy) + as_type<int4>(r2));
                                         ~~~~~~~~~~~~~~~~~ ^ ~~~~~~~~~~~~~~~~~~~~~~~~~
```

Therefore, uses of these builtins must be explicitly cast, since the
rest of the binary likely assumes that the builtin is of its declared
type.

### DIFF
--- a/spirv_msl.cpp
+++ b/spirv_msl.cpp
@@ -2917,7 +2917,14 @@ void CompilerMSL::emit_sampled_image_op(uint32_t result_type, uint32_t result_id
 // Manufacture automatic sampler arg for SampledImage texture.
 string CompilerMSL::to_func_call_arg(uint32_t id)
 {
-	string arg_str = CompilerGLSL::to_func_call_arg(id);
+	// Make sure that we use the name of the original variable, and not the parameter alias.
+	uint32_t name_id = id;
+	auto *var = maybe_get<SPIRVariable>(id);
+	if (var && var->basevariable)
+		name_id = var->basevariable;
+	// Use base to_expression in case this is a builtin. We don't want the
+	// bitcast here.
+	string arg_str = CompilerGLSL::to_expression(name_id);
 
 	// Manufacture automatic sampler arg if the arg is a SampledImage texture.
 	auto &type = expression_type(id);
@@ -3543,7 +3550,7 @@ string CompilerMSL::entry_point_args(bool append_comma)
 					ep_args += ", ";
 
 				BuiltIn bi_type = meta[var_id].decoration.builtin_type;
-				ep_args += builtin_type_decl(bi_type) + " " + to_expression(var_id);
+				ep_args += builtin_type_decl(bi_type) + " " + CompilerGLSL::to_expression(var_id);
 				ep_args += " [[" + builtin_qualifier(bi_type) + "]]";
 			}
 		}
@@ -3668,7 +3675,7 @@ string CompilerMSL::argument_decl(const SPIRFunction::Parameter &arg)
 	{
 		// Arrays of images and samplers are special cased.
 		decl += " (&";
-		decl += to_expression(name_id);
+		decl += builtin ? CompilerGLSL::to_expression(name_id) : to_expression(name_id);
 		decl += ")";
 		decl += type_to_array_glsl(type);
 	}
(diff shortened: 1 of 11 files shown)


spirv_msl.cpp
```diff
@@ -4587,6 +4594,31 @@ void CompilerMSL::bitcast_to_builtin_store(uint32_t, std::string &, const SPIRTy
...
+		// If this is one of the vector builtins that may have a different type
+		// from SPIR-V, we'll have to insert a cast.
+		switch (builtin)
+		{
+		case BuiltInGlobalInvocationId:
+		case BuiltInLocalInvocationId:
+		case BuiltInNumWorkgroups:
+		case BuiltInWorkgroupId:
+			if (type.basetype != SPIRType::UInt)
+				expr = join("as_type<", type_to_glsl(type), ">(", expr, ")");
```
HansKristian-ARM: Wouldn't a simple value cast be enough?

cdavis5e:
Sorry, no. According to the MSL spec, these builtins must have unsigned type. And indeed, Metal Clang errors on this:

```
test.metal:5:30: error: type 'int3' (aka 'vector_int3') is not valid for
      attribute 'threadgroup_position_in_grid'
kernel void test(int3 tgid [[threadgroup_position_in_grid]]) {
                             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
```

spirv_msl.cpp
cdavis5e: That works.

cdavis5e:
Hmm... Travis seems to be down. I made sure the tests pass (on my machine, at least), but I'm not sure if you're willing to take my word for it that this should test OK 😃.

spirv_msl.cpp
```diff
@@ -3668,20 +3675,20 @@ string CompilerMSL::argument_decl(const SPIRFunction::Parameter &arg)
 	{
 		// Arrays of images and samplers are special cased.
 		decl += " (&";
-		decl += to_expression(name_id);
+		decl += builtin ? CompilerGLSL::to_expression(name_id) : to_expression(name_id);
```
HansKristian-ARM: Shouldn't this just use CompilerGLSL::to_expression always? If it's not a builtin, the implementations will be equivalent anyways.

spirv_msl.cpp
```diff
@@ -2917,7 +2917,14 @@ void CompilerMSL::emit_sampled_image_op(uint32_t result_type, uint32_t result_id
 // Manufacture automatic sampler arg for SampledImage texture.
 string CompilerMSL::to_func_call_arg(uint32_t id)
 {
-	string arg_str = CompilerGLSL::to_func_call_arg(id);
+	// Make sure that we use the name of the original variable, and not the parameter alias.
```
HansKristian-ARM: This is so close to CompilerGLSL::to_func_call_arg that not being able to use it feels wrong.

spirv_msl.cpp
```diff
@@ -4587,6 +4594,31 @@ void CompilerMSL::bitcast_to_builtin_store(uint32_t, std::string &, const SPIRTy
 {
 }
 
+std::string CompilerMSL::to_expression(uint32_t id)
```
HansKristian-ARM: This feels a bit ugly, considering all the workarounds needed to pick either GLSL or MSL versions of to_expression above.

In the GLSL backend, there is bitcast_from_builtin_load and bitcast_to_builtin_store which I think would isolate this problem better. These are read-only variables, and they must be loaded first to be used, so we should be able to bitcast in OpLoad if there's a mismatch between builtin and declared type instead of hacking to_expression.

HansKristian-ARM:
You can take a look at: https://github.com/KhronosGroup/SPIRV-Cross/pull/690.
You should be able to do basically the same thing in the MSL override and call it a day.

HansKristian-ARM:
Will merge when travis passes.

PR #686, opened 2018-09-10
Fix composite array declarations in MSL.

### LINKED ISSUE #685: MSL: array initializer issue
I have a GLSL compute shader from VK-GL-CTS:

```GLSL
#version 440

layout(constant_id = 1) const int   sc0 = 1;
layout(constant_id = 2) const float sc1 = 1.0;
layout(constant_id = 3) const bool  sc2 = true;

layout (set = 0, binding = 0, std430) writeonly buffer Output {
    int result;
} sb_out;

struct Data {
    int   x;
    float y;
    bool  z;
};

Data a0[3] = Data[3](Data(sc0, 2.0, true), Data(1, sc1, true), Data(1, 2.0, sc2));

void main (void)
{
    int sum_a0 = 0;

    for (int i = 0; i < 3; ++i)
        sum_a0 += int(a0[i].x) + int(a0[i].y) + int(a0[i].z);

    sb_out.result = sum_a0;
}
```

I convert this to Metal through glslangvalidator and spirv-cross:

`../glslang-build/standalone/glslangvalidator sample.comp -V; and ../spirv-cross/spirv-cross comp.spv --msl > out.metal`
`cat out.metal`

```metal

#include <metal_stdlib>
#include <simd/simd.h>

using namespace metal;

constant int sc0_tmp [[function_constant(1)]];
constant int sc0 = is_function_constant_defined(sc0_tmp) ? sc0_tmp : 1;
constant float sc1_tmp [[function_constant(2)]];
constant float sc1 = is_function_constant_defined(sc1_tmp) ? sc1_tmp : 1.0;
constant bool sc2_tmp [[function_constant(3)]];
constant bool sc2 = is_function_constant_defined(sc2_tmp) ? sc2_tmp : true;

struct Data
{
    int x;
    float y;
    bool z;
};

struct Output
{
    int result;
};

// Implementation of an array copy function to cover GLSL's ability to copy an array via assignment.
template<typename T, uint N>
void spvArrayCopy(thread T (&dst)[N], thread const T (&src)[N])
{
    for (uint i = 0; i < N; dst[i] = src[i], i++);
}

// An overload for constant arrays.
template<typename T, uint N>
void spvArrayCopyConstant(thread T (&dst)[N], constant T (&src)[N])
{
    for (uint i = 0; i < N; dst[i] = src[i], i++);
}

kernel void main0(device Output& sb_out [[buffer(0)]])
{
    Data _24[3] = Data[]{ Data{ sc0, 2.0, true }, Data{ 1, sc1, true }, Data{ 1, 2.0, sc2 } };
    Data a0[3];
    spvArrayCopy(a0, _24);
    int sum_a0 = 0;
    for (int i = 0; i < 3; i++)
    {
        sum_a0 += ((a0[i].x + int(a0[i].y)) + int(a0[i].z));
    }
    sb_out.result = sum_a0;
}
```
Then I try to compile it:
`xcrun -sdk macosx metal out.metal -o out.air`

```
out.metal:43:19: error: 'Data' does not refer to a value
    Data _24[3] = Data[]{ Data{ sc0, 2.0, true }, Data{ 1, sc1, true }, Data{ 1, 2.0, sc2 } };
                  ^
out.metal:15:8: note: declared here
struct Data
       ^
out.metal:43:24: error: expected expression
    Data _24[3] = Data[]{ Data{ sc0, 2.0, true }, Data{ 1, sc1, true }, Data{ 1, 2.0, sc2 } };
                       ^
2 errors generated.
```

So it seems like the MSL compiler doesn't like the array initializer here. I think the `Data[]` isn't necessary but I'll look into it.

HansKristian-ARM:
Fix #685.

Removes the virtual override for constant_expression in MSL. It is no longer needed since the GLSL implementation deals with backend special cases already.

Also fixes a bug in HLSL where chained AccessChains did not work on SSBOs.

### DIFF
--- a/reference/opt/shaders-hlsl/comp/composite-array-initialization.comp
+++ b/reference/opt/shaders-hlsl/comp/composite-array-initialization.comp
@@ -1,0 +1,50 @@
+static const float X = 4.0f;
+static const uint3 gl_WorkGroupSize = uint3(2u, 1u, 1u);
+
+struct Data
+{
+    float a;
+    float b;
+};
+
+static const Data _21 = { 1.0f, 2.0f };
+static const Data _24 = { 3.0f, 4.0f };
+static const Data _25[2] = { { 1.0f, 2.0f }, { 3.0f, 4.0f } };
+static const Data _30 = { 3.0f, 5.0f };
+
+RWByteAddressBuffer _61 : register(u0);
+
+static uint3 gl_WorkGroupID;
+static uint3 gl_LocalInvocationID;
+static uint gl_LocalInvocationIndex;
+struct SPIRV_Cross_Input
+{
+    uint3 gl_WorkGroupID : SV_GroupID;
+    uint3 gl_LocalInvocationID : SV_GroupThreadID;
+    uint gl_LocalInvocationIndex : SV_GroupIndex;
+};
+
+static Data data[2];
+static Data data2[2];
+
+void comp_main()
+{
+    data = _25;
+    Data _28 = { X, 2.0f };
+    Data _31[2] = { _28, _30 };
+    data2 = _31;
+    if (gl_LocalInvocationIndex == 0u)
+    {
+        _61.Store(gl_WorkGroupID.x * 8 + 0, asuint(data[gl_LocalInvocationID.x].a + data2[gl_LocalInvocationID.x].a));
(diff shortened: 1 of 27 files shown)


grovesNL:
Thanks! This works great.

PR #692, opened 2018-09-11
Fix invalid code generation for HLSL SM 3 when fragment shader output is not a vec4

crosire:
The HLSL compiler throws the following error when targeting shader model 3 and an output with the "COLOR" semantic is not a four-component vector:
``` 
error X4529: COLOR0 must be a four-component vector
```
This fixes that by forcing fragment shader outputs to a four-component vector and padding any outputs with less components with zeroes.

Also added some user-specific Visual Studio files to the gitignore for convenience and tests to do future regression checking on the above.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,5 +15,7 @@
 *.a
 *.bc
 /external
+.vs/
+*.vcxproj.user
 
 !CMakeLists.txt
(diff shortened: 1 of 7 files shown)


HansKristian-ARM:
Thanks!

PrinceAssrielDreemurr:
hey Crosire is there a Shader that Changes: darkens & colors the shadow? i really want to see one for my borderlands 1 arid nexus edition reshade

PR #693, opened 2018-09-11
MSL: Fix OpAtomicIIncrement and OpAtomicIDecrement.

cdavis5e:
We were passing a constant '1' to `emit_atomic_func_op()`--which caused
us to refer to SPIR-V value `%1`, which is almost certainly not what we
want! What we really want is to add/subtract the literal constant '1'
to/from the memory location.

### DIFF
--- a/spirv_msl.cpp
+++ b/spirv_msl.cpp
@@ -1784,7 +1784,7 @@ void CompilerMSL::emit_instruction(const Instruction &instruction)
 		uint32_t val = ops[6];
 		uint32_t comp = ops[7];
 		emit_atomic_func_op(result_type, id, "atomic_compare_exchange_weak_explicit", mem_sem_pass, mem_sem_fail, true,
-		                    ptr, comp, true, val);
+		                    ptr, comp, true, false, val);
 		break;
 	}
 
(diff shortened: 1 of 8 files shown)


HansKristian-ARM:
Looks like you forgot to run update_test_shaders.sh before committing.

spirv_msl.cpp
```diff
@@ -2260,7 +2261,7 @@ bool CompilerMSL::maybe_emit_array_assignment(uint32_t id_lhs, uint32_t id_rhs)
 // Emits one of the atomic functions. In MSL, the atomic functions operate on pointers
 void CompilerMSL::emit_atomic_func_op(uint32_t result_type, uint32_t result_id, const char *op, uint32_t mem_order_1,
                                       uint32_t mem_order_2, bool has_mem_order_2, uint32_t obj, uint32_t op1,
-                                      bool op1_is_pointer, uint32_t op2)
+                                      bool op1_is_pointer, bool op1_is_constant, uint32_t op2)
```
HansKristian-ARM: To be consistent with SPIR-V terminology, this should be called "literal" instead of constant.

PR #699, opened 2018-09-13
MSL: Handle the SamplePosition builtin.

cdavis5e:
This is somewhat tricky, because in MSL this value is obtained through a
function, `get_sample_position()`. Since the call expression is an
rvalue, it can't be passed by reference, so functions get a copy
instead.

This was the last piece preventing us from turning on sample-rate
shading support in MoltenVK.

### DIFF
--- a/spirv_msl.cpp
+++ b/spirv_msl.cpp
@@ -56,9 +56,11 @@ CompilerMSL::CompilerMSL(const uint32_t *ir, size_t word_count, MSLVertexAttr *p
 
 void CompilerMSL::build_implicit_builtins()
 {
-	if (need_subpass_input)
+	bool need_sample_pos = active_input_builtins.get(BuiltInSamplePosition);
+	if (need_subpass_input || need_sample_pos)
 	{
 		bool has_frag_coord = false;
+		bool has_sample_id = false;
 
 		for (auto &id : ids)
 		{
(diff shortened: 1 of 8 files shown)


HansKristian-ARM:
Hi, I didn't have a chance to look at this today, I'll get to it on Monday.

spirv_msl.cpp
```diff
@@ -3680,7 +3718,13 @@ string CompilerMSL::argument_decl(const SPIRFunction::Parameter &arg)
 	if (arg.alias_global_variable && var.basevariable)
 		name_id = var.basevariable;
 
-	bool constref = !arg.alias_global_variable && type.pointer && arg.write_count == 0;
+	bool builtin = is_builtin_variable(var);
```
HansKristian-ARM: This feels a bit too invasive.

spirv_msl.cpp
```diff
@@ -4137,6 +4180,13 @@ string CompilerMSL::builtin_to_glsl(BuiltIn builtin, StorageClass storage)
 
 		break;
 
+	// This one is a strange beast, because it requires a function call in
+	// Metal--at least, in the entry point function. In other functions,
+	// we can handle it normally.
+	case BuiltInSamplePosition:
+		if (storage == StorageClassInput && current_function && (current_function->self == entry_point))
```
HansKristian-ARM: A bit too hackish I think, there should be a cleaner way.

HansKristian-ARM:
You'll want to see this PR: https://github.com/KhronosGroup/SPIRV-Cross/pull/698.
I refactored how the fixup_hook/statements work, so you probably want to rebase.

PR #706, opened 2018-09-20
MSL: Add an option to insert texture swizzles into generated shaders.

cdavis5e:
It's intended to be used with MoltenVK to support arbitrary
`VkComponentMapping` settings. The idea is that MoltenVK will pass a
buffer (which it set to some buffer index that isn't being used)
containing packed versions of the `VkComponentMapping` struct, one for
each sampled image.

Yes, this is horribly ugly. It is unfortunately necessary. Much of the
ugliness is to support swizzling gather operations, where we need to
alter the component that the gather operates on--something complicated
by the `gather()` method requiring the passed-in component to be a
constant expression. It doesn't even support swizzling gathers on depth
textures, though I could add that if it turns out we need it.

### DIFF
--- a/reference/shaders-msl-no-opt/asm/frag/texture-access.swizzle.asm.frag
+++ b/reference/shaders-msl-no-opt/asm/frag/texture-access.swizzle.asm.frag
@@ -5,6 +5,11 @@
 
 using namespace metal;
 
+struct spvAux
+{
+    uint swizzleConst[1];
+};
+
 // Returns 2D texture coords corresponding to 1D texel buffer coords
 uint2 spvTexelBufferCoord(uint tc)
 {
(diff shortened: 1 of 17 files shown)


HansKristian-ARM:
Oh ... dear. ... ._____.

Is the idea here to have two shader variants? One "clean", and one to deal with swizzling if there is at least one non-identity VkComponentMapping? Did you evaluate the possibility of having one shader variant per combination of VkComponentMapping so you could statically swizzle the texture? (I suppose that would be rather insane too).

cdavis5e:
I did. It would require us to recompile the shader every time a descriptor for an image with a non-identity swizzle were bound or pushed. And the code wouldn't look much better, since we have to contend with ZERO and ONE swizzles as well.

...I wonder if it would be best to discuss this with @billhollings.

HansKristian-ARM:
I'm fine with either approach from the aspect of SPIRV-Cross if this is truly needed, although as you say, it's extremely ugly.

I was under the impression though that the portability layers would not attempt to go to this length of emulation. Is there some important case that has changed this stance? (This might affect how we deal with odd-ball cases like textureQueryLod ...)

cdavis5e:
There is. We at CodeWeavers need stencil data from depth/stencil textures to always be in green for our D3D emulation--something which cannot be done without swizzling. There are also games that need swizzling--DOOM and Wolfenstein come to mind (no surprise there, since id Software promulgated the original `GL_ARB_texture_swizzle` extension in the first place). And then it turns out that D3D12 also added support for component swizzling.

HansKristian-ARM:
Do you know of any significant performance implication of doing dynamic swizzling like this (DOOM / Wolf)? Hopefully it's just limited to a few not very important passes.

I'll review tomorrow.

HansKristian-ARM:
And even if we went with static swizzle, we would need this fallback path for arrays of textures, so it'd get even more complicated. :( Have you filed this bug to Apple? Lack of texture swizzle is quite glaring.

cdavis5e:
I filed rdar://44287804, which turns out to be a dupe of rdar://39856026--so this is a known issue to Apple.

reference/shaders-msl-no-opt/asm/frag/texture-access.swizzle.asm.frag
```diff
@@ -0,0 +1,157 @@
...
+{
+    return static_cast<thread T&&>(x);
+}
+template<typename T> inline constexpr thread T&& spvForward(thread typename spvRemoveReference<T>::type&& x)
+{
+    return static_cast<thread T&&>(x);
+}
+
+template<typename T>
+inline T spvGetSwizzle(vec<T, 4> x, spvSwizzle s)
```
HansKristian-ARM: Does this work with int/uint textures as well?

reference/shaders-msl-no-opt/asm/frag/texture-access.swizzle.asm.frag
```diff
@@ -0,0 +1,157 @@
...
+enum class spvSwizzle : uint
+{
+    none = 0,
+    zero,
+    one,
+    red,
+    green,
+    blue,
+    alpha
+} ;
```
HansKristian-ARM: Space between } and ;

reference/shaders-msl-no-opt/asm/frag/texture-access.swizzle.asm.frag
```diff
@@ -0,0 +1,157 @@
...
+    return 0;
+}
+
+// Wrapper function that swizzles texture samples and fetches.
+template<typename T>
+inline vec<T, 4> spvTextureSwizzle(vec<T, 4> x, uint s)
+{
+    if (!s)
+        return x;
+    return vec<T, 4>(spvGetSwizzle(x, spvSwizzle((s >> 0) & 0x7)), spvGetSwizzle(x, spvSwizzle((s >> 3) & 0x7)), spvGetSwizzle(x, spvSwizzle((s >> 6) & 0x7)), spvGetSwizzle(x, spvSwizzle((s >> 9) & 0x7)));
```
HansKristian-ARM: It might make more sense to pack it in u8, bitcasting u32 to 4xu8 seems like it would allow more efficient codegen.

shaders-msl-no-opt/frag/texture-access.swizzle.frag
```diff
@@ -0,0 +1,79 @@
...
+layout(binding = 4) uniform sampler2DArray tex2dArray;
+layout(binding = 5) uniform samplerCubeArray texCubeArray;
+layout(binding = 6) uniform samplerBuffer texBuffer;
+
+layout(binding = 7) uniform sampler2DShadow depth2d;
+layout(binding = 8) uniform samplerCubeShadow depthCube;
+layout(binding = 9) uniform sampler2DArrayShadow depth2dArray;
+layout(binding = 10) uniform samplerCubeArrayShadow depthCubeArray;
+
+void main()
```
HansKristian-ARM: You'll need to test sampling from a leaf function. I don't think the swizzle buffer is being propagated properly.

spirv_msl.cpp
```diff
@@ -1584,6 +1584,113 @@ void CompilerMSL::emit_custom_functions()
...
+			statement("enum class spvSwizzle : uint");
+			begin_scope();
+			statement("none = 0,");
+			statement("zero,");
+			statement("one,");
+			statement("red,");
+			statement("green,");
+			statement("blue,");
+			statement("alpha");
+			end_scope_decl("");
```
HansKristian-ARM: This might be the code causing " ;". Maybe add a special case for empty string here.

spirv_msl.cpp
```diff
@@ -2958,6 +3102,22 @@ string CompilerMSL::to_function_args(uint32_t img, const SPIRType &imgtype, bool
...
+		if (!is_gather)
+			farg_str += ")";
+		// Get the original input variable for this image.
+		uint32_t img_var = img;
+		if (meta[img].image)
+			img_var = meta[img].image;
+		if (auto *var = maybe_get_backing_variable(img_var))
+			img_var = var->self;
+		farg_str += ", spvSwizzleConst[" +
+		            convert_to_string(get_metal_resource_index(get<SPIRVariable>(img_var), SPIRType::Image)) + "]";
```
HansKristian-ARM: For sampling inside functions, I think this won't work. We won't know this information, and we might have to pass down swizzles to leaf functions, just like we sometimes have to pass down combined samplers to leaf functions in MSL.

spirv_common.hpp
```diff
@@ -1237,6 +1237,7 @@ struct Meta
 	Decoration decoration;
 	std::vector<Decoration> members;
 	uint32_t sampler = 0;
+	uint32_t image = 0;
```
HansKristian-ARM: I think I'll need to refactor this to SPIRCombinedSampler that HLSL uses eventually ...

reference/shaders-msl-no-opt/asm/frag/texture-access.swizzle.asm.frag
cdavis5e: It should. That's why it's a template function after all.

reference/shaders-msl-no-opt/asm/frag/texture-access.swizzle.asm.frag
cdavis5e: I don't know how to get rid of that. It's generated by the `end_scope_decl()` function, which assumes its string argument is non-empty.

HansKristian-ARM: Ok, should be trivial. I can fix that later.

reference/shaders-msl-no-opt/asm/frag/texture-access.swizzle.asm.frag
HansKristian-ARM: Right, I was thinking there would be ivec/uvec ala GLSL, but that'd be weird, I see T is part of the template.

reference/shaders-msl-no-opt/asm/frag/texture-access.swizzle.asm.frag
cdavis5e: Oh, there's a variant that doesn't take an argument and which doesn't put the space in. Silly me.

reference/shaders-msl-no-opt/asm/frag/texture-access.swizzle.asm.frag
cdavis5e: I've added some tests to show that we handle this for integer textures.

spirv_msl.cpp
```diff
@@ -4422,6 +4645,59 @@ bool CompilerMSL::skip_argument(uint32_t) const
 	return false;
 }
 
+void CompilerMSL::analyze_image_and_sampler_usage()
```
HansKristian-ARM: Overriding analyze_image_and_sampler_usage seems a bit off here. This is never called from a parent class (i.e. it would need to be overridden). It seems more direct to move the swizzle_texture_samples path to its own member function in CompilerMSL, and avoid the virtual. CompilerMSL::compile() would then just call analyze_image_and_sampler_usage followed by this one.

spirv_msl.cpp
```diff
@@ -4422,6 +4645,59 @@ bool CompilerMSL::skip_argument(uint32_t) const
...
+	case OpImageGather:
+		compiler.has_sampled_images |= is_sampled_image_type(compiler.expression_type(args[2]));
+		break;
+	default:
+		break;
+	}
+	return true;
+}
+
+bool CompilerMSL::SampledImageScanner::is_sampled_image_type(const SPIRType &type)
```
HansKristian-ARM: This seems generally useful for other paths in this patch, so might make sense to hoist it out to Compiler.

HansKristian-ARM: Should probably add a check for type.basetype == SPIRType::Image || type.basetype == SPIRType::SampledImage in that case to make it general.

spirv_msl.hpp
```diff
@@ -152,9 +152,11 @@ class CompilerMSL : public CompilerGLSL
 		Platform platform = macOS;
 		uint32_t msl_version = make_msl_version(1, 2);
 		uint32_t texel_buffer_texture_width = 4096; // Width of 2D Metal textures used as 1D texel buffers
+		uint32_t aux_buffer_index = 0;
```
HansKristian-ARM: Is there any way to query if this buffer has actually been used after compilation finishes?

spirv_msl.hpp
```diff
@@ -374,10 +377,13 @@ class CompilerMSL : public CompilerGLSL
 	void emit_entry_point_declarations() override;
 	uint32_t builtin_frag_coord_id = 0;
 	uint32_t builtin_sample_id_id = 0;
+	uint32_t aux_buffer_id = 0;
 
 	void bitcast_to_builtin_store(uint32_t target_id, std::string &expr, const SPIRType &expr_type) override;
 	void bitcast_from_builtin_load(uint32_t source_id, std::string &expr, const SPIRType &expr_type) override;
 
+	void analyze_image_and_sampler_usage() override;
```
HansKristian-ARM: This shouldn't be virtual as mentioned earlier.

spirv_msl.hpp
```diff
@@ -419,6 +426,20 @@ class CompilerMSL : public CompilerGLSL
...
+	// OpcodeHandler that scans for uses of sampled images
+	struct SampledImageScanner : OpcodeHandler
+	{
+		SampledImageScanner(CompilerMSL &compiler_)
+		    : compiler(compiler_)
+		{
+		}
+
+		bool handle(spv::Op opcode, const uint32_t *args, uint32_t) override;
+		bool is_sampled_image_type(const SPIRType &type);
```
HansKristian-ARM: Can be a Compiler utility function probably.

spirv_msl.cpp
```diff
@@ -4422,6 +4645,59 @@ bool CompilerMSL::skip_argument(uint32_t) const
...
+	case OpImageSampleProjExplicitLod:
+	case OpImageSampleDrefExplicitLod:
+	case OpImageSampleProjDrefExplicitLod:
+	case OpImageSampleImplicitLod:
+	case OpImageSampleProjImplicitLod:
+	case OpImageSampleDrefImplicitLod:
+	case OpImageSampleProjDrefImplicitLod:
+	case OpImageFetch:
+	case OpImageGather:
+		compiler.has_sampled_images |= is_sampled_image_type(compiler.expression_type(args[2]));
```
HansKristian-ARM: Boolean bitwise ops creates warnings on MSVC, best to make it simpler.

spirv_msl.cpp
```diff
@@ -2680,11 +2847,42 @@ void CompilerMSL::emit_function_prototype(SPIRFunction &func, const Bitset &)
...
+		    "spvGatherSwizzle<" + type_to_glsl(get<SPIRType>(imgtype.image.type)) + ", " + type_to_glsl(imgtype);
+		// Add the arg types ourselves. Yes, this sucks, but Clang can't
+		// deduce template pack parameters in the middle of an argument list.
+		switch (imgtype.image.dim)
+		{
+		case Dim2D:
+			fname += ", float2";
+			if (imgtype.image.arrayed)
+				fname += ", uint";
+			fname += ", int2";
```
HansKristian-ARM: What is this int2 for?

spirv_msl.cpp
```diff
@@ -139,6 +141,53 @@ void CompilerMSL::build_implicit_builtins()
...
+		struct_type_ptr.pointer = true;
+		struct_type_ptr.parent_type = struct_id;
+		struct_type_ptr.storage = StorageClassUniform;
+		auto &ptr_type = set<SPIRType>(struct_ptr_id, struct_type_ptr);
+		ptr_type.self = struct_id;
+
+		set<SPIRVariable>(var_id, struct_ptr_id, StorageClassUniform);
+		set_name(var_id, "spvAuxBuffer");
+		// This should never match anything.
+		set_decoration(var_id, DecorationDescriptorSet, 0xFFFFFFFE);
```
HansKristian-ARM: Why 0xfffffffe in particular? Does it need a descriptor set decoration?

spirv_msl.cpp
```diff
@@ -28,6 +28,8 @@ using namespace std;
 static const uint32_t k_unknown_location = ~0u;
 static const uint32_t k_unknown_component = ~0u;
 
+static const uint32_t k_aux_mbr_idx_swizzle_const = 0u;
```
HansKristian-ARM: Is this used?

spirv_msl.cpp
```diff
@@ -2958,6 +3163,23 @@ string CompilerMSL::to_function_args(uint32_t img, const SPIRType &imgtype, bool
...
+		if (!is_gather)
+			farg_str += ")";
+		// Get the original input variable for this image.
+		uint32_t img_var = img;
+		if (meta[img].image)
+			img_var = meta[img].image;
+		if (auto *var = maybe_get_backing_variable(img_var))
+			img_var = var->self;
+		auto &aux_type = expression_type(aux_buffer_id);
+		farg_str += ", " + to_name(aux_buffer_id) + "." + to_member_name(aux_type, k_aux_mbr_idx_swizzle_const) + "[" +
```
HansKristian-ARM: This will still only work for images which are passed implicitly as a global, not as a parameter. But that'd be too insane to deal with for now. Could you add a check here, and error out if it cannot find a valid resource index? (i.e. it's a parameter, not a global alias).

spirv_msl.cpp
cdavis5e: Yes it is.

spirv_msl.cpp
cdavis5e: It needs a unique descriptor set so it doesn't collide with any descriptor sets specified by the client, or with the special push constant descriptor set.

spirv_msl.cpp
cdavis5e: It's the offset parameter.

spirv_msl.hpp
cdavis5e: It should only ever be emitted if it would be used. That's what the `SampledImageScanner` checks for.

HansKristian-ARM: Right, but the consumer might want to know if it should bind and create the spvAux buffer on the driver side.

spirv_msl.cpp
HansKristian-ARM: Oh, I see.

spirv_msl.cpp
HansKristian-ARM: Fair enough.

spirv_msl.hpp
HansKristian-ARM: Guess we could just add this later when we get to integration, it should be trivial to add when needed.

HansKristian-ARM:
Thanks for all the work and changes.

PR #713, opened 2018-09-27
MSL: Expand arrays of buffers passed as input.

cdavis5e:
Even as of Metal 2.1, MSL still doesn't support arrays of buffers
directly. Therefore, we must manually expand them. In the prologue, we
define arrays holding the argument pointers; these arrays are what the
transpiled code ends up referencing. We might be able to do similar
things for textures and samplers prior to MSL 2.0.

Speaking of which, also enable texture arrays on iOS MSL 1.2.


spirv_msl.cpp
```diff
@@ -4204,7 +4250,13 @@ string CompilerMSL::image_type_glsl(const SPIRType &type, uint32_t id)
 
 	if (!type.array.empty())
 	{
-		if (!msl_options.supports_msl_version(2))
+		uint32_t major = 2, minor = 0;
+		if (msl_options.is_ios())
```
HansKristian-ARM: Huh? iOS Metal 1.2 is the same as 2.0 desktop?

spirv_msl.cpp
```diff
@@ -334,6 +334,20 @@ void CompilerMSL::emit_entry_point_declarations()
...
+	for (uint32_t array_id : buffer_arrays)
+	{
+		const auto &var = get<SPIRVariable>(array_id);
+		const auto &type = get<SPIRType>(var.basetype);
+		string name = get_name(array_id);
+		statement(get_argument_address_space(var) + " " + type_to_glsl(type) + "* " + name + "[] =");
+		begin_scope();
+		for (uint32_t i = 0; i < type.array[0]; ++i)
+			statement(name + "_" + convert_to_string(i) + ",");
+		end_scope_decl();
```
HansKristian-ARM: Probably need to emit a blank line here after declaring the array for cleaner output, the declarations are smashed together.

spirv_msl.cpp
```diff
@@ -3793,10 +3807,30 @@ string CompilerMSL::entry_point_args(bool append_comma)
...
-			if (!ep_args.empty())
-				ep_args += ", ";
-			ep_args += get_argument_address_space(var) + " " + type_to_glsl(type) + "& " + r.name;
-			ep_args += " [[buffer(" + convert_to_string(r.index) + ")]]";
+			if (!type.array.empty())
+			{
+				// Metal doesn't directly support this, so we must expand the
+				// array. We'll declare a local array to hold these elements
+				// later.
+				uint32_t array_size =
```
HansKristian-ARM: Unsized array (bindless) or array-of-array won't work here. You should throw error for that case.

spirv_msl.cpp
```diff
@@ -3793,10 +3807,30 @@ string CompilerMSL::entry_point_args(bool append_comma)
 			auto &m = meta.at(type.self);
 			if (m.members.size() == 0)
 				break;
-			if (!ep_args.empty())
-				ep_args += ", ";
-			ep_args += get_argument_address_space(var) + " " + type_to_glsl(type) + "& " + r.name;
-			ep_args += " [[buffer(" + convert_to_string(r.index) + ")]]";
+			if (!type.array.empty())
+			{
+				// Metal doesn't directly support this, so we must expand the
```
HansKristian-ARM: Does this apply for array of textures as well? At least MSL 2.0 supports "proper" array of textures.

spirv_msl.cpp
```diff
@@ -3969,6 +4003,9 @@ string CompilerMSL::argument_decl(const SPIRFunction::Parameter &arg)
 	else if (is_array(type) && !type_is_image)
 	{
 		// Arrays of images and samplers are special cased.
+		if (get<SPIRVariable>(name_id).storage == StorageClassUniform ||
```
HansKristian-ARM: Probably need a comment here why this is needed.

spirv_msl.cpp
```diff
@@ -4088,6 +4125,15 @@ void CompilerMSL::replace_illegal_names()
 	CompilerGLSL::replace_illegal_names();
 }
 
+string CompilerMSL::to_member_reference(const SPIRVariable *var, const SPIRType &type, uint32_t index)
+{
+	// If this is a buffer array, we have to dereference the buffer pointers.
+	if (var && (var->storage == StorageClassUniform || var->storage == StorageClassStorageBuffer) &&
+	    !get<SPIRType>(var->basetype).array.empty())
+		return join("->", to_member_name(type, index));
+	return join(".", to_member_name(type, index));
```
HansKristian-ARM: Should put this in an else block, had to read it twice.

HansKristian-ARM:
Seems reasonable overall. Annoying that so many of these features are not properly supported by the language ...

HansKristian-ARM:
On a side note I just merged https://github.com/KhronosGroup/SPIRV-Cross/pull/714, so you might need to update your glslang build for tests.

spirv_msl.cpp
cdavis5e: Prior to MSL 2.0 on desktop (1.2 on mobile), yes. (I did mention this possibility in the initial description of this PR.)

HansKristian-ARM: Oh, ye, now I see.

spirv_msl.cpp
```diff
@@ -334,6 +334,22 @@ void CompilerMSL::emit_entry_point_declarations()
...
+		string name = get_name(array_id);
+		statement(get_argument_address_space(var) + " " + type_to_glsl(type) + "* " + name + "[] =");
+		begin_scope();
+		for (uint32_t i = 0; i < type.array[0]; ++i)
+			statement(name + "_" + convert_to_string(i) + ",");
+		end_scope_decl();
+		statement("");
+	}
+	// For some reason, without this, we end up emitting the arrays twice.
+	buffer_arrays.clear();
```
HansKristian-ARM: This is because the compiler runs in a loop if we need to compile again (if we detected funkiness). State which is built up in the compilation loop needs to be cleared out. This is typically done in ::reset(), but it's fine to do it here as well.

HansKristian-ARM:
Thanks!

PR #717, opened 2018-10-03
MSL: Swizzle gathers on depth textures as well.

cdavis5e:
Might as well.

### DIFF
--- a/reference/shaders-msl-no-opt/asm/frag/texture-access.swizzle.asm.frag
+++ b/reference/shaders-msl-no-opt/asm/frag/texture-access.swizzle.asm.frag
@@ -114,6 +114,29 @@ inline vec<T, 4> spvGatherSwizzle(sampler s, thread Tex& t, Ts... params, compon
     }
 }
 
+// Wrapper function that swizzles depth texture gathers.
+template<typename T, typename Tex, typename... Ts>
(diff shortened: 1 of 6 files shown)


reference/shaders-msl-no-opt/asm/frag/texture-access.swizzle.asm.frag
```diff
@@ -114,6 +114,29 @@ inline vec<T, 4> spvGatherSwizzle(sampler s, thread Tex& t, Ts... params, compon
...
+        switch (spvSwizzle(sw & 0xFF))
+        {
+            case spvSwizzle::none:
+            case spvSwizzle::red:
+                break;
+            case spvSwizzle::zero:
+            case spvSwizzle::green:
+            case spvSwizzle::blue:
+            case spvSwizzle::alpha:
+                return vec<T, 4>(0, 0, 0, 0);
```
HansKristian-ARM: Sure you sure alpha should be 0?

cdavis5e: No. I actually wanted to change that before making this PR, but I forgot. TBH, I'm not sure alpha = 1 is correct, either.

HansKristian-ARM: We could fix later if there's an issue (but swizzing alpha in a depth compare is pretty far out there).

PR #718, opened 2018-10-03
MSL: Handle OpImage on OpSampledImage expressions.

cdavis5e:
I have seen this happen. The included test case is one such case.


shaders-msl/asm/frag/op-image-sampled-image.asm.frag
```diff
@@ -0,0 +1,495 @@
+; SPIR-V
```
HansKristian-ARM: This is a pretty huge test shader for something that could be simple.
Is it possible to make a simpler test shader in asm?

PR #734, opened 2018-10-30
Fix Options::force_temporary to work with OpenGL GLSL

zeux:
Setting force_temporary to true produces invalid GLSL because sampler
variables are copied:

    highp sampler2D _377 = DiffuseMapTexture;

This change fixes the problem by always forwarding forwardable
variables. I also took an opportunity to restructure the code to make
it easier to read and add extra conditions to in the future.


HansKristian-ARM:
Thanks!

PR #738, opened 2018-11-01
Add support for 8- and 16-bit types to GLSL and MSL.

cdavis5e:
In GLSL, 8-bit types require `GL_EXT_shader_8bit_storage`. 16-bit types
can use either `GL_AMD_gpu_shader_int16`/`GL_AMD_gpu_shader_half_float` or
`GL_EXT_shader_16bit_storage`.

### DIFF
--- a/spirv_glsl.cpp
+++ b/spirv_glsl.cpp
@@ -3351,6 +3351,72 @@ string CompilerGLSL::constant_expression_vector(const SPIRConstant &c, uint32_t 
 		}
 		break;
 
+	case SPIRType::UShort:
+		if (splat)
+		{
+			res += convert_to_string(c.scalar(vector, 0));
+			if (is_legacy())
+			{
+				// Fake unsigned constant literals with signed ones if possible.
+				// Things like array sizes, etc, tend to be unsigned even though they could just as easily be signed.
+				if (c.scalar_i16(vector, 0) < 0)
+					SPIRV_CROSS_THROW("Tried to convert uint literal into int, but this made the literal negative.");
+			}
+			else if (backend.uint16_t_literal_suffix)
+				res += backend.uint16_t_literal_suffix;
+		}
+		else
+		{
+			for (uint32_t i = 0; i < c.vector_size(); i++)
+			{
+				if (c.vector_size() > 1 && c.specialization_constant_id(vector, i) != 0)
+					res += to_name(c.specialization_constant_id(vector, i));
+				else
+				{
+					res += convert_to_string(c.scalar(vector, i));
+					if (is_legacy())
+					{
+						// Fake unsigned constant literals with signed ones if possible.
+						// Things like array sizes, etc, tend to be unsigned even though they could just as easily be signed.
+						if (c.scalar_i16(vector, i) < 0)
+							SPIRV_CROSS_THROW(
+							    "Tried to convert uint literal into int, but this made the literal negative.");
+					}
+					else if (backend.uint16_t_literal_suffix)
+						res += backend.uint16_t_literal_suffix;
+				}
+
+				if (i + 1 < c.vector_size())
+					res += ", ";
+			}
+		}
+		break;
+
+	case SPIRType::Short:
+		if (splat)
+		{
+			res += convert_to_string(c.scalar_i16(vector, 0));
+			if (backend.int16_t_literal_suffix)
+				res += backend.int16_t_literal_suffix;
+		}
+		else
+		{
+			for (uint32_t i = 0; i < c.vector_size(); i++)
+			{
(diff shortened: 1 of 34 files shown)


spirv_glsl.cpp
```diff
@@ -366,6 +366,14 @@ void CompilerGLSL::find_static_extensions()
 
 			if (type.basetype == SPIRType::Half)
 				require_extension_internal("GL_AMD_gpu_shader_half_float");
+
+			if (type.basetype == SPIRType::Int || type.basetype == SPIRType::UInt)
```
HansKristian-ARM: Hm, this is inconsistent with how FP16 was implemented, using Half, and 64-bit types.
Either we should rename Half to Float and use width, or add SByte/SByte/Short/UShort as new types.

I think I prefer the latter, because it avoids problems where we see Int/UInt but forget to check for width explicitly and it makes failures to take width into account more obvious because we end up hitting unknown types.

cdavis5e:
> 8-bit/16-bit constants are not taken into account. Especially signed types need special care, with appropriate suffixes on the types (for 16-bit, value cast constructors for 8-bit constants).

I think we need to use constructors in all cases. `GL_EXT_shader_16bit_storage` doesn't add suffixes for 16-bit integers, though I see the AMD extension does.

spirv_glsl.cpp
```diff
@@ -3759,6 +3854,11 @@ bool CompilerGLSL::to_trivial_mix_op(const SPIRType &type, string &op, uint32_t
 	bool ret = false;
 	switch (type.basetype)
 	{
+	case SPIRType::Short:
```
HansKristian-ARM: Missing 8-bit here, or is that intentional?

spirv_glsl.cpp
```diff
@@ -3324,6 +3353,72 @@ string CompilerGLSL::constant_expression_vector(const SPIRConstant &c, uint32_t
 		}
 		break;
 
+	case SPIRType::UShort:
```
HansKristian-ARM: Missing 8-bit, or is that intentional?

HansKristian-ARM:
Thanks!

devshgraphicsprogramming:
`NV_gpu_shader5` also supports the same types that extension introduces
CC: @Przemog1

spirv_glsl.cpp
```diff
@@ -516,18 +522,34 @@ void CompilerGLSL::emit_header()
...
 			// GL_AMD_gpu_shader_half_float is a superset, so try that first.
 			statement("#if defined(GL_AMD_gpu_shader_half_float)");
 			statement("#extension GL_AMD_gpu_shader_half_float : require");
-			statement("#elif defined(GL_NV_gpu_shader5)");
-			statement("#extension GL_NV_gpu_shader5 : require");
+			if (!options.vulkan_semantics)
+			{
+				statement("#elif defined(GL_NV_gpu_shader5)");
+				statement("#extension GL_NV_gpu_shader5 : require");
+			}
```
devshgraphicsprogramming, line 535: Is `GL_NV_gpu_shader5` a superset of `GL_EXT_shader_16bit_storage` ?

spirv_glsl.cpp
```diff
@@ -516,18 +522,34 @@ void CompilerGLSL::emit_header()
 
 	for (auto &ext : forced_extensions)
 	{
-		if (ext == "GL_AMD_gpu_shader_half_float" && !options.vulkan_semantics)
+		if (ext == "GL_AMD_gpu_shader_half_float")
 		{
 			// Special case, this extension has a potential fallback to another vendor extension in normal GLSL.
 			// GL_AMD_gpu_shader_half_float is a superset, so try that first.
```
devshgraphicsprogramming, line 528: wont trying a vendor-specific extension with vulkan semantics, cause the GLSL created to require a vendor-specific extension when compiling back down to SPIR-V and hence kill the vendor neutrality the original input would have had?

spirv_glsl.cpp
```diff
@@ -516,18 +522,34 @@ void CompilerGLSL::emit_header()
...
 			statement("#else");
 			statement("#error No extension available for FP16.");
 			statement("#endif");
 		}
+		else if (ext == "GL_AMD_gpu_shader_int16")
+		{
+			// GL_AMD_gpu_shader_int16 is a superset, so try that first.
+			statement("#if defined(GL_AMD_gpu_shader_int16)");
+			statement("#extension GL_AMD_gpu_shader_int16 : require");
+			statement("#elif defined(GL_EXT_shader_16bit_storage)");
```
devshgraphicsprogramming, line 547: why no option for `GL_NV_gpu_shader5` ? it has the same 16bit types AMD and EXT introduce

PR #742, opened 2018-11-07
MSL: Emit wrapper for SSign (sign() for int types)

stenzek:
Metal does not define the sign() function for integer types, only floating-point types. GLSL defines sign() for genType, genDType, and genIType (Section 8.3 - Common Functions).

This PR generates a wrapper function which uses select() to return -1, 0, or 1 depending on the input, using the select() function. Not sure if this is the optimal method of implementing these semantics, but it seems cleaner to me than a series of ternaries, and uses templates like the surrounding wrappers.

Please let me know if I'm going about this the wrong way. If it is acceptable, should I also add a test shader for this function overload?

### DIFF
--- a/reference/opt/shaders-msl/vert/sign-int-types.vert
+++ b/reference/opt/shaders-msl/vert/sign-int-types.vert
@@ -1,0 +1,60 @@
+#pragma clang diagnostic ignored "-Wmissing-prototypes"
+
+#include <metal_stdlib>
+#include <simd/simd.h>
+
+using namespace metal;
+
+struct UBO
+{
+    float4x4 uMVP;
+    float4 uFloatVec4;
+    float3 uFloatVec3;
+    float2 uFloatVec2;
+    float uFloat;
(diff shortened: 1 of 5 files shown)


stenzek:
Tests added for both the float/int scalar and vector overloads. I made them uniforms/outputs to avoid being optimized out, is this an acceptable way of going about it?

HansKristian-ARM:
Normally, if the optimized results optimize away a lot of stuff, that's fine, there's also a test for unoptimized results which will catch any issues with the basic codegen. The optimization testing is there mostly to catch difficult cases in certain scenarios which spirv-opt will throw at us, but this is not such a case. Having more realistic input/output makes for a better test, sure, so doing it like this is fine.

Thanks for the PR.

stenzek:
I see. Will keep this in mind for the future. Thank you for the fast review/merge!

billhollings:
@stenzek  Nice implementation. Thanks!

PR #748, opened 2018-11-10
HLSL: Combine image and sampler on shader model <= 30

gongminmin:
When compiling a fragment shader to a HLSL shader model 30, an exception "Separate image and samplers not supported in legacy HLSL." fired from CompilerHLSL::emit_legacy_uniform.

This change combines image and sampler and skip the separated one, if shader model <= 30.


HansKristian-ARM:
Thanks. Fixed a few additional things on my end, should work similarly to legacy GLSL now.

PR #749, opened 2018-11-12
Deal with depth_greater/depth_less qualifiers.

### LINKED ISSUE #747: Support for SV_DepthGreaterEqual on the HLSL backend
The equivalent feature is already available on the GLSL backend ("depth_greater") and MSL backend ("depth(greater)") using the internal flag ExecutionModeDepthGreater, but it's not implemented for the HLSL backend.

Thanks

HansKristian-ARM:
Adds support on HLSL SM 5.0, and fixes bug on GLSL.
Makes sure early fragment tests is tested on MSL as well.

Fix #747.


HansKristian-ARM:
Spurious weirdness where the OSX build doesn't complete, but otherwise passes everywhere else, so merging.

PR #756, opened 2018-11-13
MSL: Also pack 2- and 4- element vectors when necessary.

cdavis5e:
This is also needed for `VK_KHR_relaxed_block_layout` support.

### DIFF
--- a/reference/shaders-msl/asm/comp/relaxed-block-layout.asm.comp
+++ b/reference/shaders-msl/asm/comp/relaxed-block-layout.asm.comp
@@ -12,9 +12,11 @@ struct foo
     packed_half2 wibble;
 };
 
-kernel void main0(device foo& _6 [[buffer(0)]], uint3 gl_LocalInvocationID [[thread_position_in_threadgroup]], uint3 gl_GlobalInvocationID [[thread_position_in_grid]])
+kernel void main0(device foo& _8 [[buffer(0)]], uint3 gl_LocalInvocationID [[thread_position_in_threadgroup]], uint3 gl_GlobalInvocationID [[thread_position_in_grid]], uint3 gl_WorkGroupID [[threadgroup_position_in_grid]], uint3 gl_NumWorkGroups [[threadgroups_per_grid]])
 {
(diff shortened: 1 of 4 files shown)


HansKristian-ARM:
I can see 2-element vectors, but why 4? Is it legal to emit a vec4 with just 4 byte alignment?

cdavis5e:
> I can see 2-element vectors, but why 4? Is it legal to emit a vec4 with just 4 byte alignment?

No--it would improperly straddle in that case. But it is legal to have a `u8vec4`, for example, do that.

spirv_msl.cpp
```diff
@@ -1231,16 +1231,20 @@ bool CompilerMSL::is_member_packable(SPIRType &ib_type, uint32_t index)
 
 	auto &mbr_type = get<SPIRType>(ib_type.member_types[index]);
 
-	// Only 3-element vectors or 3-row matrices need to be packed.
-	if (mbr_type.vecsize != 3)
+	// Only vectors or 3-row matrices need to be packed.
+	if (mbr_type.vecsize == 1 || (is_matrix(mbr_type) && mbr_type.vecsize != 3))
```
HansKristian-ARM: What about matrices at weird offsets?

reference/shaders-msl/asm/comp/relaxed-block-layout.asm.comp
```diff
@@ -7,6 +7,9 @@ struct foo
 {
     uint bar;
     packed_float3 baz;
+    uchar quux;
+    packed_uchar4 blah;
+    packed_half2 wibble;
```
HansKristian-ARM: I don't see any code trying to read from these types. We need to verify it casts to unpacked types correctly.

spirv_msl.cpp
HansKristian-ARM: As they are sort of arrays of vectors, packing rules might ban that. I guess relaxed_block_layout only relaxes alignment for non-array/matrix types?

cdavis5e: I checked the spec with and without the extension, and yeah, it only applies to vector types.

PR #757, opened 2018-11-13
MSL: Support extended arithmetic opcodes.

cdavis5e:


reference/opt/shaders-msl/desktop-only/comp/extended-arithmetic.desktop.comp
```diff
@@ -0,0 +1,177 @@
...
+{
+    int4 _m0;
+    int4 _m1;
+};
+
+kernel void main0(device SSBOUint& u [[buffer(0)]], device SSBOInt& i [[buffer(1)]])
+{
+    ResType _25;
+    _25._m0 = u.a + u.b;
+    _25._m1 = select(uint(1), uint(0), addsat(u.a, u.b) == UINT_MAX);
```
HansKristian-ARM: Sure this is correct? I would expect overflow detection to be `result < max(a, b)`.

reference/opt/shaders-msl/desktop-only/comp/extended-arithmetic.desktop.comp
```diff
@@ -0,0 +1,177 @@
...
+    u.d3 = _55._m1;
+    u.c3 = _55._m0;
+    ResType_3 _70;
+    _70._m0 = u.a4 + u.b4;
+    _70._m1 = select(uint4(1), uint4(0), addsat(u.a4, u.b4) == UINT_MAX);
+    u.d4 = _70._m1;
+    u.c4 = _70._m0;
+    ResType _79;
+    _79._m0 = u.a - u.b;
+    _79._m1 = select(uint(1), uint(0), subsat(u.a, u.b) == 0);
```
HansKristian-ARM: Don't think is correct either.

PR #767, opened 2018-11-22
call validator from spirv-cross

fjhenigman:
I know the cmake stuff isn't right, I need to figure out what to do for windows.  Clues appreciated.  :)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -35,7 +35,7 @@ endif()
 # -DCMAKE_CXX_COMPILE_FLAGS
 # However, we require the C++11 dialect.
 if (NOT "${MSVC}")
-  set(spirv-compiler-options ${spirv-compiler-options} -std=c++11 -Wall -Wextra -Werror -Wshadow)
+  set(spirv-compiler-options ${spirv-compiler-options} -std=c++11 -Wall -Wextra -Werror -Wshadow -I../external/spirv-tools -I../external/spirv-tools/include)
   set(spirv-compiler-defines ${spirv-compiler-defines} __STDC_LIMIT_MACROS)
 
   if(SPIRV_CROSS_EXCEPTIONS_TO_ASSERTIONS)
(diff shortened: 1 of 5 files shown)


grovesNL:
Why include SPIRV-Tools directly into SPIRV-Cross instead of using the CLIs separately? If SPIRV-Tools is included, it would be good to have a flag so people using SPIRV-Cross as a library could opt out of this dependency.

dj2:
It looks like this CL already includes a --validate off flag to turn off validation from the command line if desired.

HansKristian-ARM:
Right, but that assumes SPIRV-Cross in isolation needs to be robust to invalid input. From a modularity standpoint it makes more sense to run validation before going into SPIRV-Cross if the validity is uncertain, just like for a Vulkan driver. Is there a use case where invoking the validator outside SPIRV-Cross is a problem? For a system which needs to cross-compile unknown SPIR-V I would expect that application to split this process into:

- Parse SPIR-V with SPIRV-Tools and validate.
- Construct spirv_cross::ParsedIR from that.
- Pass ParsedIR to SPIRV-Cross and cross-compile.

The first two steps could easily be done outside SPIRV-Cross.

This PR also just applies validation to the CLI, shouldn't it also need to apply to the C++ API as well in that case?

billhollings:
In addition to my concerns about driver-level runtime integration...I agree with @HansKristian-ARM about whether validation needs to be integrated with SPIRV-Cross at all.

As far as CLI tooling goes...is there a reason why a two-line script could not simply invoke validation and then conversion as two distinct CLI invocations?

dneto0:
Thanks for the feedback. We'll take your suggestion and do the validation+translation integration outside of the SPIRV-Cross codebase.

HansKristian-ARM:
Ok, I think we have agreement on the direction going forward, so closing this PR.
If you run into any issues in integration where API changes to ParsedIR is needed, we can fix it.

HansKristian-ARM:
Right. Not sure exactly what those tests will look like, but sounds reasonable, since we risk breaking the integration API (ParsedIR) without specific integration testing. I assume the CMake build would look something like:

```
# Should only be checked out by CI or maintainers ala the current checkout scripts for glslang/SPIRV-Tools
if (FOUND_GOOGLETEST)
  add_test(ParserIR-integration-tests)
endif()
```
That seems totally fine to me as long as it's not something normal developers will need to build, but rather maintainers of this code and CI.

I don't have any preference on test framework, so googletest is probably fine. Of course, that limits integration testing to whatever compilers gtest supports, but I assume you're fine with that.

PR #769, opened 2018-11-22
Handle invariant decoration more robustly.

### LINKED ISSUE #761: invariant qualifier is preserved but not honored
We have a depth-pass SPIRV shader and a color-pass SPIRV shader that compute an invariant `gl_Position` in the same way. After transpiling back to GLSL, the color shader uses an intermediary variable to compute position, but the depth shader does not. This causes variance, which in turn leads to depth artifacts.

I've added our two SPIRV shaders to `shaders/asm/vert` in [this commit](https://github.com/prideout/SPIRV-Cross/commit/352eb27). This issue can likely be provoked with much simpler shaders, but I thought it best to provide the originals.

To repro, merge the above commit (or copy the two files manually), then do:

```
./test_shaders.py shaders/asm/vert
cd shaders/asm/reference/vert
grep gl_Position invariant-depthpass.asm.vert # <== no intermediary variables
grep gl_Position invariant-colorpass.asm.vert # <== uses _82 intermediary
```

HansKristian-ARM:
Avoids certain cases of variance between translation units by forcing
every dependent expression of a store to be temporary.
Should avoid the major failure cases where invariance matters.

Fix #761.

### DIFF
--- a/reference/opt/shaders/vert/invariant.vert
+++ b/reference/opt/shaders/vert/invariant.vert
@@ -1,0 +1,19 @@
+#version 310 es
+
+invariant gl_Position;
+
+layout(location = 0) in vec4 vInput0;
+layout(location = 1) in vec4 vInput1;
+layout(location = 2) in vec4 vInput2;
+layout(location = 0) invariant out vec4 vColor;
+
+void main()
+{
+    vec4 _20 = vInput1 * vInput2;
+    vec4 _21 = vInput0 + _20;
+    gl_Position = _21;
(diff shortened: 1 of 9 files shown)


prideout:
It might be a week or so before I can obtain a Pixelbook for testing, feel free to merge if you have high confidence.  Thanks for fixing this!

HansKristian-ARM:
Ok, I'll just merge then. Open a new issue if you still have issues (but if it still doesn't work, I don't think there is anything I can do, I'll declare it a driver bug)

PR #779, opened 2018-11-28
MSL: Force signedness of shader vertex attributes to match the host.

cdavis5e:
Based on a patch by Stefan Dösinger.

Metal cannot do signedness conversion on vertex attributes, and for good
reason. Putting a `uint4` into an `int4`, or a `char4` into a `uint4`,
would lose those values that are outside the range of the target type.
But putting a `uchar4` into a `short4` or an `int4`, or a `ushort4` into
an `int4`, should work. In that case, force the signedness in the shader
to match the declared type of the host.

Unfortunately, I don't really know how to automatically test this. This
remapping is done based on input parameters normally supplied by
MoltenVK. I'm not sure how we'd set this up for the command-line
`spirv-cross` tool.

### DIFF
--- a/spirv_msl.hpp
+++ b/spirv_msl.hpp
@@ -37,6 +37,8 @@ struct MSLVertexAttr
 	uint32_t msl_offset = 0;
 	uint32_t msl_stride = 0;
 	bool per_instance = false;
+	bool uint8 = false;
+	bool uint16 = false;
 	bool used_by_shader = false;
 };
 
@@ -352,6 +354,7 @@ protected:
 	uint32_t add_interface_block(spv::StorageClass storage);
 	void mark_location_as_used_by_shader(uint32_t location, spv::StorageClass storage);
 	uint32_t ensure_correct_builtin_type(uint32_t type_id, spv::BuiltIn builtin);
+	uint32_t ensure_correct_attribute_type(uint32_t type_id, uint32_t location);
 
 	void emit_custom_functions();
 	void emit_resources();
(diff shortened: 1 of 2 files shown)


HansKristian-ARM:
I think I need a concrete example to understand the problem. What API side is doing, what the shader is doing, and why the problem arises in practice.

The shader should already declare the type it wants to consume, so I'm a little confused.

cdavis5e:
We're working on porting a game to macOS. During some of its draw calls, one of the vertex attributes it declares is of format `VK_FORMAT_R8G8B8A8_UINT`. The shader, though, uses type `ivec4` for this. The corresponding Metal types would be `MTLVertexFormatUChar4` and `int4`. Because the API declared vertex attribute is unsigned, but the shader declared type is signed, Metal refuses to compile the pipeline, producing this error message:

```
Vertex attribute foo(3) of type int4 cannot be read using MTLAttributeFormatUChar4.
```

The end result of this is that large pieces of the game's UI fail to render correctly.

cdavis5e:
@billhollings 
This change is important and needs to go in soon. We're working on a tight schedule, and need to have something by January.

spirv_msl.hpp
```diff
@@ -37,6 +37,8 @@ struct MSLVertexAttr
 	uint32_t msl_offset = 0;
 	uint32_t msl_stride = 0;
 	bool per_instance = false;
+	bool uint8 = false;
```
grovesNL: Why two fields instead of having an enum?

cdavis5e: Because it was originally one boolean indicating an unsigned format. But yeah, these two are mutually exclusive, and should probably use an enum. I'll send a fix.

PR #785, opened 2018-12-11
MSL don't emit memory_scope after MSL 2.0.

billhollings:
Apple unexpectedly removed `memory_scope` in MSL 2.1.

### DIFF
--- a/spirv_msl.cpp
+++ b/spirv_msl.cpp
@@ -2671,7 +2671,7 @@ void CompilerMSL::emit_barrier(uint32_t id_exe_scope, uint32_t id_mem_scope, uin
 	else
 		bar_stmt += "mem_none";
 
-	if (msl_options.is_ios() && msl_options.supports_msl_version(2))
+	if (msl_options.is_ios() && (msl_options.supports_msl_version(2) && !msl_options.supports_msl_version(2, 1)))
 	{
 		bar_stmt += ", ";
 


LukasBanana:
It seems that the metal compiler doesn't recognize the `memory_scope` identifiers anymore even with the target API level set to `-std=ios-metal2.0`. @HansKristian-Work Is it possible to remove the output of these identifiers completely from SPIRV-Cross?

Consider the following Metal shader:
```metal
#include <metal_stdlib>
using namespace metal;

void CS(
    constant uint*  inputBuffer  [[buffer(0)]],
    constant uint&  threadCount  [[buffer(1)]],
    device uint*    outputBuffer [[buffer(2)]],
    uint            tid          [[threadgroup_position_in_grid]])
{
    outputBuffer[tid] = inputBuffer[tid];
    threadgroup_barrier(mem_flags::mem_threadgroup, memory_scope_device);
    outputBuffer[(tid + 1) % threadCount] += inputBuffer[threadCount + tid];
}
```

It does not even compile with this command line on Mac:
`xcrun -sdk macosx metal -std=ios-metal2.0 -c Example.metal -o Example.air`

I presume the `-std=ios-metal2.0` only changes compiler behavior but does not include different search paths for the `<metal_stdlib>` header file. Since Apple removed it, was it really required in the first place?

HansKristian-Work:
I made an issue to deal with this: https://github.com/KhronosGroup/SPIRV-Cross/issues/1274

PR #790, opened 2018-12-19
fix glsl tessellation evaluation shader input array declaration error

sulfurmod:


msvc/SPIRV-Cross.sln
```diff
@@ -1,26 +1,26 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Express 2013 for Windows Desktop
```
HansKristian-Work: Why all the changes to MSVC build? Was it accidentally checked in?

sulfurmod: the commit 'modify visual studio solution to vs2015' is a accidentally checked in, sorry

HansKristian-Work: Ok, please rebase and remove the commit in that case rather than revert for commit history cleanliness.

PR #791, opened 2018-12-22
Add install PHONY target in Makefile

apnavik:
Add install as a PHONY target to make it
compatible for other Linux distro such as Yocto.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,7 +35,13 @@ $(STATIC_LIB): $(OBJECTS)
 %.o: %.cpp
 	$(CXX) -c -o $@ $< $(CXXFLAGS) -MMD
 
+install:
+	mkdir -p $(TARGET_DIR_BIN)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -35,7 +35,13 @@ $(STATIC_LIB): $(OBJECTS)
 %.o: %.cpp
 	$(CXX) -c -o $@ $< $(CXXFLAGS) -MMD
 
+install:
+	mkdir -p $(TARGET_DIR_BIN)
```
HansKristian-Work: Where is TARGET_DIR_BIN defined? This does not use Unix-like conventions. The Makefile isn't really supposed to be a robust build system for SPIRV-Cross. Is it possible to use CMake instead which should already install stuff?

Also, headers and stuff are not installed either, so the static lib is not particularly useful here.

PR #802, opened 2019-01-06
Support for sampler2DRect and legacy texture2DRect() sampling function

JustSid:
More (legacy) texture stuff from me. We use this to share textures between Metal and OpenGL via an `IOSurfaceRef`, unfortunately that has a hard requirement for rectangle textures.

I wasn't exactly sure what extension requirement to pick, texture rectangle support is available via `GL_ARB_texture_rectangle` and `GL_EXT_gpu_shader4`. I've chosen `GL_ARB_texture_rectangle` for now, but if you prefer `GL_EXT_gpu_shader4` (since that's already used elsewhere for `textureLodOffset`), I'm happy to change it.


HansKristian-Work:
Looks good, thanks!

PR #807, opened 2019-01-07
MSL: Support SPV_KHR_variable_pointers.

cdavis5e:
This allows shaders to declare and use pointer-type variables. Pointers
may be loaded and stored, be the result of an `OpSelect`, be passed to
and returned from functions, and even be passed as inputs to the `OpPhi`
instruction. All types of pointers may be used as variable pointers.
Variable pointers to storage buffers and workgroup memory may even be
loaded from and stored to, as though they were ordinary variables. In
addition, this enables using an interior pointer to an array as though
it were an array pointer itself using the `OpPtrAccessChain`
instruction.

This is a rather large and involved change, mostly because this is
somewhat complicated with a lot of moving parts. It's a wonder
SPIRV-Cross's output is largely unchanged. Indeed, many of these changes
are to accomplish exactly that! Perhaps the largest source of changes
was the violation of the assumption that, when emitting types, the
pointer type didn't matter.

One of the test cases added by the change doesn't optimize very well;
the output of `spirv-opt` here is invalid SPIR-V. I need to file a bug
with SPIRV-Tools about this.

I wanted to test that variable pointers to images worked too, but I
couldn't figure out how to propagate the access qualifier properly--in
MSL, it's part of the type, so getting this right is important. I've
punted on that for now.

### DIFF
--- a/spirv_cross.cpp
+++ b/spirv_cross.cpp
@@ -1010,7 +1010,7 @@ const SPIRType &Compiler::get_type_from_variable(uint32_t id) const
 	return get<SPIRType>(get<SPIRVariable>(id).basetype);
 }
 
-uint32_t Compiler::get_non_pointer_type_id(uint32_t type_id) const
+uint32_t Compiler::get_pointee_type_id(uint32_t type_id) const
 {
 	auto *p_type = &get<SPIRType>(type_id);
 	if (p_type->pointer)
@@ -1021,7 +1021,7 @@ uint32_t Compiler::get_non_pointer_type_id(uint32_t type_id) const
 	return type_id;
 }
 
-const SPIRType &Compiler::get_non_pointer_type(const SPIRType &type) const
+const SPIRType &Compiler::get_pointee_type(const SPIRType &type) const
 {
 	auto *p_type = &type;
 	if (p_type->pointer)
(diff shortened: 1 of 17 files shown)


HansKristian-Work:
Oh boy :D

spirv_cross.cpp
```diff
@@ -977,19 +1013,18 @@ const SPIRType &Compiler::get_type_from_variable(uint32_t id) const
 uint32_t Compiler::get_non_pointer_type_id(uint32_t type_id) const
 {
 	auto *p_type = &get<SPIRType>(type_id);
-	while (p_type->pointer)
+	if (p_type->pointer)
```
HansKristian-Work: I assume this is because you want double-pointers (OpVariable of pointer) to become just a plain pointer?

spirv_msl.cpp
```diff
@@ -654,6 +656,16 @@ void CompilerMSL::extract_global_variables_from_function(uint32_t func_id, std::
...
 			}
 
+			case OpSelect:
+			{
+				uint32_t base_id = ops[3];
+				if (global_var_ids.find(base_id) != global_var_ids.end())
+					added_arg_ids.insert(base_id);
+				base_id = ops[4];
+				if (global_var_ids.find(base_id) != global_var_ids.end())
+					added_arg_ids.insert(base_id);
```
HansKristian-Work: Missing break;

spirv_cross.cpp
HansKristian-Work: Would it make sense to rename the method to be clearer about its intent?

spirv_glsl.cpp
```diff
@@ -6961,14 +7125,14 @@ void CompilerGLSL::emit_instruction(const Instruction &instruction)
 			//
 			// Including the base will prevent this and would trigger multiple reads
 			// from expression causing it to be forced to an actual temporary in GLSL.
-			auto expr = access_chain_internal(ops[2], &ops[3], length, true, true, &meta);
+			auto expr = access_chain_internal(ops[2], &ops[3], length, true, true, false, &meta);
```
HansKristian-Work: I think I'll need to refactor all these bools into a bitmask or something. Getting a bit out of hand :p

spirv_cross.cpp
cdavis5e: That is correct, and it probably would be. I just wanted to reduce the size of this change; it's already pretty big.

cdavis5e:
Yikes, that's something I did *not* consider! Good catch.

PR #810, opened 2019-01-09
Fix input array size in tessellation evaluation shaders.

HansKristian-Work:


HansKristian-Work:
Trivial change, tested locally, just merging.

PR #821, opened 2019-01-14
MSL: Fix passing a sampled image to a function.

cdavis5e:
In the past, SPIRV-Cross threw an error in this case because it couldn't
work out which swizzle from the auxiliary buffer needs to be passed.
Now, we pass the swizzle around with the texture object, like a combined
image-sampler and its associated sampler.


HansKristian-Work:
Nice!

PR #831, opened 2019-01-17
MSL: Hoist fixup hooks in entry_point_args() out of the compile loop.

cdavis5e:
Otherwise, in the event of a forced recompile, we could end up adding
them twice.


HansKristian-Work:
Seem to be missing updated reference files.

PR #845, opened 2019-01-31
Move assertions after the check for equal types.

cdavis5e:
`bitcast_glsl_op()` is sometimes called for `Boolean` types, e.g. for
specialization constants. We don't want the assert to trip if this is
going to be a no-op anyway.


HansKristian-Work:
Good point.

PR #848, opened 2019-02-06
MSL: Add a setting to capture vertex shader output to a buffer.

cdavis5e:
This will be necessary to support transform feedback, as well as
tessellation shaders.

### DIFF
--- a/reference/opt/shaders-msl/comp/force-recompile-hooks.swizzle.comp
+++ b/reference/opt/shaders-msl/comp/force-recompile-hooks.swizzle.comp
@@ -7,6 +7,7 @@ using namespace metal;
 
 struct spvAux
 {
+    uint vertexCount;
     uint swizzleConst[1];
 };
 
@@ -130,7 +131,7 @@ inline vec<T, 4> spvGatherCompareSwizzle(sampler s, const thread Tex& t, Ts... p
     return t.gather_compare(s, spvForward<Ts>(params)...);
 }
 
-kernel void main0(constant spvAux& spvAuxBuffer [[buffer(0)]], texture2d<float> foo [[texture(0)]], texture2d<float, access::write> bar [[texture(1)]], sampler fooSmplr [[sampler(0)]])
+kernel void main0(constant spvAux& spvAuxBuffer [[buffer(30)]], texture2d<float> foo [[texture(0)]], texture2d<float, access::write> bar [[texture(1)]], sampler fooSmplr [[sampler(0)]])
 {
     constant uint32_t& fooSwzl = spvAuxBuffer.swizzleConst[0];
     bar.write(spvTextureSwizzle(foo.sample(fooSmplr, float2(1.0), level(0.0)), fooSwzl), uint2(int2(0)));
--- a/reference/opt/shaders-msl/vert/basic.capture.vert
+++ b/reference/opt/shaders-msl/vert/basic.capture.vert
@@ -1,0 +1,35 @@
+#include <metal_stdlib>
+#include <simd/simd.h>
+
+using namespace metal;
+
+struct UBO
+{
+    float4x4 uMVP;
+};
+
+struct spvAux
+{
+    uint vertexCount;
+    uint swizzleConst[1];
+};
+
+struct main0_out
+{
+    float3 vNormal [[user(locn0)]];
+    float4 gl_Position [[position]];
+};
+
+struct main0_in
+{
+    float4 aVertex [[attribute(0)]];
+    float3 aNormal [[attribute(1)]];
+};
+
+vertex void main0(main0_in in [[stage_in]], constant UBO& _16 [[buffer(0)]], constant spvAux& spvAuxBuffer [[buffer(30)]], uint gl_VertexIndex [[vertex_id]], uint gl_BaseVertex [[base_vertex]], uint gl_InstanceIndex [[instance_id]], uint gl_BaseInstance [[base_instance]], device main0_out* spvOut [[buffer(29)]])
+{
+    device main0_out& out = spvOut[(gl_InstanceIndex - gl_BaseInstance) * spvAuxBuffer.vertexCount + gl_VertexIndex - gl_BaseVertex];
+    out.gl_Position = _16.uMVP * in.aVertex;
+    out.vNormal = in.aNormal;
+}
+
--- a/reference/shaders-msl-no-opt/asm/frag/texture-access.swizzle.asm.frag
+++ b/reference/shaders-msl-no-opt/asm/frag/texture-access.swizzle.asm.frag
@@ -7,6 +7,7 @@ using namespace metal;
 
 struct spvAux
 {
+    uint vertexCount;
     uint swizzleConst[1];
 };
 
@@ -136,7 +137,7 @@ inline vec<T, 4> spvGatherCompareSwizzle(sampler s, const thread Tex& t, Ts... p
     return t.gather_compare(s, spvForward<Ts>(params)...);
 }
 
-fragment void main0(constant spvAux& spvAuxBuffer [[buffer(0)]], texture1d<float> tex1d [[texture(0)]], texture2d<float> tex2d [[texture(1)]], texture3d<float> tex3d [[texture(2)]], texturecube<float> texCube [[texture(3)]], texture2d_array<float> tex2dArray [[texture(4)]], texturecube_array<float> texCubeArray [[texture(5)]], texture2d<float> texBuffer [[texture(6)]], depth2d<float> depth2d [[texture(7)]], depthcube<float> depthCube [[texture(8)]], depth2d_array<float> depth2dArray [[texture(9)]], depthcube_array<float> depthCubeArray [[texture(10)]], sampler tex1dSamp [[sampler(0)]], sampler tex2dSamp [[sampler(1)]], sampler tex3dSamp [[sampler(2)]], sampler texCubeSamp [[sampler(3)]], sampler tex2dArraySamp [[sampler(4)]], sampler texCubeArraySamp [[sampler(5)]], sampler depth2dSamp [[sampler(7)]], sampler depthCubeSamp [[sampler(8)]], sampler depth2dArraySamp [[sampler(9)]], sampler depthCubeArraySamp [[sampler(10)]])
+fragment void main0(constant spvAux& spvAuxBuffer [[buffer(30)]], texture1d<float> tex1d [[texture(0)]], texture2d<float> tex2d [[texture(1)]], texture3d<float> tex3d [[texture(2)]], texturecube<float> texCube [[texture(3)]], texture2d_array<float> tex2dArray [[texture(4)]], texturecube_array<float> texCubeArray [[texture(5)]], texture2d<float> texBuffer [[texture(6)]], depth2d<float> depth2d [[texture(7)]], depthcube<float> depthCube [[texture(8)]], depth2d_array<float> depth2dArray [[texture(9)]], depthcube_array<float> depthCubeArray [[texture(10)]], sampler tex1dSamp [[sampler(0)]], sampler tex2dSamp [[sampler(1)]], sampler tex3dSamp [[sampler(2)]], sampler texCubeSamp [[sampler(3)]], sampler tex2dArraySamp [[sampler(4)]], sampler texCubeArraySamp [[sampler(5)]], sampler depth2dSamp [[sampler(7)]], sampler depthCubeSamp [[sampler(8)]], sampler depth2dArraySamp [[sampler(9)]], sampler depthCubeArraySamp [[sampler(10)]])
 {
     constant uint32_t& tex1dSwzl = spvAuxBuffer.swizzleConst[0];
     constant uint32_t& tex2dSwzl = spvAuxBuffer.swizzleConst[1];
--- a/reference/shaders-msl/vert/basic.capture.vert
+++ b/reference/shaders-msl/vert/basic.capture.vert
@@ -1,0 +1,35 @@
+#include <metal_stdlib>
+#include <simd/simd.h>
+
+using namespace metal;
+
+struct UBO
+{
+    float4x4 uMVP;
+};
+
+struct spvAux
+{
+    uint vertexCount;
+    uint swizzleConst[1];
+};
+
+struct main0_out
+{
+    float3 vNormal [[user(locn0)]];
+    float4 gl_Position [[position]];
+};
+
+struct main0_in
+{
+    float4 aVertex [[attribute(0)]];
+    float3 aNormal [[attribute(1)]];
+};
+
+vertex void main0(main0_in in [[stage_in]], constant UBO& _16 [[buffer(0)]], constant spvAux& spvAuxBuffer [[buffer(30)]], uint gl_VertexIndex [[vertex_id]], uint gl_BaseVertex [[base_vertex]], uint gl_InstanceIndex [[instance_id]], uint gl_BaseInstance [[base_instance]], device main0_out* spvOut [[buffer(29)]])
+{
+    device main0_out& out = spvOut[(gl_InstanceIndex - gl_BaseInstance) * spvAuxBuffer.vertexCount + gl_VertexIndex - gl_BaseVertex];
+    out.gl_Position = _16.uMVP * in.aVertex;
+    out.vNormal = in.aNormal;
+}
+
--- a/main.cpp
+++ b/main.cpp
@@ -490,6 +490,7 @@ struct CLIArguments
 	bool yflip = false;
 	bool sso = false;
 	bool support_nonzero_baseinstance = true;
+	bool msl_capture_output_to_buffer = false;
 	bool msl_swizzle_texture_samples = false;
 	bool msl_ios = false;
 	bool msl_pad_fragment_output = false;
@@ -545,6 +546,7 @@ static void print_help()
 	                "\t[--cpp-interface-name <name>]\n"
 	                "\t[--msl]\n"
 	                "\t[--msl-version <MMmmpp>]\n"
+	                "\t[--msl-capture-output]\n"
 	                "\t[--msl-swizzle-texture-samples]\n"
 	                "\t[--msl-ios]\n"
 	                "\t[--msl-pad-fragment-output]\n"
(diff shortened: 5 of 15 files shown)


reference/shaders-msl-no-opt/asm/frag/texture-access.swizzle.asm.frag
```diff
@@ -7,6 +7,7 @@ using namespace metal;
 
 struct spvAux
 {
+    uint vertexCount;
```
HansKristian-Work: I'm a bit concerned about this magic AuxBuffer. I suppose, we'll need to make sure MoltenVK stays in sync with these changes? @billhollings, is there any integration testing in place to catch changes like these when you pull in latest SPIRV-Cross?

reference/opt/shaders-msl/vert/basic.capture.vert
```diff
@@ -0,0 +1,35 @@
...
+
+struct spvAux
+{
+    uint vertexCount;
+    uint swizzleConst[1];
+};
+
+struct main0_out
+{
+    float3 vNormal [[user(locn0)]];
```
HansKristian-Work: Should outputs have locations/semantics when they are captured?

HansKristian-Work: Do we have any way of dealing with XfbStride/XfbOffset for VK_EXT_transform_feedback?

reference/opt/shaders-msl/comp/force-recompile-hooks.swizzle.comp
```diff
@@ -7,6 +7,7 @@ using namespace metal;
 
 struct spvAux
 {
+    uint vertexCount;
```
HansKristian-Work: Is this enough to deal with indirect draw calls?

reference/shaders-msl/vert/basic.capture.vert
```diff
@@ -0,0 +1,35 @@
...
+
+struct main0_in
+{
+    float4 aVertex [[attribute(0)]];
+    float3 aNormal [[attribute(1)]];
+};
+
+vertex void main0(main0_in in [[stage_in]], constant UBO& _16 [[buffer(0)]], constant spvAux& spvAuxBuffer [[buffer(30)]], uint gl_VertexIndex [[vertex_id]], uint gl_BaseVertex [[base_vertex]], uint gl_InstanceIndex [[instance_id]], uint gl_BaseInstance [[base_instance]], device main0_out* spvOut [[buffer(29)]])
+{
+    device main0_out& out = spvOut[(gl_InstanceIndex - gl_BaseInstance) * spvAuxBuffer.vertexCount + gl_VertexIndex - gl_BaseVertex];
```
HansKristian-Work: Have you tested this with nested functions? If we need to pass down global variables to leaf function, they'll now need to have device storage space and not thread.

reference/shaders-msl-no-opt/asm/frag/texture-access.swizzle.asm.frag
billhollings: @HansKristian-Work You raise a good point...no...there is no explicit API integration testing at the buffer-contents level.

@cdavis5e What do you think? I expect `spvAux` will evolve slowly...and that we would make changes to both SPIRV-Cross and MoltenVK simultaneously...so I am not **_too_** concerned about this right now. But in the larger scope..where SPIRV-Cross MSL is influenced by more than MoltenVK...there is the possibility for silent surprises.

One option I can think of is to enable changes to the `spvAux` struct through API Option switches. Each member of `spvAux` could have a corresponding API option bool (eg. an `CompilerMSL::Options::spvAuxFeatures` which looks the same as `spvAux`but with all bools)...and SPIRV-Cross would only emit each member of `spvAux` if the parallel Options struct member enabled it.

But would that grow to be a huge PITA?

Another option would be to create a really simple test GLSL or SPIR-V shader file within the MoltenVK repo that uses the aux buffer...and run `test-shaders.py` on it from the MoltenVK `fetchDependencies` script. That would identify that the struct has changed when a new version of SPIRV-Cross is brought into MoltenVK.

billhollings: @cdavis5e Regarding my last option...a test shader within MoltenVK...if you can provide me with a simple GLSL shader and resulting MSL reference shader that uses `spvAux`...I can put them into a new `spirv-test` directory within MoltenVK and modify `fetchDependencies` to run `test-shaders.py` on that directory whenever SPIRV-Cross is updated.

We can then add additional shaders as needed when the SPIRV-Cross API changes in an similarly hidden ways.

HansKristian-Work: I think some kind of query would make sense so that the consumer of SPIRV-Cross can know the ABI layout of this structure. Happy to support that, but I'd rather the consumers of this API propose it. From SPIRV-Cross' point of view, this should be as simple as bumping a few constants whenever the AuxBuffer layout changes, it's the downstream which has to deal with it.

reference/opt/shaders-msl/vert/basic.capture.vert
cdavis5e: > Should outputs have locations/semantics when they are captured?

Probably not, since they aren't used.

> Do we have any way of dealing with XfbStride/XfbOffset for VK_EXT_transform_feedback?

Not yet. That's definitely something we'll need to add, though.

reference/shaders-msl-no-opt/asm/frag/texture-access.swizzle.asm.frag
cdavis5e: > What do you think? I expect `spvAux` will evolve slowly...and that we would make changes to both SPIRV-Cross and MoltenVK simultaneously

As do I.

> One option I can think of is to enable changes to the `spvAux` struct through API Option switches.
> But would that grow to be a huge PITA?

I don't expect this to have too many more members. The only one I can think of off the top of my head right now that I haven't added yet is `BuiltInDrawIndex`. In the past, I mentioned `BuiltInViewIndex`, but I'm considering deriving that from the `BuiltInInstanceIndex`; the way `VK_KHR_multiview` is specified implies this is an acceptable way to implement this.

> if you can provide me with a simple GLSL shader and resulting MSL reference shader that uses `spvAux`...I can put them into a new `spirv-test` directory within MoltenVK and modify `fetchDependencies` to run `test_shaders.py` on that directory whenever SPIRV-Cross is updated.

The GLSL shader:
```glsl
#version 450

layout(binding = 0) uniform sampler2D t0;

layout(location = 0) out vec4 FragColor;

void main()
{
    FragColor = texture(t0, gl_FragCoord.xy);
}
```

Compiled with:
```console
$ glslangValidator -V --target-env vulkan1.1 spv-aux-buffer.frag -o spv-aux-buffer.spv
$ spirv-cross --entry main --output spv-aux-buffer.frag.metal spv-aux-buffer.spv --msl --msl-version 10200 --msl-swizzle-texture-samples
```

The Metal shader:
```c++

#include <metal_stdlib>
#include <simd/simd.h>

using namespace metal;

struct spvAux
{
    uint vertexCount;
    uint swizzleConst[1];
};

struct main0_out
{
};

enum class spvSwizzle : uint
{
    none = 0,
    zero,
    one,
    red,
    green,
    blue,
    alpha
};

template<typename T> struct spvRemoveReference { typedef T type; };
template<typename T> struct spvRemoveReference<thread T&> { typedef T type; };
template<typename T> struct spvRemoveReference<thread T&&> { typedef T type; };
template<typename T> inline constexpr thread T&& spvForward(thread typename spvRemoveReference<T>::type& x)
{
    return static_cast<thread T&&>(x);
}
template<typename T> inline constexpr thread T&& spvForward(thread typename spvRemoveReference<T>::type&& x)
{
    return static_cast<thread T&&>(x);
}

template<typename T>
inline T spvGetSwizzle(vec<T, 4> x, T c, spvSwizzle s)
{
    switch (s)
    {
        case spvSwizzle::none:
            return c;
        case spvSwizzle::zero:
            return 0;
        case spvSwizzle::one:
            return 1;
        case spvSwizzle::red:
            return x.r;
        case spvSwizzle::green:
            return x.g;
        case spvSwizzle::blue:
            return x.b;
        case spvSwizzle::alpha:
            return x.a;
    }
}

// Wrapper function that swizzles texture samples and fetches.
template<typename T>
inline vec<T, 4> spvTextureSwizzle(vec<T, 4> x, uint s)
{
    if (!s)
        return x;
    return vec<T, 4>(spvGetSwizzle(x, x.r, spvSwizzle((s >> 0) & 0xFF)), spvGetSwizzle(x, x.g, spvSwizzle((s >> 8) & 0xFF)), spvGetSwizzle(x, x.b, spvSwizzle((s >> 16) & 0xFF)), spvGetSwizzle(x, x.a, spvSwizzle((s >> 24) & 0xFF)));
}

template<typename T>
inline T spvTextureSwizzle(T x, uint s)
{
    return spvTextureSwizzle(vec<T, 4>(x, 0, 0, 1), s).x;
}

// Wrapper function that swizzles texture gathers.
template<typename T, typename Tex, typename... Ts>
inline vec<T, 4> spvGatherSwizzle(sampler s, const thread Tex& t, Ts... params, component c, uint sw) METAL_CONST_ARG(c)
{
    if (sw)
    {
        switch (spvSwizzle((sw >> (uint(c) * 8)) & 0xFF))
        {
            case spvSwizzle::none:
                break;
            case spvSwizzle::zero:
                return vec<T, 4>(0, 0, 0, 0);
            case spvSwizzle::one:
                return vec<T, 4>(1, 1, 1, 1);
            case spvSwizzle::red:
                return t.gather(s, spvForward<Ts>(params)..., component::x);
            case spvSwizzle::green:
                return t.gather(s, spvForward<Ts>(params)..., component::y);
            case spvSwizzle::blue:
                return t.gather(s, spvForward<Ts>(params)..., component::z);
            case spvSwizzle::alpha:
                return t.gather(s, spvForward<Ts>(params)..., component::w);
        }
    }
    switch (c)
    {
        case component::x:
            return t.gather(s, spvForward<Ts>(params)..., component::x);
        case component::y:
            return t.gather(s, spvForward<Ts>(params)..., component::y);
        case component::z:
            return t.gather(s, spvForward<Ts>(params)..., component::z);
        case component::w:
            return t.gather(s, spvForward<Ts>(params)..., component::w);
    }
}

// Wrapper function that swizzles depth texture gathers.
template<typename T, typename Tex, typename... Ts>
inline vec<T, 4> spvGatherCompareSwizzle(sampler s, const thread Tex& t, Ts... params, uint sw) 
{
    if (sw)
    {
        switch (spvSwizzle(sw & 0xFF))
        {
            case spvSwizzle::none:
            case spvSwizzle::red:
                break;
            case spvSwizzle::zero:
            case spvSwizzle::green:
            case spvSwizzle::blue:
            case spvSwizzle::alpha:
                return vec<T, 4>(0, 0, 0, 0);
            case spvSwizzle::one:
                return vec<T, 4>(1, 1, 1, 1);
        }
    }
    return t.gather_compare(s, spvForward<Ts>(params)...);
}

fragment main0_out main0(constant spvAux& spvAuxBuffer [[buffer(30)]], texture2d<float> t0 [[texture(0)]], sampler t0Smplr [[sampler(0)]], float4 gl_FragCoord [[position]])
{
    main0_out out = {};
    constant uint32_t& t0Swzl = spvAuxBuffer.swizzleConst[0];
    out.FragColor = spvTextureSwizzle(t0.sample(t0Smplr, gl_FragCoord.xy), t0Swzl);
    return out;
}
```

reference/opt/shaders-msl/comp/force-recompile-hooks.swizzle.comp
cdavis5e: Now you've got me a bit concerned. In that case, we wouldn't necessarily be able to extract the `vertexCount` from the indirect buffer, especially if it's in device-private memory. We'd have to copy it to the auxiliary buffer, in which case we would no longer be able to use Metal's `setXxxBytes:` method; or, we'd have to pass the indirect buffer itself to the shader.

reference/shaders-msl/vert/basic.capture.vert
cdavis5e: This actually occurred to me, which is why I added code to `get_argument_address_space()` to handle this case. I'll add a test to make sure this is actually working.

reference/opt/shaders-msl/comp/force-recompile-hooks.swizzle.comp
HansKristian-Work: Yes, I think the only real solution for indirect is to bind the indirect buffer as a `uint *`, not the indirect struct (because offset for vkCmdIndirectDraw can align to just 4 bytes). Not sure how to deal with difference between vkCmdDraw and vkCmdDrawIndexed, because they have different structs, but for indirect, it is not mandatory to support indexed draw calls I think ... The AuxBuffer could store the indirect draw offset and stride (for multi-draw-indirect) into the indirect buffer. This only be relevant for vertex output capture, so the extra overhead should be minimal.

I don't think we need any copies, that would be nightmare fuel.

reference/shaders-msl-no-opt/asm/frag/texture-access.swizzle.asm.frag
HansKristian-Work: If we're changing AuxBuffer, could we just add reserved fields for all the relevant driver data which may or may not be needed going forward? Avoiding churn on the ABI versions sounds like a good idea to me.

reference/shaders-msl/vert/basic.capture.vert
HansKristian-Work: Right, I thought I saw that, just wanted to make sure.

reference/opt/shaders-msl/vert/basic.capture.vert
HansKristian-Work: Ok, don't think it's necessary now at any rate.

reference/shaders-msl-no-opt/asm/frag/texture-access.swizzle.asm.frag
HansKristian-Work: AuxBuffer should also only contain data which is known at CPU recording time, FWIW. How would DrawIndex be supported? Unrolled on CPU side?

cdavis5e: > AuxBuffer should also only contain data which is known at CPU recording time, FWIW. How would DrawIndex be supported? Unrolled on CPU side?

That's how multi draws are already implemented in MoltenVK, precisely because Metal doesn't really support them directly.

reference/opt/shaders-msl/comp/force-recompile-hooks.swizzle.comp
cdavis5e: > The AuxBuffer could store the indirect draw offset and stride (for multi-draw-indirect) into the indirect buffer.

I'm not even sure we need that. Metal has methods for updating only the offset of a buffer resource:

```objective-c
[renderCmdEnc setVertexBufferOffset: 32 atIndex: 0];
[renderCmdEnc setFragmentBufferOffset: 0xf00 atIndex: 5];
[computeCmdEnc setBufferOffset: 48 atIndex: 31];
```

Since multi draws are unrolled anyway, we could just update the offset of the indirect buffer passed to the shader.

HansKristian-Work: The concern is alignment of those offsets. In Vulkan on nVidia for example, the alignment for an SSBO offset is 256 bytes or something like that. Does Metal have no such restrictions?

cdavis5e: SSBOs (buffers in the `device` address space) have natural alignment. Uniform buffers (in the `constant` address space) also have natural alignment on iOS. On Mac, they are aligned to 256 bytes.

cdavis5e: The relevant bit from the docs:

> For buffers in the *device* address space, align the offset to the data type consumed by the function (which is always less than or equal to 16 bytes).
>
> For buffers in the *constant* address space, align the offset to 256 bytes in macOS. In iOS, align the offset to the maximum of either the data type consumed by the function, or 4 bytes. A 16-byte alignment is safe in iOS if you don't need to consider the data type.

reference/shaders-msl-no-opt/asm/frag/texture-access.swizzle.asm.frag
cdavis5e: I have added a structure to hold optional members of the aux buffer.

billhollings: On the subject of tracking structural changes to `spvAux`...let's just take a KISS approach and define a macro in `spriv_msl.hpp`:

    #define SPIRV_CROSS_MSL_AUX_BUFFER_STRUCT_VERSION  2

which gets incremented whenever `spvAux` changes.

We can then include similar `XXX_SUPPORTED_MSL_AUX_BUFFER_STRUCT_VERSION` macros in MoltenVK & gfx, and test them during builds.

cdavis5e: Agreed. Done.

billhollings: @cdavis5e Sorry...it looks like our messages crossed in cyber-space. On reflection I think my initial suggestion of a feature struct was unnecessarily complicated.

reference/opt/shaders-msl/comp/force-recompile-hooks.swizzle.comp
HansKristian-Work: I see. Then we shouldn't need any offset/stride for indirect command arguments.

cdavis5e: I've added something to deal with indirect draws. And yeah, this gets rid of the extra field from the aux buffer, rendering that entire discussion moot. :smile That discussion probably needed to happen anyway.

PR #851, opened 2019-02-06
Remove some obsolete build systems.

### LINKED ISSUE #847: MSVC standalone build reported to fail in debug, should just be removed
Apparently, it fails to link in debug mode. The msvc/ build has never been tested in CI, so this is not surprising. The likely solution is to just remove that old build system which does not serve much of a purpose anymore.

HansKristian-Work:
They are unmaintained and untested for years.
Standalone MSVC build was broken, and was hardcoded to settings which refused to build on my system.

Fix #847.


HansKristian-Work:
Non-functional issue, just merging.

PR #852, opened 2019-02-07
MSL: Add support for tessellation control shaders.

cdavis5e:
These are transpiled to kernel functions that write the output of the
shader to three buffers: one for per-vertex varyings, one for per-patch
varyings, and one for the tessellation levels. This structure is
mandated by the way Metal works, where the tessellation factors are
supplied to the draw method in their own buffer, while the per-patch and
per-vertex varyings are supplied as though they were vertex attributes;
since they have different step rates, they must be in separate buffers.

The kernel is expected to be run in a workgroup whose size is the
greater of the number of input or output control points. It uses Metal's
support for vertex-style stage input to a compute shader to get the
input values; therefore, at least one instance must run per input point.
Meanwhile, Vulkan mandates that it run at least once per output point.
Overrunning the output array is a concern, but any values written should
either be discarded or overwritten by subsequent patches. I'm probably
going to put some slop space in the buffer when I integrate this into
MoltenVK to be on the safe side.

### DIFF
--- a/reference/shaders-msl-no-opt/vert/pass-array-by-value.vert
+++ b/reference/shaders-msl-no-opt/vert/pass-array-by-value.vert
@@ -31,7 +31,7 @@ void spvArrayCopyFromConstant1(thread T (&dst)[N], constant T (&src)[N])
     for (uint i = 0; i < N; dst[i] = src[i], i++);
 }
 
-float4 consume_constant_arrays2(thread const float4 (&positions)[4], thread const float4 (&positions2)[4], thread int& Index1, thread int& Index2)
+float4 consume_constant_arrays2(thread const float4 (&positions)[4], thread const float4 (&positions2)[4], int Index1, int Index2)
 {
     float4 indexable[4];
     spvArrayCopyFromStack1(indexable, positions);
@@ -40,7 +40,7 @@ float4 consume_constant_arrays2(thread const float4 (&positions)[4], thread cons
     return indexable[Index1] + indexable_1[Index2];
 }
 
-float4 consume_constant_arrays(thread const float4 (&positions)[4], thread const float4 (&positions2)[4], thread int& Index1, thread int& Index2)
+float4 consume_constant_arrays(thread const float4 (&positions)[4], thread const float4 (&positions2)[4], int Index1, int Index2)
 {
     return consume_constant_arrays2(positions, positions2, Index1, Index2);
 }
--- a/reference/shaders-msl-no-opt/vert/pass-array-by-value.vert
+++ b/reference/shaders-msl-no-opt/vert/pass-array-by-value.vert
@@ -31,7 +31,7 @@ void spvArrayCopyFromConstant1(thread T (&dst)[N], constant T (&src)[N])
     for (uint i = 0; i < N; dst[i] = src[i], i++);
 }
 
-float4 consume_constant_arrays2(thread const float4 (&positions)[4], thread const float4 (&positions2)[4], int Index1, int Index2)
+float4 consume_constant_arrays2(thread const float4 (&positions)[4], thread const float4 (&positions2)[4], thread int& Index1, thread int& Index2)
 {
     float4 indexable[4];
     spvArrayCopyFromStack1(indexable, positions);
@@ -40,7 +40,7 @@ float4 consume_constant_arrays2(thread const float4 (&positions)[4], thread cons
     return indexable[Index1] + indexable_1[Index2];
 }
 
-float4 consume_constant_arrays(thread const float4 (&positions)[4], thread const float4 (&positions2)[4], int Index1, int Index2)
+float4 consume_constant_arrays(thread const float4 (&positions)[4], thread const float4 (&positions2)[4], thread int& Index1, thread int& Index2)
 {
     return consume_constant_arrays2(positions, positions2, Index1, Index2);
 }
--- a/spirv_cross.cpp
+++ b/spirv_cross.cpp
@@ -1094,6 +1094,22 @@ const SPIRType &Compiler::get_variable_data_type(const SPIRVariable &var) const
 	return get<SPIRType>(get_variable_data_type_id(var));
 }
 
+SPIRType &Compiler::get_variable_element_type(const SPIRVariable &var)
+{
+	SPIRType *type = &get_variable_data_type(var);
+	if (is_array(*type))
+		type = &get<SPIRType>(type->parent_type);
+	return *type;
+}
+
+const SPIRType &Compiler::get_variable_element_type(const SPIRVariable &var) const
+{
+	const SPIRType *type = &get_variable_data_type(var);
+	if (is_array(*type))
+		type = &get<SPIRType>(type->parent_type);
+	return *type;
+}
+
 bool Compiler::is_sampled_image_type(const SPIRType &type)
 {
 	return (type.basetype == SPIRType::Image || type.basetype == SPIRType::SampledImage) && type.image.sampled == 1 &&
@@ -1183,6 +1199,14 @@ void Compiler::set_extended_decoration(uint32_t id, ExtendedDecorations decorati
 	case SPIRVCrossDecorationPackedType:
 		dec.extended.packed_type = value;
 		break;
+
+	case SPIRVCrossDecorationInterfaceMemberIndex:
+		dec.extended.ib_member_index = value;
+		break;
+
+	case SPIRVCrossDecorationInterfaceOrigID:
+		dec.extended.ib_orig_id = value;
+		break;
 	}
 }
 
@@ -1201,6 +1225,14 @@ void Compiler::set_extended_member_decoration(uint32_t type, uint32_t index, Ext
 	case SPIRVCrossDecorationPackedType:
 		dec.extended.packed_type = value;
 		break;
+
+	case SPIRVCrossDecorationInterfaceMemberIndex:
+		dec.extended.ib_member_index = value;
+		break;
+
+	case SPIRVCrossDecorationInterfaceOrigID:
+		dec.extended.ib_orig_id = value;
+		break;
 	}
 }
 
@@ -1218,6 +1250,12 @@ uint32_t Compiler::get_extended_decoration(uint32_t id, ExtendedDecorations deco
 
 	case SPIRVCrossDecorationPackedType:
 		return dec.extended.packed_type;
+
+	case SPIRVCrossDecorationInterfaceMemberIndex:
+		return dec.extended.ib_member_index;
+
+	case SPIRVCrossDecorationInterfaceOrigID:
+		return dec.extended.ib_orig_id;
 	}
 
 	return 0;
@@ -1240,6 +1278,12 @@ uint32_t Compiler::get_extended_member_decoration(uint32_t type, uint32_t index,
 
 	case SPIRVCrossDecorationPackedType:
 		return dec.extended.packed_type;
+
+	case SPIRVCrossDecorationInterfaceMemberIndex:
+		return dec.extended.ib_member_index;
+
+	case SPIRVCrossDecorationInterfaceOrigID:
+		return dec.extended.ib_orig_id;
 	}
 
 	return 0;
@@ -1259,6 +1303,12 @@ bool Compiler::has_extended_decoration(uint32_t id, ExtendedDecorations decorati
 
 	case SPIRVCrossDecorationPackedType:
 		return dec.extended.packed_type != 0;
+
+	case SPIRVCrossDecorationInterfaceMemberIndex:
+		return dec.extended.ib_member_index != (uint32_t)-1;
+
+	case SPIRVCrossDecorationInterfaceOrigID:
+		return dec.extended.ib_orig_id != 0;
 	}
 
 	return false;
@@ -1281,6 +1331,12 @@ bool Compiler::has_extended_member_decoration(uint32_t type, uint32_t index, Ext
 
 	case SPIRVCrossDecorationPackedType:
 		return dec.extended.packed_type != 0;
+
+	case SPIRVCrossDecorationInterfaceMemberIndex:
+		return dec.extended.ib_member_index != (uint32_t)-1;
+
+	case SPIRVCrossDecorationInterfaceOrigID:
+		return dec.extended.ib_orig_id != 0;
 	}
 
 	return false;
@@ -1298,6 +1354,14 @@ void Compiler::unset_extended_decoration(uint32_t id, ExtendedDecorations decora
 	case SPIRVCrossDecorationPackedType:
 		dec.extended.packed_type = 0;
 		break;
+
+	case SPIRVCrossDecorationInterfaceMemberIndex:
+		dec.extended.ib_member_index = -1;
+		break;
+
+	case SPIRVCrossDecorationInterfaceOrigID:
+		dec.extended.ib_orig_id = 0;
+		break;
 	}
 }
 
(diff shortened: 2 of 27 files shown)


HansKristian-Work:
... It begins.

cdavis5e:
Yeah, this patch is a lot bigger than I'd like. I promise the change to implement tessellation evaluation shaders will be simpler.

reference/shaders-msl-no-opt/vert/pass-array-by-value.vert
```diff
@@ -31,7 +31,7 @@ void spvArrayCopyFromConstant1(thread T (&dst)[N], constant T (&src)[N])
     for (uint i = 0; i < N; dst[i] = src[i], i++);
 }
 
-float4 consume_constant_arrays2(thread const float4 (&positions)[4], thread const float4 (&positions2)[4], thread int& Index1, thread int& Index2)
```
HansKristian-Work: The address space qualifiers are gone here. Is this intentional?

reference/shaders-msl-no-opt/vert/pass-array-by-value.vert
```diff
@@ -40,7 +40,7 @@ float4 consume_constant_arrays2(thread const float4 (&positions)[4], thread cons
     return indexable[Index1] + indexable_1[Index2];
 }
 
-float4 consume_constant_arrays(thread const float4 (&positions)[4], thread const float4 (&positions2)[4], thread int& Index1, thread int& Index2)
```
HansKristian-Work: Same here.

spirv_cross.cpp
```diff
@@ -1094,6 +1094,22 @@ const SPIRType &Compiler::get_variable_data_type(const SPIRVariable &var) const
 	return get<SPIRType>(get_variable_data_type_id(var));
 }
 
+SPIRType &Compiler::get_variable_element_type(const SPIRVariable &var)
+{
+	SPIRType *type = &get_variable_data_type(var);
+	if (is_array(*type))
```
HansKristian-Work: The result will still be an array if the type is array-of-arrays. Is this intentional?

spirv_msl.cpp
```diff
@@ -320,12 +379,31 @@ static string create_sampler_address(const char *prefix, MSLSamplerAddress addr)
...
 
+SPIRType &CompilerMSL::get_patch_stage_out_struct_type()
+{
+	auto &so_var = get<SPIRVariable>(patch_stage_out_var_id);
+	return get_variable_data_type(so_var);
+}
+
+std::string CompilerMSL::get_tess_factor_struct_name()
+{
+	if (get_entry_point().flags.get(ExecutionModeTriangles))
```
HansKristian-Work: Is this known in the control shader? IIRC, this is only known in the evaluation shader.

HansKristian-Work: What about line tessellation?

HansKristian-Work:
Hehe, you were not kidding, this is a massive patch. Must have taken forever.

I found one bug while testing a bit myself:

```
#version 450

layout(vertices = 4) out;
layout(location = 0) patch out vec3 vPatch[2];
layout(location = 2) out vec3 vVertex[];
layout(location = 0) in vec3 vInput[];

void main()
{
        vVertex[gl_InvocationID] =
                vInput[gl_InvocationID] +
                vInput[gl_InvocationID ^ 1];

        barrier();

        if (gl_InvocationID == 0)
        {
                vPatch[0] = vec3(10.0);
                vPatch[1] = vec3(20.0);
                gl_TessLevelOuter[0] = 1.0;
                gl_TessLevelOuter[1] = 2.0;
                gl_TessLevelOuter[2] = 3.0;
                gl_TessLevelOuter[3] = 4.0;
                gl_TessLevelInner[0] = 1.0;
                gl_TessLevelInner[1] = 2.0;
        }
}
```
compiles to:
```
kernel void main0(main0_in in [[stage_in]], uint gl_InvocationID [[thread_index_in_threadgroup]], uint gl_PrimitiveID [[threadgroup_position_in_grid]], device main0_out* spvOut [[buffer(28)]], device uint* spvIndirectParams [[buffer(29)]], device main0_patchOut* spvPatchOut [[buffer(27)]], device MTLQuadTessellationFactorsHalf* spvTessLevel [[buffer(26)]], threadgroup main0_in* gl_in [[threadgroup(0)]])
{
    float3 vPatch[2] = {};
    device main0_out* gl_out = &spvOut[gl_PrimitiveID * 4];
    device main0_patchOut& patchOut = spvPatchOut[gl_PrimitiveID];
    if (gl_InvocationID < spvIndirectParams[0])
        gl_in[gl_InvocationID] = in;
    threadgroup_barrier(mem_flags::mem_threadgroup);
    gl_out[gl_InvocationID].vVertex = gl_in[gl_InvocationID].vInput + gl_in[gl_InvocationID ^ 1].vInput;
    threadgroup_barrier(mem_flags::mem_device);
    if (gl_InvocationID == 0)
    {
        vPatch[0] = float3(10.0);
        vPatch[1] = float3(20.0);
        spvTessLevel[gl_PrimitiveID].edgeTessellationFactor[0] = half(1.0);
        spvTessLevel[gl_PrimitiveID].edgeTessellationFactor[1] = half(2.0);
        spvTessLevel[gl_PrimitiveID].edgeTessellationFactor[2] = half(3.0);
        spvTessLevel[gl_PrimitiveID].edgeTessellationFactor[3] = half(4.0);
        spvTessLevel[gl_PrimitiveID].insideTessellationFactor[0] = half(1.0);
        spvTessLevel[gl_PrimitiveID].insideTessellationFactor[1] = half(2.0);
    }
    patchOut.vPatch_0 = vPatch[0];
    patchOut.vPatch_1 = vPatch[1];
}
```
The problem is that the flattening is not isolated to whatever thread wrote to it. Slapping a thread_id == 0 isn't going to cut it either since it might not be alive, and it might not have been the thread which wrote to the patch output ... Not sure how we solve this cleanly.
Isn't this a pure buffer though? Do we actually need to flatten the array here?

HansKristian-Work:
I'm seeing:
```
    if (gl_InvocationID < spvIndirectParams[0])
        gl_in[gl_InvocationID] = in;
```
The input struct appears to use attributes even in compute. Does this work similarly to vertex shading, where the InvocationID applies a stride or something?

spirv_msl.cpp
```diff
@@ -1004,13 +1137,15 @@ void CompilerMSL::add_plain_variable_to_interface_block(StorageClass storage, co
 		set_member_decoration(ib_type.self, ib_mbr_idx, DecorationCentroid);
 	if (is_sample)
 		set_member_decoration(ib_type.self, ib_mbr_idx, DecorationSample);
+
+	set_extended_member_decoration(ib_type.self, ib_mbr_idx, SPIRVCrossDecorationInterfaceOrigID, var.self);
```
HansKristian-Work: What is this OrigID used for? I cannot find any "get" of it.

spirv_msl.cpp
```diff
@@ -3016,6 +3322,121 @@ void CompilerMSL::emit_instruction(const Instruction &instruction)
...
+
+				indices.push_back(ops[3]);
+
+				uint32_t const_mbr_id = next_id++;
+				uint32_t index = get_extended_decoration(ops[2], SPIRVCrossDecorationInterfaceMemberIndex);
+				if (var->storage == StorageClassInput ||
+				    has_decoration(get_variable_element_type(*var).self, DecorationBlock))
+				{
+					uint32_t i = 4;
+					if (index == (uint32_t)-1)
```
HansKristian-Work: Prefer C++-style arithmetic cast `uint32_t(-1)` instead.

reference/shaders-msl-no-opt/vert/pass-array-by-value.vert
HansKristian-Work: Dug around to see why this is happening. This seems to be related to the `ensure_correct_attribute_type` changes made a while back.

https://github.com/KhronosGroup/SPIRV-Cross/blob/master/spirv_msl.cpp#L970

This dropping of references only happens on vertex shaders not fragment. I guess it's not a real problem since Input storage class is constant after all, but the pointer type is dropped from var.basetype, there might be some other problems related to that. I think we can fix that later.

spirv_cross.cpp
HansKristian-Work: From reading the code which uses it, it seems intentional, just want to confirm this in case there are some bizzarro cases for array-of-arrays ...

spirv_cross.cpp
```diff
@@ -3527,7 +3599,7 @@ bool Compiler::ActiveBuiltinHandler::handle(spv::Op opcode, const uint32_t *args
 		auto *type = &compiler.get_variable_data_type(*var);
 
 		auto &flags =
-		    type->storage == StorageClassInput ? compiler.active_input_builtins : compiler.active_output_builtins;
+		    var->storage == StorageClassInput ? compiler.active_input_builtins : compiler.active_output_builtins;
```
HansKristian-Work: Did this matter at some point?

spirv_cross.cpp
cdavis5e: Yeah, this is intentional. It's to strip off the first layer of array, since per-vertex I/O in a tessellation control shader is done to arrays.

spirv_msl.cpp
cdavis5e: > Is this known in the control shader? IIRC, this is only known in the evaluation shader.

The Vulkan spec says that either one can have it set. The intent is that, if it's set on the eval shader, MoltenVK will copy this to the control shader. I don't know what to do about offline compilation.

> What about line tessellation?

That's actually unsupported by Metal.

cdavis5e:
> Isn't this a pure buffer though? Do we actually need to flatten the array here?

We should not. I already changed it not to flatten outputs in certain cases because of this. I should probably expand that.

cdavis5e:
> I'm seeing:
> 
> ```
>     if (gl_InvocationID < spvIndirectParams[0])
>         gl_in[gl_InvocationID] = in;
> ```
> The input struct appears to use attributes even in compute. Does this work similarly to vertex shading, where the InvocationID applies a stride or something?

In fact, it's exactly like vertex input.

spirv_msl.cpp
cdavis5e: I noticed that, but forgot to go back and fix that...

spirv_cross.cpp
cdavis5e: It did. I think one of the tests I added broke without this change.

spirv_msl.cpp
cdavis5e: It's used after the input struct is sorted, to map the member back to its original variable. We need to do this in order to set the member index, so we can get access chains right.

cdavis5e: [Here](https://github.com/KhronosGroup/SPIRV-Cross/pull/852/files#diff-6a4029d3c45d50818e6f80e915867c64R1614). I used it directly since I have the meta, instead of going through the function.

spirv_msl.cpp
HansKristian-Work: Hm, right. I guess worst case we just need some offline override options, but this is niche enough. Could we have an error if line tessellation is used at least?

spirv_msl.cpp
HansKristian-Work: Oh, right.

spirv_msl.cpp

reference/shaders-msl-no-opt/vert/pass-array-by-value.vert
cdavis5e: I've changed it to pass the pointer type to `ensure_correct_attribute_type()`. That will stop us from dropping that.

spirv_msl.cpp
```diff
@@ -3016,6 +3327,121 @@ void CompilerMSL::emit_instruction(const Instruction &instruction)
...
+					if (index == uint32_t(-1))
+					{
+						// Maybe this is a struct type in the input class, in which case
+						// we put it as a decoration on the corresponding member.
+						index =
+						    get_extended_member_decoration(ops[2], ops[4], SPIRVCrossDecorationInterfaceMemberIndex);
+						assert(index != uint32_t(-1));
+						i++;
+					}
+					// In this case, we flattened structures and arrays, so now we have to
```
HansKristian-Work: Does this assumption still hold?

spirv_msl.cpp
```diff
@@ -3016,6 +3327,121 @@ void CompilerMSL::emit_instruction(const Instruction &instruction)
...
+
+						auto &c = get_constant(ops[i]);
+						index += c.scalar();
+						if (type->parent_type)
+							type = &get<SPIRType>(type->parent_type);
+					}
+					set<SPIRConstant>(const_mbr_id, type_id, index, false);
+					indices.push_back(const_mbr_id);
+
+					indices.insert(indices.end(), &ops[i], &ops[length]);
```
HansKristian-Work: This is broken on a test shader I have.

HansKristian-Work:
```
#version 450

struct Boo
{
        vec3 a;
        vec3 b;
};

layout(vertices = 4) out;
layout(location = 0) out Boo vVertex[];
layout(location = 0) in Boo vInput[];

void main()
{
        vVertex[gl_InvocationID] = vInput[gl_InvocationID];
        gl_TessLevelOuter[0] = 1.0;
        gl_TessLevelOuter[1] = 2.0;
        gl_TessLevelOuter[2] = 3.0;
        gl_TessLevelOuter[3] = 4.0;
        gl_TessLevelInner[0] = 1.0;
        gl_TessLevelInner[1] = 2.0;
}
```
i is 5 here, yet length is 4. Also, I think dereferencing out of bounds is not technically valid, please change it to ops + i and ops + length to avoid asan/valgrind/debug STL complaining.

spirv_msl.cpp
```diff
@@ -3016,6 +3327,121 @@ void CompilerMSL::emit_instruction(const Instruction &instruction)
...
+				}
+				else
+				{
+					set<SPIRConstant>(const_mbr_id, type_id, index, false);
+					indices.push_back(const_mbr_id);
+
+					indices.insert(indices.end(), &ops[4], &ops[length]);
```
HansKristian-Work: Ditto here with range.

spirv_msl.cpp
HansKristian-Work: I tried hacking in this patch:
```
@@ -3368,7 +3368,7 @@ void CompilerMSL::emit_instruction(const Instruction &instruction)
                                                index =
                                                    get_extended_member_decoration(ops[2], ops[4], SPIRVCrossDecorationInterfaceMemberIndex);
                                                assert(index != uint32_t(-1));
-                                               i++;
+                                               //i++;
                                        }
```
That lets me compile, but it does expose another issue I think. The code-gen is:
```
    gl_out[gl_InvocationID].vVertex = gl_in[gl_InvocationID].Boo_a;
```
I think we need to do the unflattening step which we currently do for I/O structs in normal vertex shading to be able to access them as whole structs.

spirv_msl.cpp
cdavis5e: I'm pretty sure it still holds. We should only hit that for per-vertex input.

HansKristian-Work: Ok, there's still the bug I ran into though. Seems tricky.

spirv_msl.cpp
cdavis5e: After this is all done, I've half a mind to file a radar requesting support for composite types other than vectors in stage-in structs. This flattening/unflattening stuff is getting out of hand.

HansKristian-Work: Yup, it's getting nightmarish. It's painful enough without tesc. Feel free to just drop support for composites in tesc if it gets out of hand as long as we can detect it and just fail gracefully.

spirv_msl.cpp
cdavis5e: Nope, you were right and I was wrong. The test case you posted violated this assumption.

spirv_msl.cpp
cdavis5e: OK, I finally got this figured out. (I think.)

HansKristian-Work:
Thanks a lot.

PR #857, opened 2019-02-14
MSL: Add support for tessellation evaluation shaders.

### LINKED ISSUE #120: Support tesselation shaders in Metal backend..
Hi,
just pointing it out.. Not much usage right now since MoltenVK still doesn't support tesselation stage which is new in OSX 10.12 and IOS 10..
Can't ask about geometry shaders as this stage isn't supported on Metal..
Would be nice if a detailed issue is created detailing all Metal shading features and if currently supported by SPIR-V cross similar to this (HLSL support in glslang):
https://github.com/KhronosGroup/glslang/issues/362

cdavis5e:
These are mapped to Metal's post-tessellation vertex functions. The
semantic difference is much less here, so this change should be simpler
than the previous one. There are still some hairy parts, though.

In MSL, the array of control point data is represented by a special
type, `patch_control_point<T>`, where `T` is a valid stage-input type.
This object must be embedded inside the patch-level stage input. For
this reason, I've added a new type to the type system to represent this.

On Mac, the number of input control points to the function must be
specified in the `patch()` attribute. This is optional on iOS.
SPIRV-Cross takes this from the `OutputVertices` execution mode; the
intent is that if it's not set in the shader itself, MoltenVK will set
it from the tessellation control shader. If you're translating these
offline, you'll have to update the control point count manually, since
this number must match the number that is passed to the
`drawPatches:...` family of methods.

Fixes #120.

### DIFF
--- a/reference/opt/shaders-msl/desktop-only/tese/triangle.desktop.sso.tese
+++ b/reference/opt/shaders-msl/desktop-only/tese/triangle.desktop.sso.tese
@@ -1,0 +1,27 @@
+#include <metal_stdlib>
+#include <simd/simd.h>
+
+using namespace metal;
+
+struct main0_out
+{
+    float4 gl_Position [[position]];
+};
+
+struct main0_in
+{
+    float4 gl_Position [[attribute(0)]];
+};
+
+struct main0_patchIn
+{
+    patch_control_point<main0_in> gl_in;
+};
+
+[[ patch(triangle, 0) ]] vertex main0_out main0(main0_patchIn patchIn [[stage_in]], float3 gl_TessCoord [[position_in_patch]])
+{
+    main0_out out = {};
+    out.gl_Position = ((patchIn.gl_in[0].gl_Position * gl_TessCoord.x) + (patchIn.gl_in[1].gl_Position * gl_TessCoord.y)) + (patchIn.gl_in[2].gl_Position * gl_TessCoord.z);
+    return out;
+}
+
--- a/spirv_cross.hpp
+++ b/spirv_cross.hpp
@@ -360,6 +360,9 @@ public:
 	uint32_t get_execution_mode_argument(spv::ExecutionMode mode, uint32_t index = 0) const;
 	spv::ExecutionModel get_execution_model() const;
 
+	static bool is_tessellation_shader(spv::ExecutionModel model);
+	bool is_tessellation_shader() const;
+
 	// In SPIR-V, the compute work group size can be represented by a constant vector, in which case
 	// the LocalSize execution mode is ignored.
 	//
--- a/spirv_cross.hpp
+++ b/spirv_cross.hpp
@@ -360,7 +360,6 @@ public:
 	uint32_t get_execution_mode_argument(spv::ExecutionMode mode, uint32_t index = 0) const;
 	spv::ExecutionModel get_execution_model() const;
 
-	static bool is_tessellation_shader(spv::ExecutionModel model);
 	bool is_tessellation_shader() const;
 
 	// In SPIR-V, the compute work group size can be represented by a constant vector, in which case
@@ -616,6 +615,7 @@ protected:
 
 	const SPIREntryPoint &get_entry_point() const;
 	SPIREntryPoint &get_entry_point();
+	static bool is_tessellation_shader(spv::ExecutionModel model);
 
 	virtual std::string to_name(uint32_t id, bool allow_alias = true) const;
 	bool is_builtin_variable(const SPIRVariable &var) const;
--- a/reference/opt/shaders-msl/tese/input-array.tese
+++ b/reference/opt/shaders-msl/tese/input-array.tese
@@ -1,0 +1,28 @@
+#include <metal_stdlib>
+#include <simd/simd.h>
+
+using namespace metal;
+
+struct main0_out
+{
+    float4 gl_Position [[position]];
+};
+
+struct main0_in
+{
+    float4 Floats [[attribute(0)]];
+    float4 Floats2 [[attribute(2)]];
+};
+
+struct main0_patchIn
+{
+    patch_control_point<main0_in> gl_in;
+};
+
+[[ patch(quad, 0) ]] vertex main0_out main0(main0_patchIn patchIn [[stage_in]], float2 gl_TessCoord [[position_in_patch]])
+{
+    main0_out out = {};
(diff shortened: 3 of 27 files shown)


spirv_cross.hpp
```diff
@@ -360,6 +360,9 @@ class Compiler
 	uint32_t get_execution_mode_argument(spv::ExecutionMode mode, uint32_t index = 0) const;
 	spv::ExecutionModel get_execution_model() const;
 
+	static bool is_tessellation_shader(spv::ExecutionModel model);
```
HansKristian-Work: Does this need to be public?

reference/opt/shaders-msl/desktop-only/tese/triangle.desktop.sso.tese
```diff
@@ -0,0 +1,27 @@
...
+{
+    float4 gl_Position [[attribute(0)]];
+};
+
+struct main0_patchIn
+{
+    patch_control_point<main0_in> gl_in;
+};
+
+[[ patch(triangle, 0) ]] vertex main0_out main0(main0_patchIn patchIn [[stage_in]], float3 gl_TessCoord [[position_in_patch]])
```
HansKristian-Work: Where is the difference between CW and CCW winding handled?

HansKristian-Work:
I played around with a test shader, and found something sketchy:
```
#version 450

layout(ccw, quads, fractional_even_spacing) in;

layout(location = 0) in vec4 vColor[];
layout(location = 1) patch in vec4 vColors;
layout(location = 2) in Block
{
        vec4 a;
        vec4 b;
} blocks[];

struct Foo
{
        vec4 a;
        vec4 b;
};
layout(location = 4) patch in Foo vFoo;

void set_from_function()
{
        gl_Position = blocks[0].a;
        gl_Position += blocks[0].b;
        gl_Position += blocks[1].a;
        gl_Position += blocks[1].b;
        gl_Position += vColor[0];
        gl_Position += vColor[1];
        gl_Position += vColors;
        gl_Position += vFoo.a;
        gl_Position += vFoo.b;
}

void main()
{
        set_from_function();
}
```
Ends up generating:
```
struct Block
{
    float4 a;
    float4 b;
};

struct Foo
{
    float4 a;
    float4 b;
};

struct main0_out
{
};

struct main0_in
{
    float4 vColor [[attribute(0)]];
    float4 Block_a [[attribute(2)]];
    float4 Block_b [[attribute(3)]];
};

struct main0_patchIn
{
    float4 vColors [[attribute(1)]];
    float4 Foo_a [[attribute(4)]];
    float4 Foo_b [[attribute(5)]];
    patch_control_point<main0_in> gl_in;
};

void set_from_function(thread float4& gl_Position, thread patch_control_point<main0_in>& gl_in, thread float4& vColors, thread Foo& vFoo)
{
    gl_Position = gl_in[0].Block_a;
    gl_Position += gl_in[0].Block_b;
    gl_Position += gl_in[1].Block_a;
    gl_Position += gl_in[1].Block_b;
    gl_Position += gl_in[0].vColor;
    gl_Position += gl_in[1].vColor;
    gl_Position += vColors;
    gl_Position += vFoo.a;
    gl_Position += vFoo.b;
}

[[ patch(quad, 0) ]] vertex main0_out main0(main0_patchIn patchIn [[stage_in]])
{
    main0_out out = {};
    Foo vFoo = {};
    blocks.a = in.Block_a;
    blocks.b = in.Block_b;
    vFoo.a = patchIn.Foo_a;
    vFoo.b = patchIn.Foo_b;
    set_from_function(out.gl_Position, patchIn.gl_in, patchIn.vColors, vFoo);
    return out;
}
```
The problem is 
```
    blocks.a = in.Block_a;
    blocks.b = in.Block_b;
```
This looks like a left-over unflatten operating which is normally done for I/O blocks, but perhaps it should not be done here. It seems like codegen is okay otherwise.

HansKristian-Work:
Here's a more complete test shader which exposes various bugs in the current codegen. Not sure if all cases here are feasible to fix (array of structs, flattening that is not supported at all atm ...)

```
#version 450

layout(ccw, quads, fractional_even_spacing) in;

// Try to use the whole taxonomy of input methods.

// Per-vertex vector.
layout(location = 0) in vec4 vColor[];
// Per-patch vector.
layout(location = 1) patch in vec4 vColors;
// Per-patch vector array.
layout(location = 2) patch in vec4 vColorsArray[2];

// I/O blocks, per patch and per control point.
layout(location = 4) in Block
{
        vec4 a;
        vec4 b;
} blocks[];

layout(location = 6) patch in PatchBlock
{
        vec4 a;
        vec4 b;
} patch_block;

// Composites.
struct Foo
{
        vec4 a;
        vec4 b;
};
layout(location = 8) patch in Foo vFoo;
layout(location = 10) patch in Foo vFooArray[2];

// Per-control point struct.
layout(location = 14) in Foo vFoos[];

void set_from_function()
{
        gl_Position = blocks[0].a;
        gl_Position += blocks[0].b;
        gl_Position += blocks[1].a;
        gl_Position += blocks[1].b;
        gl_Position += patch_block.a;
        gl_Position += patch_block.b;
        gl_Position += vColor[0];
        gl_Position += vColor[1];
        gl_Position += vColors;

        Foo foo = vFoo;
        gl_Position += foo.a;
        gl_Position += foo.b;

        foo = vFooArray[0];
        gl_Position += foo.a;
        gl_Position += foo.b;

        foo = vFooArray[1];
        gl_Position += foo.a;
        gl_Position += foo.b;

        foo = vFoos[0];
        gl_Position += foo.a;
        gl_Position += foo.b;

        foo = vFoos[1];
        gl_Position += foo.a;
        gl_Position += foo.b;
}

void main()
{
        set_from_function();
}
```

reference/opt/shaders-msl/desktop-only/tese/triangle.desktop.sso.tese
cdavis5e: In Metal, winding order and partitioning are properties of the pipeline state.

cdavis5e:
There's definitely something broken with per-patch blocks. Right now, this is putting them in the per-vertex input instead of the per-patch input like it should.

spirv_cross.hpp
cdavis5e: Probably not. I've moved it down.

cdavis5e:
All right, I think I fixed most of the issues you found. I've punted on arrays of structures of per-patch inputs for now. I'm pretty sure array-of-struct input was already broken for vertex shaders anyway.

cdavis5e:
Oh great... the test I added in 6b79880 won't pass without the change from 08863c1. This won't affect CI, but it might pose a problem for bisection.

PR #860, opened 2019-02-18
MSL: Add an option to set the tessellation domain origin.

cdavis5e:
This is intended to be used to support `VK_KHR_maintenance2`'s
tessellation domain origin feature. If `tess_domain_origin_lower_left`
is `true`, the `v` coordinate will be inverted with respect to the
domain. Additionally, in `Triangles` mode, the `v` and `w` coordinates
will be swapped. This is because the winding order is interpreted
differently in lower-left mode.

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -494,6 +494,7 @@ struct CLIArguments
 	bool msl_swizzle_texture_samples = false;
 	bool msl_ios = false;
 	bool msl_pad_fragment_output = false;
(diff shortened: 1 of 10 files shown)


cdavis5e:
I'm actually not sure that swapping `v` and `w` is correct.

cdavis5e:
Yes I am. Just as soon as I get the changes to MoltenVK building...

PR #866, opened 2019-02-20
MSL: Some fixes for tessellation shaders.

cdavis5e:
All of these were from running the CTS.

### DIFF
--- a/spirv_msl.cpp
+++ b/spirv_msl.cpp
@@ -3105,8 +3105,8 @@ bool CompilerMSL::emit_tessellation_access_chain(const uint32_t *ops, uint32_t l
 	if (get_execution_model() == ExecutionModelTessellationControl && var && m &&
 	    m->decoration.builtin_type == BuiltInTessLevelInner && get_entry_point().flags.get(ExecutionModeTriangles))
 	{
-		if (auto *c = maybe_get<SPIRConstant>(ops[3]))
-			if (c->scalar() == 1) return false;
+		auto *c = maybe_get<SPIRConstant>(ops[3]);
+		if (c && c->scalar() == 1) return false;
 		auto &dest_var = set<SPIRVariable>(ops[1], *var);
 		dest_var.basetype = ops[0];
 		ir.meta[ops[1]] = ir.meta[ops[2]];
@@ -3117,7 +3117,8 @@ bool CompilerMSL::emit_tessellation_access_chain(const uint32_t *ops, uint32_t l
 	return false;
 }
 
-bool CompilerMSL::is_out_of_bounds_tessellation_level(uint32_t id_lhs) {
+bool CompilerMSL::is_out_of_bounds_tessellation_level(uint32_t id_lhs)
+{
 	if (!get_entry_point().flags.get(ExecutionModeTriangles))
 		return false;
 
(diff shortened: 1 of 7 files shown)


spirv_msl.cpp
```diff
@@ -3113,6 +3117,28 @@ bool CompilerMSL::emit_tessellation_access_chain(const uint32_t *ops, uint32_t l
 	return false;
 }
 
+bool CompilerMSL::is_out_of_bounds_tessellation_level(uint32_t id_lhs) {
```
HansKristian-Work: Formatting.

spirv_msl.cpp
```diff
@@ -3099,10 +3099,14 @@ bool CompilerMSL::emit_tessellation_access_chain(const uint32_t *ops, uint32_t l
 	// drop the last index. It isn't an array in this case, so we can't have an
 	// array reference here. We need to make this ID a variable instead of an
 	// expression so we don't try to dereference it as a variable pointer.
+	// Don't do this if the index is a constant 1, though. We need to drop stores
+	// to that one.
 	auto *m = ir.find_meta(var ? var->self : 0);
 	if (get_execution_model() == ExecutionModelTessellationControl && var && m &&
 	    m->decoration.builtin_type == BuiltInTessLevelInner && get_entry_point().flags.get(ExecutionModeTriangles))
 	{
+		if (auto *c = maybe_get<SPIRConstant>(ops[3]))
```
HansKristian-Work: Please move c to its own line, this pattern is not found anywhere else I think in the code base.

spirv_msl.cpp
```diff
@@ -7147,6 +7176,14 @@ void CompilerMSL::bitcast_from_builtin_load(uint32_t source_id, std::string &exp
...
 	if (expected_type != expr_type.basetype)
 		expr = bitcast_expression(expr_type, expected_type, expr);
+
+	if (builtin == BuiltInTessCoord && get_entry_point().flags.get(ExecutionModeQuads) && expr_type.vecsize == 3)
+	{
+		// In SPIR-V, this is always a vec3, even for quads. In Metal, though, it's a float2 for quads.
+		// The code is expecting a float3, so we need to widen this.
+		expr = join("float3(", expr, ", 0)");
+	}
+
```
HansKristian-Work: Redundant newline.

spirv_msl.cpp
```diff
@@ -3113,6 +3117,28 @@ bool CompilerMSL::emit_tessellation_access_chain(const uint32_t *ops, uint32_t l
...
+	// levels must be stored to a dedicated buffer in a particular format that depends
+	// on the patch type. Therefore, in Triangles mode, any access to the second
+	// inner level or the fourth outer level must be dropped.
+	const auto *e = maybe_get<SPIRExpression>(id_lhs);
+	if (!e || !e->access_chain) return false;
+	BuiltIn builtin = BuiltIn(get_decoration(e->loaded_from, DecorationBuiltIn));
+	if (builtin != BuiltInTessLevelInner && builtin != BuiltInTessLevelOuter) return false;
+	auto *c = maybe_get<SPIRConstant>(e->implied_read_expressions[1]);
+	if (!c) return false;
+	return (builtin == BuiltInTessLevelInner && c->scalar() == 1) || (builtin == BuiltInTessLevelOuter && c->scalar() == 3);
```
HansKristian-Work: I suppose we cannot deal with dynamically accessed tessellation levels?

cdavis5e: Don't think so. How would we prove that the missing levels are accessed in that case?

HansKristian-Work: Right. I think in the worst case we'll have to dynamically bounds check the store I think, but not necessary unless we run into it.

PR #869, opened 2019-02-21
Add a stable C API for SPIRV-Cross.

### LINKED ISSUE #611: C Interface
Not really an issue, would be nice to have C interface for SPIRV-Cross so we can call from external, for example C#.

Maybe I can send PR tho.

HansKristian-Work:
This adds a new C API for SPIRV-Cross which is intended to be stable,
    both API and ABI wise.
    
    The C++ API has been refactored a bit to make the C wrapper easier and
    cleaner to write. Especially the vertex attribute / resource interfaces
    for MSL has been rewritten to avoid taking mutable pointers into the
    interface. This would be very annoying to wrap and it didn't fit well
    with the rest of the C++ API to begin with. While doing this, I went
    ahead and removed all the old deprecated interfaces.
    
    The CMake build system has also seen an overhaul.
    It is now possible to build static/shared/CLI separately with -D
    options.
    The shared library only exposes the C API, as it is the only ABI-stable
    API. pkg-configs as well as CMake modules are exported and installed for
    the shared library configuration.

Fix #611.


spirv_msl.hpp
```diff
@@ -49,72 +49,70 @@ struct MSLVertexAttr
...
-// compilation of SPIR-V to MSL if the shader makes use of this vertex attribute.
+// descriptor used in a particular shading stage.
 struct MSLResourceBinding
 {
-	spv::ExecutionModel stage;
+	spv::ExecutionModel stage = spv::ExecutionModelMax;
 	uint32_t desc_set = 0;
 	uint32_t binding = 0;
-
-	uint32_t msl_buffer = 0;
```
HansKristian-Work: While refactoring, this seemed a bit redundant. @billhollings @cdavis5e, is it needed to have 3 different bindings? Bindings of different types cannot alias in SPIR-V. Are you okay with this change?

spirv_msl.hpp
```diff
@@ -269,6 +255,44 @@ class CompilerMSL : public CompilerGLSL
...
+	explicit CompilerMSL(std::vector<uint32_t> spirv);
+	CompilerMSL(const uint32_t *ir, size_t word_count);
+	explicit CompilerMSL(const ParsedIR &ir);
+	explicit CompilerMSL(ParsedIR &&ir);
+
+	// attr is a vertex attribute binding used to match
+	// vertex content locations to MSL attributes. If vertex attributes are provided,
+	// is_msl_vertex_attribute_used() will return true after calling ::compile() if
+	// the location was used by the MSL code.
+	void add_msl_vertex_attribute(const MSLVertexAttr &attr);
```
HansKristian-Work: To make the C API wrapping easier, I change this around to be a bit more "set, and query", rather than relying on mutable pointers being set deep in the implementation. @billhollings @cdavis5e, are you okay with this change?

floooh:
...just my 2 cent from looking at spirv_cross_c.h, but all just subjective "style issues", so feel free to ignore :)

- In my C headers I'm wrapping things like ```SPVC_PUBLIC_API``` in an ```#if !defined(SPVC_PUBLIC_API)``` so that users can override those with whatever they want (for instance I'm not sure whether the ```__declspec(export)``` or ```__attribute__(visibility("default"))``` would disable dead-code-elimination in case of static linking, so that code for unused functions is removed from the exe.

- In my C headers I typedef'ed public structs the same way you're declaring enums, e.g.
    ```c
    typedef my_struct {
        ...
    } my_struct;
    ```
    That way're they're both forward-declarable, and can be written without the struct keyword, e.g. 
    ```my_struct bla;``` instead of ```struct my_struct bla;```.

- I would probably rename the ```*_defaults()``` functions to ```*_init()``` (e.g. see the pthread init functions which basically do the same thing: http://man7.org/linux/man-pages/man3/pthread_attr_init.3.html)

- Related to default initialization, in my C APIs, I got rid of any ```*_init()``` constructor functions by reserving 0 for default values, so that a zero-initialized struct is in a useful default state (basically what would be the result of the ```*_defaults()``` call. The downsize is that this may require extra handling in functions which take such zero-initialized struct arguments, the upside is that C99 designated initialization can be used (but the API can still be C89, it's a 'user-side' thing).

For instance one could write in C99:
```
spvc_compiler_msl_add_vertex_attribute(compiler, &(spvc_msl_vertex_attribute) {
    .location = 1,
    .format = SPVC_MSL_VERTEX_FORMAT_UINT16
});
```
...assuming the other struct members have meaningful default values. This is a nice-to-have thing though, I know that the zero-init-means-default can be tricky, especially when there's underlying C++ code for which the zero-values have no special meaning :)

I think that's it :)

HansKristian-Work:
Thanks for the feedback. I think all of those are reasonable, except the "0 defaults" for C99 initializers. The structs need to match up with the C++ structs, and those don't always have zero initialized values. E.g. for SpvBuiltIn, 0 is Position not "Unknown", which would have been a nicer design. The sentinel needs to be non-zero sadly. Otherwise I'll change it.

spirv_msl.hpp
cdavis5e: What about combined image-samplers? Those consume both a texture and a sampler binding.

spirv_msl.hpp
cdavis5e: I have no objections. Can't speak for @billhollings, though.

billhollings:
> Requesting some feedback from @billhollings and @cdavis5e.

I'll review on Monday...if it can wait that long.

HansKristian-Work:
@billhollings No problem, this isn't urgent.

spirv_msl.hpp
HansKristian-Work: Oh, right, that's a good catch. I'll change it back.

spirv_msl.hpp
billhollings: Yup. Looks fine. I've modified and tested MoltenVK to use the updated "set and query" API wrt vertex attributes and resource bindings...and it all works...and is actually a bit cleaner usage.

PR #890, opened 2019-03-14
MSL: Add support for Metal 2 indirect argument buffers.

### LINKED ISSUE #293: Metal indirect argument buffers
### Feature
SPIRV-Cross needs to be able to generate argument buffer descriptions from given bindings. For example, given the SPIRV generated from the following GLSL:
```glsl
layout(set = 0, binding = 0) uniform texture2D u_Texture;
layout(set = 0, binding = 1) uniform sampler u_Sampler;
```

The Metal-1 version of the PS entry point (using those bindings) would look like:
```cpp
fragment float4 ps_main(VsOutput in [[stage_in]], 
    texture2d<float> u_Texture [[ texture(0) ]],
    sampler u_Sampler [[ sampler(1) ]])
{...}
```

With Metal-2 and the indirect argument buffers, we'd need the following code generated:
```cpp
struct PixelArg {
  texture2d<float> u_Texture [[ id(0) ]];
  sampler u_Sampler [[ id(1) ]];
};
fragment float4 ps_main(VsOutput in [[stage_in]],
  device PixelArg* pixelArg [[ buffer(0) ]]) // note the buffer index
{...}
```

Using indirect argument buffers provides a fast path of binding resources on Metal2, similar to how descriptor sets are bound in Vulkan/DX12.

### Issues
One implementation detail that is not clear is what buffer index to assign to. A vertex program may still have vertex attributes provided in buffers (while everything else will go into the indirect arguments), so a collision is possible. I propose an option to specify the starting indirect buffer binding:
```
spirv-cross --msl-indirect-argument-buffers <start_index>
```

### Background
gfx-rs has the indirect argument code path implemented for the descriptor set binding implementation on Metal2. Here is [the switch](https://github.com/gfx-rs/gfx/blob/cae88c4f1403ff1b7c864e2b21f4bb1c9c393865/examples/core/quad/Cargo.toml#L10) in the core/quad example. We currently provide MSL code manually, which is obviously different between [regular](https://github.com/gfx-rs/gfx/blob/cae88c4f1403ff1b7c864e2b21f4bb1c9c393865/examples/core/quad/shader/quad.metal) and [indirect argument](https://github.com/gfx-rs/gfx/blob/cae88c4f1403ff1b7c864e2b21f4bb1c9c393865/examples/core/quad/shader/quad_indirect.metal) paths. We'd like both versions to be generated by SPIRV-Cross.

HansKristian-Work:
Fix #293.


billhollings:
Hmmm...I wonder what this means in the [Apple arg buffer doc](https://developer.apple.com/documentation/metal/buffers/about_argument_buffers?language=objc)?

>*Writable textures are not supported within an argument buffer.

Does it mean that a texture that a shader will write to must be bound outside an argument buffer...and that its descriptor set would need to use "classic" discrete bindings?

If so...is the `add_push_descriptor_set()` function sufficient for the API to specify that any particular descriptor set should be handled as "classic" bindings? And should it therefore be renamed to something like `add_discrete_descriptor_set()` or `exclude_descriptor_set_from_argument_buffer()`.

@HansKristian-Work Do any of your unit tests include a writable texture to confirm through MSL compilation whether writable textures are allowed in argument buffers...despite that above comment from the docs?

cdavis5e:
> Hmmm...I wonder what this means in the [Apple arg buffer doc](https://developer.apple.com/documentation/metal/buffers/about_argument_buffers?language=objc)?
> 
> > *Writable textures are not supported within an argument buffer.
> 
> Does it mean that a texture that a shader will write to must be bound outside an argument buffer...and that its descriptor set would need to use "classic" discrete bindings?

Only on iOS. Desktop argument buffers *do* support writable textures.

HansKristian-Work:
Yes, there is a unit test for it. It compiled fine for me, but the default is to target macOS, so that makes sense. Yes, calling add_push_descriptor_set for some descriptor set, is enough to force "classic" bindings for any resource which use that DecorationDescriptorSet. You should be able to hoist out single resources if you desire by rewriting its decoration with reflection API before calling compile().

The reason I called it add_push_descriptor_set was to make it more explicit it's supposed to be used for VK_KHR_push_descriptor, but I can change the name if it's going to be used for fallback paths as well.

reference/opt/shaders-msl/comp/argument-buffers-image-load-store.msl2.argument.comp
```diff
@@ -0,0 +1,17 @@
+#include <metal_stdlib>
+#include <simd/simd.h>
+
+using namespace metal;
+
+struct spvDescriptorSetBuffer0
+{
+    texture2d<float, access::write> uImage [[id(1)]];
```
HansKristian-Work: @billhollings This doesn't work on iOS? ._.

billhollings:
@cdavis5e 

> Only on iOS. Desktop argument buffers _do_ support writable textures.

Is this for all macOS...or just Tier 2...meaning only on discrete GPU's...which is not all macOS devices?

In any case...we will need to support iOS with this API.

@HansKristian-Work 

>I can change the name if it's going to be used for fallback paths as well.

I think that makes sense.

>You should be able to hoist out single resources if you desire by rewriting its decoration with reflection API before calling compile().

Interesting. So we could hoist out a single writable texture from a descriptor set...and leave everything else in the argument buffer? That sounds good.

cdavis5e:
> > Only on iOS. Desktop argument buffers _do_ support writable textures.
> 
> Is this for all macOS...or just Tier 2...meaning only on discrete GPU's...which is not all macOS devices?

This is for all macOS.

reference/opt/shaders-msl/comp/argument-buffers-image-load-store.msl2.argument.comp
billhollings: Yes...on iOS...we'll have to pull writable textures out via `add_discrete_descriptor_set()`. 

I think we can handle that in the API...especially since the two have to be synchronized. It should be sufficient if SPRIV-Cross detects writability and throw an error.

spirv_cross_c.cpp
```diff
@@ -699,6 +703,19 @@ spvc_result spvc_compiler_msl_add_resource_binding(spvc_compiler compiler,
 	return SPVC_SUCCESS;
 }
 
+spvc_result spvc_compiler_msl_add_push_descriptor_set(spvc_compiler compiler, unsigned desc_set)
```
billhollings: Can we change this to `add_discrete_descriptor_set()`...as discussed in main conversation?

HansKristian-Work:
It should be possible to hoist out a single resource indeed. Just reassign the set for the resource you want to some magic set index which is disabled for use with argument buffers, and use that set index in the MSLResourceBinding. Any set index above the 8 set limit is always assumed to be a "classic" binding FWIW.

spirv_cross_c.cpp
HansKristian-Work: Yes.

HansKristian-Work: Renamed now.

reference/opt/shaders-msl/comp/argument-buffers-image-load-store.msl2.argument.comp
HansKristian-Work: Added a check for this case.

billhollings:
I'm okay with this now. I think we can flush out any further API or behaviour tweaks as @cdavis5e does the [implementation](https://github.com/KhronosGroup/MoltenVK/issues/523) in MoltenVK.

PR #907, opened 2019-03-25
GLSL: Add support for SPV_NV_ray_tracing

pmoursnv:
This adds support for all the SPV_NV_ray_tracing tokens so that SPIR-V modules with that extension are successfully parsed by SPIRV-cross and generate valid GLSL output.


spirv_glsl.cpp
```diff
@@ -410,6 +410,17 @@ void CompilerGLSL::find_static_extensions()
...
 
+	case ExecutionModelRayGenerationNV:
+	case ExecutionModelIntersectionNV:
+	case ExecutionModelAnyHitNV:
+	case ExecutionModelClosestHitNV:
+	case ExecutionModelMissNV:
+	case ExecutionModelCallableNV:
+		if (options.es || options.version < 460)
+			SPIRV_CROSS_THROW("Ray tracing shaders require non-es profile with version 460 or above.");
+		require_extension("GL_NV_ray_tracing");
```
HansKristian-Work: Use `require_extension_internal()`

spirv_glsl.cpp
```diff
@@ -8980,6 +9034,25 @@ void CompilerGLSL::emit_instruction(const Instruction &instruction)
 		GLSL_BFOP(unsupported_FUnordGreaterThanEqual);
 		break;
 
+	case OpReportIntersectionNV:
+		statement("reportIntersectionNV(", to_func_call_arg(ops[0]), ", ", to_func_call_arg(ops[1]), ");");
```
HansKristian-Work: Just use `to_expression` here. `to_func_call_arg` is used for OpFunctionCall is certain scenarios.

HansKristian-Work: Same with OpTraceNV and OpExecuteCallableNV.

pmoursnv:
Thanks! Things should be fixed now. I also added reflection support to query for acceleration structures and ray tracing execution model types.

pmoursnv:
I was a bit surprised that the regression test scripts also created reference shaders with GL-semantics, even though those shaders should only be used with Vulkan-semantics (hence the `.vk` in the file name, or so I thought). Not sure if that is intentional?

HansKristian-Work:
Thanks. There is an option to disable GL files being generated, tag the files with `.nocompat.` IIRC. I'll review the changes tomorrow.

HansKristian-Work:
Thanks! Merged. Fixed up some minor issues with the .nocompat. stuff and reflection for C API and stuff like that.

PR #919, opened 2019-03-28
MSL: Declare gl_WorkGroupSize constant with [[maybe_unused]].

### LINKED ISSUE #915: MSL: gl_WorkGroupSize warning spam
Almost all compute shaders currently trigger warnings in the MSL compiler when gl_WorkGroupSize is not accessed.

HansKristian-Work:
Avoids ugly warnings on nearly every compute shader.
We could do analysis to detect whether we need to emit this constant,
but it's a bit tedious to figure out if an OpConstantComponent is
actually used by opcodes, so just make it simple.

Fix #915.


HansKristian-Work:
Travis is taking forever, and verified locally.

PR #920, opened 2019-03-28
Fix crash when backend.int16_t_literal_suffix set to null.

billhollings:
The design of backend.int16_t_literal_suffix and backend.uint16_t_literal_suffix
allows them to be set to null, but that was not always tested for.
I have removed the expectation that they can be null and set
backend.int16_t_literal_suffix to "" when no suffix is needed.
That has the same effect, and seemed to be a more usable and defensive approach.

### DIFF
--- a/spirv_msl.cpp
+++ b/spirv_msl.cpp
@@ -563,7 +563,7 @@ string CompilerMSL::compile()
 	backend.null_pointer_literal = "nullptr";
 	backend.float_literal_suffix = false;
 	backend.uint32_t_literal_suffix = true;
-	backend.int16_t_literal_suffix = nullptr;
+	backend.int16_t_literal_suffix = "";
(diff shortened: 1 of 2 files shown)


spirv_msl.cpp
```diff
@@ -563,7 +563,7 @@ string CompilerMSL::compile()
 	backend.null_pointer_literal = "nullptr";
 	backend.float_literal_suffix = false;
 	backend.uint32_t_literal_suffix = true;
-	backend.int16_t_literal_suffix = nullptr;
+	backend.int16_t_literal_suffix = "";
```
HansKristian-Work: Needs fixup in HLSL backend as well.

HansKristian-Work: nvm, fixed locally.

HansKristian-Work:
Fixed trivial one-liner locally.

PR #922, opened 2019-04-02
Improve performance by reducing global allocations.

### LINKED ISSUE #923: Reduce global allocation pressure

HansKristian-Work:
SPIRV-Cross is rather heavy on global allocations with the use of data structures required to implement the codegen. There are many, many opportunities to pool them, avoid them, allocate more on stack, etc.

This PR is about reducing global allocations as much as is reasonable. It is not about rewriting lots of code significantly, but replace data structures where it makes sense.

- Replace ostringstream with custom implementation. This is the biggest win.
  ~30% performance uplift on vector-shuffle-oom test.
  Allocations are measurably reduced in Valgrind.
    
- Replace std::vector with SmallVector.
  Classic malloc optimization, small vectors are backed by inline data.
  ~ 7-8% gain on vector-shuffle-oom on GCC 8 on Linux.
    
- Use an object pool for IVariant type.
  We generally allocate a lot of SPIR* objects. We can amortize these
  allocations neatly by pooling them.

- Add a way to pass down --iterations down to spirv-cross in test_shaders.py. This allows for large performance sweeps over a shader test set.
    
- ~15% overall uplift on ./test_shaders.py --iterations 10000 shaders/. Which has lots of other overhead and tiny shaders.

I spent some time on trying to do similar optimizations for unordered_map and friends but I failed to implement a working std::allocator replacement. The only way forward here is to write custom hashmap implementations I think, but that is a lot of work we can punt.

Fix #923.


spirv_common.hpp
```diff
@@ -105,6 +104,310 @@ class CompilerError : public std::runtime_error
...
+	T *data()
+	{
+		return nullptr;
+	}
+};
+
+// Simple vector which supports up to N elements inline, without malloc/free.
+// We use a lot of throwaway vectors all over the place which triggers allocations.
+template <typename T, size_t N = 8>
+class SmallVector
```
cdavis5e: Consider deriving this from a base. Then you can pass references to the base around, and it won't matter how many inline elements the actual object uses. See discussion at KhronosGroup/MoltenVK#376.

HansKristian-Work: Right.

HansKristian-Work:
@billhollings @cdavis5e This should stave off a lot of the overhead. I tried building this commit in MoltenVK without issue. If you have some performance data points from real games, that'd be nice to see!

billhollings:
@HansKristian-Work 

Thanks for all the effort on this!

MoltenVK has the `MVK_CONFIG_PERFORMANCE_TRACKING` env var...which if run with a Debug build of MoltenVK...will cause shader conversion performance statistics to be logged (I should change the Debug build requirement for logging to an env var too at some point). With a Release build...the performance stats are not logged...but are queryable through the API.

I'll try running a comparison against Dota2 when I get a chance.

HansKristian-Work:
Cool. Hopefully we see a substantial gain.

billhollings:
@HansKristian-Work 

Hmmm...in Dota2 I am not seeing any appreciable change in the average conversion time across 350 shaders. 

What kind of numbers are you seeing...in other words...what should I expect to see? I'm suspicious that I'm getting more or less identical results between Before vs After. Perhaps I have screwed up my testing.

And a live game is pretty noisy (as is evident from the variations across runs from a single build) and may not be an accurate measure of numerical improvements.

The values below are 3 separate runs of Dota2 startup for each of Before and After...and include just the SPIRV-Cross conversions (no loading or MSL compilation). Each run converts about 350 shaders.

BEFORE:

```
[mvk-info] Performance to convert SPIR-V to MSL source code curr: 4.818 ms, avg: 3.212 ms, min: 0.304 ms, max: 33.480 ms, count: 335
[mvk-info] Performance to convert SPIR-V to MSL source code curr: 4.443 ms, avg: 5.116 ms, min: 0.300 ms, max: 94.172 ms, count: 344
[mvk-info] Performance to convert SPIR-V to MSL source code curr: 5.518 ms, avg: 5.602 ms, min: 0.395 ms, max: 114.684 ms, count: 347
```

AFTER:

```
[mvk-info] Performance to convert SPIR-V to MSL source code curr: 3.826 ms, avg: 5.436 ms, min: 0.448 ms, max: 162.127 ms, count: 350
[mvk-info] Performance to convert SPIR-V to MSL source code curr: 3.748 ms, avg: 3.593 ms, min: 0.384 ms, max: 99.548 ms, count: 340
[mvk-info] Performance to convert SPIR-V to MSL source code curr: 3.793 ms, avg: 5.653 ms, min: 0.505 ms, max: 110.268 ms, count: 348
```

HansKristian-Work:
Thanks for testing. That's very erratic performance, doesn't seem like we can draw any conclusions. I'll try to consider how we can isolate this from the application with a specialized Fossilize replayer perhaps ... That max timing value looks very suspicious.

billhollings:
Yeah...there is a lot of multithreading going on in shader and pipeline prep in Dota2. So I suspect the max values are outliers that get caught being delayed because of thread starvation.

I could probably add similar timing tracking to the MoltenVKShaderConverter tool and run it across an entire directory tree. Or something similar could be added to `spirv-cross main` to do the same?

PR #928, opened 2019-04-03
Update to latest glslang/SPIRV-Tools.

fjhenigman:


HansKristian-Work:
Any particular reason for this update other than staying up-to-date with master?

fjhenigman:
> Any particular reason for this update other than staying up-to-date with master?

I need to roll spirv-tools in shaderc to pick up a fix, but then shaderc's spirv-cross wrapper (which I test with the spirv-cross tests) fails, so I need to update the spirv-cross tests.

Is there a reason not to do this update?

Thanks.

HansKristian-Work:
That makes sense. I see there's some change in the reference files. I'll investigate that tomorrow.

fjhenigman:
Appreciate the quick merge!  Makes my life easier.

PR #932, opened 2019-04-06
MSL: Fix infinite CAS loop on atomic_compare_exchange_weak_explicit().

billhollings:
The CAS loop on atomic_compare_exchange_weak_explicit() became infinite if the atomic comparison was not equal. See further notes in the in-code comments.


cdavis5e:
Well no wonder that test caused a GPU hang!

HansKristian-Work:
Fun, fun :P

PR #934, opened 2019-04-08
Add OpUndef instruction to block's instruction list for completeness.

lifpan:


HansKristian-Work:
Undefs are hoisted out separately. Is there any reason why these have to be part of the block ops? Have to tested that it works correctly? None of the backends expect this.

lifpan:
This would not impact translation from SPIR-V to high level language. I will check the failures.
My application traverses all instructions and build symbol for each of them. The symbol table helps to translate SPIR-V to other IR. If OpUndef is not in the instruction list, no symbol is built for it, and symbol table lookup failure will occur.

PR #940, opened 2019-04-09
Use constant address space for SPIR-V parameters when generating tessellation control shader.

mbarriault:
In MoltenVK, `MVKCommandEncoder::setComputeBytes` can call either `setBytes` or `setBuffer`, but Metal only allows `setBytes` to be used with the `constant` address space. As a small read-only buffer, this should be a safe change for both cases.


cdavis5e:
Only one problem: The `constant` address space imposes an alignment requirement of 256 bytes. For an indirect draw, this buffer comes from the indirect draw parameters, which we can't guarantee will be 256-byte aligned. I think this needs to be fixed MoltenVK-side.

mbarriault:
I have a number of other fixes related to tessellation in MoltenVK and a few bug reports for issues I couldn't solve, I'll file that bug report there too and leave this open for now.

mbarriault:
@cdavis5e Do you have documentation on this 256-byte requirement? I can't find it in the MSL spec nor MTLComputeCommandEncoder documentation.

cdavis5e:
> @cdavis5e Do you have documentation on this 256-byte requirement? I can't find it in the MSL spec nor MTLComputeCommandEncoder documentation.

From `-[MTLComputeCommandEncoder setBuffer:atOffset:]`:

> For buffers in the constant address space, the offset must be aligned to 256 bytes in macOS.

So iOS does not have this restriction, but macOS does.

mbarriault:
Okay, noted, thanks!

HansKristian-Work:
My primary concern here is alignment indeed. I'll let @cdavis5e have the final word on this PR.

mbarriault:
The pull request has been updated to fix the tests for this change.

spvpatch.patch
```diff
@@ -0,0 +1,39 @@
+diff --git a/spirv_msl.cpp b/spirv_msl.cpp
```
HansKristian-Work: ?

mbarriault: Whoops, that's another change I've been toying with but don't have the full understanding yet to commit. Shouldn't have been included, not sure how it got caught up, will update pull request shortly to remove it.

cdavis5e:
So, I've thought about this, and I think what we should do is: the address space should be changed to `constant`, but *only* for tessellation control shaders. MoltenVK always uses `setComputeBytes()` here, even in indirect draws, so we don't have the issue with buffer alignment here.

But for vertex shaders, we need to keep using the `device` AS.

mbarriault:
This change is indeed only for tessellation control shaders. However MoltenVK doesn't strictly use `setComputeBytes`: if dynamic Metal buffer feature isn't enabled, it falls back on `setComputeBuffer`.

cdavis5e:
> This change is indeed only for tessellation control shaders.

Are you sure? The changes to the tests indicate that it also affects vertex shaders.

> However MoltenVK doesn't strictly use `setComputeBytes`: if dynamic Metal buffer feature isn't enabled, it falls back on `setComputeBuffer`.

Yes, but in that case it gets a temporary buffer and ensures that the buffer offset is aligned correctly for the `constant` AS.

mbarriault:
Yes, you're right. In `CompilerMSL::entry_point_args_builtin` and `MVKCmdDrawIndirect::encode`. What are the performance characteristics of `device` vs `constant`? My intuition (OpenCL background) suggests that, for these buffers, `constant` should be preferred, but as an array of `uint32_t`s that would be rather wasteful in memory.

I'll update to your suggestion and only use `constant` in the tessellation control case.

mbarriault:
Hm, nope, that won't do either. The non-indirect case still uses `setVertexBytes` and crashes with `device`.

cdavis5e:
KhronosGroup/MoltenVK#572 should fix that.

mbarriault:
Updated pull request for merged KhronosGroup/MoltenVK#572.

spirv_msl.cpp
```diff
@@ -5608,13 +5608,20 @@ void CompilerMSL::entry_point_args_builtin(string &ep_args)
...
+		if (get_execution_model() == ExecutionModelTessellationControl)
 		{
 			if (!ep_args.empty())
 				ep_args += ", ";
 			ep_args +=
-			    join("device uint* spvIndirectParams [[buffer(", msl_options.indirect_params_buffer_index, ")]]");
-		}
+			    join("constant uint* spvIndirectParams [[buffer(", msl_options.indirect_params_buffer_index, ")]]");
+		}
+        else if (stage_out_var_id)
```
cdavis5e: Formatting. This project uses tabs for indentation. (If you have `clang-format(1)` installed, you can use the `format_all.sh` script to automate this.)

mbarriault: Corrected.

mbarriault:
With KhronosGroup/MoltenVK#572 merged, is this clear to merge in now? MoltenVK tessellation is still broken without this commit.

cdavis5e:
I have no objections. @HansKristian-Work?

HansKristian-Work:
Then I have none either. Merged. Thanks!

Issue #342, opened 2017-11-23
HLSL: unhandled storage image format "unknown"


TiemoJung:
Hi,

the to HLSL translator throws an exception if it encounters a storage image with unknown target format.

HansKristian-ARM:
Hm ... I'm unsure how this would be implemented in a sensible way as the format seems to have to be defined in RWBuffer<T> in most cases. Would float4/int4/uint4 fallback work?

TiemoJung:
In our case its mostly float4, i also take a look at our compiler pipeline to emit the proper format info, if possible.

PR #945, opened 2019-04-12
Don't apply vertex attribute remapping to other interface blocks

ashleyharris-maptek-com-au:
HLSLVertexAttributeRemap appears intended to remap Vertex Attributes to new names, however currently it applies to all interface blocks. This leads to HLSL code like:

Vertex Shader:
```
struct SPIRV_Cross_Input
{
    float3 VertexCoord : VertexCoord;  // Position 0
    float3 VertexNorm : VertexNorm;   // Position 1
    float4 VertexColour : VertexColour;   // Position 2
};

struct SPIRV_Cross_Output
{
    float4 FragColour : VertexCoord;  // Position 0. Thus also tagged as VertexCoord
    float4 gl_Position : SV_Position;
};
```

This is in itself not a dealbreaker, only weird. Where it becomes an issue is when the fragment shader is converted independently of the vertex shader, and thus loses the vertex attributes:

Fragment Shader:
```
struct SPIRV_Cross_Input
{
    float4 FragColour : TEXCOORD0;
};
```

This gives the error:
D3D11 ERROR: ID3D11DeviceContext::Draw: Vertex Shader - Pixel Shader linkage error: Signatures between stages are incompatible. The input stage requires Semantic/Index (TEXCOORD,0) as input, but it is not provided by the output stage. [ EXECUTION ERROR #342: DEVICE_SHADER_LINKAGE_SEMANTICNAME_NOT_FOUND]

### DIFF
--- a/spirv_hlsl.cpp
+++ b/spirv_hlsl.cpp
@@ -744,17 +744,25 @@ string CompilerHLSL::to_interpolation_qualifiers(const Bitset &flags)
 	return res;
 }
 
-std::string CompilerHLSL::to_semantic(uint32_t vertex_location)
+std::string CompilerHLSL::to_semantic(uint32_t location, ExecutionModel em, StorageClass sc)
 {
-	for (auto &attribute : remap_vertex_attributes)
-		if (attribute.location == vertex_location)
-			return attribute.semantic;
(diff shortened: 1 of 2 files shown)


HansKristian-Work:
Thanks!

PR #946, opened 2019-04-12
TestShaders now looks for spirvcross, and fix clone time-out error in checkout_glslang_spirv_tools.sh

ashleyharris-maptek-com-au:
These are some changes I had to make to the testing scripts to get the tests running:

So "git clone git://github.com/KhronosGroup/glslang.git" and friends times out on my machine. Changing it to https gets it working for me. Since the path github first gives on the web interface is the https one, I'm assuming it's preferred. 

test_shaders.py looks for a ./spirv-cross in the current directory to execute. This appears to need to be a symlink, it appears you need to manually create it, and the need for this isn't referenced in the documentation, and it fails messily if it can't find it. I've changed the script so that if it can't find this symlink, it falls back to the output paths for the visual studio project.

### DIFF
--- a/checkout_glslang_spirv_tools.sh
+++ b/checkout_glslang_spirv_tools.sh
@@ -13,7 +13,7 @@ else
 	echo "Cloning glslang revision $GLSLANG_REV."
 	mkdir -p external
 	cd external
-	git clone git://github.com/KhronosGroup/glslang.git
+	git clone https://github.com/KhronosGroup/glslang.git
 	cd glslang
 	git checkout $GLSLANG_REV
 fi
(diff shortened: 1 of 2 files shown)


HansKristian-Work:
Right, there is some maintenance I need to do to support using the CMake build with the test_shader scripts. It's too hard coded for this purpose (tied in with the convenience Makefile build), but I think the appropriate solution will look quite different. I needed to do similar things when supporting MSVC fully on Travis CI, but these scripts have traditionally only been run locally by contributors.

ashleyharris-maptek-com-au:
Ok - if there's a better solution then by all means. Did you want me to promote it from draft so there's something in the meantime? Or we wait for the proper solution?

With this patch the steps in the readme under "Contributing/Testing" work (for me at least). But an alternative to patching the python script may be to update the documentation to include the part about getting the executable into the current directory.

HansKristian-Work:
I'll try to get in a solution today.

HansKristian-Work:
Both issues should be addressed now on master. Check out README. Verify cmake-built test_shaders.sh locally on Windows.

PR #947, opened 2019-04-12
Can build SPIRV-Cross as a library (.Lib)

ashleyharris-maptek-com-au:
Adds a new visual studio project allowing spirv-cross to be built as a library (.lib) and compilers to be linked in directly with user C++ code, rather than only compiling as a stand-alone .exe.


HansKristian-Work:
Is this targeting a very old commit? I see some diffs against msvc/ build the MSVC build was removed a while back (it had been broken for over a year and noone reported it), and that is intentional. I will not maintain a standalone MSVC build anymore. The main build system is CMake and that can build static, shared and CLI targets with proper install targets.

PR #961, opened 2019-04-26
Fix a copy-pasto.

cdavis5e:
(I don't know why those two variable pointer tests suddenly need
updating...)


HansKristian-Work:
Looks good. I think you need to update the local glslang/spirv-opt, the reference commits changed not that long ago.

cdavis5e:
I thought I did that. I guess I didn't rebuild them.

PR #973, opened 2019-05-09
MSL: Support image swizzle workaround with argument buffers

### LINKED ISSUE #972: MSL: Texture swizzle aux buffer does not work with indirect argument buffers
The binding space for aux buffers assumes a flat binding space.

HansKristian-Work:
There is no good reason to expand the aux buffer, so name it
appropriately.

Make the code cleaner by emitting a straight pointer to uint rather than
a dummy struct which only contains a single unsized array member anyways.

Implement support for argument buffers when using image swizzle workaround.

Fix #972.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -267,7 +267,7 @@ endif()
 
 if (SPIRV_CROSS_SHARED)
 	set(spirv-cross-abi-major 0)
-	set(spirv-cross-abi-minor 7)
+	set(spirv-cross-abi-minor 8)
 	set(spirv-cross-abi-patch 0)
 	set(SPIRV_CROSS_VERSION ${spirv-cross-abi-major}.${spirv-cross-abi-minor}.${spirv-cross-abi-patch})
 	set(SPIRV_CROSS_INSTALL_LIB_DIR ${CMAKE_INSTALL_PREFIX}/lib)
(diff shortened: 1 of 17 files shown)


HansKristian-Work:
This is a preparation patch to support OpArrayLength in MSL properly. The code paths for buffer sizes will be very similar to this code path.

Ping @billhollings @cdavis5e.

Some changes:
- Rather than having a buffer called "aux" I think it makes sense to rename this to swizzle buffer. There is very little reason to make it a struct. Rather than a struct, a `constant uint *` array is emitted instead. The GPU ABI doesn't change since it was a struct with unsized array.
- To support argument buffers, I added a `const uint *` buffer to the argument buffer, which holds all swizzle constants for all textures in that set. The indexing is based on resource index `[[id(N)]]` within the argument buffer, just like the old-school descriptors.

I'm not sure if you've started working on argument buffer support in MoltenVK yet, but please have a look at this to make sure it's something which could work in Metal.

cdavis5e:
I'm wondering if we could instead inline the swizzle constants for images in argument buffers into the argument buffer. I don't know, though, if that would be better or worse than keeping them in an array and slurping that whole into a buffer, like we do for push descriptors.

HansKristian-Work:
I don't think that's feasible because of array indexing. I just added fixes for swizzle with arrays-of-textures on master, and you need to dynamically index the swizzle buffer for that.

HansKristian-Work:
Doesn't sound like there are any significant issues doing this, so merging. Code-gen can always be fixed if it doesn't work out.

PR #975, opened 2019-05-11
GLSL: Add option to disable buffer blocks regardless of version

alpqr:
Introduce a way to force treating all GLSL versions as "legacy" with regards to buffer blocks. This can be very useful for renderers that must work with multiple GLSL versions in their OpenGL backend due to having to support operating both in core and compatibility (or 2.x) contexts, the choice being outside of the renderer's control.  This is typically achieved by having (generating) two or more versions of the GLSL source (for example, 120 for compatibility/2.1 and 150 for core), but it would then be ideal if all the GLSL variants could use the same (legacy) approach for uniform blocks, instead of forcing the renderer to support both plain old uniforms and uniform buffers.

### DIFF
--- a/spirv_cross_c.h
+++ b/spirv_cross_c.h
@@ -33,7 +33,7 @@ extern "C" {
 /* Bumped if ABI or API breaks backwards compatibility. */
 #define SPVC_C_API_VERSION_MAJOR 0
 /* Bumped if APIs or enumerations are added in a backwards compatible way. */
-#define SPVC_C_API_VERSION_MINOR 6
+#define SPVC_C_API_VERSION_MINOR 7
 /* Bumped if internal implementation details change. */
 #define SPVC_C_API_VERSION_PATCH 0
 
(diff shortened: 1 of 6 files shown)


spirv_cross_c.h
```diff
@@ -33,7 +33,7 @@ extern "C" {
 /* Bumped if ABI or API breaks backwards compatibility. */
 #define SPVC_C_API_VERSION_MAJOR 0
 /* Bumped if APIs or enumerations are added in a backwards compatible way. */
-#define SPVC_C_API_VERSION_MINOR 6
+#define SPVC_C_API_VERSION_MINOR 7
```
HansKristian-Work: You need to update the version in the CMakeLists.txt as well.

spirv_glsl.cpp
```diff
@@ -1461,6 +1461,8 @@ string CompilerGLSL::layout_for_variable(const SPIRVariable &var)
 
 	// GL 3.0/GLSL 1.30 is not considered legacy, but it doesn't have UBOs ...
 	bool can_use_buffer_blocks = (options.es && options.version >= 300) || (!options.es && options.version >= 140);
+        // pretend no UBOs when options say so
+	can_use_buffer_blocks &= options.enable_buffer_blocks;
```
HansKristian-Work: Use logical operations here. This hidden bool -> arithmetic and -> bool will trigger warnings on many compilers.

spirv_glsl.hpp
```diff
@@ -99,6 +99,10 @@ class CompilerGLSL : public Compiler
 		// In non-Vulkan GLSL, emit push constant blocks as UBOs rather than plain uniforms.
 		bool emit_push_constant_as_uniform_buffer = false;
 
+		// Enable using buffer blocks. If disabled, uniform/shader storage/push constant blocks are never
+		// emitted as buffer blocks, regardless the GLSL version.
+		bool enable_buffer_blocks = true;
```
HansKristian-Work: Shader storage buffers must be banned if using this option, since it's not possible to use legacy uniforms with that feature. There needs to be some kind of check for that, or SSBOs should ignore this option, and the option name must be clearer, like "enable_uniform_buffer_blocks" or something.

alpqr: Ok, renamed to emit_uniform_buffer_as_plain_uniforms with false by default (this also follows the pattern set by emit_push_constant_as_uniform_buffer better), and it is now ignored for pc and ssb blocks.

PR #982, opened 2019-05-15
MSL: Add support for subgroup operations.

cdavis5e:
Some support for subgroups is present starting in Metal 2.0 on both iOS
and macOS. macOS gains more complete support in 10.14 (Metal 2.1).

Some restrictions are present. On iOS and on macOS 10.13, the
implementation of `OpGroupNonUniformElect` is incorrect: if thread 0 has
already terminated or is not executing a conditional branch, the first
thread that *is* will falsely believe itself not to be. Unfortunately,
this operation is part of the "basic" feature set; without it, subgroups
cannot be supported at all.

The `SubgroupSize` and `SubgroupLocalInvocationId` builtins are only
available in compute shaders (and, by extension, tessellation control
shaders), despite SPIR-V making them available in all stages. This
limits the usefulness of some of the subgroup operations in fragment
shaders.

Although Metal on macOS supports some clustered, inclusive, and
exclusive operations, it does not support them all. In particular,
inclusive and exclusive min, max, and, or, and xor; as well as cluster
sizes other than 4 are not supported. If this becomes a problem, they
could be emulated, but at a significant performance cost due to the need
for non-uniform operations.

### DIFF
--- a/spirv_msl.cpp
+++ b/spirv_msl.cpp
@@ -7059,7 +7059,6 @@ void CompilerMSL::emit_subgroup_op(const Instruction &i)
 		{
 		default:
 			SPIRV_CROSS_THROW("iOS only supports quad-group operations.");
-		case OpGroupNonUniformElect:
 		case OpGroupNonUniformBroadcast:
 		case OpGroupNonUniformShuffle:
 		case OpGroupNonUniformShuffleXor:
@@ -7077,7 +7076,6 @@ void CompilerMSL::emit_subgroup_op(const Instruction &i)
 		{
 		default:
 			SPIRV_CROSS_THROW("Subgroup ops beyond broadcast and shuffle on macOS require Metal 2.0 and up.");
-		case OpGroupNonUniformElect:
 		case OpGroupNonUniformBroadcast:
 		case OpGroupNonUniformShuffle:
 		case OpGroupNonUniformShuffleXor:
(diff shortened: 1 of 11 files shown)


spirv_msl.cpp
```diff
@@ -6748,6 +7041,245 @@ string CompilerMSL::image_type_glsl(const SPIRType &type, uint32_t id)
...
+	switch (op)
+	{
+	case OpGroupNonUniformElect:
+		// Vulkan spec says we have to support this if we support subgroups at all.
+		// But Metal prior to macOS 10.14 doesn't have the simd_is_first() function, and
+		// iOS doesn't have it at all. So we fake it by comparing the subgroup-local
+		// ID to 0. This isn't quite correct: this is supposed to return if we're the
+		// lowest *active* thread, but we'll otherwise be unable to support subgroups
+		// on macOS 10.13 or iOS.
+		if (msl_options.is_macos() && msl_options.supports_msl_version(2, 1))
```
HansKristian-Work: Can't we implement the elect on older targets in terms of simd_broadcast_first(subgroup_id) == subgroup_id?

cdavis5e: Unfortunately, no. Neither `simd_broadcast_first()` nor `quad_broadcast_first()` is available on iOS or macOS 10.13. I guess we won't be able to enable subgroup support for these devices.

oscarbg:
@cdavis5e this is nice! one step closer to Vulkan 1.1 for MoltenVK..but you haven't contributed needed MoltenVK changes to advertise subgroup ops, right? (at least updating SPIRV-Cross dep for having this commit)

PR #990, opened 2019-05-22
MSL: Stop exiting early from extra tesc instances.

cdavis5e:
Unfortunately, my worst fears were in fact confirmed. Some hardware,
particularly NVIDIA hardware, doesn't handle the early exits well. Upon
executing a barrier, the GPU will often hang, causing weird graphical
glitches at best and kernel panics at worst.

Instead, we'll do what I should've done and wanted to do in the first
place: guard all writes to externally visible memory with an `if` block,
to ensure the extra spawned invocations have no visible effect.
Unfortunately, this will keep the GPU busy executing code that will
ultimately have no effect, but it's necessary to prevent GPU hangs by
ensuring that every thread hits all the barriers.

The formatting is... not ideal. We really want to stick consecutive
writes in the same `if` block. But I'm not sure how to do that yet.

### DIFF
--- a/reference/opt/shaders-msl/asm/tesc/tess-fixed-input-array-builtin-array.invalid.asm.tesc
+++ b/reference/opt/shaders-msl/asm/tesc/tess-fixed-input-array-builtin-array.invalid.asm.tesc
@@ -52,22 +52,38 @@ kernel void main0(main0_in in [[stage_in]], uint gl_InvocationID [[thread_index_
     if (gl_InvocationID < spvIndirectParams[0])
         gl_in[gl_InvocationID] = in;
     threadgroup_barrier(mem_flags::mem_threadgroup);
-    if (gl_InvocationID >= 3)
-        return;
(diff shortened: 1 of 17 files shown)


cdavis5e:
> Is it possible to abuse 2 or 3 dimensional dispatches and strides to hack around attribute offsets?

I think that will only work if the input count is an integer multiple of the output count.

HansKristian-Work:
Is it possible to just read from the vertex output directly in tesc rather than relying on attributes? That way you could just loop N iterations in worst case to do the initial copy, or don't bother with the copy at all, just index gl_in somehow ...

cdavis5e:
At this point, I'm no longer certain that this was in fact the cause of the issues I was seeing. The game we were seeing problems in uses equal numbers of input and output control points, so it couldn't be hitting this. I'm considering withdrawing this PR altogether.

HansKristian-Work:
I would be in favor of that. Merging this PR is the last resort in my book, and I would definitely like to avoid merging this workaround if possible.

PR #996, opened 2019-05-28
Support emitting OpLine directive.

### LINKED ISSUE #989: Suggestion: emit #line statements for SPIR-V OpLine opcode
It would be useful if there's an option to emit ```#line``` statements in the generated source code guided by the SPIR-V OpLine  opcode.

With this, a shader-cross-compiler tool can map errors and warnings from the Metal- and HLSL-compilers back to the original source file (at least roughly).

For example:

The Metal compiler might generate warnings like this on a piece of generated Metal source code, such "is used uninitialized" cases don't generate a warning on the original GLSL when compiled with glslang:

```bash
test.metal:34:13: warning: variable 'param' is used uninitialized whenever 'if' condition is false
      [-Wsometimes-uninitialized]
test.metal:74:23: note: uninitialized use occurs here
        float4 _593 = param;
```

Manually adding random ```#line``` statements like this:
```cpp
        float _636 = _656;
#line 12 "bla.glsl"
        float4 param;
#line 13 "bla.glsl"
        if (_656 < 0.100000001490116119384765625)
```
...changes the warning output to:

```cpp
bla.glsl:13:13: warning: variable 'param' is used uninitialized whenever 'if' condition is false
      [-Wsometimes-uninitialized]
```

It seems that one ```#line``` statement needs to be emitted for each generated source code line, without one, the compiler seems to increment the internal line counter for each parsed source code line, and a ```#line``` statement just resets this counter.

HansKristian-Work:
Facilitates easier mapping from source language to cross-compiled output
in tooling.

Fix #989.

### DIFF
--- a/spirv_glsl.hpp
+++ b/spirv_glsl.hpp
@@ -103,6 +103,10 @@ public:
 		// Does not apply to shader storage or push constant blocks.
 		bool emit_uniform_buffer_as_plain_uniforms = false;
 
+		// Emit OpLine directives if present in the module.
+		// May not correspond exactly to original source, but should be a good approximation.
+		bool emit_line_directives = true;
+
 		enum Precision
 		{
 			DontCare,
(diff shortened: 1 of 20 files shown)


spirv_glsl.hpp
```diff
@@ -103,6 +103,10 @@ class CompilerGLSL : public Compiler
 		// Does not apply to shader storage or push constant blocks.
 		bool emit_uniform_buffer_as_plain_uniforms = false;
 
+		// Emit OpLine directives if present in the module.
+		// May not correspond exactly to original source, but should be a good approximation.
+		bool emit_line_directives = true;
```
grovesNL: Is it intentional to default to `true` here but `false` in the CLI?

HansKristian-Work:
Oops. Yes, it should be false. Thanks for notifying.

grovesNL:
Thanks!

PR #1001, opened 2019-05-31
MSL: Support SPV_KHR_multiview.

cdavis5e:
This is needed to support `VK_KHR_multiview`, which is in turn needed
for Vulkan 1.1 support. Unfortunately, Metal provides no native support
for this, and Apple is once again less than forthcoming, so we have to
implement it all ourselves.

Tessellation and geometry shaders are deliberately unsupported for now.
The problem is that the current implementation encodes the `ViewIndex`
as part of the `InstanceIndex`, which in the SPIR-V environment at least
only exists in the vertex shader. So we need to work out a way to pass
the view index along to the later stages.

This implementation runs vertex shaders for all views up to the highest
bit set in the view mask, even those whose bits are clear. The fragments
for the inactive views are then discarded. Avoiding this is difficult:
calculating the view indices becomes far more complicated if we can only
run for those views which are set in the mask.

### DIFF
--- a/reference/opt/shaders-msl/vulkan/frag/basic.multiview.nocompat.vk.frag
+++ b/reference/opt/shaders-msl/vulkan/frag/basic.multiview.nocompat.vk.frag
@@ -1,0 +1,32 @@
+#include <metal_stdlib>
+#include <simd/simd.h>
+
+using namespace metal;
+
+struct main0_out
+{
+    float4 FragColor [[color(0)]];
+};
+
+struct main0_in
+{
+    float4 vColor [[user(locn0)]];
+    float2 vTex_0 [[user(locn1)]];
+    float2 vTex_1 [[user(locn2)]];
+    float2 vTex_2 [[user(locn3)]];
+    float2 vTex_3 [[user(locn4)]];
+};
+
+fragment main0_out main0(main0_in in [[stage_in]], constant uint* spvViewMask [[buffer(24)]], texture2d<float> uTex [[texture(0)]], sampler uTexSmplr [[sampler(0)]], uint gl_ViewIndex [[render_target_array_index]])
+{
+    main0_out out = {};
+    float2 vTex[4] = {};
+    vTex[0] = in.vTex_0;
+    vTex[1] = in.vTex_1;
+    vTex[2] = in.vTex_2;
+    vTex[3] = in.vTex_3;
+    gl_ViewIndex += spvViewMask[0];
+    out.FragColor = in.vColor * uTex.sample(uTexSmplr, vTex[int(gl_ViewIndex)]);
+    return out;
+}
+
--- a/reference/opt/shaders-msl/vulkan/vert/multiview.multiview.nocompat.vk.vert
+++ b/reference/opt/shaders-msl/vulkan/vert/multiview.multiview.nocompat.vk.vert
@@ -1,0 +1,31 @@
+#include <metal_stdlib>
+#include <simd/simd.h>
+
+using namespace metal;
+
+struct MVPs
+{
+    float4x4 MVP[2];
+};
+
+struct main0_out
+{
+    float4 gl_Position [[position]];
+    uint gl_Layer [[render_target_array_index]];
+};
+
+struct main0_in
+{
+    float4 Position [[attribute(0)]];
+};
+
+vertex main0_out main0(main0_in in [[stage_in]], constant uint* spvViewMask [[buffer(24)]], constant MVPs& _19 [[buffer(0)]], uint gl_InstanceIndex [[instance_id]])
+{
+    main0_out out = {};
+    uint gl_ViewIndex = spvViewMask[0] + gl_InstanceIndex % spvViewMask[1];
+    gl_InstanceIndex /= spvViewMask[1];
+    out.gl_Position = _19.MVP[int(gl_ViewIndex)] * in.Position;
+    out.gl_Layer = gl_ViewIndex - spvViewMask[0];
+    return out;
+}
+
--- a/reference/shaders-msl/vulkan/frag/basic.multiview.nocompat.vk.frag
+++ b/reference/shaders-msl/vulkan/frag/basic.multiview.nocompat.vk.frag
@@ -1,0 +1,32 @@
+#include <metal_stdlib>
+#include <simd/simd.h>
+
+using namespace metal;
+
+struct main0_out
+{
+    float4 FragColor [[color(0)]];
+};
+
+struct main0_in
+{
+    float4 vColor [[user(locn0)]];
+    float2 vTex_0 [[user(locn1)]];
+    float2 vTex_1 [[user(locn2)]];
+    float2 vTex_2 [[user(locn3)]];
+    float2 vTex_3 [[user(locn4)]];
+};
+
+fragment main0_out main0(main0_in in [[stage_in]], constant uint* spvViewMask [[buffer(24)]], texture2d<float> uTex [[texture(0)]], sampler uTexSmplr [[sampler(0)]], uint gl_ViewIndex [[render_target_array_index]])
+{
+    main0_out out = {};
+    float2 vTex[4] = {};
+    vTex[0] = in.vTex_0;
+    vTex[1] = in.vTex_1;
+    vTex[2] = in.vTex_2;
+    vTex[3] = in.vTex_3;
+    gl_ViewIndex += spvViewMask[0];
+    out.FragColor = in.vColor * uTex.sample(uTexSmplr, vTex[int(gl_ViewIndex)]);
+    return out;
+}
+
--- a/spirv_msl.cpp
+++ b/spirv_msl.cpp
@@ -105,8 +105,10 @@ void CompilerMSL::build_implicit_builtins()
 	    active_input_builtins.get(BuiltInSubgroupLtMask);
 	bool need_subgroup_ge_mask = !msl_options.is_ios() && (active_input_builtins.get(BuiltInSubgroupGeMask) ||
 	                                                       active_input_builtins.get(BuiltInSubgroupGtMask));
+	bool need_multiview = get_execution_model() == ExecutionModelVertex &&
+	                      (msl_options.multiview || active_input_builtins.get(BuiltInViewIndex));
 	if (need_subpass_input || need_sample_pos || need_subgroup_mask || need_vertex_params || need_tesc_params ||
-	    needs_subgroup_invocation_id)
+	    need_multiview || needs_subgroup_invocation_id)
 	{
 		bool has_frag_coord = false;
 		bool has_sample_id = false;
@@ -118,6 +120,7 @@ void CompilerMSL::build_implicit_builtins()
 		bool has_primitive_id = false;
 		bool has_subgroup_invocation_id = false;
 		bool has_subgroup_size = false;
+		bool has_view_idx = false;
 
 		ir.for_each_typed_id<SPIRVariable>([&](uint32_t, SPIRVariable &var) {
 			if (var.storage != StorageClassInput || !ir.meta[var.self].decoration.builtin)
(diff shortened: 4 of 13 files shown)


reference/opt/shaders-msl/vulkan/vert/multiview.multiview.nocompat.vk.vert
```diff
@@ -0,0 +1,32 @@
...
+
+struct main0_in
+{
+    float4 Position [[attribute(0)]];
+};
+
+vertex main0_out main0(main0_in in [[stage_in]], constant uint* spvViewMask [[buffer(24)]], constant MVPs& _19 [[buffer(0)]], uint gl_InstanceIndex [[instance_id]])
+{
+    main0_out out = {};
+    uint spvMaxViewIndex = 31 - clz(*spvViewMask);
```
HansKristian-Work: I think it makes more sense to pass an array of view indices here rather than clz which can have gaps. I expect something like:
```
uint spvMaxView = spvViewMask[0];
uint gl_ViewIndex = spvViewMask[InstanceIndex % spvMaxViewIndex + 1];
```

reference/shaders-msl/vulkan/frag/basic.multiview.nocompat.vk.frag
```diff
@@ -0,0 +1,25 @@
...
+struct main0_in
+{
+    float4 vColor [[user(locn0)]];
+    float2 vTex [[user(locn1)]];
+};
+
+fragment main0_out main0(main0_in in [[stage_in]], constant uint* spvViewMask [[buffer(24)]], texture2d<float> uTex [[texture(0)]], sampler uTexSmplr [[sampler(0)]], uint gl_ViewIndex [[render_target_array_index]])
+{
+    main0_out out = {};
+    if (!(*spvViewMask & (1 << gl_ViewIndex)))
```
HansKristian-Work: This is really bad, and we have to avoid this. This forces late-Z for everything, and I think this is completely avoidable if we pass down the view mask as proposed above.

reference/opt/shaders-msl/vulkan/vert/multiview.multiview.nocompat.vk.vert
```diff
@@ -0,0 +1,32 @@
...
+    uint gl_ViewIndex = gl_InstanceIndex % spvMaxViewIndex;
```
HansKristian-Work: There is an alternative here to render all instances for a layer, and this style renders all views in one go. Not sure which style is best from a caching POV, but optimizing for attribute read caching which this code does might be best here.

reference/opt/shaders-msl/vulkan/vert/multiview.multiview.nocompat.vk.vert
```diff
@@ -0,0 +1,32 @@
...
+    gl_InstanceIndex /= spvMaxViewIndex;
```
HansKristian-Work: How do we handle indirect here? I assume the indirect buffer must be modified by multiplying by number of views?

HansKristian-Work: Is it possible to unroll the multiview by just emitting N draw calls? I imagine the implementation for that would be far simpler than fixing this issue.

HansKristian-Work:
Tessellation and geometry shaders with multiview is optional, and I think there is no reason to support that in MSL. Multiview is a feature hailing from mobile VR, and that never supported tessellation with multiview either.

reference/shaders-msl/vulkan/frag/basic.multiview.nocompat.vk.frag
HansKristian-Work: The alternative is to set w = -1 in vertex shaders which should mask rasterization if it's not possible to do it like I proposed above.

HansKristian-Work:
There is an edge case with subpassInput where we would need to be able to index the right view's texture somehow. Wonder if that is solvable without unrolling per-view draw calls ... I wouldn't worry about this case yet though.

reference/opt/shaders-msl/vulkan/vert/multiview.multiview.nocompat.vk.vert
HansKristian-Work: The view mask is fixed to the render pass, so it should be possible to make this constant data rather than loading it from a buffer and having to manage that from the API side.

HansKristian-Work:
Does Metal support masking which layers are actually rendered to? There is some ambiguity what happens if you have loadOp = CLEAR and a subset of the layers are actually part of pViewMask. Traditional layered rendering might not be correct, and it might be the case that the only way to implement multiview on Metal is to unroll draw calls for N render passes ...

cdavis5e:
Sorry about the long delay. Things have been somewhat busy around here. Now that things have settled down a bit, I can come back and finish this.

I think this new version should address most of your concerns.

reference/opt/shaders-msl/vulkan/vert/multiview.multiview.nocompat.vk.vert
cdavis5e: Done.

reference/shaders-msl/vulkan/frag/basic.multiview.nocompat.vk.frag
cdavis5e: I've decided to split the subpass into multiple passes in case there's a gap in the mask.

spirv_msl.cpp
```diff
@@ -6489,6 +6549,31 @@ void CompilerMSL::fix_up_shader_inputs_outputs()
...
+						          msl_options.multiview_base_view, " + ", to_expression(builtin_instance_idx_id), " % ",
+						          msl_options.multiview_view_count, ";");
+						statement(to_expression(builtin_instance_idx_id), " /= ", msl_options.multiview_view_count,
+						          ";");
+					});
+					// In addition to setting the variable itself, we also need to
+					// set the render_target_array_index with it on output.
+					entry_func.fixup_hooks_out.push_back(
+					    [=]() { statement(to_expression(builtin_layer_id), " = ", to_expression(var_id), ";"); });
+				}
```
HansKristian-Work: Missing a break; here.

spirv_msl.cpp
```diff
@@ -6489,6 +6549,31 @@ void CompilerMSL::fix_up_shader_inputs_outputs()
...
+						statement("const ", builtin_type_decl(bi_type), " ", to_expression(var_id), " = 0;");
+					});
+				}
+				else if (get_execution_model() == ExecutionModelVertex)
+				{
+					// Metal provides no special support for multiview, so we smuggle
+					// the view index in the instance index.
+					entry_func.fixup_hooks_in.push_back([=]() {
+						statement(builtin_type_decl(bi_type), " ", to_expression(var_id), " = ",
+						          msl_options.multiview_base_view, " + ", to_expression(builtin_instance_idx_id), " % ",
```
HansKristian-Work: So, I guess the idea here is that you want to detect all contiguous ranges in the view mask, and render all those in one go?

As layered rendering is used here, how will this work with loadOp = CLEAR? Can you restrict the render pass to just base layer + count? Is the [[render_target_index]] shifted to accomodate this?

HansKristian-Work: There's also the issue of indirect draws ...

cdavis5e:
It just occurred to me that, if we pass the view range as compile-time constants, but there's a gap in the view mask, then we'll have to compile the shaders (and thus, the pipeline) multiple times. I don't know if that's worth the gains from hardcoding the view range at compile time.

HansKristian-Work:
I have a feeling that the only reasonable implementation is to unroll the multiview draw calls in multiple passes and then pass down gl_ViewIndex as a constant uint and don't bother with layered rendering at all.

spirv_msl.cpp
cdavis5e: > So, I guess the idea here is that you want to detect all contiguous ranges in the view mask, and render all those in one go?

Pretty much.

> As layered rendering is used here, how will this work with loadOp = CLEAR? Can you restrict the render pass to just base layer + count? Is the [[render_target_index]] shifted to accomodate this?

I can set the `renderTargetArrayLength` property in the `MTLRenderPassDescriptor` to set the number of views to be rendered, and the `slice` property of the `MTLRenderPassAttachmentDescriptor` to set the base view. This is the same as regular layered drawing.

> Does this mean we need different shaders to deal with different parts of the render pass where the base layer needs to change?

I'm considering reverting the change to use hard-coded constants.

> There's also the issue of indirect draws ...

We already have to munge the indirect buffer for tessellated draws. I can probably do that for multiview, too.

HansKristian-Work: Ok, I guess if you're munging the indirect buffers, this will work.

HansKristian-Work:
Think the only missing piece now would be to make base_view not a compile time constant, but sourced from a buffer?

spirv_msl.hpp
```diff
@@ -167,6 +167,11 @@ static const uint32_t kBufferSizeBufferBinding = ~(2u);
 // will start at max(kArgumentBufferBinding) + 1.
 static const uint32_t kArgumentBufferBinding = ~(3u);
 
+// Special constant used in a MSLResourceBinding binding
+// element to indicate the buffer binding for the current subpass view mask, to
+// support SPV_KHR_multiview.
+static const uint32_t kViewMaskBufferBinding = ~(4u);
```
HansKristian-Work: This is not relevant, since it's part of msl_options.

spirv_msl.cpp
```diff
@@ -428,6 +480,16 @@ void CompilerMSL::build_implicit_builtins()
...
 	}
+
+	if (needs_view_mask_buffer())
+	{
+		uint32_t var_id = build_constant_uint_array_pointer();
+		set_name(var_id, "spvViewMask");
+		// This should never match anything.
+		set_decoration(var_id, DecorationDescriptorSet, kViewMaskBufferBinding);
+		set_decoration(var_id, DecorationBinding, msl_options.view_mask_buffer_index);
+		view_mask_buffer_id = var_id;
```
HansKristian-Work: This needs to have set_extended_decoration() with SPIRVCrossResourceIndexPrimary now as well. I had to fix the automatic resource binding assignment scheme, and it now assigns resources automatically since fallbacks to DecorationBinding are not correct in all cases.

reference/opt/shaders-msl/vulkan/frag/basic.multiview.nocompat.vk.frag
```diff
@@ -0,0 +1,32 @@
...
+struct main0_in
+{
+    float4 vColor [[user(locn0)]];
+    float2 vTex_0 [[user(locn1)]];
+    float2 vTex_1 [[user(locn2)]];
+    float2 vTex_2 [[user(locn3)]];
+    float2 vTex_3 [[user(locn4)]];
+};
+
+fragment main0_out main0(main0_in in [[stage_in]], constant uint* spvViewMask [[buffer(0)]], texture2d<float> uTex [[texture(0)]], sampler uTexSmplr [[sampler(0)]], uint gl_ViewIndex [[render_target_array_index]])
```
HansKristian-Work: This automatic binding for spvViewMask looks wrong.

reference/opt/shaders-msl/vulkan/vert/multiview.multiview.nocompat.vk.vert
```diff
@@ -0,0 +1,32 @@
...
+
+struct main0_in
+{
+    float4 Position [[attribute(0)]];
+};
+
+vertex main0_out main0(main0_in in [[stage_in]], constant uint* spvViewMask [[buffer(1)]], constant MVPs& _19 [[buffer(0)]], uint gl_InstanceIndex [[instance_id]])
+{
+    main0_out out = {};
+    uint spvMaxViewIndex = spvViewMask[1];
```
HansKristian-Work: No need to emit a temporary here.

HansKristian-Work:
Probably a good idea to rebase as well since there is a merge commit in this PR.

HansKristian-Work:
Thanks!

PR #1005, opened 2019-06-05
Add some missing features for NV_ray_tracing

### LINKED ISSUE #999: GLSL: NV_ray_tracing support seems forgot shaderRecordNV.
Hi guys-
We already have GLSL_NV_ray_tracing support but it looks that we forgot shaderRecordNV. It turns out to be nothing but enums if I search "ShaderRecordBufferNV".

pmoursnv:
This should fix #999 and add all the remaining features from GLSL_NV_ray_tracing, with the exception of ray flags (which are just printed as integer constants, rather than as a combination of named flags).

### DIFF
--- a/reference/shaders/vulkan/rgen/shader_record_buffer.nocompat.vk.rgen.vk
+++ b/reference/shaders/vulkan/rgen/shader_record_buffer.nocompat.vk.rgen.vk
@@ -1,7 +1,7 @@
 #version 460
 #extension GL_NV_ray_tracing : require
 
-layout(shaderRecordNV) buffer sbt
+layout(shaderRecordNV, std430) buffer sbt
 {
     vec3 direction;
     float tmax;
--- a/reference/shaders/vulkan/rgen/shader_record_buffer.nocompat.vk.rgen.vk
+++ b/reference/shaders/vulkan/rgen/shader_record_buffer.nocompat.vk.rgen.vk
@@ -1,0 +1,17 @@
+#version 460
+#extension GL_NV_ray_tracing : require
+
+layout(shaderRecordNV) buffer sbt
+{
+    vec3 direction;
+    float tmax;
+} _20;
+
+layout(set = 0, binding = 0) uniform accelerationStructureNV as;
+layout(location = 0) rayPayloadNV float payload;
+
+void main()
+{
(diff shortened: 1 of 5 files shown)


reference/shaders/vulkan/rgen/shader_record_buffer.nocompat.vk.rgen.vk
```diff
@@ -0,0 +1,17 @@
+#version 460
+#extension GL_NV_ray_tracing : require
+
+layout(shaderRecordNV) buffer sbt
```
HansKristian-Work: What about packing qualifiers like std140 and std430, do they apply here?

spirv_glsl.cpp
```diff
@@ -1862,6 +1864,14 @@ const char *CompilerGLSL::to_storage_qualifiers_glsl(const SPIRVariable &var)
 	{
 		return "hitAttributeNV ";
 	}
+	else if (var.storage == StorageClassCallableDataNV)
+	{
+		return "callableDataNV";
```
HansKristian-Work: Missing a space at the end? I don't see any test case for this.

spirv_glsl.cpp
```diff
@@ -1862,6 +1864,14 @@ const char *CompilerGLSL::to_storage_qualifiers_glsl(const SPIRVariable &var)
...
+	}
+	else if (var.storage == StorageClassIncomingCallableDataNV)
+	{
+		return "callableDataInNV";
```
HansKristian-Work: Ditto.

reference/shaders/vulkan/rgen/shader_record_buffer.nocompat.vk.rgen.vk
pmoursnv: `shaderRecordNV` cannot be used together with any other layout qualifiers (https://github.com/KhronosGroup/GLSL/blob/master/extensions/nv/GLSL_NV_ray_tracing.txt#L465).

HansKristian-Work: I see, so it's effectively "logical" storage?

pmoursnv: You are right, my mistake. It does indeed accept packing qualifiers.

PR #1022, opened 2019-06-12
MSL: Support Invariant qualifier on position.

### LINKED ISSUE #1017: [MSL] Support the invariant qualifier on position in MSL 2.1
MSL 2.1 finally adds an `invariant` attribute, but it only applies to the position, not to other vertex outputs. Regardless, it should at least be possible to pass any `Invariant` decoration on the output `Position` through to Metal.

HansKristian-Work:
Fix #1017.


mabaro:
related post-> https://github.com/KhronosGroup/SPIRV-Cross/issues/2420

PR #1031, opened 2019-06-18
MSL: Support 64-bit integers.

HansKristian-Work:
MSL compiler is complaining about that you cannot place long/ulong in buffers, which does seem like a bug.

### DIFF
--- a/reference/opt/shaders-msl/comp/int64.invalid.msl22.comp
+++ b/reference/opt/shaders-msl/comp/int64.invalid.msl22.comp
@@ -3,51 +3,22 @@
 
 using namespace metal;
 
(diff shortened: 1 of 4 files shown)


cdavis5e:
I remember reading somewhere that this was intentional.

/me looks

Ah, here it is. From the MSL 2.2 spec, section 2.7:

> Note: Metal does *not* support buffers that contain `long` or `ulong` data types.

HansKristian-Work:
What?! Welp, no Int64 capability for MoltenVK I guess.

HansKristian-Work:
FP64 is not supported at all, no.

PR #1035, opened 2019-06-20
Use the member index, not the resource ID, as the Metal argument buffer index.

pcwalton:
In SPIR-V, IDs of different resources can overlap, but they can't for Metal
argument buffers.

The existing code was causing compile errors in my shaders because samplers, textures, and buffers were being assigned duplicate IDs, and this change fixed it. I'm not sure this is the right fix, though.

### DIFF
--- a/spirv_msl.cpp
+++ b/spirv_msl.cpp
@@ -8799,8 +8799,11 @@ void CompilerMSL::analyze_argument_buffers()
 				}
 			}
 
+			// We can't in general use the resource ID as the argument buffer index because they might conflict. For
+			// example, a texture and a sampler might have the same resource ID, but they must have different argument
+			// buffer IDs.
 			set_extended_member_decoration(buffer_type.self, member_index, SPIRVCrossDecorationArgumentBufferID,
-			                               resource.index);
+			                               member_index);
 			set_extended_member_decoration(buffer_type.self, member_index, SPIRVCrossDecorationInterfaceOrigID,
 			                               var.self);
 			member_index++;


HansKristian-Work:
This is the wrong approach I think. You are supposed to use the add_msl_resource_binding() method to assign your own IDs/MSL bindings from set/binding. If you let SPIRV-Cross do it, it's going to be very unpredictable at any rate how ids get assigned.

cdavis5e:
I'm kinda worried that we won't be able to support this properly anyway. MoltenVK has to work with descriptor set and binding numbers; if we have multiple SPIR-V resources binding to the same descriptor, but we can't assign them the same `id` in a Metal argument buffer, how do we know which one to fill out when it comes time to encode the argument buffer? I suspect that if bound resource and its shader declaration don't match up exactly, then Metal's going to throw a fit and refuse to accept it.

HansKristian-Work:
In the descriptor set layout, you can translate binding to ID, and when it comes to record an argument buffer update, use that. I think the translation should be the exact same as for the non-argument buffer path.

HansKristian-Work:
@pcwalton You have a concrete shader example?

pcwalton:
I can't use explicit bindings in my GLSL because the GLSL compilers I'm using won't accept them. So I auto-assign bindings. It doesn't matter that the IDs aren't stable, because I reflect on the member names in the argument buffer struct to determine which ID to use.

HansKristian-Work:
Ok, so this is not related to aliasing multiple resources on the same set and binding, but you basically need a robust auto-ID assigner that works as expected in all cases?

pcwalton:
It is related to aliasing multiple resources, because I can automatically assign IDs in the SPIR-V compiler, but those automatically-assigned IDs don't work with argument buffers due to aliasing. (It works without argument buffers. Argument buffers seem to have different rules than regular Metal arguments. But I can't use the non-argument-buffer path because that doesn't handle some uniforms properly either, and thus fails to compile.)

Here is a problematic example:

    #version 330

    precision highp float;

    uniform sampler2D uTexture;
    uniform vec4 uColor;

    in vec2 vTexCoord;

    out vec4 oFragColor;

    void main() {
        float alpha = texture(uTexture, vTexCoord).r * uColor.a;
        oFragColor = alpha * vec4(uColor.rgb, 1.0);
    }

First I compile to SPIR-V like this, automatically assigning locations:

    glslangValidator --auto-map-locations -G330 -S frag -o debug_texture.fs.spv debug_texture.fs.glsl

Then I compile the SPIR-V to Metal:

    spirv-cross --msl --msl-version 020100 --msl-argument-buffers debug_texture.fs.spv

That produces the following Metal code:


    #include <metal_stdlib>
    #include <simd/simd.h>

    using namespace metal;

    struct spvDescriptorSetBuffer0
    {
        constant float4* uColor [[id(0)]];
        texture2d<float> uTexture [[id(0)]];
        sampler uTextureSmplr [[id(1)]];
    };

    struct main0_out
    {
        float4 oFragColor [[color(0)]];
    };

    struct main0_in
    {
        float2 vTexCoord [[user(locn0)]];
    };

    fragment main0_out main0(main0_in in [[stage_in]], constant spvDescriptorSetBuffer0& spvDescriptorSet0 [[buffer(0)]])
    {
        main0_out out = {};
        float alpha = spvDescriptorSet0.uTexture.sample(spvDescriptorSet0.uTextureSmplr, in.vTexCoord).x * (*spvDescriptorSet0.uColor).w;
        out.oFragColor = float4((*spvDescriptorSet0.uColor).xyz, 1.0) * alpha;
        return out;
    }

However, the Metal compiler refuses to compile this because `id(0)` is used more than once in the argument buffer.

pcwalton:
I don't need a consistent translation from SPIR-V descriptor set ID to argument buffer ID, because I can use Metal reflection to match on names instead. I understand many users don't want to use reflection, but there's a mismatch here between what SPIR-V allows and what Metal allows :(

HansKristian-Work:
Right, ok. Seems like you're using the GL semantics model here and not Vulkan where bindings are per-resource type, unlike Vulkan. Currently, the fallback in SPIRV-Cross is to use bindings to assign ids, but clearly it cannot work in all cases. I think the fix will be for me to do a proper pass over the resource ID generation and not rely blindly on bindings and shaky fallbacks. add_msl_resource will always trump auto-assigned bindings though, but the implementation should be friendlier to these use cases as well.

This PR would need to be quite more sophisticated I think (need to think what actually needs to change), but it's probably better for me to do it. I'll try to have a look at it tomorrow.

HansKristian-Work:
A fix for this is merged to master. I added two new reflections methods to get the automatically chosen resource ID/bindings, but I guess the direct MSL reflection also works, up to you.

pcwalton:
Your fix works great, thank you!

PR #1046, opened 2019-06-27
MSL: Fix sampling with FP16 coordinates.

### LINKED ISSUE #1044: HLSL does not support case fallthrough
Noticed failure in FXC, glslang HLSL allowed it.

HansKristian-Work:
MSL fails to resolve overloads if you don't pass in FP32 coordinates exactly.
Also fixes an unrelated FXC failure with HLSL.

Fix #1044.


HansKristian-Work:
Verified locally, merging early.

PR #1060, opened 2019-07-04
Parse SPIR-V debug information extended instructions, as well as OpNoLine

lifpan:
No impact on result shader string.


lifpan:
Here is an example shader:
[spv_debug_info_simple.spv.txt](https://github.com/KhronosGroup/SPIRV-Cross/files/3360934/spv_debug_info_simple.spv.txt)
[spv_debug_info_simple.zip](https://github.com/KhronosGroup/SPIRV-Cross/files/3360936/spv_debug_info_simple.zip)

PR #1067, opened 2019-07-10
MSL: Support scalar block layout.

cdavis5e:
Relaxed block layout relaxed the restrictions on vector alignment,
allowing them to be aligned on scalar boundaries. Scalar block layout
relaxes this further, allowing *any* member to be aligned on a scalar
boundary. The requirement that a vector not improperly straddle a
16-byte boundary is also relaxed.

I've also added a test showing that `std430` layout works with UBOs.

I'm troubled by the dual meaning of the `Packed` extended decoration. In
some instances (struct, `float[]`, and `vec2[]` members), it actually
means the exact opposite, that the member needs extra padding. This is
especially problematic for `vec2[]`, because now we need to distinguish
the two cases by checking the array stride. I wonder if this should
actually be split into two decorations.

I also don't like that there are now four places where we set the
decoration on struct members.

### DIFF
--- a/spirv_msl.cpp
+++ b/spirv_msl.cpp
@@ -2420,7 +2420,7 @@ void CompilerMSL::align_struct(SPIRType &ib_type)
 		{
 			set_extended_member_decoration(ib_type_id, mbr_idx, SPIRVCrossDecorationPacked);
 			set_extended_member_decoration(ib_type_id, mbr_idx, SPIRVCrossDecorationPackedType,
-			                               ib_type.member_types[mbr_idx]);
+			                               get_member_packed_type(ib_type, mbr_idx));
 		}
 
 		// Align current offset to the current member's default alignment.
@@ -2480,7 +2480,7 @@ bool CompilerMSL::is_member_packable(SPIRType &ib_type, uint32_t index, uint32_t
 			{
 				set_extended_member_decoration(mbr_type.self, i, SPIRVCrossDecorationPacked);
 				set_extended_member_decoration(mbr_type.self, i, SPIRVCrossDecorationPackedType,
-				                               mbr_type.member_types[i]);
+				                               get_member_packed_type(mbr_type, i));
 			}
 		}
 		size_t declared_struct_size = get_declared_struct_size(mbr_type);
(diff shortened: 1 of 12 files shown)


HansKristian-Work:
Right, now that we have the extended decoration system in place, it does seem a bit awkward to overload the Packed decoration in this way. We should refactor this and distinguish between "packed" and "padded" at the very least. If you have some ideas feel free to look into it, otherwise I'll look at it eventually.

HansKristian-Work:
get_declared_struct_member_alignment() does not seem to be correct. It will return 16 for a Packed array of vec3, because it hits the special case for array of float[] and float2[].

spirv_msl.cpp
```diff
@@ -2490,9 +2540,10 @@ bool CompilerMSL::is_member_packable(SPIRType &ib_type, uint32_t index)
 	if (mbr_type.vecsize == 1 || (is_matrix(mbr_type) && mbr_type.vecsize != 3))
 		return false;
 
-	// Only row-major matrices need to be packed.
-	if (is_matrix(mbr_type) && !has_member_decoration(ib_type.self, index, DecorationRowMajor))
-		return false;
+	// Pack if the member's offset doesn't conform to the type's usual
+	// alignment. For example, a float3 at offset 4.
+	if (mbr_offset_curr % get_declared_struct_member_alignment(ib_type, index))
```
HansKristian-Work: I do wonder if this needs to come before the check above for matrix type and vecsize != 3. I think it's still possible we hit the case where a mat2 has an offset of 4. Try this declaration:

```
layout(scalar) buffer {
    float a;
    mat2 b; <-- is declared as float2x2, not packed_float2x2
};
```

cdavis5e: Well, OK, but I had to completely redo support for packed matrices.

cdavis5e:
> get_declared_struct_member_alignment() does not seem to be correct. It will return 16 for a Packed array of vec3, because it hits the special case for array of float[] and float2[], and there is no array stride check like with the other paths.

Fixed.

HansKristian-Work:
Thanks! I do like the idea to use array of vectors instead to deal with matrices. That should also allow us to support mat2 in std140 where we need to pad out the matrix and similar shenanigans. I've made a maintenance issue https://github.com/KhronosGroup/SPIRV-Cross/issues/1074 to rethink how we deal with packed types going forward. I hope to get around to it eventually.

PR #1075, opened 2019-07-11
Remove unreasonable assertion for OpTypeImage Sampled parameter.

lifpan:
**OpTypeImage** Sampled parameter must be one of the following values:
0 indicates this is only known at run time, not at compile time
1 indicates will be used with sampler
2 indicates will be used without a sampler (a storage image)


HansKristian-Work:
For parsing this is fair enough, but I don't think it's possible to do codegen this way though.

PR #1078, opened 2019-07-11
Support the SPV_KHR_post_depth_coverage extension.

cdavis5e:
Using the `PostDepthCoverage` mode specifies that the `gl_SampleMaskIn`
variable is to contain the computed coverage mask following the early
fragment tests, which this mode requires and implicitly enables.

Note that unlike Vulkan and OpenGL, Metal places this on the sample mask
input itself, and furthermore does *not* implicitly enable early
fragment testing. If it isn't enabled explicitly with an
`[[early_fragment_tests]]` attribute, the compiler will error out. So we
have to enable that mode explicitly if `PostDepthCoverage` is enabled
but `EarlyFragmentTests` isn't.

For Metal, only iOS supports this; for some reason, Apple has yet to
implement it on macOS, even though many desktop cards support it.

### DIFF
--- a/reference/opt/shaders-msl/frag/post-depth-coverage.ios.msl2.frag
+++ b/reference/opt/shaders-msl/frag/post-depth-coverage.ios.msl2.frag
@@ -1,0 +1,17 @@
+#include <metal_stdlib>
+#include <simd/simd.h>
+
+using namespace metal;
+
+struct main0_out
+{
+    float4 FragColor [[color(0)]];
+};
+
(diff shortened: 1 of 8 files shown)


Degerz:
> Apple has yet to implement it on macOS, even though many desktop cards support it.

It's because AMD GPUs [don't](https://github.com/GPUOpen-Drivers/llpc/commit/dcf44d7a060d5e130bb82518c40ced914c7a4bb1) support post-depth coverage. Not even on the latest gfx10/Navi GPU ... 

Even if it is supported on Intel GPUs, Apple don't care about them ...

HansKristian-Work:
Thanks for implementing GLSL side as well!

PR #1083, opened 2019-07-13
Update external repos.

cdavis5e:
Fix fallout from changes.

There's a bug in glslang that prevents `float16_t`, `[u]int16_t`, and
`[u]int8_t` constants from adding the corresponding SPIR-V capabilities.
SPIRV-Tools, meanwhile, tightened validation so that these constants are
only valid if the corresponding `Float16`, `Int16`, and `Int8` caps are
on. This affects the `16bit-constants.frag` test for GLSL and MSL.

I've also updated the local copy of the SPIR-V headers.


HansKristian-Work:
Nice, seems like the nonuniform output works now on latest SPIRV-Tools/glslang again, so updating repos is fine now.

PR #1087, opened 2019-07-17
Support the SPV_EXT_demote_to_helper_invocation extension.

### LINKED ISSUE #1052: Support SPV_EXT_demote_to_helper_invocation
Not sure if it's implementable in MSL, but should be implementable in at least GLSL (pending glslang update) and HLSL (discard implies demotion, not exit).

cdavis5e:
This extension provides a new operation which causes a fragment to be
discarded without terminating the fragment shader invocation. The
invocation for the discarded fragment becomes a helper invocation, so
that derivatives will remain defined. The old `HelperInvocation` builtin
becomes undefined when this occurs, so a second new instruction queries
the current helper invocation status.

This is only fully supported for GLSL. HLSL doesn't support the
`IsHelperInvocation` operation and MSL doesn't support the
`DemoteToHelperInvocation` op.

Fixes #1052.

### DIFF
--- a/spirv_glsl.cpp
+++ b/spirv_glsl.cpp
@@ -9789,7 +9789,7 @@ void CompilerGLSL::emit_instruction(const Instruction &instruction)
 		if (!options.vulkan_semantics)
 			SPIRV_CROSS_THROW("GL_EXT_demote_to_helper_invocation is only supported in Vulkan GLSL.");
 		require_extension_internal("GL_EXT_demote_to_helper_invocation");
-		emit_op(ops[0], ops[1], "helperInvocationEXT()", true);
+		emit_op(ops[0], ops[1], "helperInvocationEXT()", false);
 		break;
 
 	default:
--- a/spirv_glsl.cpp
+++ b/spirv_glsl.cpp
@@ -9778,6 +9778,20 @@ void CompilerGLSL::emit_instruction(const Instruction &instruction)
 	case OpNoLine:
 		break;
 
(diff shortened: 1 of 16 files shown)


spirv_glsl.cpp
```diff
@@ -9778,6 +9778,20 @@ void CompilerGLSL::emit_instruction(const Instruction &instruction)
...
+			SPIRV_CROSS_THROW("GL_EXT_demote_to_helper_invocation is only supported in Vulkan GLSL.");
+		require_extension_internal("GL_EXT_demote_to_helper_invocation");
+		statement(backend.demote_literal, ";");
+		break;
+
+	case OpIsHelperInvocationEXT:
+		if (!options.vulkan_semantics)
+			SPIRV_CROSS_THROW("GL_EXT_demote_to_helper_invocation is only supported in Vulkan GLSL.");
+		require_extension_internal("GL_EXT_demote_to_helper_invocation");
+		emit_op(ops[0], ops[1], "helperInvocationEXT()", true);
```
HansKristian-Work: This expression cannot be forwarded. If this is forwarded, and we then call Demote(), and then consume the forwarded expression, we'll query IsHelperInvocation after the fact. Alternatively, DemoteToHelperInvocation() must invalidate all IsHelperInvocationEXT expressions, but that's a bit overkill.

HansKristian-Work: There should be a test which exposes this.

PR #1089, opened 2019-07-18
MSL: Refactor buffer packing logic from ground up.

### LINKED ISSUE #1074: Refactor MSL buffer packing implementation
The code to handle buffer packing in the MSL backend is reaching a critical mass of complexity. We need to simplify as much as we can.

### LINKED ISSUE #1088: MSL: why is a std140 mat3 compiled to packed_float3x3?
Given the following GLSL:

```
#version 410

layout(std140) uniform ObjectUniforms
{
    mat3 worldFromModelNormalMatrix;
} objectUniforms;

layout(location = 6) out vec3 vertex_worldTangent;

void main()
{
    vertex_worldTangent = objectUniforms.worldFromModelNormalMatrix * vec3(1.0, 0.0, 0.0);
}
```

`SPIRV-cross` on the latest [master](fb71a337a2899eed2d1dbc2f5cd21cb35aba67b0) now emits a `packed_float3x3` for `worldFromModelNormalMatrix` along with a `typedef`:

```
typedef packed_float3 packed_float3x3[3];
struct ObjectUniforms
{
    packed_float3x3 worldFromModelNormalMatrix;
};
```

**Isn't the correct type here a MSL`float3x3`?** Previously `SPIRV-cross` output `float3x3`, which has the same packing as GLSL's `mat3` under `std140`. [This commit](https://github.com/KhronosGroup/SPIRV-Cross/commit/28454facbb3d1b3ecdb29749d8d87ba7b31c3e2c) seems suspect.

Command-line:
```
$ glslangValidator -G -o test.spv test.frag
$ spirv-cross --msl test.spv
```

HansKristian-Work:
This PR rewrites how buffer packing works in MSL. This change is rather massive, but it fixes a lot of bugs, and adds support for a lot more use cases. We also have exhaustive buffer packing tests for scalar, std140 and std430 layouts. I am now far, far more confident in the buffer layout in MSL.

First, we add two type modifying decorations.
- PhysicalTypeID: Overrides the actual type of the member. E.g. a float[] can be turned into a float4[], or float2x2 can be turned into a float2x4. This exists to support std140 layouts.
- Packed: This declares that a type is to be packed_T. This is orthogonal with PhysicalTypeID. Now we can cleanly represent all sorts of packing/padding scenarios.

Struct padding at the end of the struct was not supported previously, but we can now support padding structs for e.g. array of std140 structs. Rather than using the crufty StructMemberKey hashmap, we just make this a dedicated decoration instead now that we have the extended decoration system in place.

Structs can also be placed at "scalar" layout offsets. To sort this out, we need to analyze struct types recursively before we enter align_struct. Problematic structs are marked as Packed, i.e. all members are packed. Structs in MSL inherit their alignment based on their members alone, so in order to reduce struct alignment, we have to make everything packed.

We add MSL size queries of all types. This is critical since we did not have this before, and we had no real way to be sure that the MSL we generated conformed to our packing rules. These checks take PhysicalID remapping, packed and row-major state into account.

In order to align a struct, we go through a very straight forward algorithm which removes most of the nastiest cruft we accumulated in the earlier implementation.

For each member, we look at the SPIRV Offset, the effective MSL size and the next member.
If the MSL variable is too large such that it spills over next offset, e.g. float3 + float, we fail validation.
Alignment rules, array stride and matrix stride is also checked.

If any rule fails validation, we have no recourse but to try packed_ types. We check validation again. If that also fails, we are kind of screwed, but we have two possible fallbacks to try. One is to check ArrayStride of vectors and create padded vectors, and similar with MatrixStride for matrices. We check validation a last time. If that fails, we cannot do anything, so we fail gracefully, but this should work in all reasonable cases.

We also now support a crazy amount of edge cases with matrices. Matrices can be row-major, packed, padded and all sorts of complication. I had to add a lot of complexity to unpack_expression and emit_store_statement since we need to handle all sorts of complicated cases where we're either packed or we have remapped the physical type. We attempt to emit the optimal code for these cases.

I also reworked how row-major works with access chains. If we try to access a column inside a matrix by means of indexing, we previously had to transpose() the matrix before accessing it. Now however, we support deferring this transpose, because otherwise we cannot store to a row-major column in MSL. If we access a scalar from this access chain, we just flip the array indices, if we Load or Store to this access chain, we can unroll the load or store. This is similar to how UAV row-major stores work in HLSL already. All of this is exhaustively tested now for all matrix sizes, packing layouts and col-major/row-major ...

The transposition code is greatly simplified. Now legacy GLSL and MSL goes through the same logic. I also added support for multiplying two transposed matrices together in an optimal way.

These fixes also exposed a lot of other minor bugs which I had to work through.

There is a lot of commits in this PR, but I will not squash them because it provides a larger context of why certain things were done in a certain way.

Fix #1074.
Fix #1088.


HansKristian-Work:
@cdavis5e Having quite steady progress, if you haven't gotten around to fixing mat3 case yet, it might not be worth the effort as it would likely be replaced very soon.

HansKristian-Work:
Exhaustively tested load/store/column/element access on all types in std140, std430, scalar, phew!

HansKristian-Work:
I think this is good to go, I've tested a lot of cases, but need to go over all the changes and document a bit more.

Tested it on Granite and DOTA in MoltenVK, and runs as expected.

billhollings:
@HansKristian-Work 

Thanks for all the effort on this!

> Tested it on Granite and DOTA in MoltenVK, and runs as expected.

What is _Granite_?

HansKristian-Work:
My personal engine.

cdavis5e:
Thanks for cleaning this up. I know I certainly didn't help things :). And thanks for putting in the extra effort to get this merged so we can ship the SDK with `VK_EXT_scalar_block_layout`.

HansKristian-Work:
A regression and the rewrite was inevitable, don't worry about it. :)

PR #1097, opened 2019-07-24
MSL: Adjust BuiltInWorkgroupId for vkCmdDispatchBase().

cdavis5e:
This command allows the caller to set the base value of
`BuiltInWorkgroupId`, and thus of `BuiltInGlobalInvocationId`. Metal
provides no direct support for this... but it does provide a builtin,
`[[grid_origin]]`, normally used to pass the base values for the stage
input region, which we will now abuse to pass the dispatch base and
avoid burning a buffer binding.

`[[grid_origin]]`, as part of Metal's support for compute stage input,
requires MSL 1.2. For 1.0 and 1.1, we're forced to provide a buffer.

(Curiously, this builtin was undocumented until the MSL 2.2 release. Go
figure.)

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -516,6 +516,7 @@ struct CLIArguments
 	bool msl_texture_buffer_native = false;
 	bool msl_multiview = false;
 	bool msl_view_index_from_device_index = false;
+	bool msl_dispatch_base = false;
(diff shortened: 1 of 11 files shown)


HansKristian-Work:
I'll need to work around MSVC 2013 issues before I can merge. Should be a trivial one-liner.

PR #1105, opened 2019-07-26
MSL: Unify the get_*_address_space() methods.

cdavis5e:
These methods have largely the same logic, with minor differences. That
I felt compelled to duplicate the logic into another method was one of
the things that bothered me about the variable pointers change. This
cleans that part of the code up; now we don't have two places to change.


HansKristian-Work:
Thanks for the cleanup!

Issue #1108, opened 2019-07-30
Performance regression with spirv-cross compilation to GLSL


dnovillo:
We are experiencing compile-time slowdowns of about 75x for the attached SPIRV input file
[shader_test.spirv.zip](https://github.com/KhronosGroup/SPIRV-Cross/files/3447657/shader_test.spirv.zip)

I've bisected the changes and I think the slowdown was caused by commit https://github.com/KhronosGroup/SPIRV-Cross/commit/18bcc9b79077fe179318f607d348ff35afd24f6a

To reproduce, compile and run this code:

```c++
#include <sys/time.h>

#include <cinttypes>
#include <cstdlib>
#include <memory>
#include <spirv_glsl.hpp>
#include <sstream>

int main(int argc, char **argv) {
  std::string input;

  const int argc_check = argc - 1;
  for (int i = 1; i < argc; ++i) {
    if (!strcmp("-i", argv[i]) && i < argc_check) {
      input = argv[++i];
    }
  }
  if (input.empty()) {
    printf("\nNo input shader has been specified.\n");
    return -1;
  }

  FILE *raw_file_ptr = fopen(input.c_str(), "rb");
  if (raw_file_ptr == nullptr) {
    printf("\nFailed to read shader blob.\n");
    return -1;
  }
  fseek(raw_file_ptr, SEEK_SET, SEEK_END);

  const size_t file_size = static_cast<size_t>(ftell(raw_file_ptr));
  rewind(raw_file_ptr);

  std::vector<uint32_t> spirv_blob;
  spirv_blob.resize(file_size / 4);
  fread(spirv_blob.data(), 1, file_size, raw_file_ptr);

  timeval tv_start;
  gettimeofday(&tv_start, NULL);

  spirv_cross::CompilerGLSL glslCompiler(spirv_blob);
  glslCompiler.compile();

  timeval tv_end;
  gettimeofday(&tv_end, NULL);

  int64_t milliseconds = (tv_end.tv_sec - tv_start.tv_sec) * 1000;
  milliseconds += (tv_end.tv_usec - tv_start.tv_usec) / 1000;
  printf("Shader compiled in %d ms\n", (int)milliseconds);

  return 0;
}
```

Before that commit, the output is:

```
$ ./compile -i shader_test.spirv
Shader compiled in 38 ms
```

After the commit, the output is:

```
$ ./compile -i shader_test.spirv
Shader compiled in 2923 ms
```

[...]

PR #1109, opened 2019-07-31
Revert "Do not disable temporary forwarding when we suppress usage tracking"

dnovillo:
This reverts commit 18bcc9b79077fe179318f607d348ff35afd24f6a.

This is a temporary workaround to address the performance slowdown
reported in https://github.com/KhronosGroup/SPIRV-Cross/issues/1108.

Tested with `./test_shaders.sh` and confirmed that compile times are back to 40ms for the test case in #1108.


HansKristian-Work:
Try reverting just this branch: https://github.com/KhronosGroup/SPIRV-Cross/pull/1109/files#diff-09b7cdf207013ada0b4e38970bc27158L7553. That might be it.

dnovillo:
> Try reverting just this branch: https://github.com/KhronosGroup/SPIRV-Cross/pull/1109/files#diff-09b7cdf207013ada0b4e38970bc27158L7553. That might be it, or, hmm. Maybe not.

No, that was the first branch I reverted when initially tracking this down.  It was not it.  I believe that the new `suppressed_usage_tracking` set caused the slowdown.  The hashing functions in these sets is now overwhelming.

I have not noticed any regressions after this commit.  I'll try applying it locally because I don't really know this codebase and a principled fix would take more cycles than I have available.

HansKristian-Work:
Fair enough, it does seem like there is a problem where invalidate ID can be called over and over again on the same ID, which then trickles down to all expression dependencies recursively, and certain pathological shaders would explode in compile time. I guess I'll spend a little time on this tomorrow after all.

dnovillo:
> Fair enough, it does seem like there is a problem where invalidate ID can be called over and over again on the same ID, which then trickles down to all expression dependencies recursively, and certain pathological shaders would explode in compile time. I guess I'll spend a little time on this tomorrow after all.

Thanks!

HansKristian-Work:
Fixed it now (almost 10x faster for me compared to before the performance regression).

dnovillo:
> Fixed it now (almost 10x faster for me compared to before the performance regression). The performance bug was always there.

Excellent.  Thanks for the quick fix!

PR #1113, opened 2019-08-02
MSL: Add support for sampler Y'CbCr conversion.

cdavis5e:
This change introduces functions and in one case, a class, to support
the `VK_KHR_sampler_ycbcr_conversion` extension. Except in the case of
GBGR8 and BGRG8 formats, for which Metal natively supports implicit
chroma reconstruction, we're on our own here. We have to do everything
ourselves. Much of the complexity comes from the need to support
multiple planes, which must now be passed to functions that use the
corresponding combined image-samplers. The rest is from the actual
Y'CbCr conversion itself, which requires additional post-processing of
the sample retrieved from the image.

Passing sampled images to a function was a particular problem. To
support this, I've added a new class which is emitted to MSL shaders
that pass sampled images with Y'CbCr conversions attached around. It
can handle sampled images with or without Y'CbCr conversion. This is an
awful abomination that should not exist, but I'm worried that there's
some shader out there which does this. This support requires Metal 2.0
to work properly, because it uses default-constructed texture objects,
which were only added in MSL 2. I'm not even going to get into arrays of
combined image-samplers--that's a whole other can of worms.  They are
deliberately unsupported in this change.

I've taken the liberty of refactoring the support for texture swizzling
while I'm at it. It's now treated as a post-processing step similar to
Y'CbCr conversion. I'd like to think this is cleaner than having
everything in `to_function_name()`/`to_function_args()`. It still looks
really hairy, though. I did, however, get rid of the explicit type
arguments to `spvGatherSwizzle()`/`spvGatherCompareSwizzle()`.

Update the C API. In addition to supporting this new functionality, add
some compiler options that I added in previous changes, but for which I
neglected to update the C API. Unfortunately, this is an ABI-breaking
change. If this is unacceptable, let me know.

(Yes, this is a terrible change to review. I apologize in advance for
inflicting this upon the hapless reviewer.)


HansKristian-Work:
I'm on holiday, can this wait until I'm back?

cdavis5e:
Yeah, this can wait.

spirv_cross_c.h
```diff
@@ -386,9 +435,21 @@ typedef struct spvc_msl_constexpr_sampler
 	float lod_clamp_max;
 	int max_anisotropy;
 
+	/* Sampler Y'CbCr conversion parameters */
```
HansKristian-Work: ABI break is not allowed. This needs to be split into a separate YCbCr struct.

HansKristian-Work:
I'm back now, so picking this up again.

This is ... quite a large and complex PR. Even larger than tessellation. Going to take a few review passes to feel comfortable merging something like this. The complexity scares me.

Is this purely to satisfy a Vulkan 1.1 requirement or are there actual use cases for this extension?

HansKristian-Work:
Not sure if splitting it up would make it that easier, just need to take it slow and steady. Will also help to run some of my own shaders to see how things behave.

spirv_cross_c.h
```diff
@@ -550,6 +625,8 @@ SPVC_PUBLIC_API spvc_bool spvc_compiler_msl_is_resource_used(spvc_compiler compi
                                                              unsigned binding);
 SPVC_PUBLIC_API spvc_result spvc_compiler_msl_remap_constexpr_sampler(spvc_compiler compiler, spvc_variable_id id, const spvc_msl_constexpr_sampler *sampler);
 SPVC_PUBLIC_API spvc_result spvc_compiler_msl_remap_constexpr_sampler_by_binding(spvc_compiler compiler, unsigned desc_set, unsigned binding, const spvc_msl_constexpr_sampler *sampler);
+SPVC_PUBLIC_API spvc_result spvc_compiler_msl_remap_constexpr_sampler_2(spvc_compiler compiler, spvc_variable_id id, const spvc_msl_constexpr_sampler *sampler, const spvc_msl_sampler_ycbcr_conversion *conv);
```
HansKristian-Work: Probably more obvious to call it sampler_ycbcr rather than _2.

spirv_msl.hpp
```diff
@@ -349,6 +423,14 @@ class CompilerMSL : public CompilerGLSL
 	// sampler's binding is returned instead. For any other resource type, -1 is returned.
 	uint32_t get_automatic_msl_resource_binding_secondary(uint32_t id) const;
 
+	// Same as get_automatic_msl_resource_binding, but should only be used for combined image samplers for multiplanar images,
+	// in which case the second plane's binding is returned instead. For any other resource type, -1 is returned.
+	uint32_t get_automatic_msl_resource_binding_tertiary(uint32_t id) const;
```
HansKristian-Work: Need to update `get_default_extended_decoration` to get this default -1 behavior.

spirv_msl.cpp
```diff
@@ -3109,6 +3126,37 @@ void CompilerMSL::emit_custom_functions()
...
+			spv_function_implementations.insert(SPVFuncImplGatherSwizzle);
+		for (uint32_t i = SPVFuncImplChromaReconstructNearest2Plane;
+		     i <= SPVFuncImplChromaReconstructLinear420XMidpointYMidpoint3Plane; i++)
+			spv_function_implementations.insert(static_cast<SPVFuncImpl>(i));
+		spv_function_implementations.insert(SPVFuncImplExpandITUFullRange);
+		spv_function_implementations.insert(SPVFuncImplExpandITUNarrowRange);
+		spv_function_implementations.insert(SPVFuncImplConvertYCbCrBT709);
+		spv_function_implementations.insert(SPVFuncImplConvertYCbCrBT601);
+		spv_function_implementations.insert(SPVFuncImplConvertYCbCrBT2020);
+	}
```
HansKristian-Work: Please add some new-lines between the blocks, hard to read otherwise.

spirv_msl.cpp
```diff
@@ -3456,11 +3510,14 @@ void CompilerMSL::emit_custom_functions()
...
 			statement("");
+			break;
+
+		case SPVFuncImplGatherSwizzle:
 			statement("// Wrapper function that swizzles texture gathers.");
-			statement("template<typename T, typename Tex, typename... Ts>");
-			statement(
-			    "inline vec<T, 4> spvGatherSwizzle(sampler s, const thread Tex& t, Ts... params, component c, uint sw) "
-			    "METAL_CONST_ARG(c)");
+			statement("template<typename T, template<typename, access = access::sample, typename = void> class Tex, "
```
HansKristian-Work: Didn't think I would ever see templates of templates in shader code ...

spirv_msl.cpp
```diff
@@ -3646,6 +3707,704 @@ void CompilerMSL::emit_custom_functions()
 			statement("");
 			break;
 
+		case SPVFuncImplChromaReconstructNearest2Plane:
+			statement("template<typename T, typename... LodOptions>");
+			statement("inline vec<T, 4> spvChromaReconstructNearest(texture2d<T> plane0, texture2d<T> plane1, sampler "
```
HansKristian-Work: I don't follow most of these implementations. Care to elaborate how these implementations are found?

spirv_msl.cpp
```diff
@@ -3646,6 +3707,704 @@ void CompilerMSL::emit_custom_functions()
...
+			statement("vec<T, 4> rgba;");
+			statement("rgba.rgb = vec<T, 3>(spvBT2020Factors * ycbcr.gbr);");
+			statement("rgba.a = ycbcr.a;");
+			statement("return rgba;");
+			end_scope();
+			statement("");
+			break;
+
+		case SPVFuncImplDynamicImageSampler:
+			statement("enum class spvFormatResolution");
```
HansKristian-Work: ... ye. Just going to assume it works. :)

HansKristian-Work:
Starting to grok it.

spirv_msl.cpp
cdavis5e: They're based on [the descriptions of explicit chroma reconstruction in the Vulkan standard](https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/chap15.html#textures-explicit-reconstruction).

spirv_msl.cpp
cdavis5e: Like I said: this is an Awful Abomination That Should Not Exist.

spirv_msl.cpp
cdavis5e: To be honest, I wasn't even sure that would work.

spirv_msl.cpp
HansKristian-Work: I see. Sounds good then.

HansKristian-Work:
Thanks a lot for implementing this. This must have been pure hell to implement :p

PR #1138, opened 2019-08-28
Support the SPV_EXT_fragment_shader_interlock extension.

### LINKED ISSUE #1002: Request [GLSL,MSL,HLSL]: Support for new SPV_EXT_fragment_shader_interlock
Hi,
after my last year request (https://github.com/KhronosGroup/Vulkan-Ecosystem/issues/27) to expose ROV functionality, finally new Vulkan 1.1.110 got released with VK_EXT_fragment_shader_interlock support and associated SPIR-V ext.: SPV_EXT_fragment_shader_interlock..
right now SPIR-V ext is here:
https://github.com/KhronosGroup/SPIRV-Registry/pull/31/commits/065daf4d19a4be2da7dc78d525768668d4188a68

As all other modern APIs (Metal2 and D3D12 and even GL4) support this functionality to various degrees (I mean for ex. the interactions with Variable rate shading are missing in Metal not sure how interacts with new D3D12 VRS support) would be nice if SPIRV-Cross can support it in all shading languages: GLSL output (it's the same existing https://www.khronos.org/registry/OpenGL/extensions/ARB/ARB_fragment_shader_interlock.txt), Metal and HLSL..

EDIT: forgot to note we have drv support from NV (0-day) and even Mesa Anvil patches for Intel: https://gitlab.freedesktop.org/mesa/mesa/merge_requests/989

thanks..

cdavis5e:
This was straightforward to implement in GLSL. The
`ShadingRateInterlockOrderedEXT` and `ShadingRateInterlockUnorderedEXT`
modes aren't implemented yet, because we don't support
`SPV_NV_shading_rate` or `SPV_EXT_fragment_invocation_density` yet.

HLSL and MSL were more interesting. They don't support this directly,
but they do support marking resources as "rasterizer ordered," which
does roughly the same thing. So this implementation scans all accesses
inside the critical section and marks all storage resources found
therein as rasterizer ordered. They also don't support the fine-grained
controls on pixel- vs. sample-level interlock and disabling ordering
guarantees that GLSL and SPIR-V do, but that's OK. "Unordered" here
merely means the order is undefined; that it just so happens to be the
same as rasterizer order is immaterial. As for pixel- vs. sample-level
interlock, Vulkan explicitly states:

> With sample shading enabled, [the `PixelInterlockOrderedEXT` and
> `PixelInterlockUnorderedEXT`] execution modes are treated like
> `SampleInterlockOrderedEXT` or `SampleInterlockUnorderedEXT`
> respectively.

and:

> If [the `SampleInterlockOrderedEXT` or `SampleInterlockUnorderedEXT`]
> execution modes are used in single-sample mode they are treated like
> `PixelInterlockOrderedEXT` or `PixelInterlockUnorderedEXT`
> respectively.

So this will DTRT for MoltenVK and gfx-rs, at least.

MSL additionally supports multiple raster order groups; resources that
are not accessed together can be placed in different ROGs to allow them
to be synchronized separately. A more sophisticated analysis might be
able to place resources optimally, but that's outside the scope of this
change. For now, we assign all resources to group 0, which should do for
our purposes.

`glslang` doesn't support the `RasterizerOrdered` UAVs this
implementation produces for HLSL, so the test case needs `fxc.exe`.

It also insists on GLSL 4.50 for `GL_ARB_fragment_shader_interlock`,
even though the spec says it needs either 4.20 or
`GL_ARB_shader_image_load_store`; and it doesn't support the
`GL_NV_fragment_shader_interlock` extension at all. So I haven't been
able to test those code paths.

Fixes #1002.

### DIFF
--- a/spirv_hlsl.cpp
+++ b/spirv_hlsl.cpp
@@ -203,7 +203,7 @@ static string image_format_to_type(ImageFormat fmt, SPIRType::BaseType basetype)
 	}
 }
 
-string CompilerHLSL::image_type_hlsl_modern(const SPIRType &type, uint32_t)
+string CompilerHLSL::image_type_hlsl_modern(const SPIRType &type, uint32_t id)
 {
 	auto &imagetype = get<SPIRType>(type.image.type);
 	const char *dim = nullptr;
@@ -235,7 +235,12 @@ string CompilerHLSL::image_type_hlsl_modern(const SPIRType &type, uint32_t)
 		if (type.image.sampled == 1)
 			return join("Buffer<", type_to_glsl(imagetype), components, ">");
 		else if (type.image.sampled == 2)
+		{
+			if (interlocked_resources.count(id))
+				return join("RasterizerOrderedBuffer<", image_format_to_type(type.image.format, imagetype.basetype),
+				            ">");
 			return join("RWBuffer<", image_format_to_type(type.image.format, imagetype.basetype), ">");
+		}
 		else
 			SPIRV_CROSS_THROW("Sampler buffers must be either sampled or unsampled. Cannot deduce in runtime.");
 	case DimSubpassData:
@@ -248,6 +253,8 @@ string CompilerHLSL::image_type_hlsl_modern(const SPIRType &type, uint32_t)
 	const char *arrayed = type.image.arrayed ? "Array" : "";
 	const char *ms = type.image.ms ? "MS" : "";
 	const char *rw = typed_load ? "RW" : "";
+	if (typed_load && interlocked_resources.count(id))
+		rw = "RasterizerOrdered";
 	return join(rw, "Texture", dim, ms, arrayed, "<",
 	            typed_load ? image_format_to_type(type.image.format, imagetype.basetype) :
 	                         join(type_to_glsl(imagetype), components),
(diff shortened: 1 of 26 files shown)


spirv_hlsl.cpp
```diff
@@ -4850,6 +4867,8 @@ string CompilerHLSL::compile()
 	validate_shader_model();
 	update_active_builtins();
 	analyze_image_and_sampler_usage();
+	if (get_execution_model() == ExecutionModelFragment)
+		analyze_interlocked_resource_usage();
```
HansKristian-Work: Should only be used if ROV execution modes are active. No need to analyze here unless we need to.

HansKristian-Work:
@cdavis5e not sure how well you know the CFG part of SPIRV-Cross. I can probably continue on the PR and implement the analysis if you want.

HansKristian-Work:
@cdavis5e There is nothing particularly urgent to do right now, so picking this up tomorrow unless you've started on the rework already.

cdavis5e:
I haven't. I made the other change you asked for, and I've thought a bit about what I wanted to do, but if you have nothing better to do, be my guest.

HansKristian-Work:
Alright, thanks! Hopefully this will land tomorrow then.

HansKristian-Work:
Merged with the fixes on top. 🎉

PR #1141, opened 2019-08-29
MSL: Inline all non-entry-point functions

troughton:
This PR ensures that all emitted MSL functions are marked as inline, enabling multiple output files to be bundled together into a metallib without duplicate symbols.

The commits are probably best viewed separately; the last one just updates all the tests, while the first two make the functionality change.

### DIFF
--- a/spirv_msl.cpp
+++ b/spirv_msl.cpp
@@ -5368,9 +5368,10 @@ void CompilerMSL::emit_function_prototype(SPIRFunction &func, const Bitset &)
 		add_function_overload(func);
 
 	local_variable_names = resource_names;
-	string decl;
 
 	processing_entry_point = (func.self == ir.default_entry_point);
+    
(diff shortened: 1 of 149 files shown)


troughton:
(As an alternative, marking all functions as `static` would also serve the same purpose).

spirv_msl.cpp
```diff
@@ -5368,9 +5368,10 @@ void CompilerMSL::emit_function_prototype(SPIRFunction &func, const Bitset &)
 		add_function_overload(func);
 
 	local_variable_names = resource_names;
-	string decl;
 
 	processing_entry_point = (func.self == ir.default_entry_point);
+    
+    string decl = processing_entry_point ? "" : "inline ";
```
HansKristian-Work: Formatting (tabs vs spaces). There is a .clang-format in the repo.

PR #1144, opened 2019-09-04
MSL: Force storage images on iOS to use discrete descriptors.

cdavis5e:
Writable textures cannot use argument buffers on iOS. They must be
passed as arguments directly to the shader function. Since we won't know
if a given storage image will have the `NonWritable` decoration at the
time we encode the argument buffer, we must therefore pass all storage
images as discrete arguments. Previously, we were throwing an error if
we encountered an argument buffer with a writable texture in it on iOS.

### DIFF
--- a/spirv_msl.cpp
+++ b/spirv_msl.cpp
@@ -8161,15 +8161,19 @@ void CompilerMSL::entry_point_args_discrete_descriptors(string &ep_args)
 
 	SmallVector<Resource> resources;
 
-	ir.for_each_typed_id<SPIRVariable>([&](uint32_t, SPIRVariable &var) {
+	ir.for_each_typed_id<SPIRVariable>([&](uint32_t var_id, SPIRVariable &var) {
 		if ((var.storage == StorageClassUniform || var.storage == StorageClassUniformConstant ||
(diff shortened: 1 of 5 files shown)


spirv_msl.cpp
```diff
@@ -11033,10 +11042,21 @@ void CompilerMSL::analyze_argument_buffers()
...
-				add_resource_name(var_id);
-				resources_in_set[desc_set].push_back(
-				    { &var, to_name(var_id), type.basetype, get_metal_resource_index(var, type.basetype), 0 });
+				if (msl_options.is_ios() && type.basetype == SPIRType::Image && type.image.sampled == 2)
+				{
+					// Very specifically, image load-store in argument buffers are disallowed on MSL on iOS.
+					// But we won't know when the argument buffer is encoded whether this image will have
+					// a NonWritable decoration. So just use discrete arguments for all storage images
+					// on iOS.
+					discrete_storage_images.insert(var_id);
```
HansKristian-Work: Going through a std::unordered_set causes MSVC to emit different code. get_metal_resource_index() is called in flipped order, so the resources get index 1, 0 instead of 0, 1. Is it possible to check for this twice with positive test and negative test instead of going through a map?

PR #1145, opened 2019-09-04
Fix ParsedIR::mark_used_as_array_length(uint32_t id)

lifpan:
Need take care of arguments specially for OpCompositeExtract and OpCompositeInsert.
[OpGroupNonUniformElect.spv.zip](https://github.com/KhronosGroup/SPIRV-Cross/files/3576775/OpGroupNonUniformElect.spv.zip)


HansKristian-Work:
Nice catch!

PR #1160, opened 2019-09-18
Update external/ to SPIR-V 1.5

zoddicus:
Rolled the hashes used for glslang, SPIRV-Tools, and SPIRV-Headers to
HEAD, which includes the update to 1.5.

Added passing '--amb' to glslang, so I didn't have to explicitly set
bindings in a large number of test shaders that currently don't, and
now glslang considers them invalid.

Ran update_test_shaders.sh to update all of the reference files.

Marked all shaders that no longer pass spirv-val as .invalid.


zoddicus:
PTAL, this is my naive attempt up updating spirv-cross's dependencies.

The last update to glslang's hash in external/ is to a version from June, so now there is difficulties integrating newer versions of spirv-tools & glslang into projects that also depend on spirv-cross.

HansKristian-Work:
Seems reasonable, that's the right approach. I'll need to verify tomorrow that the codegen changes make sense, some are a little suspect.

HansKristian-Work:
This change exposed a SPIRV-Cross bug which caused the test failure, I'll fix that locally and push the fix.

PR #1162, opened 2019-09-20
Lots of changes and fixes with new test cases to make SPIRV-Cross work in UE4 for Metal

LukasBanana:
This pull request is the result of a bunch of work from my former colleague and some adjustments by myself in order to make SPIRV-Cross work in conjunction with ShaderConductor in Unreal Engine 4.

Several new test cases were added to justify the changes that were made.
I also renamed the respective test shaders to match the guideline of this project.
Most of the changed files in this pull request are the updated reference shaders.

One of the most notable changes is the introduction of the `spvUnsafeArray<>` template struct to make arrays a value type in Metal.
It's called "unsafe" because there are no boundary checks and it's really just a wrapper with respective `operator []` functions around a native array to make them compatible with how structured and byte-address buffers work.
The `spvArrayCopy*` functions can now be avoided most of the time to let the Metal compiler do a better optimization. However, `spvArrayCopy*` wrappers are still required when arrays are declared in `threadgroup` address space.

Another change worth mentioning is that the translation of HLSL's `SV_InstanceID` and `SV_VertexID` works correct now, since these system values always start at 0.
Metal's `[[instance_id]]`/`[[vertex_id]]` and GLSL's `gl_VertexIndex`/`gl_InstanceIndex`, however, always start at the base index specified in the draw command. Both Metal and GLSL support system values for these base indices which are now exposed to make the translation of `SV_InstanceID` and `SV_VertexID` behave in the same way.

We also had to re-enable some features that were removed recently, like the user defined decoration binding, which can now be enabled again with the respective boolean in `CompilerMSL::Options` but it's disabled by default, to keep the behavior as it was without these changes.

We hope to get these changes into Khronos' repository to simplify our update iterations.
Any feedback is highly welcome.


HansKristian-Work:
Thanks! This looks like a rather massive PR, so will take a while to get through the review.

LukasBanana:
The main contributor to these changes replied to my request and he will sort it out in the next days.
So yes, everyone will be able to sign it.

spirv_common.hpp
```diff
@@ -1415,12 +1415,12 @@ class Variant
 			group->pools[type]->free_opaque(holder);
 		holder = nullptr;
 
-		if (!allow_type_rewrite && type != TypeNone && type != new_type)
+		/*if (!allow_type_rewrite && type != TypeNone && type != new_type)
```
dneto0: Drive-by, as this is not my project(!). 

I'd prefer too see the code deleted entirely.
(In a pinch use #if 0 .. #endif)

spirv_msl.cpp
```diff
@@ -10682,68 +11957,529 @@ void CompilerMSL::OpCodePreprocessor::check_resource_write(uint32_t var_id)
...
-
-	case OpFunctionCall:
+	// Storage buffer robustness
+	case OpInBoundsAccessChain:
+	case OpAccessChain:
+	case OpPtrAccessChain:
 	{
-		auto &return_type = compiler.get<SPIRType>(args[0]);
-		if (return_type.array.size() > 1)
+		if (compiler.msl_options.enforce_storge_buffer_bounds)
```
dneto0: FYI. 

For Google's WebGPU work, we too needed a way to force sandboxing of memory accesses. 

We wrote a SPIRV-Tools transform to inject index clamping for all access chains (assuming no variable pointers, just plain logical addressing). The transform does a superset of robust-buffer-access, as it also protects internal memory accesses as well.

In our WebGPU flow on top of Metal, the plan is: parse the SPIR-V, validate it, transform it with this sandboxing pass, then feed it to SPIRV-Cross.

You can play with it via spirv-opt --graphics-robust-access

The code is at https://github.com/KhronosGroup/SPIRV-Tools/blob/master/source/opt/graphics_robust_access_pass.cpp 

I have not compared this implementation with the SPIRV-Tools transform.   But I wanted to make you aware of the other work.   We could probably both benefit from looking at each others test cases.

LukasBanana: Thanks for the hint. I will have a closer look at this spirv-opt pass next week. Would be definitely great to find a common denominator for as many cases as possible instead of divergence.

reference/opt/shaders-msl/asm/comp/atomic-decrement.asm.comp
```diff
@@ -13,7 +14,8 @@ struct u0_counters
 };
 
 // Returns 2D texture coords corresponding to 1D texel buffer coords
-inline uint2 spvTexelBufferCoord(uint tc)
+static inline __attribute__((always_inline))
```
HansKristian-Work: Is it possible for MSL compiler to actually emit function calls here?

spirv_common.hpp
HansKristian-Work: This is a sanity check to make sure an ID is not tagged with different types over the course of compiling. If this check triggers, it's most likely a bug.

spirv_cross.cpp
```diff
@@ -3910,10 +3928,40 @@ void Compiler::CombinedImageSamplerUsageHandler::add_hierarchy_to_comparison_ids
 {
 	// Traverse the variable dependency hierarchy and tag everything in its path with comparison ids.
 	comparison_ids.insert(id);
+	
+	// If the underlying resource has been used for comparison then duplicate loads of that resource must be too.
+	if (!compiler.supports_combined_samplers())
+	{
+		for (auto it = dependency_hierarchy.begin(); it != dependency_hierarchy.end(); ++it)
```
HansKristian-Work: Use C++11 for loop here `(for auto &blah : blahs)`.

spirv_cross.cpp
```diff
@@ -3910,10 +3928,40 @@ void Compiler::CombinedImageSamplerUsageHandler::add_hierarchy_to_comparison_ids
...
+	}
+	
 	for (auto &dep_id : dependency_hierarchy[id])
 		add_hierarchy_to_comparison_ids(dep_id);
 }
 
+// If the underlying resource has been used for comparison then duplicate loads of that resource must be too.
+bool Compiler::CombinedImageSamplerUsageHandler::dependent_used_for_comparison(uint32_t id) const
+{
+	if (compiler.supports_combined_samplers())
```
HansKristian-Work: Similar here to samplers. Is there a concrete test case for this in the PR?

spirv_cross.hpp
```diff
@@ -683,7 +686,7 @@ class Compiler
 	bool interface_variable_exists_in_entry_point(uint32_t id) const;
 
 	SmallVector<CombinedImageSampler> combined_image_samplers;
-
+
```
HansKristian-Work: Formatting

spirv_glsl.cpp
```diff
@@ -3394,11 +3398,19 @@ string CompilerGLSL::constant_expression(const SPIRConstant &c)
 	{
 		// Handles Arrays and structures.
 		string res;
+		// Allow Metal to use the array<T> template to make arrays a value type
+		bool bTrailingBracket = false;
```
HansKristian-Work: No hungarian notation.

spirv_glsl.cpp
```diff
@@ -333,7 +336,8 @@ void CompilerGLSL::reset()
 
 	ir.for_each_typed_id<SPIRVariable>([&](uint32_t, SPIRVariable &var) { var.dependees.clear(); });
 
-	ir.reset_all_of_type<SPIRExpression>();
+    // Track write-throughs for loop variables - dxc likes to generate them
```
HansKristian-Work: What does this mean?

spirv_glsl.cpp
```diff
@@ -7585,7 +7618,7 @@ bool CompilerGLSL::remove_unity_swizzle(uint32_t base, string &op)
 	auto &type = expression_type(base);
 
 	// Sanity checking ...
-	assert(type.columns == 1 && type.array.empty());
+	assert(type.columns == 1); //  && type.array.empty()
```
HansKristian-Work: Why remove array check?

spirv_glsl.cpp
```diff
@@ -4009,6 +4024,18 @@ string CompilerGLSL::constant_expression_vector(const SPIRConstant &c, uint32_t
 		}
 		break;
 
+	// Metal tessellation likes empty structs which are then constant expressions.
+	case SPIRType::Struct:
+		if (type.member_types.size() == 0)
```
HansKristian-Work: Struct constant expression are emitted by going through subconstants of SPIRConstant. Is this path ever hit?

spirv_glsl.hpp
```diff
@@ -523,7 +527,10 @@ class CompilerGLSL : public Compiler
 	std::string to_unpacked_expression(uint32_t id, bool register_expression_read = true);
 	std::string to_unpacked_row_major_matrix_expression(uint32_t id);
 	std::string to_enclosed_unpacked_expression(uint32_t id, bool register_expression_read = true);
-	std::string to_dereferenced_expression(uint32_t id, bool register_expression_read = true);
+	
+    // Metal expands float[]/float2[] members inside structs to float4[] so we must unpack
```
HansKristian-Work: Formatting, there is a clang format file, FWIW.

HansKristian-Work: Was there a need to change this? Unpacking from float1/2 UBO array should already work.

spirv_msl.hpp
```diff
@@ -253,6 +253,10 @@ class CompilerMSL : public CompilerGLSL
 			macOS = 1
 		} Platform;
 
+		// Provide the Metal bindings as part of the options structure as that is more convenient.
```
HansKristian-Work: Please remove this, this will mess with the C API.

spirv_msl.hpp
```diff
@@ -320,6 +333,18 @@ class CompilerMSL : public CompilerGLSL
 	void set_msl_options(const Options &opts)
 	{
 		msl_options = opts;
+		
+		// Provide the Metal bindings as part of the options structure as that is more convenient.
```
HansKristian-Work: Please remove this. This messes with the C API. There shouldn't be two paths to set this.

spirv_msl.hpp
```diff
@@ -266,6 +270,7 @@ class CompilerMSL : public CompilerGLSL
 		uint32_t dynamic_offsets_buffer_index = 23;
 		uint32_t shader_input_wg_index = 0;
 		uint32_t device_index = 0;
+		uint32_t argument_buffer_offset = 0; // Allow the caller to specify the Metal translation should use argument buffers
```
HansKristian-Work: What does this do? "should use argument buffers" is another option.

spirv_msl.hpp
```diff
@@ -282,6 +287,14 @@ class CompilerMSL : public CompilerGLSL
 		// Fragment output in MSL must have at least as many components as the render pass.
 		// Add support to explicit pad out components.
 		bool pad_fragment_output_components = false;
+		bool ios_support_base_vertex_instance = false; // Provide the Metal bindings as part of the options structure as that is more convenient.
```
HansKristian-Work: Comment seems wrong.

spirv_msl.hpp
```diff
@@ -460,6 +485,9 @@ class CompilerMSL : public CompilerGLSL
 	// to use for a particular location. The default is 4 if number of components is not overridden.
 	void set_fragment_output_components(uint32_t location, uint32_t components);
 
+	// Returns false, because Metal does not support combined texture-samplers.
+	virtual bool supports_combined_samplers() const override;
```
HansKristian-Work: Move this to protected, it's not a public API.

spirv_msl.hpp
```diff
@@ -282,6 +287,14 @@ class CompilerMSL : public CompilerGLSL
...
 		// Add support to explicit pad out components.
 		bool pad_fragment_output_components = false;
+		bool ios_support_base_vertex_instance = false; // Provide the Metal bindings as part of the options structure as that is more convenient.
+		bool ios_use_framebuffer_fetch_subpasses = true; // Use Metal's native frame-buffer fetch API for subpass inputs.
+		bool enforce_storge_buffer_bounds = false; // Storage buffer robustness - clamps access to SSBOs to the size of the buffer
+		
+		bool invariant_float_math = false;
+		
+		bool emulate_cube_array = false; // Emulate texturecube_array with texture2d_array for iOS where this type is not available
+		bool enable_decoration_binding = false; // Allow user to enable decoration binding
```
HansKristian-Work: What does this do?

spirv_msl.hpp
```diff
@@ -483,9 +511,21 @@ class CompilerMSL : public CompilerGLSL
...
+		SPVFuncImplImage2DAtomicCoords, // Emulate texture2D atomic operations
+		SPVFuncImplStorageBufferCoords, // Storage buffer robustness
+		SPVFuncImplFMul,
+		SPVFuncImplFAdd,
+		SPVFuncImplCubemapTo2DArrayFace,
+		SPVFuncImplUnsafeArray, // Allow Metal to use the array<T> template to make arrays a value type
 		SPVFuncImplInverse4x4,
 		SPVFuncImplInverse3x3,
 		SPVFuncImplInverse2x2,
+		SPVFuncImplRowMajor2x3,
```
HansKristian-Work: Why is this reintroduced?

spirv_msl.hpp
```diff
@@ -789,6 +852,7 @@ class CompilerMSL : public CompilerGLSL
 
 	std::unordered_set<uint32_t> buffers_requiring_array_length;
 	SmallVector<uint32_t> buffer_arrays;
+    std::set<SPIRVariable *> atomic_vars; // Emulate texture2D atomic operations
```
HansKristian-Work: Does this have to be a set and not unordered_set?

spirv_msl.cpp
```diff
@@ -1033,10 +1083,14 @@ void CompilerMSL::preprocess_op_codes()
 
 	suppress_missing_prototypes = preproc.suppress_missing_prototypes;
 
+    // Disable warning about missing braces for array<T> template to make arrays a value type
+    add_pragma_line("#pragma clang diagnostic ignored \"-Wmissing-braces\"");
```
HansKristian-Work: Can this be added conditionally to avoid getting changes for 700++ files in this PR?

spirv_msl.cpp
```diff
@@ -2303,11 +2372,22 @@ uint32_t CompilerMSL::add_interface_block(StorageClass storage, bool patch)
 					{
 						// The first member of the indirect buffer is always the number of vertices
 						// to draw.
-						statement("device ", to_name(ir.default_entry_point), "_", ib_var_ref, "& ", ib_var_ref, " = ",
-						          output_buffer_var_name, "[(", to_expression(builtin_instance_idx_id), " - ",
-						          to_expression(builtin_base_instance_id), ") * spvIndirectParams[0] + ",
-						          to_expression(builtin_vertex_idx_id), " - ", to_expression(builtin_base_vertex_id),
-						          "];");
+						// We zero-base the InstanceID & VertexID variables for HLSL emulation elsewhere, so don't do it twice
+						if (ir.source.hlsl == true)
```
HansKristian-Work: Cannot rely on OpSource. It's a debug instruction.

spirv_msl.cpp
```diff
@@ -2303,11 +2372,22 @@ uint32_t CompilerMSL::add_interface_block(StorageClass storage, bool patch)
...
+						{
+							statement("device ", to_name(ir.default_entry_point), "_", ib_var_ref, "& ", ib_var_ref, " = ",
+									  output_buffer_var_name, "[", to_expression(builtin_instance_idx_id), " * spvIndirectParams[0] + ",
+									  to_expression(builtin_vertex_idx_id),
+									  "];");
+						}
+						else
+						{
+							statement("device ", to_name(ir.default_entry_point), "_", ib_var_ref, "& ", ib_var_ref, " = ",
```
HansKristian-Work: Why is base vertex subtracted here?

spirv_msl.cpp
```diff
@@ -3097,6 +3177,32 @@ void CompilerMSL::emit_store_statement(uint32_t lhs_expression, uint32_t rhs_exp
 	}
 }
 
+// Metal expands float[]/float2[] members inside structs to float4[] so we must unpack
+string CompilerMSL::to_dereferenced_expression(uint32_t id, bool register_expression_read)
```
HansKristian-Work: Is it necessary to create an override here? Handling these cases should already work without a redundant override.

spirv_msl.cpp
```diff
@@ -3097,6 +3177,32 @@ void CompilerMSL::emit_store_statement(uint32_t lhs_expression, uint32_t rhs_exp
...
+			packed_type = &get<SPIRType>(packed_type_id);
+		
+		if (packed_type && is_array(*packed_type) && is_scalar(*packed_type))
+			return to_unpacked_expression(id, register_expression_read);
+	}
+	
+	return CompilerGLSL::to_dereferenced_expression(id, register_expression_read);
+}
+
+static bool expression_ends_with(string const& expr_str, std::string const& ending)
```
HansKristian-Work: `const string &expr_str, const string &ending`

spirv_msl.cpp
```diff
@@ -3097,6 +3177,32 @@ void CompilerMSL::emit_store_statement(uint32_t lhs_expression, uint32_t rhs_exp
...
+{
+	if (expr_str.length() >= ending.length())
+		return (0 == expr_str.compare(expr_str.length() - ending.length(), ending.length(), ending));
```
HansKristian-Work: no yoda-conditionals (`0 == expr`).

spirv_msl.cpp
```diff
@@ -3118,7 +3224,7 @@ string CompilerMSL::unpack_expression_type(string expr_str, const SPIRType &type
 	};
 
 	// std140 array cases for vectors.
-	if (physical_type && is_vector(*physical_type) && is_array(*physical_type) && physical_type->vecsize > type.vecsize)
+	if (physical_type && is_vector(*physical_type) && is_array(*physical_type) && physical_type->vecsize > type.vecsize && !expression_ends_with(expr_str, swizzle_lut[type.vecsize - 1]))
```
HansKristian-Work: Why is it necessary to string compare here?

spirv_msl.cpp
```diff
@@ -3207,7 +3313,63 @@ void CompilerMSL::add_typedef_line(const string &line)
 		force_recompile();
 }
 
+// Template struct like spvUnsafeArray<> need to be declared *before* any resources are declared
+void CompilerMSL::emit_custom_templates()
+{
+	for (const auto& spv_func : spv_function_implementations)
```
HansKristian-Work: `const auto &spv_func`

spirv_msl.cpp
```diff
@@ -3246,15 +3408,16 @@ void CompilerMSL::emit_custom_functions()
...
+	
+	for (const auto& spv_func : spv_function_implementations)
 	{
 		switch (spv_func)
 		{
 		case SPVFuncImplMod:
 			statement("// Implementation of the GLSL mod() function, which is slightly different than Metal fmod()");
 			statement("template<typename Tx, typename Ty>");
-			statement("inline Tx mod(Tx x, Ty y)");
+			statement(force_inline);
```
HansKristian-Work: Why does this have to be force inline?

spirv_msl.cpp
```diff
@@ -3533,19 +3824,87 @@ void CompilerMSL::emit_custom_functions()
...
+			statement("template<typename T>");
+			statement(force_inline);
+			statement("constexpr thread T&& spvForward(thread typename spvRemoveReference<T>::type&& x)");
 			begin_scope();
 			statement("return static_cast<thread T&&>(x);");
 			end_scope();
 			statement("");
 			break;
 
+		case SPVFuncImplRowMajor2x3:
```
HansKristian-Work: This was removed in the grand buffer packing refactor. There should be no reason to reintroduce it.

spirv_msl.cpp
```diff
@@ -4509,14 +4873,15 @@ void CompilerMSL::declare_undefined_values()
 	bool emitted = false;
 	ir.for_each_typed_id<SPIRUndef>([&](uint32_t, SPIRUndef &undef) {
 		auto &type = this->get<SPIRType>(undef.basetype);
-		statement("constant ", variable_decl(type, to_name(undef.self), undef.self), " = {};");
+		statement("constant ", CompilerGLSL::variable_decl(type, to_name(undef.self), undef.self), " = {};");
```
HansKristian-Work: Why jump up to parent class here?

spirv_msl.cpp
```diff
@@ -5643,6 +6207,228 @@ void CompilerMSL::emit_instruction(const Instruction &instruction)
 	previous_instruction_opcode = opcode;
 }
 
+// If the underlying resource has been used for comparison then duplicate loads of that resource must be too
+static inline bool image_opcode_is_sample_no_dref(Op op)
```
HansKristian-Work: This function probably does not exist anymore.

spirv_msl.cpp
```diff
@@ -5643,6 +6207,228 @@ void CompilerMSL::emit_instruction(const Instruction &instruction)
...
+	case OpImageSparseRead:
+		return true;
+		
+	default:
+		return false;
+	}
+}
+
+void CompilerMSL::emit_texture_op(const Instruction &i)
+{
```
HansKristian-Work: Was this rebased on the code before YCbCr support was added a few weeks ago? This code should not exist anymore.

spirv_msl.cpp
```diff
@@ -5707,29 +6496,6 @@ void CompilerMSL::emit_barrier(uint32_t id_exe_scope, uint32_t id_mem_scope, uin
 			bar_stmt += "mem_flags::mem_none";
 	}
 
-	if (msl_options.is_ios() && (msl_options.supports_msl_version(2) && !msl_options.supports_msl_version(2, 1)))
```
HansKristian-Work: Is this okay to remove?

spirv_msl.cpp
```diff
@@ -9909,19 +10977,104 @@ string CompilerMSL::builtin_to_glsl(BuiltIn builtin, StorageClass storage)
 	switch (builtin)
 	{
 
+	// Handle HLSL-style 0-based vertex/instance index.
 	// Override GLSL compiler strictness
 	case BuiltInVertexId:
-		return "gl_VertexID";
+		ensure_builtin(StorageClassInput, BuiltInVertexId);
+		if ((ir.source.hlsl == true) && msl_options.supports_msl_version(1, 1) && (msl_options.ios_support_base_vertex_instance || msl_options.is_macos()))
```
HansKristian-Work: Cannot rely on OpSource to be declared.

spirv_msl.cpp
```diff
@@ -10655,6 +11883,53 @@ bool CompilerMSL::OpCodePreprocessor::handle(Op opcode, const uint32_t *args, ui
 		uint32_t result_type = args[0];
 		uint32_t id = args[1];
 		uint32_t ptr = args[2];
+		
+		// Fix tessellation patch function processing
+		if(compiler.get_execution_model() == ExecutionModelTessellationControl)
+		{
+			uint32_t source_id = args[3];
+			bool isIndexedByInvocation = variables_indexed_by_invocation.find(ptr) != variables_indexed_by_invocation.end() || invocation_ids.find(source_id) != invocation_ids.end();
```
HansKristian-Work: no camel case

spirv_msl.cpp
```diff
@@ -10682,68 +11957,529 @@ void CompilerMSL::OpCodePreprocessor::check_resource_write(uint32_t var_id)
 		uses_resource_write = true;
 }
 
+// Storage buffer robustness
+// Fix loads from tessellation control inputs not being forwarded to the gl_in structure array
+// Fix loads from tessellation evaluation inputs not being forwarded to the stage_in structure array
+std::string CompilerMSL::access_chain_internal(uint32_t base, const uint32_t *indices, uint32_t count, AccessChainFlags flags, AccessChainMeta *meta)
```
HansKristian-Work: Completely duplicating the function seems a bit overkill.

spirv_msl.cpp
HansKristian-Work: I'd prefer if robustness checks can be done in SPIRV-Tools. No need to duplicate this effort.

reference/opt/shaders-msl/asm/comp/atomic-decrement.asm.comp
LukasBanana: Which function call do you mean?
The `static inline __attribute__((always_inline))` part is to enforce inlining. Otherwise, linking errors might occur due to duplicate symbols when multiple Metal shaders are linked into one library.

spirv_cross.cpp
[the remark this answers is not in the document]
LukasBanana: Since most of the changes are not from me, I still need to grasp a few things myself. Maybe you can clarify a few things for me, for example what is the type's meaning of `dependency_hierarchy`? It was already part of SPIRV-Cross, but the code doesn't document what this map-to-map type is supposed to accomplish.

Regarding your question:
As far as I remember, this fixed an issue from a Vulkan GLSL to ESSL translation when a texture was used with a regular sampler and a comparison sampler which needs to be merged into combined-texture-samplers for ESSL. But I'm not 100% sure.

reference/opt/shaders-msl/asm/comp/atomic-decrement.asm.comp
HansKristian-Work: I mean, doesn't static inline already make sure functions are inlined? From a linking perspective, inline should fix it. If the toolchain is buggy, I guess we'll have to do this ...

spirv_cross.cpp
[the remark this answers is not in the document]
HansKristian-Work: The dependency hierarchy is used in combined image sampler analysis to figure out which samplers and textures need to be marked as such. We need to be able to trace back from a sampling instruction or OpSampledImage all the way back to the original image, even through call stacks. I.e. a call-site -> image traversal.

From what I can understand of this, it seems to propagate this information the other way, from image to call-site. From image back up to the combined sampler call-site?

spirv_glsl.cpp
LukasBanana: Some of those changes might be solved elsewhere now. Updating all the test shaders with this changed being reverted didn't alter any of the reference shaders, so this might now longer be needed.
I'm preparing all necessary changes and submit it when I'm done.

spirv_glsl.cpp
LukasBanana: Looks like the same case as with the removal of `ir.reset_all_of_type<SPIRExpression>();` to me. Updated shaders work with this change reverted.

spirv_glsl.cpp
LukasBanana: This condition is triggered in `shaders-msl/asm/tesc/hs-incorrect-base-type.asm.tesc` - one of the new test shaders.

spirv_msl.hpp
LukasBanana: `argument_buffer_offset` is added to the buffer binding slot in `CompilerMSL::entry_point_args_argument_buffer`, to provide a way to adjust the binding slot of argument buffers. The commentary is misleading here, which might be the result of my rebasing on top of a more recent revision. Let me update this.

spirv_msl.hpp
LukasBanana: Agreed

spirv_msl.hpp
LukasBanana: `invariant_float_math` enables the use of [`fma`](https://developer.apple.com/metal/Metal-Shading-Language-Specification.pdf#page=122) builtin function in Metal.

`emulate_cube_array` adds a function to calculate cube-map-array texture coordinates for a 2D array texture. This is required for iOS versions that do not support cube-map arrays, but 2D array textures.

`enable_decoration_binding` adds the way resource bindings were exposed in a previous revision. Made optional and disabled by default, to not change the behavior.

spirv_msl.cpp
LukasBanana: I think that check was to avoid duplicate swizzle operators like `v.xy.xy`.
The string comparison can handle input expressions in `expr_str` that are unknown at that point. And the maximum string it is compared to is `".xyz"` so that shouldn't be a performance issue.

spirv_msl.cpp
LukasBanana: Seems to be unnecessary for templates. It was probably added to be on the safe side.

spirv_msl.cpp
LukasBanana: You're right, this is my fault. I didn't remove it properly when I rebased the changes from an older revision into a more recent one. It didn't break anything but indeed seems to be unnecessary now. I'll remove that.

spirv_msl.cpp
LukasBanana: The `variable_decl` function is overloaded with two virtual functions in `CompilerGLSL`, but `CompilerMSL` only overrides one of them. Omitting the base class prefix here will fail to compile with clang.

reference/opt/shaders-msl/asm/comp/atomic-decrement.asm.comp
LukasBanana: I'm afraid the only language "specification" for Metal we have is the [Metal-Shading-Language-Specification.pdf](https://developer.apple.com/metal/Metal-Shading-Language-Specification.pdf) document, but this doesn't document all compiler features that are inherited from clang (as Metal is a subset of a superset of C++14). As far as I know, even `static inline` is not a guarantee that functions always get inlined. With a quick search I found [this mailing list](https://lore.kernel.org/patchwork/patch/801329/) discussing that compilers don't always honor `inline`. `__attribute__((always_inline))`, however, is a language extension in clang that enforces it, so why not make use of it.

spirv_msl.hpp
LukasBanana: As mentioned below, this is indeed no longer used.

spirv_msl.hpp
LukasBanana: Thanks for pointing that out. This set is only used to check if a given key exists, so it can also be an unordered set.

spirv_msl.cpp
LukasBanana: Good point.
I moved that code to `CompilerMSL::emit_header()` where `-Wmissing-prototypes` is emitted, too. At that point we can check if `SPVFuncImplUnsafeArray` is contained in `spv_function_implementations`.

spirv_msl.cpp
LukasBanana: Another accidental removal during the rebase. Sorry for that.

reference/opt/shaders-msl/asm/comp/atomic-decrement.asm.comp
HansKristian-Work: Alright, if it's really needed I'm find keeping it for the relevant cases. Not using for templates should remove 90% of them already, so that's nice.

HansKristian-Work:
Alright, I'll await further updates. Thanks.

spirv_msl.cpp
LukasBanana: I would simply add another boolean member to `msl_options` then, like `bool enable_base_index_zero`

spirv_common.hpp
LukasBanana: I remember this check being an issue during the rebase. It seems to be fine now.

spirv_cross.cpp
LukasBanana: The generated Metal code of the test shader `shaders-msl/asm/frag/depth-compare.asm.frag` will fail to compile without this change.
A `depthcube<float>` texture will be sampled with `sample_compare` and also with `sample`. For both sampling functions the return value is `float` and not `float4` because its a depth texture even though it's sampled with a non-comparison sample function in the second case.
This case (and the above section) fixes that issue. Otherwise a `.x` suffix is added, which is not allowed for floating point types.

spirv_msl.cpp
LukasBanana: This is the behavior as it was before the change and as the comment above the if-statement implies, this subtraction is handled elsewhere now, i.e. in `CompilerMSL::builtin_to_glsl` to be precise.
The upper condition is to avoid duplicate subtraction like:
`gl_VertexID - gl_BaseVertex - gl_BaseVertex`

spirv_cross.cpp
HansKristian-Work: That should already work, but I guess you found some edge cases. I'll check it out in more detail.

spirv_glsl.hpp
LukasBanana: You might be right with that, too. I'm still doing internal testing which takes some time.

spirv_msl.cpp
LukasBanana: The first update reduced the amount of changed files to 436

spirv_msl.cpp
LukasBanana: Replaced this by `bool enable_base_index_zero` in `msl_options`.

spirv_msl.cpp
LukasBanana: Removed this part. Seems to work without this change.

spirv_msl.cpp
LukasBanana: I updated all touched source files with clang-format

spirv_msl.cpp
LukasBanana: I reverted static inlining for all template functions.

spirv_msl.cpp
LukasBanana: This is also removed in the first update.

spirv_common.hpp
LukasBanana: I re-enabled this check with the first update.

reference/opt/shaders-hlsl/comp/access-chains.comp
```diff
@@ -1,5 +1,3 @@
-static const uint3 gl_WorkGroupSize = uint3(1u, 1u, 1u);
```
LukasBanana: @HansKristian-Work Could it be that these test shaders haven't been updated in a while? Because when I run `update_test_shaders.sh` with a recent KhronosGroup/SPIRV-Cross clone that I have locally (from Sep. 5th), all `(1u, 1u, 1u)` and `uint3(1u)` variants of `gl_WorkGroupSize` disappear just like this with PR.

spirv_msl.cpp
LukasBanana: I simplified this overriden function to also make use of `to_texture_op`, but still preserve the use of the new option `ios_use_framebuffer_fetch_subpasses`.

spirv_msl.cpp
LukasBanana: It was used in `CompilerMSL::emit_texture_op` but after I simplified this function to use `to_texture_op` I reverted it.

reference/opt/shaders-hlsl/comp/access-chains.comp
HansKristian-Work: The dependencies were recently updated. Try ./checkout_glslang_spirv_tools.sh and build new deps.

HansKristian-Work:
Thanks! I'll check out the updates tomorrow morning.

spirv_cross.cpp
HansKristian-Work: Hm, I see no difference running master and this PR on that file both with and without optimization. I'll try to create an isolated reproducer, because I don't really understand this particular change.

HansKristian-Work: Ok, have a reproducer which this PR partially fixes:
```
#version 450

layout(set = 0, binding = 0) uniform texture2D uTexture;
layout(set = 0, binding = 1) uniform sampler uSampler;
layout(set = 0, binding = 2) uniform samplerShadow uSamplerShadow;

layout(location = 0) out float FragColor;
layout(location = 0) in vec3 vUV;

float sample_normal2(texture2D tex)
{
	return texture(sampler2D(tex, uSampler), vUV.xy).x;
}

float sample_normal(texture2D tex)
{
	return sample_normal2(tex);
}

float sample_comp(texture2D tex)
{
	return texture(sampler2DShadow(tex, uSamplerShadow), vUV);
}

void main()
{
	FragColor = sample_normal(uTexture);
	FragColor += sample_comp(uTexture);
}
```
The generated code becomes:
```

#include <metal_stdlib>
#include <simd/simd.h>

using namespace metal;

struct main0_out
{
    float FragColor [[color(0)]];
};

struct main0_in
{
    float3 vUV [[user(locn0)]];
};

static inline __attribute__((always_inline))
float sample_normal2(thread const texture2d<float> tex, thread sampler uSampler, thread float3& vUV)
{
    return tex.sample(uSampler, vUV.xy).x;
}

static inline __attribute__((always_inline))
float sample_normal(thread const depth2d<float> tex, thread sampler uSampler, thread float3& vUV)
{
    return sample_normal2(tex, uSampler, vUV);
}

static inline __attribute__((always_inline))
float sample_comp(thread const depth2d<float> tex, thread float3& vUV, thread sampler uSamplerShadow)
{
    return tex.sample_compare(uSamplerShadow, vUV.xy, vUV.z);
}

fragment main0_out main0(main0_in in [[stage_in]], depth2d<float> uTexture [[texture(0)]], sampler uSampler [[sampler(0)]], sampler uSamplerShadow [[sampler(1)]])
{
    main0_out out = {};
    out.FragColor = sample_normal(uTexture, uSampler, in.vUV);
    out.FragColor += sample_comp(uTexture, in.vUV, uSamplerShadow);
    return out;
}
```
The parameters have to end up as depth2d as well, but we only get one link of the hierarchy resolved. I'll try to think of something here ...

spirv_cross.hpp
```diff
@@ -487,6 +487,9 @@ class Compiler
 	// The most common use here is to check if a buffer is readonly or writeonly.
 	Bitset get_buffer_block_flags(VariableID id) const;
 
+	// Returns true if the target language supports combined texture-samplers. Returns fasle by default.
+	virtual bool supports_combined_samplers() const;
```
HansKristian-Work: Should be protected.

spirv_glsl.cpp
HansKristian-Work: Alright I see now what's going on. The struct is empty, so subconstants is empty, so it falls down to the constant_expression_vector path. Please move this path to CompilerGLSL::constant_expression as an else if which would be clearer.

spirv_msl.hpp
```diff
@@ -279,10 +280,30 @@ class CompilerMSL : public CompilerGLSL
...
+
 		// Fragment output in MSL must have at least as many components as the render pass.
 		// Add support to explicit pad out components.
 		bool pad_fragment_output_components = false;
 
+		// Specifies whether the iOS target version supports the [[base_vertex]] and [[base_instance]] attributes.
+		bool ios_support_base_vertex_instance = false;
+
+		// Use Metal's native frame-buffer fetch API for subpass inputs.
+		bool ios_use_framebuffer_fetch_subpasses = true;
```
HansKristian-Work: Make this false by default. This might break MoltenVK.

spirv_msl.cpp
```diff
@@ -5485,14 +5864,30 @@ void CompilerMSL::emit_instruction(const Instruction &instruction)
...
 		for (uint32_t col = 0; col < type.columns; col++)
 		{
-			expr += to_enclosed_expression(a);
-			expr += " * ";
-			expr += to_extract_component_expression(b, col);
+			if (msl_options.invariant_float_math && type_a && type_b &&
+			    ((is_matrix(*type_a) && is_matrix(*type_b)) || (is_matrix(*type_a) && is_vector(*type_b)) ||
+			     (is_vector(*type_a) && is_matrix(*type_b))))
+			{
+				expr += "spvFMulMatrixMatrix(";
```
HansKristian-Work: Is this correct? MatrixMatrix implies 4x4 matrix multiply, this is a vector times scalar thing.

spirv_msl.cpp
```diff
@@ -5503,6 +5898,89 @@ void CompilerMSL::emit_instruction(const Instruction &instruction)
...
+		// If the matrix needs transpose, just flip the multiply order.
+		auto *e = maybe_get<SPIRExpression>(ops[opcode == OpMatrixTimesVector ? 2 : 3]);
+		if (e && e->need_transpose)
+		{
+			e->need_transpose = false;
+			string expr;
+
+			if (opcode == OpMatrixTimesVector)
+			{
+				expr = join(to_enclosed_unpacked_expression(ops[3]), " * ",
```
HansKristian-Work: invariant float math is not handled here.

HansKristian-Work: Also, no need to enclose expressions which go into functions, this is only used for binary ops like `*`, e.g. `1 + 2 * 3 + 4` might be fixed up to `(1 + 2) * (3 + 4)` but `foo(1 + 2, 3 + 4)` is fine.

spirv_msl.cpp
HansKristian-Work: `shaders-msl/asm/frag/frame-buffer-fetch-subpass-input.asm.frag` is not tagged with `.ios.` so it is compiled as macOS, and the test is a bit moot. See `cross_compile_msl` in `test_shaders.py` how arguments are passed to the CLI tool when running tests.

I think this function can be simplified quite a bit. There is no need to inherit expression dependencies (it's just an input value) and control dependent expressions (no LOD). Probably best to do something like `if (frame buffer fetch case) { do_thing; } else { CompilerGLSL::emit_texture_op(i); }`

spirv_msl.cpp
```diff
@@ -6599,18 +7152,33 @@ string CompilerMSL::to_function_args(VariableID img, const SPIRType &imgtype, bo
...
 		alt_coord_component = 1;
 		break;
 
 	case DimSubpassData:
-		if (imgtype.image.ms)
-			tex_coords = "uint2(gl_FragCoord.xy)";
-		else
-			tex_coords = join("uint2(gl_FragCoord.xy), 0");
+		// Use Metal's native frame-buffer fetch API for subpass inputs.
+		if (!msl_options.is_ios() || !msl_options.ios_use_framebuffer_fetch_subpasses)
```
HansKristian-Work: This path shouldn't even be hit with the emit_texture_op override.

spirv_msl.cpp
```diff
@@ -7435,7 +8034,7 @@ string CompilerMSL::to_struct_member(const SPIRType &type, uint32_t member_type_
 	SPIRType row_major_physical_type;
 	const SPIRType *declared_type = &physical_type;
 
-	if (member_is_packed_physical_type(type, index))
+	if (member_is_packed_physical_type(type, index) && validate_member_packing_rules_msl(type, index))
```
HansKristian-Work: Why validate member packing rules again? This should be done already by the time we come here.

HansKristian-Work: There is a regression here:
```
#version 450
#extension GL_EXT_scalar_block_layout : require

layout(set = 0, binding = 0, scalar) buffer SSBO
{
	vec3 v[16];
};

void main()
{
	v[1] = v[0];
}
```

ends up as:

```
#pragma clang diagnostic ignored "-Wmissing-braces"

#include <metal_stdlib>
#include <simd/simd.h>

using namespace metal;

template<typename T, size_t Num>
struct spvUnsafeArray
{
    T elements[Num ? Num : 1];
    
    thread T& operator [] (size_t pos) thread
    {
        return elements[pos];
    }
    constexpr const thread T& operator [] (size_t pos) const thread
    {
        return elements[pos];
    }
    
    device T& operator [] (size_t pos) device
    {
        return elements[pos];
    }
    constexpr const device T& operator [] (size_t pos) const device
    {
        return elements[pos];
    }
    
    constexpr const constant T& operator [] (size_t pos) const constant
    {
        return elements[pos];
    }
    
    threadgroup T& operator [] (size_t pos) threadgroup
    {
        return elements[pos];
    }
    constexpr const threadgroup T& operator [] (size_t pos) const threadgroup
    {
        return elements[pos];
    }
};

struct SSBO
{
    packed_spvUnsafeArray<float3, 16> v;
};

kernel void main0(device SSBO& _13 [[buffer(0)]])
{
    spvUnsafeArray<float3, 16>(_13.v[1]) = float3(spvUnsafeArray<float3, 16>(_13.v[0]));
}
```
it should be using `packed_float3 v[16]` instead. There's a lot of packing tests in `shaders-msl-no-opt/packing/`.

spirv_msl.cpp
```diff
@@ -7505,7 +8123,10 @@ void CompilerMSL::emit_struct_member(const SPIRType &type, uint32_t member_type_
 		statement("char _m", index, "_pad", "[", pad_len, "];");
 	}
 
+	// Handle HLSL-style 0-based vertex/instance index.
+	builtin_declaration = true;
```
HansKristian-Work: Is vertex indices emitted in structs?

spirv_msl.cpp
```diff
@@ -9909,23 +10733,114 @@ string CompilerMSL::builtin_to_glsl(BuiltIn builtin, StorageClass storage)
...
-		return "gl_InstanceIndex";
-	case BuiltInBaseVertex:
-		return "gl_BaseVertex";
-	case BuiltInBaseInstance:
-		return "gl_BaseInstance";
-	case BuiltInDrawIndex:
-		SPIRV_CROSS_THROW("DrawIndex is not supported in MSL.");
-
-	// When used in the entry function, output builtins are qualified with output struct name.
+		ensure_builtin(StorageClassInput, BuiltInVertexId);
```
HansKristian-Work: Considered adding a fixup lambda to the entry point which just `-= gl_BaseVertex` the VertexID rather than subtracting every time it's used?

spirv_msl.cpp
```diff
@@ -10655,6 +11645,55 @@ bool CompilerMSL::OpCodePreprocessor::handle(Op opcode, const uint32_t *args, ui
 		uint32_t result_type = args[0];
 		uint32_t id = args[1];
 		uint32_t ptr = args[2];
+
+		// Fix tessellation patch function processing
+		if (compiler.get_execution_model() == ExecutionModelTessellationControl)
```
HansKristian-Work: I don't understand what this is trying to accomplish.

spirv_msl.cpp
```diff
@@ -10682,68 +11721,541 @@ void CompilerMSL::OpCodePreprocessor::check_resource_write(uint32_t var_id)
 		uses_resource_write = true;
 }
 
-// Returns an enumeration of a SPIR-V function that needs to be output for certain Op codes.
-CompilerMSL::SPVFuncImpl CompilerMSL::OpCodePreprocessor::get_spv_func_impl(Op opcode, const uint32_t *args)
+// Storage buffer robustness
+// Fix loads from tessellation control inputs not being forwarded to the gl_in structure array
+// Fix loads from tessellation evaluation inputs not being forwarded to the stage_in structure array
+std::string CompilerMSL::access_chain_internal(uint32_t base, const uint32_t *indices, uint32_t count,
```
HansKristian-Work: Does this function have to override the entire access_chain_internal from GLSL? Consider adding just a few virtuals to deal with robustness if necessary.

spirv_msl.cpp
```diff
@@ -11192,7 +12724,9 @@ void CompilerMSL::analyze_argument_buffers()
 				auto &var = set<SPIRVariable>(var_id, uint_ptr_type_id, StorageClassUniformConstant);
 				set_name(var_id, "spvBufferSizeConstants");
 				set_decoration(var_id, DecorationDescriptorSet, desc_set);
-				set_decoration(var_id, DecorationBinding, kBufferSizeBufferBinding);
+				/* UE Begin Change: Move this to the front of IABs for convenience of the runtime */
```
HansKristian-Work: hm?

spirv_msl.cpp
```diff
@@ -10755,6 +12267,26 @@ CompilerMSL::SPVFuncImpl CompilerMSL::OpCodePreprocessor::get_spv_func_impl(Op o
...
 		if (tid && compiler.get<SPIRType>(tid).image.dim == DimBuffer && !compiler.msl_options.texture_buffer_native)
 			return SPVFuncImplTexelBufferCoords;
+
+		if (opcode == OpImageFetch && compiler.msl_options.swizzle_texture_samples)
+			return SPVFuncImplTextureSwizzle;
+
+		break;
+	}
+
+	case OpImageSampleExplicitLod:
```
HansKristian-Work: These are already handled I think.

spirv_msl.cpp
HansKristian-Work: The add_msl_resource_binding API lets you rebind these things to an appropriate `[[id()]]` you'd like, even for these internally created resources. Please remove this hardcoding.

spirv_msl.cpp
```diff
@@ -8193,6 +8824,9 @@ string CompilerMSL::entry_point_args_argument_buffer(bool append_comma)
 				buffer_binding = i;
 		}
 
+		// Allow the caller to specify an offset for argument buffer binding slots
+		buffer_binding += msl_options.argument_buffer_offset;
```
HansKristian-Work: This is actually kinda redundant. You can use add_msl_resource_binding to pick where IABs themselves are placed.

HansKristian-Work:
Seeing a warning:
```
spirv_msl.cpp: In member function ‘virtual std::string spirv_cross::CompilerMSL::image_type_glsl(const spirv_cross::SPIRType&, uint32_t)’:
spirv_msl.cpp:10400:4: warning: this statement may fall through [-Wimplicit-fallthrough=]
10400 |    if (msl_options.is_ios() && msl_options.ios_use_framebuffer_fetch_subpasses)
      |    ^~
spirv_msl.cpp:10402:3: note: here
10402 |   case Dim2D:
```

spirv_cross.cpp
LukasBanana: Great, thanks for the effort. I'd be happy to see a simpler solution that doesn't have O(n^2) complexity as you pointed out.

HansKristian-Work: Now, I could implement this on master tomorrow and you can rebase against that, but if rebasing is too hard, I can do a trampoline branch where I merge your branch, add my fix on top and push that to master. Which method do you prefer?

LukasBanana:
> Seeing a warning:
> 
> ```
> spirv_msl.cpp: In member function ‘virtual std::string spirv_cross::CompilerMSL::image_type_glsl(const spirv_cross::SPIRType&, uint32_t)’:
> spirv_msl.cpp:10400:4: warning: this statement may fall through [-Wimplicit-fallthrough=]
> 10400 |    if (msl_options.is_ios() && msl_options.ios_use_framebuffer_fetch_subpasses)
>       |    ^~
> spirv_msl.cpp:10402:3: note: here
> 10402 |   case Dim2D:
> ```

The case fallthrough is intended, but the `[[fallthrough]]` specifier was first introduced in C++17. What do you suggest for this case?

HansKristian-Work:
Writing it without fallthrough in that case, merge the SubpassInput/2D case labels and check for subpass input inside the case perhaps. CI builds with Werror, and only C++11 should be required.

marksatt-pitbull:
I've signed the CLA for you all now. Note that I am no longer with Epic and thus can't contribute any more code but I might be able to answer questions.

With regard to the changes; in general I made whatever modifications were necessary to make UE4 shaders work. I wasn't initially concerned with how that might affect any other project. As such there are some changes which are incomplete, or in light of work from upstream possible even outright wrong. I was up against a deadline to get it to work - and it was compiling all the Mac & iOS shaders I could throw at it when I left.

The packing changes are likely to be one of the most likely to be wrong - every now and then a UE4 shader packs something in ways that don't translate neatly.

There are some really important tessellation fixes - one temporary array is declared as a regular variable when it should be threadlocal storage for example - but I doubt the changes were made in the ideal way.

I also implemented texture atomic emulation which is required for UE4 by binding a buffer alongside texture_buffer/texture2d objects on which atomics are used. This requires the runtime to allocate & bind appropriately which UE4 already does. Unfortunately I didn't get to make this work properly with IABs. I only implemented experimental IAB support in UE4's Metal backend and didn't have any shaders that used texture-atomics on a resource within an IAB so it got left as a TODO. Epic will need to maintain and iterate on this work until it can be accepted upstream as its the only way to implement this shader feature in Metal.

LukasBanana:
Thanks a lot @marksatt-pitbull, I'll continue with the necessary changes and will consider to remove a few things temporarily just for the pull request to be able to work on it later, like the storage buffer bound checks for instance as this might take a bit more time to work using SPIRV-Tools in UE4.

HansKristian-Work:
Thanks for the changes. I'll be traveling the rest of the week, but I'll try to look through changes when I have some downtime.

spirv_msl.cpp
LukasBanana: This was either a sanity check or simply a leftover before `OpVectorTimesMatrix` differentiation. I removed that now and also added a new test case as well as `--msl-invariant-float-math` compiler option.

LukasBanana:
No worries, I pushed to simplify my workflow to merge the latest changes in this repository. I'm still working on some other changes. I'll keep you posted when I (think) am done :)

spirv_cross.cpp
LukasBanana: I'm sorry I missed your reply here. If you have a better solution I'd be happy to replace it. Either way works for me. Since you mentioned you might want to make a few adjustments to this PR yourself, a side branch might be a good way to handle it.

spirv_glsl.cpp
LukasBanana: Done in [spirv_glsl.cpp:3440](https://github.com/KhronosGroup/SPIRV-Cross/pull/1162/files#diff-09b7cdf207013ada0b4e38970bc27158R3440)

spirv_msl.cpp
LukasBanana: Removed the `enclose_expression` for one of the parameters as the other one uses `to_enclosed_unpacked_expression`.

spirv_msl.cpp
LukasBanana: You're right. I added two tests to that folder:
- `array-of-vec3.comp` which is basically your example
- `array-of-vec4.comp` to verify `spvUnsafeArray` is still used when packing is not necessary.

spirv_msl.cpp
LukasBanana: I replaced that by overriding only the [`append_index` function](https://github.com/KhronosGroup/SPIRV-Cross/pull/1162/files#diff-6a4029d3c45d50818e6f80e915867c64R11738).

HansKristian-Work:
One question, when I try to go through the changed reference shaders, I want to isolate myself to the existing shaders. Would it be possible to hoist out the new UE4 test shaders to its own folder? I.e. shaders-ue4 and reference/shaders-ue4. These shaders aren't really maintainable, and their main purpose is there to just make sure we don't break the shaders.

reference/shaders-msl-no-opt/packing/matrix-3x2-scalar.comp
```diff
@@ -5,7 +5,7 @@
 
 using namespace metal;
 
-typedef packed_float3 packed_rm_float3x2[2];
+typedef float2x3 packed_rm_float3x2;
```
HansKristian-Work: Still wrong here with packing.

spirv_cross.hpp
```diff
@@ -612,6 +612,9 @@ class Compiler
 	void register_read(uint32_t expr, uint32_t chain, bool forwarded);
 	void register_write(uint32_t chain);
 
+	// Returns true if the target language supports combined texture-samplers. Returns fasle by default.
+	virtual bool supports_combined_samplers() const;
```
HansKristian-Work: This needs to be moved to spirv_glsl.hpp, virtual backend methods do it belong here.

spirv_msl.hpp
```diff
@@ -266,6 +266,7 @@ class CompilerMSL : public CompilerGLSL
 		uint32_t dynamic_offsets_buffer_index = 23;
 		uint32_t shader_input_wg_index = 0;
 		uint32_t device_index = 0;
+		uint32_t argument_buffer_offset = 0; // Allow the caller to specify an offset for argument buffer binding slots
```
HansKristian-Work: This should not be needed. Are you using the set_msl_resource_binding API at all?

reference/shaders-msl-no-opt/packing/matrix-3x2-scalar.comp
HansKristian-Work: Overall, would it be okay to simply remove any use of spvUnsafeArray for any buffer-related arrays?

HansKristian-Work:
The test suite is now only failing on Windows, I'm going to guess that it has something to do with non-deterministic order from unordered_set/map somewhere. Any ideas?

spirv_msl.cpp
HansKristian-Work: This still looks like a hack.

spirv_msl.cpp
```diff
@@ -10664,6 +11615,42 @@ bool CompilerMSL::OpCodePreprocessor::handle(Op opcode, const uint32_t *args, ui
...
+					break;
+				}
+			}
+		}
+		break;
+	}
+
+	case OpControlBarrier:
+	{
+		passed_control_barrier = true;
```
HansKristian-Work: Trying to understand this part of the PR now ...
I believe this one is trying to detect when the shader has entered its "patch constant" phase.

spirv_msl.cpp
```diff
@@ -10672,6 +11659,55 @@ bool CompilerMSL::OpCodePreprocessor::handle(Op opcode, const uint32_t *args, ui
...
+						break;
+					}
+				}
+			}
+
+			if (is_indexed_by_invocation)
+			{
+				if (passed_control_barrier)
+				{
+					auto *var = compiler.maybe_get_backing_variable(ptr);
```
HansKristian-Work: Ok, so I have a feeling this is an unnecessary hack for invalid SPIR-V. Looking at `shaders-msl/asm/tesc/tess-factor-must-be-threadgroup.asm.tesc`.

GLSL output is:
```
...
    FFlatTessellationHSToDS temp_var_hullMainRetVal[3]; // <-- Huh?
    temp_var_hullMainRetVal[gl_InvocationID] = FFlatTessellationHSToDS(param_var_I[gl_InvocationID], _125, Material.Material_ScalarExpressions[0].x, 1.0);
    barrier();
    if (gl_InvocationID == 0u)
    {
        vec4 _151 = _85;
        _151.x = 0.5 * (temp_var_hullMainRetVal[1u].TessellationMultiplier + temp_var_hullMainRetVal[2u].TessellationMultiplier);
        vec4 _157 = _151;
        _157.y = 0.5 * (temp_var_hullMainRetVal[2u].TessellationMultiplier + temp_var_hullMainRetVal[0u].TessellationMultiplier);
        vec4 _162 = _157;
        _162.z = 0.5 * (temp_var_hullMainRetVal[0u].TessellationMultiplier + temp_var_hullMainRetVal[1u].TessellationMultiplier);
        vec4 _169 = _162;
        _169.w = 0.333000004291534423828125 * ((temp_var_hullMainRetVal[0u].TessellationMultiplier + temp_var_hullMainRetVal[1u].TessellationMultiplier) + temp_var_hullMainRetVal[2u].TessellationMultiplier);
...
```
This is the kind of thing the MSL patch will check for, and that array is forced into a groupshared array instead, but the SPIR-V is just wrong. For the other cases where patch constant phase reads from Output variables, they are already in device memory, not threadgroup, and this "fix" actually breaks those shaders.

This is very likely a DXC/glslang bug if anything.

HansKristian-Work: Consider a reduced case for GLSL which does something similar:

```
#version 450

layout(vertices = 2) out;

layout(location = 0) patch out vec4 vFoo;
layout(location = 1) out vec4 vOutput[2];

void main()
{
	vOutput[gl_InvocationID] = vec4(1.0);
	barrier();
	if (gl_InvocationID == 0)
	{
		vFoo = vOutput[0] + vOutput[1];
		gl_TessLevelInner[0] = 1.0;
		gl_TessLevelOuter[1] = 2.0;
	}
}
```

This is now converted into:
```
#include <metal_stdlib>
#include <simd/simd.h>

using namespace metal;

struct main0_patchOut
{
    float4 vFoo;
};

kernel void main0(uint gl_InvocationID [[thread_index_in_threadgroup]], uint gl_PrimitiveID [[threadgroup_position_in_grid]], constant uint* spvIndirectParams [[buffer(29)]], device main0_patchOut* spvPatchOut [[buffer(27)]], device MTLQuadTessellationFactorsHalf* spvTessLevel [[buffer(26)]])
{
    threadgroup float4 vOutput[2]; // <-- Wrong.
    device main0_patchOut& patchOut = spvPatchOut[gl_PrimitiveID];
    vOutput[gl_InvocationID] = float4(1.0);
    threadgroup_barrier(mem_flags::mem_device | mem_flags::mem_threadgroup);
    if (gl_InvocationID == 0)
    {
        patchOut.vFoo = vOutput[0] + vOutput[1];
        spvTessLevel[gl_PrimitiveID].insideTessellationFactor[0] = half(1.0);
        spvTessLevel[gl_PrimitiveID].edgeTessellationFactor[1] = half(2.0);
    }
}
```

spirv_msl.cpp
```diff
@@ -10699,6 +11735,97 @@ void CompilerMSL::OpCodePreprocessor::check_resource_write(uint32_t var_id)
 		uses_resource_write = true;
 }
 
+void CompilerMSL::access_chain_internal_append_index(std::string &expr, uint32_t base, const SPIRType *type,
+                                                     AccessChainFlags flags, bool &access_chain_is_arrayed,
+                                                     uint32_t index)
+{
+	bool index_is_literal = (flags & ACCESS_CHAIN_INDEX_IS_LITERAL_BIT) != 0;
+	bool register_expression_read = (flags & ACCESS_CHAIN_SKIP_REGISTER_EXPRESSION_READ_BIT) == 0;
```
HansKristian-Work: I think I figured out what the purpose of this function is. The case mostly happens when trying to OpLoad an array-of-array of inputs in TESE. This happens when TESC emits an array per-control point.

This function then handles the interleave-like operation which happens when trying to index into the loaded array-of-array with CompositeExtract.

This smells rather fishy, because input arrays should be unrolled already. Perhaps this is a special case which is only needed for array-of-array because I believe there are some missing cases for the current input unrolling code.

I'd rather fix that problem than adding a crude hack like this. The comment below also suggests this is a workaround.

HansKristian-Work: I'll make a separate issue out of this, and try to fix it in a cleaner way, then, this override should be possible to remove altogether.

marksatt-pitbull:
That definitely shows that something is wrong. In GLSL the temp_var_hullMainRetVal variable should be an output which in Metal must become a thread group variable (making it a 'device' variable would require yet another temporary allocation and binding - both of which are undesirable).

I openly admitted to doing what was needed to get UE4 running without regard for whether it was ideal or perfect and this is an example. I worked around the problem here in order to keep moving forward. Lukas didn't make these changes.

The source of this issue will be in DXC and/or SPIRVTools - IIRC this particular variable is created by the function inlining pass as DXC restructures the Hull shader to be invoked N times within the main() body.

Mark

________________________________

marksatt-pitbull:
The problem is that DXC and SPIRVTools are intentionally ignorant of the way Metal restructures input and output data for Hull and Domain shaders, so whatever they do there will need to be some fairly complex code in the Metal backend to rewire the independent arrays of SPIRV into an array of structures, with additional special patch templates for Metal. Therefore I just fixed it up in the access chain code as that was by the easiest way for me at the time. I just didn't know what or where the correct solution would be

Mark

________________________________

reference/shaders-msl-no-opt/packing/matrix-3x2-scalar.comp
LukasBanana: Sorry about that, I fixed that for the next commit.

spirv_cross.hpp
LukasBanana: This function is used by `Compiler::CombinedImageSamplerUsageHandler::add_hierarchy_to_comparison_ids` and a few others in `spirv_cross.cpp`. What do you recommend to move it outside of the `Compiler` class?

spirv_msl.hpp
LukasBanana: I think the `add_msl_resource_binding` API isn't used in our pipeline yet, maybe because it was added after the changes for UE4 were made. I removed this parameter and will add your recommended API on our side.

LukasBanana:
@HansKristian-Work Regarding your question about a separated folder for UE4 test shaders: I'm waiting for approval from our legal department. As soon as I get the okay, I'll take care of that.

@marksatt-pitbull Thank you for clarification and your time to get back on this topic.

> I'd rather fix that problem than adding a crude hack like this. The comment below also suggests this is a workaround.

I agree that it is desirable to have a proper fix for that, but as Mark pointed out, this is very likely an issue in DXC or SPIRV-Tools. As long as the workaround does not break the test bench I would prefer to keep the workaround in there.
If you insist to keep this workaround out, I can remove it and keep it in our internal repository.

HansKristian-Work:
I would prefer to keep a workaround for bad SPIR-V out as much as possible. The tessellation code is at the breaking point of complexity already.

HansKristian-Work:
> The problem is that DXC and SPIRVTools are intentionally ignorant of the way Metal restructures input and output data for Hull and Domain shaders, so whatever they do there will need to be some fairly complex code in the Metal backend to rewire the independent arrays of SPIRV into an array of structures, with additional special patch templates for Metal. Therefore I just fixed it up in the access chain code as that was by the easiest way for me at the time. I just didn't know what or where the correct solution would be Mark
> ________________________________ From: Hans-Kristian Arntzen <notifications@github.com> Sent: Wednesday, 23 October 2019, 11:13 To: KhronosGroup/SPIRV-Cross Cc: Mark Satterthwaite; Mention Subject: Re: [KhronosGroup/SPIRV-Cross] Lots of changes and fixes with new test cases to make SPIRV-Cross work in UE4 for Metal (#1162) @HansKristian-Work commented on this pull request.
> ________________________________ In spirv_msl.cpp<https://eur04.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2FKhronosGroup%2FSPIRV-Cross%2Fpull%2F1162%23discussion_r338111274&data=02%7C01%7C%7Cc0beff7163814146a7b308d757cb9f83%7C84df9e7fe9f640afb435aaaaaaaaaaaa%7C1%7C0%7C637074404370884269&sdata=b58bWvyBJF5X3pDARTKWeejeQlHo7kc4AHuyom11%2BLY%3D&reserved=0>:
> @@ -10699,6 +11735,97 @@ void CompilerMSL::OpCodePreprocessor::check_resource_write(uint32_t var_id)
> uses_resource_write = true; } +void CompilerMSL::access_chain_internal_append_index(std::string &expr, uint32_t base, const SPIRType *type, + AccessChainFlags flags, bool &access_chain_is_arrayed, + uint32_t index) +{ + bool index_is_literal = (flags & ACCESS_CHAIN_INDEX_IS_LITERAL_BIT) != 0; + bool register_expression_read = (flags & ACCESS_CHAIN_SKIP_REGISTER_EXPRESSION_READ_BIT) == 0; I think I figured out what the purpose of this function is. The case mostly happens when trying to OpLoad an array-of-array of inputs in TESE. This happens when TESC emits an array per-control point. This function then handles the interleave-like operation which happens when trying to index into the loaded array-of-array with CompositeExtract. This smells rather fishy, because input arrays should be unrolled already. Perhaps this is a special case which is only needed for array-of-array because I believe there are some missing cases for the current input unrolling code. I'd rather fix that problem than adding a crude hack like this. The comment below also suggests this is a workaround. — You are receiving this because you were mentioned. Reply to this email directly, view it on GitHub<https://eur04.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2FKhronosGroup%2FSPIRV-Cross%2Fpull%2F1162%3Femail_source%3Dnotifications%26email_token%3DABXDN4S5MIY6747PJH46RS3QQBS3HA5CNFSM4IYZF63KYY3PNVWWK3TUL52HS4DFWFIHK3DMKJSXC5LFON2FEZLWNFSXPKTDN5WW2ZLOORPWSZGOCI6MSSQ%23pullrequestreview-305973578&data=02%7C01%7C%7Cc0beff7163814146a7b308d757cb9f83%7C84df9e7fe9f640afb435aaaaaaaaaaaa%7C1%7C0%7C637074404370884269&sdata=GM0zItSyHpqXJukCUnaF5OLpDfStDz2kpFn3zzGUsbg%3D&reserved=0>, or unsubscribe<https://eur04.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Fnotifications%2Funsubscribe-auth%2FABXDN4WD6YRXEPXDA2V6OPDQQBS3HANCNFSM4IYZF63A&data=02%7C01%7C%7Cc0beff7163814146a7b308d757cb9f83%7C84df9e7fe9f640afb435aaaaaaaaaaaa%7C1%7C0%7C637074404370894278&sdata=W%2BXW%2FoKTEpeg8URFOfz6QH21Db9BFW9BXdKYbUyUOE8%3D&reserved=0>.

If SPIRV-Tools inlines a variable like that, shouldn't it become either Workgroup or Output storage class? Right now the variable is declared as Function, and there is no way for that to be sharable across threads. Output variables should be shareable as they are placed in `device spvOut*` already.

spirv_cross.hpp
HansKristian-Work: In that case, it's best to just pass a bool into CombinedImageSamplerUsageHandler if it's supported or not. The base class should not know about backend specifics.

reference/shaders-msl-no-opt/packing/matrix-3x2-scalar.comp
HansKristian-Work: I mean arrays which are declared inside buffers: `struct UBO { float4 arr[16]; }; kernel void main(constant UBO &ubo) {}`.

The struct type has "Offset" decorations if it's used as a "buffer-like" type, so I think that might be usable.

From what I understand, spvUnsafeArray is only really used for variables outside buffer objects to improve copying arrays around as value-types, but copying array types from buffers doesn't really happen afaik. At any rate, it's not well tested.

If there is a reliable way to get `spvUnsafeArray<packed_float3, N>` emitted instead without too much hackery, that might be an option, but it doesn't seem easy.

marksatt-pitbull:
I expect this variable will be function scope because it is the return result from a function call that is invoked inside a loop setup by DXC like:

main() {

  T result[]
  For (NumCPs) {
    result = HullShader()
 }

  PatchFunction(result)
}

For whatever reason either DXC isn't marking it as output or it gets lost along the way.

Once it goes into Metal it shouldn't be a device variable as that would require the runtime to allocate and bind yet another buffer that is not present in the HLSL which is both dramatically slower and more complicated than pushing this temporary result into thread group storage - which is effectively what it is in HLSL/GLSL.

Mark
________________________________

HansKristian-Work:
Right, so the correct solution is for somewhere in DXC/SPIRV-Tools to turn that Function variable which is used as per-CP output into Workgroup storage instead, and that appears to be what the workaround is doing. Has this bug been reported somewhere?

However, if the per-CP output is actually part of the output interface, it should be Output, and placed in the `device spvOut *` struct. Which has to be allocated anyways. There is no "special side-allocation" required at least.

I'm a little confused by the loop there though ... Is it an actual loop running in single thread, even when there are multiple output control points? In that case, using Workgroup instead of Function doesn't make much sense ...

Thanks for pitching in by the way.

marksatt-pitbull:
Lukas and I didn't get to sorting things out to upstream my changes to DXC or SPIRVTools, I'm sure he will in time. This problem should be added to the list to address in DXC.

I apologise, I misspoke and confused the issue, the loop is conceptual, not actually part of the SPIRV. The HullShader code will use InvocationID to read/write the correct slot as it is executed NumCP times and the patch function only gets run for InvocationID == 0.

It'll look like:

output X hullOutput

main(... ) {

T result[NumCps]
result[InvocationID] = HullShader(... )

hullOutput = convertToX(result[InvocationID])

If InvocationID == 0 {
  PatchFunc(result,...)
}

------------------

I can't recall how hullOutput's type is expressed in SPIRV but I think it is different to that used by HLSL which is what 'T' will be. Thus DXC uses 'result' as a temporary storage buffer to simplify how data is passed around. It all needs someone to go back and look at the SPIRV emitted by DXC prior to legalization and optimisation so you can figure out what the *ideal* solution is.

Mark
________________________________

HansKristian-Work:
Thanks, I see why it would go wrong here. It lowers the CP output later to Output, but it's the intermediate, temporary output in Function which is consumed by per-patch output. That's the bug then. The better formulation in DXC would be:

```
out T2 out_results[NumCP];
T result = HullShader()
groupshared T results[NumCP];
results[InvocationID] = result;
out_results[InvocationID] = ConvT2(result);
barrier();
if (invocation == 0)
{
    PatchFunc(results);
}
```

HansKristian-Work:
I've started a local merge and clean-up branch: https://github.com/KhronosGroup/SPIRV-Cross/tree/ue4-merge.

HansKristian-Work:
Getting close ... The tessellation cleanup is taking a while.

LukasBanana:
Thank you very much for your support. Hopefully, future pull requests from our side will get a lot smaller.

HansKristian-Work:
Merged, phew. Please look over the changes I made. https://github.com/KhronosGroup/SPIRV-Cross/pull/1185/commits/1d5eba67f88cc23549584ef62964060cfab4b581 is the commit where I revert the patch processing hack for DXC, you might have to maintain that one yourself until DXC fixes the issue. The UE4 output changed a bit, but it still compiles, so I assume it's fine.

PR #1184, opened 2019-10-24
MSL: Support option for treating 1D textures as 2D textures of height 1.

billhollings:
Add `CompilerMSL::Options::texture_1D_as_2D`.

Metal imposes significant restrictions on 1D textures, including not being
renderable, clearable, or permitting mipmaps. This option allows SPIR-V 1D
textures to be treated as 2D textures to permit this additional behaviour.
App must of course supply the textures to Metal as 2D textures.

### DIFF
--- a/spirv_msl.cpp
+++ b/spirv_msl.cpp
@@ -6596,6 +6596,12 @@ string CompilerMSL::to_function_args(VariableID img, const SPIRType &imgtype, bo
 		else if (sampling_type_needs_f32_conversion(coord_type))
 			tex_coords = convert_to_f32(tex_coords, 1);
 
+		if (msl_options.texture_1D_as_2D)
+		{
+			string base_type = is_fetch ? "uint" : "float";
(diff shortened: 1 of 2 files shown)


spirv_msl.cpp
```diff
@@ -6596,6 +6596,12 @@ string CompilerMSL::to_function_args(VariableID img, const SPIRType &imgtype, bo
 		else if (sampling_type_needs_f32_conversion(coord_type))
 			tex_coords = convert_to_f32(tex_coords, 1);
 
+		if (msl_options.texture_1D_as_2D)
+		{
+			string base_type = is_fetch ? "uint" : "float";
+			tex_coords = base_type + "2(" + tex_coords + ", 0)";
```
HansKristian-Work: Needs to be y = 0.5 for float UV to avoid potential issues with sampling on edges.

billhollings: Good catch! Fixed now.

PR #1192, opened 2019-11-01
Added BUILD.gn file to be used by Chromium's ANGLE project

kakashidinho:
CC @null77

FYI: I'm an ANGLE project's contributor. I have no affiliation with Google.

### DIFF
--- a/gn/BUILD.gn
+++ b/gn/BUILD.gn
@@ -1,0 +1,81 @@
+# Copyright (C) 2019 Google, Inc.
+#
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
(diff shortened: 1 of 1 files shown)


gn/BUILD.gn
```diff
@@ -0,0 +1,81 @@
+# Copyright (C) 2019 Google, Inc.
+#
+# All rights reserved.
```
HansKristian-Work: There is a license mismatch here. This project uses Apache 2 as per Khronos CLA.

kakashidinho: done

kakashidinho:
@null77 Do you want to include the license in this file?
Note that I just copied the license comments from https://github.com/KhronosGroup/glslang/blob/master/BUILD.gn

PR #1197, opened 2019-11-05
Only run spirv-opt if the spirv is valid.

dj2:
This CL updates the test runner to only run spirv-opt if the generated
SPIR-V is valid. If validation is skipped it's possible to hit aborts
and other memory errors in the optimizer as it assumes the SPIR-V is
valid.


test_shaders.py
```diff
@@ -161,6 +161,7 @@ def validate_shader_msl(shader, opt):
         raise RuntimeError('Failed to compile Metal shader')
 
 def cross_compile_msl(shader, spirv, opt, iterations, paths):
+    print("Compile mMSL")
```
HansKristian-Work: Typo.

PR #1203, opened 2019-11-06
MSL: extract vector into swizzle for matrix

### LINKED ISSUE #1200: MSL: Incorrect convert about mat2
When converting mat2 in MSL, it says that we cannot constuct a vec<float,2) from vec<float, 4>, we find mat2 is stored as float2x4 because the MatrixStride is required to be 16.

For example:
```
#version 450

layout (set = 0, binding = 0) uniform vertexUniformBuffer1 {
    mat2 transform1;
};
layout (set = 1, binding = 0) uniform vertexUniformBuffer2{
    mat2 transform2;
};
void main() {
    const vec2 pos[3] = vec2[3](vec2(-1.f, -1.f), vec2(1.f, -1.f), vec2(-1.f, 1.f));
    gl_Position = vec4((transform1 + transform2) * pos[0], 0.f, 1.f);
}
```
MSL:
```
#include <metal_stdlib>
#include <simd/simd.h>

using namespace metal;

struct vertexUniformBuffer1
{
    float2x4 transform1;
};

struct vertexUniformBuffer2
{
    float2x4 transform2;
};

struct main0_out
{
};

vertex main0_out main0(constant vertexUniformBuffer1& _20 [[buffer(0)]], constant vertexUniformBuffer2& _26 [[buffer(1)]])
{
    main0_out out = {};
    out.gl_Position = float4(float2x2(float2(_20.transform1[0]) + float2(_26.transform2[0]), float2(_20.transform1[1]) + float2(_26.transform2[1])) * float2(-1.0), 0.0, 1.0);
    return out;
}
```

Here `_20.transform[0]` is a vector with 4 floats, but  we only need first two floats in our case. The vector in matrix need to be converted into the components following its actuall size if the size is smaller than its phyiscal size.

haoxli:
For mat2 in MSL, it will be stored as float2x4 because the MatrixStride is required to be 16. But we cannot constuct a vec<float,2) from vec<float, 4>, so extract it into swizzle if actual size is smaller than
its pyhiscal size.

Fix #1200

### DIFF
--- a/reference/shaders-msl/vert/matrix.swizzle.vert
+++ b/reference/shaders-msl/vert/matrix.swizzle.vert
@@ -1,0 +1,27 @@
+#include <metal_stdlib>
+#include <simd/simd.h>
+
+using namespace metal;
+
+struct vertexUniformBuffer1
+{
+    float2x4 transform1;
+};
+
+struct vertexUniformBuffer2
+{
+    float2x4 transform2;
+};
+
+struct main0_out
+{
+    float4 gl_Position [[position]];
+};
+
+vertex main0_out main0(constant vertexUniformBuffer1& _20 [[buffer(0)]], constant vertexUniformBuffer2& _26 [[buffer(1)]])
+{
+    main0_out out = {};
+    out.gl_Position = float4(float2x2(float2(_20.transform1[0].xy) + float2(_26.transform2[0].xy), float2(_20.transform1[1].xy) + float2(_26.transform2[1].xy)) * float2(-1.0), 0.0, 1.0);
+    return out;
+}
+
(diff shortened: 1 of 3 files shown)


haoxli:
@Kangz  @Jiawei-Shao

HansKristian-Work:
There is a problem here where the fixup is applied in the access chain which means it becomes an r-value expression too early. The correct place to put it would be in the unpack expression code. As the fix turned out to be rather simple, I went ahead and made a PR which supersedes this one. https://github.com/KhronosGroup/SPIRV-Cross/pull/1204/files. Thanks for the test case though!

Kangz:
Thanks for taking this on!

PR #1222, opened 2019-11-28
Fix sign handling for S/UToF.

HansKristian-Work:
Always deal with input casting when output is floating point here.


HansKristian-Work:
Trivial fix, tested locally.

PR #1227, opened 2019-12-03
Share test shader script

dj2:
This CL adds the ability to pass options to test_shader.sh. The
update_test_shaders.sh script is then changed to call test_shaders.sh
with the --update flag.

This removes the duplication of code between test_shaders.sh and
update_test_shaders.sh.

### DIFF
--- a/test_shaders.sh
+++ b/test_shaders.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+OPTS=$@
+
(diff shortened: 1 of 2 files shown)


dj2:
This seems to be doing the right thing on my Linux machine and should remove the possibility of forgetting to update one of the lists?

HansKristian-Work:
Seems okay. Should have done this earlier.

PR #1228, opened 2019-12-03
Fix MSVC warnings when building without exceptions.

### LINKED ISSUE #1225: Lot of warnings due to SPIRV_CROSS_THROW
Hi,

There are lot of warnings when using SPIRV_CROSS_THROW in methods returning a value, this warning appears in VisualStudio 2019, I could suspend the warning by defining the "/wd4715" but I think would be better to avoid warning supression.

I can submit PR in case.

HansKristian-Work:
Mark report_and_abort as noreturn.

Fix #1225.


HansKristian-Work:
Trivial, tested locally.

PR #1239, opened 2019-12-18
MSL: Support inline uniform blocks in argument buffers.

cdavis5e:
Here, the inline uniform block is explicit: we instantiate the buffer
block itself in the argument buffer, instead of a pointer to the buffer.
I just hope this will work with the `MTLArgumentDescriptor` API...

Note that Metal recursively assigns individual members of embedded
structs IDs. This means for automatic assignment that we have to
calculate the binding stride for a given buffer block. For MoltenVK,
we'll simply increment the ID by the size of the inline uniform block.
Then the later IDs will never conflict with the inline uniform block. We
can get away with this because Metal doesn't require that IDs be
contiguous, only monotonically increasing.

### DIFF
--- a/spirv_msl.cpp
+++ b/spirv_msl.cpp
@@ -9856,31 +9856,6 @@ void CompilerMSL::fix_up_shader_inputs_outputs()
 	});
 }
 
-// Returns the number of IDs that Metal would assign to an inline uniform block in an argument buffer.
-uint32_t CompilerMSL::get_inline_uniform_block_binding_stride(SPIRType &type)
-{
-	// We need this information now.
-	mark_scalar_layout_structs(type);
-	unordered_set<uint32_t> aligned_structs;
-	align_struct(type, aligned_structs);
-	uint32_t binding_stride = 0;
-	for (uint32_t i = 0; i < uint32_t(type.member_types.size()); i++)
-	{
-		uint32_t member_type_id = type.member_types[i];
-		uint32_t member_binding_stride = 1;
-		auto &member_type = get<SPIRType>(member_type_id);
-		if (member_type.basetype == SPIRType::Struct)
-			member_binding_stride = get_inline_uniform_block_binding_stride(member_type);
-		else if (member_is_packed_physical_type(type, i) && is_matrix(member_type))
-			// Packed matrices are represented as arrays, so they get multiple IDs.
-			member_binding_stride = member_type.columns;
-		for (uint32_t j = 0; j < uint32_t(member_type.array.size()); j++)
-			member_binding_stride *= to_array_size_literal(member_type, j);
-		binding_stride += member_binding_stride;
-	}
-	return binding_stride;
-}
-
 // Returns the Metal index of the resource of the specified type as used by the specified variable.
 uint32_t CompilerMSL::get_metal_resource_index(SPIRVariable &var, SPIRType::BaseType basetype, uint32_t plane)
 {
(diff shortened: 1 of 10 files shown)


HansKristian-Work:
Is this urgent? I'd like to look at this after the holidays.

spirv_msl.cpp
```diff
@@ -9688,6 +9694,31 @@ void CompilerMSL::fix_up_shader_inputs_outputs()
 	});
 }
 
+// Returns the number of IDs that Metal would assign to an inline uniform block in an argument buffer.
+uint32_t CompilerMSL::get_inline_uniform_block_binding_stride(SPIRType &type)
```
HansKristian-Work: Would this be necessary if we make sure to place the uniform buffer block at the highest ID? Then stride becomes irrelevant. I think relying on this layout is a little too sketchy for my liking.

spirv_msl.cpp
```diff
@@ -9761,20 +9799,11 @@ uint32_t CompilerMSL::get_metal_resource_index(SPIRVariable &var, SPIRType::Base
...
-		desc_set = get_decoration(var.self, DecorationDescriptorSet);
-		allocate_argument_buffer_ids = descriptor_set_is_argument_buffer(desc_set);
-	}
-
 	if (allocate_argument_buffer_ids)
 	{
 		// Allocate from a flat ID binding space.
-		resource_index = next_metal_resource_ids[desc_set];
-		next_metal_resource_ids[desc_set] += binding_stride;
+		resource_index = next_metal_resource_ids[var_desc_set];
```
HansKristian-Work: Why this change?

cdavis5e:
> My main worry here is how we assign data to ids. For a pipeline layout to work in MoltenVK, we need to be able to assign a UBO blob to a single ID, irrespective of the underlying data type in the shader. If Metal requires us to assign data per sub-ID (one call for each vector) in the argument buffer, I don't think it makes sense to support this feature in SPIRV-Cross, but I don't have an Apple device handy until next year.

From the docs for `-[MTLArgumentEncoder constantDataAtIndex:]`:

> Constants declared contiguously in the Metal shading language (in an array or struct) are contiguous in memory. You can encode contiguous ranges of inlined constant data through a pointer to the first constant.

So this should be OK.

spirv_msl.cpp
cdavis5e: I imagine this isn't a commonly used feature. Really, I'm only adding it because support for this extension was added to MoltenVK, so now I feel like I have to support it with argument buffers. I guess this would work OK.

spirv_msl.cpp
cdavis5e: Isn't the `desc_set` variable redundant?

HansKristian-Work: Looks like it, yes. Seems fine then.

spirv_msl.cpp
HansKristian-Work: Alright, I think the missing piece here is to move the (single) inline uniform block to the back of the argument buffer, and make sure it works for automatically allocated IDs as well as the add_msl_resource API.

HansKristian-Work: With that change I think this method is redundant.

cdavis5e:
Sorry about the long delay. I left on vacation shortly after this, and when I got back, I found myself with work elsewhere that needed to be done.

HansKristian-Work:
No problem, thanks!

PR #1240, opened 2019-12-20
MSL Argument Buffers: preserve all resources

troughton:
In MSL, the unused contents of an argument buffer are important if that argument buffer is to be used across multiple entry points; unused members still contribute to the layout of the argument buffer. To accomodate use cases where an argument buffer is reused across multiple entry points (e.g. bound to the vertex and fragment stages at the same time, with not all resources used in each stage), don't strip unused resources when argument buffers are enabled.

This is a tentative change for review. If the basic idea looks good then I can go through and update the tests. Otherwise, if this needs to be behind a flag, let me know and I will alter the patch accordingly.

### DIFF
--- a/spirv_msl.cpp
+++ b/spirv_msl.cpp
@@ -1003,6 +1003,31 @@ string CompilerMSL::compile()
 	fixup_image_load_store_access();
 
 	set_enabled_interface_variables(get_active_interface_variables());
+	
+	if (msl_options.argument_buffers)
+ 	{
+ 		// If we're using argument buffers, we need to emit all bindings that 
+		// could be contained within an argument buffer.
+ 		const ShaderResources resources = get_shader_resources();
+ 		for (auto &resource : resources.uniform_buffers)
+ 			active_interface_variables.insert(resource.id);
+
+  		for (auto &resource : resources.storage_buffers)
+ 			active_interface_variables.insert(resource.id);
+
+  		for (auto &resource : resources.storage_images)
+ 			active_interface_variables.insert(resource.id);
+
+  		for (auto &resource : resources.sampled_images)
+ 			active_interface_variables.insert(resource.id);
+
+  		for (auto &resource : resources.separate_images)
+ 			active_interface_variables.insert(resource.id);
+
+  		for (auto &resource : resources.separate_samplers)
+ 			active_interface_variables.insert(resource.id);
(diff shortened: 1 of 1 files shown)


HansKristian-Work:
Isn't the layout determined from `[[id()]]` attributes?

troughton:
Partially, but I think `[[id()]]` attributes only control the ordering and not the padding. As a simple example using inline members:

```
struct ArgBuffer0 {
    constant void* someBuffer [[ id(0) ]];
    uint x [[ id(1) ]];
    float y [[ id(2) ]];
    constant void* someOtherBuffer [[ id(3) ]];
};
```

must have a different layout than:

```
struct ArgBuffer1 {
    constant void* someBuffer [[ id(0) ]];
    constant void* someOtherBuffer [[ id(3) ]];
};
```

unless each `id` has a fixed size. However, it's guaranteed that for `ArgBuffer0`, `constantBytesAt: 1` must point to `uint x; float y;` in contiguous memory, which limits the size of an `id` element to at most four bytes. That's not enough for a texture or buffer + offset.

(I haven't double-checked this implementation-wise, but that's my reasoning).

HansKristian-Work:
There has to be some kind of analog to a pipeline layout/root signature for argument buffers to even make sense. I'll look further into this after the holidays (https://developer.apple.com/documentation/metal/mtlargumentdescriptor seems like what you need), but overall this seems like a hack which cannot work in the general case anyways since "unused" resources simply might not even be declared in the shader at all.

troughton:
It works in the specific case where all resources are declared in the shaders, which you can guarantee by e.g. including shared headers which contain the resource bindings and using `--preserve-bindings` in any optimisation passes (see https://github.com/KhronosGroup/SPIRV-Tools/pull/2693). I wouldn't regard this as a hack.

With that said, you're right that `MTLArgumentDescriptor` may be a more general solution; however, it's also a lot more engineering effort for my specific use case. This can easily wait until after the holidays, though, and I appreciate you having taken the time you have to look into it.

HansKristian-Work:
Alright, I made a new issue which I'll follow up at some point.

HansKristian-Work:
I made a PR which implements something similar to this PR. There is now an option to force-activate any resources which are part of argument buffers (but leaves discrete descriptors alone).

PR #1264, opened 2020-01-16
MSL: Add support for force-activating IAB resources.

HansKristian-Work:
Important for ABI compatibility on MSL in certain cases.

Supersedes https://github.com/KhronosGroup/SPIRV-Cross/pull/1240.


troughton:
Thanks for picking this up.

PR #1275, opened 2020-02-06
Remove old memory_scope flag from iOS barriers.

### LINKED ISSUE #1274: MSL: Look into memory scope qualifiers
Apparently they are not supported anymore.

HansKristian-Work:
I cannot find any reference to this flag ever having existed in older
MSL spec documents, and it breaks compilation on any recent SDK for any
iOS/macOS Metal version. Just remove it.

Fix #1274.


HansKristian-Work:
Ping @billhollings in case there is another solution.

HansKristian-Work:
Actually, there is one reference to it here: https://github.com/KhronosGroup/SPIRV-Cross/pull/785#issuecomment-582093640. But the memory_scope argument seems to be optional anyways?

Issue #719, opened 2018-10-03
Support multiple SPIR-V modules compiled together into one MSL shader


cdavis5e:
This is unfortunately necessary for #120. Tessellation works differently in Metal vs. Vulkan (and D3D). In Vulkan, two shader stages (tessellation control and tessellation evaluation) and a fixed function stage (tessellation proper) are inserted between the vertex shader and rasterization (or the geometry shader, if present). In Metal, however, the vertex and tess. control stages are omitted entirely. Instead, you are expected to run a compute shader that populates the buffer(s) used by the tessellator. The tess. evaluation stage is then run as a "post-tessellation vertex function." Therefore, for Vulkan->Metal translation, as MoltenVK is doing, the vertex shader and tess. control shader must be compiled together into one Metal compute kernel, which is responsible for running the original vertex and tess. control functions and using their output to populate the tessellation input buffers. (See also [Apple's guidance for porting D3D11 shaders to Metal](https://developer.apple.com/library/archive/documentation/Miscellaneous/Conceptual/MetalProgrammingGuide/Tessellation/Tessellation.html#//apple_ref/doc/uid/TP40014221-CH15-DontLinkElementID_51).)

Geometry shaders are another place where this may be necessary, since Metal does not yet support them natively. The geometry shader must thus run as a compute kernel, feeding its output through a buffer and a pass-through vertex shader to the rasterizer. But the original vertex shader must run prior to the geometry shader. If the original vertex shader and geometry shader (assuming no tessellation) were run normally, it would require one render pipeline with `rasterizationDisabled`, followed by a compute pipeline running the geometry shader, followed by yet another render pipeline for the rest of the rendering process. (And indeed, this is what we'll likely end up doing for combined geometry and tessellation shaders.) But if the vertex and geometry shaders were combined into one compute kernel, the first render pipeline could be omitted (along with a buffer needed to hold the output of the vertex shader to pass to the geometry shader).

cdavis5e:
I should probably add why I think we need this specifically instead of just decompiling the shaders separately and concatenating the source strings.

[...]

PR #1294, opened 2020-03-10
MSL: Add support for processing more than one patch per workgroup.

cdavis5e:
This should hopefully reduce underutilization of the GPU, especially on
GPUs where the thread execution width is greater than the number of
control points.

This also simplifies initialization by reading the buffer directly
instead of using Metal's vertex-attribute-in-compute support. It turns
out the only way in which shader stages are allowed to differ in their
interfaces is in the number of components per vector; the base type must
be the same. Since we are using the raw buffer instead of attributes, we
can now also emit arrays and matrices directly into the buffer, instead
of flattening them and then unpacking them. Structs are still flattened,
however; this is due to the need to handle vectors with fewer components
than were output, and I think handling this while also directly emitting
structs could get ugly.

Another advantage of this scheme is that the extra invocations needed to
read the attributes when there were more input than output points are
now no more. The number of threads per workgroup is now lcm(SIMD-size,
output control points). This should ensure we always process a whole
number of patches per workgroup. Additionally, if the device supports
non-uniform workgroups, the boundary checks are completely unnecessary.

To avoid complexity handling indices in the tessellation control shader,
I've also changed the way vertex shaders for tessellation are handled.
They are now compute kernels using Metal's support for vertex-style
stage input. This lets us always emit vertices into the buffer in order
of vertex shader execution. Now we no longer have to deal with indexing
in the tessellation control shader. This also fixes a long-standing
issue where if an index were greater than the number of vertices to
draw, the vertex shader would wind up writing outside the buffer, and
the vertex would be lost.

This is a breaking change, and I know SPIRV-Cross has other clients, so
I've hidden this behind an option for now. In the future, I want to
remove this option and make it the default.


cdavis5e:
I hate Visual C++ so much.

cdavis5e:
The Travis VM that builds for VS15 is missing something--that's why it broke.

HansKristian-Work:
> It turns
out the only way in which shader stages are allowed to differ in their
interfaces is in the number of components per vector; the base type must
be the same.

I don't think this is correct. Can you refer to some spec language?

main.cpp
```diff
@@ -823,6 +831,8 @@ static string compile_iteration(const CLIArguments &args, std::vector<uint32_t>
 			msl_comp->add_dynamic_buffer(v.first, v.second, i++);
 		for (auto &v : args.msl_inline_uniform_blocks)
 			msl_comp->add_inline_uniform_block(v.first, v.second);
+		for (auto &v : args.msl_vertex_attribs)
+			msl_comp->add_msl_vertex_attribute(v);
```
HansKristian-Work: Is this adding input attributes? I would expect we would need these things for vertex outputs <-> tesc input if anything?

reference/opt/shaders-msl/asm/tesc/tess-level-overrun.multi-patch.asm.tesc
```diff
@@ -0,0 +1,26 @@
...
+    float outer0;
+    float outer1;
+    float outer2;
+    float outer3;
+};
+
+kernel void main0(const device TessLevels& sb_levels [[buffer(0)]], uint3 gl_GlobalInvocationID [[thread_position_in_grid]], constant uint* spvIndirectParams [[buffer(29)]], device MTLTriangleTessellationFactorsHalf* spvTessLevel [[buffer(26)]])
+{
+    if (gl_GlobalInvocationID.x / 1 >= spvIndirectParams[1])
+        return;
```
HansKristian-Work: We always had the potential issue of barrier in tesc after returning, which should be undefined. Do we avoid that problem if partial workgroups are supported?

spirv_msl.hpp
```diff
@@ -317,6 +321,31 @@ class CompilerMSL : public CompilerGLSL
...
+		bool vertex_for_tessellation = false;
+
+		typedef enum
+		{
+			NoIndex,
+			Index16,
+			Index32
+		} IndexType;
+
+		// The type of index in the index buffer, if present. For a compute shader, Metal
```
HansKristian-Work: I assume this only works because the PATCH primitive type does not allow primitive restart?

HansKristian-Work: Does Metal require this (does Metal actually support using a compute shader as a vertex shader?), or isn't it just our implementation that requires it?

spirv_msl.hpp
```diff
@@ -317,6 +321,31 @@ class CompilerMSL : public CompilerGLSL
...
+		// single workgroup. This requires changes to the way the InvocationId and PrimitiveId
+		// builtins are processed, but should result in more efficient usage of the GPU.
+		bool multi_patch_workgroup = false;
+
+		// If set, a vertex shader will be compiled as part of a tessellation pipeline.
+		// It will be translated as a compute kernel, so it can use the global invocation ID
+		// to index the output buffer.
+		bool vertex_for_tessellation = false;
+
+		typedef enum
```
HansKristian-Work: This could be an enum class.

spirv_msl.hpp
HansKristian-Work: Looking at the code I'm only seeing the index buffer being used to resolve gl_VertexIndex, does Metal support using index buffers to guide compute kernels?

reference/opt/shaders-msl/tesc/load-control-point-array.multi-patch.tesc
```diff
@@ -0,0 +1,71 @@
...
+struct main0_in
+{
+    float4 vInputs;
+    ushort2 m_43;
+};
+
+kernel void main0(uint3 gl_GlobalInvocationID [[thread_position_in_grid]], device main0_out* spvOut [[buffer(28)]], constant uint* spvIndirectParams [[buffer(29)]], device MTLQuadTessellationFactorsHalf* spvTessLevel [[buffer(26)]], constant main0_in* spvIn [[buffer(22)]])
+{
+    device main0_out* gl_out = &spvOut[gl_GlobalInvocationID.x - gl_GlobalInvocationID.x % 4];
+    constant main0_in* gl_in = &spvIn[gl_GlobalInvocationID.x / 4 * spvIndirectParams[0]];
```
HansKristian-Work: Is this supposed to be constant and not `const device *`?

spirv_cross_c.cpp
```diff
@@ -992,6 +1012,34 @@ spvc_result spvc_compiler_msl_add_vertex_attribute(spvc_compiler compiler, const
 #endif
 }
 
+spvc_result spvc_compiler_msl_add_vertex_attribute_2(spvc_compiler compiler, const spvc_msl_vertex_attribute_2 *va)
```
HansKristian-Work: I see that the vertex shaders are still using attributes in the compute kernel variant. Is vecsize actually needed? Doesn't attribute deal with it in compute?

spirv_cross_c.h
```diff
@@ -257,15 +257,24 @@ typedef enum spvc_msl_platform
 	SPVC_MSL_PLATFORM_MAX_INT = 0x7fffffff
 } spvc_msl_platform;
 
+/* Maps to C++ API. */
+typedef enum spvc_msl_index_type
+{
+	SPVC_MSL_NO_INDEX = 0,
+	SPVC_MSL_INDEX_16 = 1,
+	SPVC_MSL_INDEX_32 = 2
```
HansKristian-Work: Must have a SPVC_MSL_INDEX_TYPE_MAX_INT to force 32-bit enum.

spirv_cross_c.h
```diff
@@ -257,15 +257,24 @@ typedef enum spvc_msl_platform
 	SPVC_MSL_PLATFORM_MAX_INT = 0x7fffffff
 } spvc_msl_platform;
 
+/* Maps to C++ API. */
+typedef enum spvc_msl_index_type
+{
+	SPVC_MSL_NO_INDEX = 0,
```
HansKristian-Work: Enum name should be SPVC_MSL_INDEX_TYPE_NONE, INDEX_TYPE_16, INDEX_TYPE_32 to match the enum name.

spirv_msl.cpp
```diff
@@ -147,10 +148,15 @@ void CompilerMSL::build_implicit_builtins()
 	bool need_multiview = get_execution_model() == ExecutionModelVertex && !msl_options.view_index_from_device_index &&
 	                      (msl_options.multiview || active_input_builtins.get(BuiltInViewIndex));
 	bool need_dispatch_base =
-	    msl_options.dispatch_base && get_execution_model() == ExecutionModelGLCompute &&
-	    (active_input_builtins.get(BuiltInWorkgroupId) || active_input_builtins.get(BuiltInGlobalInvocationId));
+	    (msl_options.dispatch_base && get_execution_model() == ExecutionModelGLCompute &&
```
HansKristian-Work: Can this be split up into different statements?

spirv_msl.cpp
```diff
@@ -1319,7 +1331,7 @@ void CompilerMSL::extract_global_variables_from_function(uint32_t func_id, std::
 				{
 					if (added_in)
 						continue;
-					name = input_wg_var_name;
```
HansKristian-Work: Why this change?

spirv_msl.cpp
```diff
@@ -1434,9 +1446,32 @@ void CompilerMSL::mark_as_packable(SPIRType &type)
 }
 
 // If a vertex attribute exists at the location, it is marked as being used by this shader
-void CompilerMSL::mark_location_as_used_by_shader(uint32_t location, StorageClass storage)
+void CompilerMSL::mark_location_as_used_by_shader(uint32_t location, const SPIRType &type, StorageClass storage)
 {
-	if ((get_execution_model() == ExecutionModelVertex || is_tessellation_shader()) && (storage == StorageClassInput))
+	if ((get_execution_model() != ExecutionModelVertex && !is_tessellation_shader()) || (storage != StorageClassInput))
```
HansKristian-Work: This is a little hard to follow. Possible to make it a positive boolean check and then `if (!blah) return;`?

spirv_msl.cpp
```diff
@@ -2588,17 +2674,38 @@ uint32_t CompilerMSL::add_interface_block(StorageClass storage, bool patch)
...
-					entry_func.fixup_hooks_in.push_back([=]() {
-						statement("device ", to_name(ir.default_entry_point), "_", ib_var_ref, "& ", ib_var_ref, " = ",
-						          patch_output_buffer_var_name, "[", to_expression(builtin_primitive_id_id), "];");
-					});
+				if (msl_options.multi_patch_workgroup)
+				{
+					// We cannot use PrimitiveId here, because the hook may not have run yet.
+					if (patch)
+						entry_func.fixup_hooks_in.push_back([=]() {
+							statement("device ", to_name(ir.default_entry_point), "_", ib_var_ref, "& ", ib_var_ref,
```
HansKristian-Work: These are multi-line branches and should have braces around them.

spirv_msl.cpp
```diff
@@ -2588,17 +2674,38 @@ uint32_t CompilerMSL::add_interface_block(StorageClass storage, bool patch)
...
+				}
 				else
-					entry_func.fixup_hooks_in.push_back([=]() {
-						statement("device ", to_name(ir.default_entry_point), "_", ib_var_ref, "* gl_out = &",
-						          output_buffer_var_name, "[", to_expression(builtin_primitive_id_id), " * ",
-						          get_entry_point().output_vertices, "];");
-					});
+				{
+					if (patch)
+						entry_func.fixup_hooks_in.push_back([=]() {
```
HansKristian-Work: Multi-line branch. (Seems like original code didn't have them for whatever reason, but it should not be like that.

spirv_msl.cpp
```diff
@@ -2625,6 +2732,47 @@ uint32_t CompilerMSL::add_interface_block(StorageClass storage, bool patch)
 		add_variable_to_interface_block(storage, ib_var_ref, ib_type, *p_var, meta);
 	}
 
+	if (get_execution_model() == ExecutionModelTessellationControl && msl_options.multi_patch_workgroup &&
+	    storage == StorageClassInput)
+	{
+		// For tessellation control inputs, add all outputs from the vertex shader to ensure
```
HansKristian-Work: Do we ensure that the tessellation control inputs are declared in an expected order?

HansKristian-Work: Since vertex shader uses attributes to write varying, I don't see how we guarantee that tessellation inputs are declared at the expected Offset.

cdavis5e:
> > It turns
> > out the only way in which shader stages are allowed to differ in their
> > interfaces is in the number of components per vector; the base type must
> > be the same.
> 
> I don't think this is correct. Can you refer to some spec language? I think it must be an exact match.

From [Vulkan 14.1.3 - Interface Matching][https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/chap14.html#interfaces-iointerfaces-matching]:

> Additionally, scalar and vector inputs are well-defined if there is a corresponding output satisfying all of the following conditions:
> * the input and output match exactly in decoration,
> * the output is a vector with the same basic type and has at least as many components as the input, and
> * the common component type of the input and output is 16-bit integer or floating-point, or 32-bit integer or floating-point (64-bit component types are excluded).

main.cpp
cdavis5e: I'm using "vertex attributes" during the tessellation control stage to fill in the input struct.

reference/opt/shaders-msl/asm/tesc/tess-level-overrun.multi-patch.asm.tesc
cdavis5e: Unfortunately, we can't. I wanted to add a knob that omits this check if non-uniform workgroups are supported, but this doesn't work for indirect draws: there's no `-[MTLComputeCommandEncoder dispatchThreadsWithIndirectBuffer:...]` method.

spirv_msl.hpp
cdavis5e: > I assume this only works because the PATCH primitive type does not allow primitive restart?

Yep.



> Does Metal require this (does Metal actually support using a compute shader as a vertex shader?), or isn't it just our implementation that requires it?

It's just the current implementation that requires it, and only (as you later discovered) to resolve `gl_VertexIndex`. Note that you can't specify a compute function as the `vertexFunction` in a render pipeline, but you can do vertex-style stage input to a compute function. Ostensibly, this was intended to be used to run the vertex stage of a tessellation pipeline that was ported from D3D/OpenGL/Vulkan.

> Looking at the code I'm only seeing the index buffer being used to resolve gl_VertexIndex, does Metal support using index buffers to guide compute kernels?

Yes, but you have to specify the index buffer's location and index size when you create the pipeline object. (We had this issue with `tesc` shaders in the old implementation.)

reference/opt/shaders-msl/tesc/load-control-point-array.multi-patch.tesc
cdavis5e: If we can ensure that it meets the required alignment for buffers in the `constant` space, which we do in MoltenVK, we should be able to get away with using the `constant` AS.

spirv_cross_c.cpp
cdavis5e: Like I said above, this is used to fill in the holes in the input buffer in a tessellation control shader, in case an output from the previous stage is not declared. We have to do this in order to make sure the `struct` layout matches.

spirv_cross_c.h
cdavis5e: Ugh, it's not like we have to build 16-bit code--that's why DX had those `FORCE_DWORD` enums in the first place. All right, fine... /me grumbles

spirv_msl.cpp
cdavis5e: Because in the new implementation, the parameter name to the entry point is no longer the same as the name that gets used in the generated shader code.

spirv_msl.cpp
cdavis5e: They should be sorted by location in both shaders.

reference/opt/shaders-msl/asm/tesc/tess-level-overrun.multi-patch.asm.tesc
cdavis5e: Come to think of it, I have been seeing weirdness in the `dEQP-VK.tessellation.shader_input_output.barrier` test. The geometry isn't offset correctly, because the per-vertex varyings aren't written correctly in the `tesc` shader. There was a bump at 4-5 which was closer to what it should be, which led me to suspect that the driver was miscompiling the shader around a conditional branch for vertex 5. But this might explain some of the weirdness.

spirv_msl.hpp
cdavis5e: Done.

spirv_cross_c.h
cdavis5e: Done.

spirv_msl.cpp
cdavis5e: Sure.

spirv_msl.cpp

spirv_msl.cpp

spirv_msl.cpp

reference/opt/shaders-msl/asm/tesc/tess-level-overrun.multi-patch.asm.tesc
cdavis5e: I've wondered if, like in Apple's example ported hull shader, I should just wrap the whole shader in an `if` statement. But then the shaders that aren't supposed to be running won't hit the barriers, which will cause the other shaders to hang.

HansKristian-Work: Not sure that would help. If we know that the workgroup is <= SIMD width, it might happen to work though. I think we need to have a solution for this problem in a new implementation though.

spirv_msl.hpp
HansKristian-Work: So to be clear, attribute fetching can be indexed by an index buffer in a Metal compute kernel? And that means the vertex shader will transparently unroll VBO + IBO -> to a flat control point array, which TESC then consumes?

reference/opt/shaders-msl/tesc/load-control-point-array.multi-patch.tesc
HansKristian-Work: Ok.

spirv_cross_c.cpp
HansKristian-Work: Ok, so the general idea here is to first look at the vertex shader, figure out all its outputs and capture their types as well.

Then, we want to replicate the exact buffer layout for tessellation control inputs?

spirv_cross_c.h
```diff
@@ -257,15 +257,25 @@ typedef enum spvc_msl_platform
...
 {
 	SPVC_MSL_VERTEX_FORMAT_OTHER = 0,
 	SPVC_MSL_VERTEX_FORMAT_UINT8 = 1,
-	SPVC_MSL_VERTEX_FORMAT_UINT16 = 2
+	SPVC_MSL_VERTEX_FORMAT_UINT16 = 2,
+	SPVC_MSL_VERTEX_FORMAT_UINT32 = 3,
 } spvc_msl_vertex_format;
 
-/* Maps to C++ API. */
+/* Maps to C++ API. Deprecated; use spvc_msl_vertex_attribute2. */
```
HansKristian-Work: It doesn't feel right to deprecate this if vecsize is only relevant for a very specific tessellation control input implementation.

Would it make sense to add a separate entry point which just adds vecsize information ala
`spvc_msl_vertex_attribute_explicit_vecsize(comp, location, vecsize)` or something like that?

reference/opt/shaders-msl/asm/tesc/tess-level-overrun.multi-patch.asm.tesc
HansKristian-Work: Since number of control points is <= 32, if SIMD width >= 32, we should be able to use 64-wide workgroups on AMD as well with simd barriers.

HansKristian-Work:
@LukasBanana probably should be in the loop here.

spirv_msl.hpp
cdavis5e: Yes.

reference/opt/shaders-msl/asm/tesc/tess-level-overrun.multi-patch.asm.tesc
cdavis5e: MSL 1.2. Oh, and SIMD-group barriers require MSL 2.0 on Mac.

spirv_cross_c.cpp
cdavis5e: That might be a workable strategy.

reference/opt/shaders-msl/asm/tesc/tess-level-overrun.multi-patch.asm.tesc
cdavis5e: There's nothing in the MSL spec to suggest that barriers are undefined after a return statement. It only says:

> If `threadgroup_barrier` (or `simdgroup_barrier`) is inside a conditional statement and if any thread enters the conditional statement and executes the barrier function, then all threads in the threadgroup (or SIMD-group) must enter the conditional and execute the barrier function.
>
> If `threadgroup_barrier` (or `simdgroup_barrier`) is inside a loop, for each iteration of the loop, all threads in the threadgroup (or SIMD-group) must execute the barrier function before any threads continue execution beyond the barrier function.

I wonder if the behavior here is under-documented.

HansKristian-Work: I would be surprised if it has different semantics than Vulkan or D3D here, "all threads must enter" means that any returned threads cannot enter and that is undefined behavior in those APIs. If the semantics are actually to make this well-behaved in Metal, then that is a very pleasant surprise. Needs to be well tested though.

spirv_msl.hpp
HansKristian-Work: That's kinda cool actually, good to know.

spirv_cross_c.cpp
cdavis5e: > What happens if the application is expecting to load a float2, but the type was mapped to float4 in this scheme?

If there is a next member, its offset should be set as if the previous member were a `float4`.

reference/opt/shaders-msl/asm/tesc/tess-level-overrun.multi-patch.asm.tesc
cdavis5e: Oh, and the other thing about SIMD-groups, which I mentioned on #719. The driver is allowed to vary the SIMD-group size based on register usage--like in Vulkan's `VK_EXT_subgroup_size_control` extension, only it can't be turned off. So I can't necessarily guarantee that the SIMD-group size will be >= the number of control points.

HansKristian-Work: Is it possible to query the minimum SIMD-width though?

cdavis5e: No, not even with any of the secret undocumented methods.

cdavis5e: All right, I've made a test program which will reproduce the problem with the test. It replicates roughly what MoltenVK does in response to the test's actions. As you can see, I commented out the return, and... it had no effect. The output is still wrong. (You can see the output by running it under Xcode's control. It will automatically trigger a GPU capture with no action on your part.)

(Gaah! GitHub is being dumb. It won't let me upload the file. Sorry, I can't share it with you. You'll just have to take my word for it.)

HansKristian-Work: One test to try is using a large workgroup and then do this early return, doesn't have to be tess, just some compute kernel. I figure that for AMD you'll see 1 SIMD group, and I doubt s_barrier will care in that case, but it might for Intel and iOS where you usually get much smaller SIMD groups.

spirv_cross_c.cpp
HansKristian-Work: Do you want to keep using vecsize in that case, or use offset directly?

reference/opt/shaders-msl/asm/tesc/tess-level-overrun.multi-patch.asm.tesc
cdavis5e: On AMD, it works fine, even if some of the threads return early.

On Intel, I saw hangs.

Well that sucks. I can't have the extra threads return early apparently, but I can't have them actually write anything, or they might write out of bounds and cause a GPU page fault. What do I do now?

HansKristian-Work: Ouch, this was expected, but unfortunate. I think in 99.9% of cases, barriers() happen outside of control flow in hull shaders. In that case it should be possible to do a transform like:

```
void main()
{
    bool active_lane = ?;
    if (active_lane) { do_stuff(); }
    barrier();
    if (active_lane) { do_other_stuff(); }
}
```
However this assumption breaks down if the app somehow emits barriers in control flow. HLSL will never do this, and GLSL bans it, but SPIR-V technically supports it I think (not sure though! It might be our saving grace if not).

Another approach is padding. I think we can make it work just fine if we allocate padded space for the intermediate output buffers. To avoid problems with unintentional out-of-bounds reads we can clamp the reported gl_PrimitiveID to the application. That way we get duplicated work at the tail end of the dispatch. We also have to ban side-effects in tessellation so that application cannot observe duplicated execution of the tessellation control shader (it's a separate feature bit), which shouldn't be a big issue.

LukasBanana: Hi, I'm late to the party and catching up with posts and emails after my vacation.
We had an issue in a compute kernel where a barrier blocked the entire GPU because some threads had an early exit path. We filed an issue at Apple and some days later they reported the driver is fine, our compute kernel just fall into undefined behavior because of the conditional barriers.

HansKristian-Work:
It's been a while. Is this PR still on the backburner?

cdavis5e:
> It's been a while. Is this PR still on the backburner?

Unfortunately, yes. I still have other things I need to look at for our upcoming CrossOver 20 release, mostly related to getting D3D11 games working at all. I should have more time to work on this after CX20 is released.

spirv_cross_c.cpp
cdavis5e: Actually, it turns out I need to do this for fragment shaders, too. Unlike Vulkan, Metal complains when the types don't match exactly. I'll try to make a separate change.

cdavis5e:
> > It's been a while. Is this PR still on the backburner?
> 
> Unfortunately, yes. I still have other things I need to look at for our upcoming CrossOver 20 release, mostly related to getting D3D11 games working at all. I should have more time to work on this after CX20 is released.

It turns out, I need this to get a game working.

spirv_cross_c.h
cdavis5e: This has already been replaced with `spvc_msl_shader_input`.

main.cpp
cdavis5e: "Vertex attributes" have been replaced by "shader inputs."

reference/opt/shaders-msl/asm/tesc/tess-level-overrun.multi-patch.asm.tesc
cdavis5e: I've decided to just remove the boundary checks and have MoltenVK pad the buffers. Adding boundary checks around everything except barriers just seemed too complex.

spirv_msl.cpp
```diff
@@ -1466,9 +1480,32 @@ void CompilerMSL::mark_as_packable(SPIRType &type)
...
 // If a shader input exists at the location, it is marked as being used by this shader
-void CompilerMSL::mark_location_as_used_by_shader(uint32_t location, StorageClass storage)
+void CompilerMSL::mark_location_as_used_by_shader(uint32_t location, const SPIRType &type, StorageClass storage)
 {
-	if (storage == StorageClassInput)
+	if (storage != StorageClassInput)
+		return;
+	if (is_array(type))
+	{
+		for (uint32_t i = 0; i < type.array.size(); i++)
```
HansKristian-Work, line 1489: Seems a bit cleaner to multiply together all dimensions and then loop over number of dimensions separately.

spirv_msl.cpp
```diff
@@ -2570,17 +2637,29 @@ uint32_t CompilerMSL::add_interface_block(StorageClass storage, bool patch)
...
-				statement("threadgroup_barrier(mem_flags::mem_threadgroup);");
-				statement("if (", to_expression(builtin_invocation_id_id), " >= ", get_entry_point().output_vertices,
-				          ")");
-				statement("    return;");
+				// Can't use PatchVertices, PrimitiveId, or InvocationId yet; the hooks for those may not have run yet.
+				if (msl_options.multi_patch_workgroup)
+				{
+					// n.b. builtin_invocation_id_id here is the dispatch global invocation ID,
+					// not the TC invocation ID.
+					statement("constant ", to_name(ir.default_entry_point), "_", ib_var_ref, "* gl_in = &",
```
HansKristian-Work, line 2647: Is constant safe (w.r.t buffer size limitations)? I would expect `const device` address space for inputs coming from vertex shader.

spirv_msl.cpp
```diff
@@ -9732,6 +10045,16 @@ string CompilerMSL::entry_point_args_classic(bool append_comma)
 
 void CompilerMSL::fix_up_shader_inputs_outputs()
 {
+	auto &entry_func = this->get<SPIRFunction>(ir.default_entry_point);
+
+	// Emit a guard to ensure we don't execute beyond the last vertex.
+	if (get_execution_model() == ExecutionModelVertex && msl_options.vertex_for_tessellation)
```
HansKristian-Work, line 10051: Wrap this in `{}`.

spirv_msl.cpp
```diff
@@ -11648,6 +12056,9 @@ string CompilerMSL::builtin_qualifier(BuiltIn builtin)
 
 	// Tess. control function in
 	case BuiltInInvocationId:
+		if (msl_options.multi_patch_workgroup)
```
HansKristian-Work, line 12059: Wrap this in `{}`.

spirv_msl.cpp
```diff
@@ -11656,6 +12067,9 @@ string CompilerMSL::builtin_qualifier(BuiltIn builtin)
 		switch (execution.model)
 		{
 		case ExecutionModelTessellationControl:
+			if (msl_options.multi_patch_workgroup)
```
HansKristian-Work, line 12070: Wrap this in `{}`.

HansKristian-Work:
Hm, I don't recall most of this patch anymore, so I'll need some time to figure it out again.
There are still some concerns I had when looking over this, though.

For the padded control points, do we have any control over the input? I am a bit concerned what could happen if the shader gets to read garbage input, and it then uses control flow which depends on that, it might hang the GPU if the vertex shader emits a loop counter of some kind. What we could do is to simply clamp the input index and write duplicate data. Since we're now rendering the same control point multiple times, this might affect tessellation with side-effects, but that's at least an optional feature.

Barriers could get weird if one control point have all their threads return early for whatever reason, and there's a barrier() later. I'm not sure if this is even legal, and I don't expect this to show up in any code. Code coming from HLSL will generally wrap their two phases in separate functions and the barrier is implied. In GLSL, you can only place a barrier outside any control flow I think, so it might be just fine to just pray that this will work out.

spirv_msl.cpp
cdavis5e: Not on Mac GPUs, I think. We should use the `device` AS anyway: Apple recommends it for data that can vary between threads.

I wanted to use `const device`, but since `gl_in` is technically a variable pointer, I couldn't get SPIRV-Cross to emit a reference to a `const device` pointer.

cdavis5e:
> For the padded control points, do we have any control over the input?

Yes we do. The input buffer to the tessellation control stage is the output buffer from the vertex shader stage. We should thus be able to add the requisite padding easily. We might want to zero it, though, before passing it. This is yet another situation where I wish Metal supported the Direct3D semantics for out-of-bounds raw buffer access. Or we could do as you suggested, and clamp the invocation ID on the extra invocations. We don't support `vertexPipelineStoresAndAtomics` in MoltenVK anyway, because Metal vertex shaders don't let you simultaneously write to a buffer *and* pass data to the rasterizer.

HansKristian-Work:
Clamping input data index in the TESC seems safer to me, since all-zero data is still technically "magic, hallucinated data" that might cause some problems. If stores and atomics are disabled, that should be the most robust solution.

main.cpp
```diff
@@ -747,9 +749,15 @@ static void print_help_msl()
...
 	                "\t[--msl-no-clip-distance-user-varying]:\n\t\tDo not emit user varyings to emulate gl_ClipDistance in fragment shaders.\n"
 	                "\t[--msl-shader-input <index> <format> <size>]:\n\t\tSpecify the format of the shader input at <index>.\n"
-	                "\t\t<format> can be 'u16', 'u8', or 'other', to indicate a 16-bit unsigned integer, 8-bit unsigned integer, "
+	                "\t\t<format> can be 'any32', 'any16', 'u16', 'u8', or 'other', to indicate a 32-bit opaque value, 16-bit opaque value, 16-bit unsigned integer, 8-bit unsigned integer, "
 	                "or other-typed variable. <size> is the vector length of the variable, which must be greater than or equal to that declared in the shader.\n"
-	                "\t\tUseful if shader stage interfaces don't match up, as pipeline creation might otherwise fail.\n");
+	                "\t\tUseful if shader stage interfaces don't match up, as pipeline creation might otherwise fail.\n"
+	                "\t[--msl-multi-patch-workgroup]:\n\t\tUse the new style of tessellation control processing, where multiple patches are processed per workgroup.\n"
```
HansKristian-Work, line 755: Formatting here is off with the rest of the code around it.

spirv_msl.cpp
```diff
@@ -1484,13 +1521,38 @@ uint32_t CompilerMSL::get_target_components_for_fragment_location(uint32_t locat
...
+	type->array.clear();
+	type->array_size_literal.clear();
+	type->pointer = false;
+
+	if (is_array(old_type))
+	{
+		uint32_t array_type_id = ir.increase_bound_by(1);
+		type = &set<SPIRType>(array_type_id, *type);
+		type->parent_type = new_type_id;
+		std::copy(old_type.array.begin(), old_type.array.end(), type->array.begin());
```
HansKristian-Work, line 1540: This is sketchy, what's the size of `type->array` here?

spirv_msl.cpp
```diff
@@ -1466,9 +1480,32 @@ void CompilerMSL::mark_as_packable(SPIRType &type)
...
 {
-	if (storage == StorageClassInput)
+	if (storage != StorageClassInput)
+		return;
+	if (is_array(type))
+	{
+		uint64_t dim = 1;
+		for (uint32_t i = 0; i < type.array.size(); i++)
+			dim *= to_array_size_literal(type, i);
+		for (uint64_t i = 0; i < dim; i++)
```
HansKristian-Work, line 1492: uint64_t seems a bit overkill.

HansKristian-Work:
Tried this code:
```
#version 450

layout(location = 0) in mat4 v;
layout(location = 0) out mat4 vMat;

void main()
{
	vMat = v;
	gl_Position = vec4(float(gl_InstanceIndex + gl_VertexIndex));
}
```

and getting output code with `--msl-vertex-for-tessellation`:

```
#include <metal_stdlib>
#include <simd/simd.h>

using namespace metal;

struct main0_out
{
    float4x4 vMat [[user(locn0)]];
};

struct main0_in
{
    float4 v_0 [[attribute(0)]];
    float4 v_1 [[attribute(1)]];
    float4 v_2 [[attribute(2)]];
    float4 v_3 [[attribute(3)]];
};

kernel void main0(main0_in in [[stage_in]], uint3 gl_GlobalInvocationID [[thread_position_in_grid]], uint3 spvStageInputSize [[grid_size]], uint3 spvDispatchBase [[grid_origin]], device main0_out* spvOut [[buffer(28)]])
{
    float4x4 v = {};
    device main0_out& out = spvOut[gl_GlobalInvocationID.y * spvStageInputSize.x + gl_GlobalInvocationID.x];
    v[0] = in.v_0;
    v[1] = in.v_1;
    v[2] = in.v_2;
    v[3] = in.v_3;
    if (any(gl_GlobalInvocationID >= spvStageInputSize))
        return;
    uint gl_InstanceIndex = gl_GlobalInvocationID.y + spvDispatchBase.y;
    uint gl_VertexIndex = gl_GlobalInvocationID.x + spvDispatchBase.x;
    out.vMat = v;
    out.gl_Position = float4(float(int(gl_InstanceIndex) + int(gl_VertexIndex)));
}
```

Is it intentional that the buffer output should receive a user location?

spirv_msl.cpp
HansKristian-Work: Looking at this code again, I don't understand this. It seems like in other places, input_wg_var_name is not used either. Is this variable even necessary? Is there a reason for it to be != "gl_in"?

spirv_msl.cpp
```diff
@@ -9437,8 +9746,12 @@ void CompilerMSL::entry_point_args_builtin(string &ep_args)
 			{
 				if (!ep_args.empty())
 					ep_args += ", ";
-				ep_args += join("threadgroup ", type_to_glsl(get_stage_in_struct_type()), "* ", input_wg_var_name,
-				                " [[threadgroup(", convert_to_string(msl_options.shader_input_wg_index), ")]]");
+				if (msl_options.multi_patch_workgroup)
```
HansKristian-Work, line 9749: Wrap in `{}`.

spirv_msl.cpp
```diff
@@ -9819,6 +10144,35 @@ void CompilerMSL::fix_up_shader_inputs_outputs()
 					statement(builtin_type_decl(bi_type), " ", to_expression(var_id), " = simd_is_helper_thread();");
 				});
 				break;
+			case BuiltInInvocationId:
+				if (get_execution_model() == ExecutionModelTessellationControl)
```
HansKristian-Work, line 10148: Could probably merge the two branches?

spirv_msl.cpp
```diff
@@ -9819,6 +10144,35 @@ void CompilerMSL::fix_up_shader_inputs_outputs()
...
+					entry_func.fixup_hooks_in.push_back([=]() {
+						statement(builtin_type_decl(bi_type), " ", to_expression(var_id), " = ",
+						          to_expression(builtin_invocation_id_id), ".x % ",
+						          this->get_entry_point().output_vertices, ";");
+					});
+				}
+				break;
+			case BuiltInPrimitiveId:
+				// This is natively supported by fragment and tessellation evaluation shaders.
+				if (get_execution_model() == ExecutionModelTessellationControl)
```
HansKristian-Work, line 10163: Could probably merge the two branches?

spirv_msl.cpp
```diff
@@ -9339,7 +9607,12 @@ void CompilerMSL::entry_point_args_builtin(string &ep_args)
 			     bi_type != BuiltInViewIndex) &&
 			    (get_execution_model() == ExecutionModelGLCompute ||
 			     (get_execution_model() == ExecutionModelFragment && msl_options.supports_msl_version(2, 2)) ||
-			     (bi_type != BuiltInSubgroupLocalInvocationId && bi_type != BuiltInSubgroupSize)))
+			     (bi_type != BuiltInSubgroupLocalInvocationId && bi_type != BuiltInSubgroupSize)) &&
```
HansKristian-Work, line 9610: This branch has to be setting some kind of record for unreadability at this point ...

spirv_msl.cpp
```diff
@@ -308,24 +316,18 @@ void CompilerMSL::build_implicit_builtins()
...
 
 			// Create gl_SampleID.
-			SPIRType uint_type;
-			uint_type.basetype = SPIRType::UInt;
-			uint_type.width = 32;
-			set<SPIRType>(type_id, uint_type);
-
 			SPIRType uint_type_ptr;
-			uint_type_ptr = uint_type;
+			uint_type_ptr = get_uint_type();
```
HansKristian-Work, line 325: It would be great if this get_uint_type() refactor could be done in a separate PR to keep this PR as managable as possible.

HansKristian-Work:
The test suite is failing on Windows: https://travis-ci.org/github/KhronosGroup/SPIRV-Cross/jobs/709064740#L7416.

I think there is an ordering problem with main0_in struct declaration. The members are out of order, so seems to be some kind of unstable sorting going on perhaps.

spirv_msl.cpp
cdavis5e: I agree. So I refactored it in #1430.

spirv_msl.cpp
cdavis5e: #1430.

spirv_msl.cpp
cdavis5e: Replaced with `uint32_t`.

main.cpp
cdavis5e: I think I fixed it.

spirv_msl.cpp
cdavis5e: Probably not the same as `old_type.array`. I changed it to an assign.

cdavis5e:
> Is it intentional that the buffer output should receive a user location, similar with position?

No. I've changed it not to emit those.

cdavis5e:
> The test suite is failing on Windows: https://travis-ci.org/github/KhronosGroup/SPIRV-Cross/jobs/709064740#L7416.
> 
> I think there is an ordering problem with main0_in struct declaration. The members are out of order, so seems to be some kind of unstable sorting going on perhaps.

I figured out what's wrong. I was iterating over the `inputs_by_location` container, which is an `unordered_map`. So the extra members were added in an undefined order. Using an ordered `map` should solve the issue.

PR #1307, opened 2020-04-03
MSL: mark BuiltInFragCoord as implicitly used for subpass reads

troughton:
Without this patch, `gl_FragCoord` is not output for subpass reads in certain cases where `gl_FragCoord` is not used elsewhere in the shader.

I'm not sure why this isn't caught by existing tests (e.g. [input-attachment.frag](shaders-msl/frag/input-attachment.frag)), but I encountered this issue in code generated by DXC and passed through spire-opt.


HansKristian-Work:
Right, the problem is that FragCoord is declared in the SPIR-V, but unused and that confused SPIRV-Cross as bit, since we need to make sure it is treated as a used builtin. I think this affects other builtins as well, so I'll fix that as well.

PR #1309, opened 2020-04-03
Expose query if a resource was used as a comparison/depth resource

### LINKED ISSUE #1305: Expose Compiler::image_is_comparison as public
WebGPU's validation requires that on pipeline compilation, comparison/shadow samplers used in the shader must match the binding types specified in the pipeline's bind group layout(s).

In order to implement this in Dawn, we need to differentiate between comparison samplers and non-comparison samplers to set the proper enums. I see there is already `Compiler::comparison_ids` and `Compiler::image_is_comparison` but these are private and protected, respectively.

Could SPIRV-Cross expose a public API to query whether a `Resource::ID` is a comparison sampler? This could be as simple as making `Compiler::image_is_comparison` public.

HansKristian-Work:
Fix #1305.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -323,7 +323,7 @@ if (SPIRV_CROSS_STATIC)
 endif()
 
 set(spirv-cross-abi-major 0)
-set(spirv-cross-abi-minor 29)
+set(spirv-cross-abi-minor 30)
 set(spirv-cross-abi-patch 0)
 
 if (SPIRV_CROSS_SHARED)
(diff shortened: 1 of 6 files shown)


austinEng:
Hi @HansKristian-Work, I see that `variable_is_depth_or_compare` was added onto `CompilerGLSL`. Is there a reason it couldn't be put on the base `Compiler` class? In our Vulkan backend we're only using the base `Compiler` class to extract binding information. From what I understand, we should be able to call `Compiler::analyze_image_and_sampler_usage` after parsing the IR and then query `Compiler::comparison_ids`. I guess `analyze_image_and_sampler_usage` would need to be public as well, or lazily called.

zoddicus:
@austinEng Sorry for the confusion with my comment in shaderc. For Vulkan you should be able to use CompilerReflection, which is a subclass of CompilerGLSL.

HansKristian-Work:
That information is only known *after* compilation, so it doesn't make sense to expose it in base class.

PR #1319, opened 2020-04-10
MSL: Add options to control emission of fragment outputs.

cdavis5e:
Like with `point_size` when not rendering points, Metal complains when
writing to a variable using the `[[depth]]` qualifier when no depth
buffer be attached. In that case, we must avoid emitting `FragDepth`,
just like with `PointSize`.

I assume it will also complain if there be no stencil attachment and the
shader write to `[[stencil]]`, or it write to `[[color(n)]]` but there
be no color attachment at n.


cdavis5e:
Testing was a good call. I caught a problem compiling the resulting MSL and fixed it.

spirv_msl.hpp
```diff
@@ -870,6 +873,8 @@ class CompilerMSL : public CompilerGLSL
 	// Must be ordered since array is in a specific order.
 	std::map<SetBindingPair, std::pair<uint32_t, uint32_t>> buffers_requiring_dynamic_offset;
 
+	std::unordered_set<uint32_t> disabled_frag_outputs;
```
HansKristian-Work, line 876: These variables need to be emitted in order to make test suite stable, so I suggest a vector here since there are generally few elements anyways.

spirv_msl.cpp
```diff
@@ -2467,6 +2479,17 @@ uint32_t CompilerMSL::add_interface_block(StorageClass storage, bool patch)
 		if (bi_type == BuiltInClipDistance)
 			hidden = false;
 
+		// It's not enough to simply avoid marking fragment outputs if the pipeline won't
+		// accept them. We can't put them in the struct at all, or otherwise the compiler
+		// complains that the outputs weren't explicitly marked.
+		if ((is_builtin && ((bi_type == BuiltInFragDepth && !msl_options.enable_frag_depth_builtin) ||
```
HansKristian-Work, line 2485: Shouldn't we check storage class == Output here?

spirv_msl.cpp
```diff
@@ -965,6 +965,17 @@ void CompilerMSL::emit_entry_point_declarations()
 	}
 	// For some reason, without this, we end up emitting the arrays twice.
 	buffer_arrays.clear();
+
+	// Emit disabled fragment outputs.
+	for (uint32_t var_id : disabled_frag_outputs)
```
HansKristian-Work, line 970: Needs std::vector to ensure stable output (or std::map, but very few elements, so doesn't make much sense to use a hammer for this task)

spirv_msl.cpp
cdavis5e: And `patch`. Done.

PR #1324, opened 2020-04-16
MSL: Force disabled fragment builtins to have the right name.

cdavis5e:
DXVK emits SPIR-V where fragment shader builtins have names derived from
DXBC assembly, e.g. `oDepth` for `FragDepth`. When we declared the
disabled output, we used this name, but when referencing it, we
continued to use the GLSL name. This breaks compilation.


HansKristian-Work:
The Travis failure is a common trivial one where I would normally restart the job, but I can't do it now for some reason, so just merging.

Issue #1325, opened 2020-04-16
missing `unpackUint2x32` when convering uint64_t to uvec2


Inori:
```
        %151 = OpLoad %uint %exec_lo
        %152 = OpLoad %uint %exec_hi
        %153 = OpCompositeConstruct %v2uint %152 %151
       %exec = OpBitcast %ulong %153
        %155 = OpBitcast %v2uint %exec
        %156 = OpCompositeExtract %uint %155 0
        %157 = OpCompositeExtract %uint %155 1
               OpStore %s28 %157
               OpStore %s29 %156
```
will tranlate to:
```
    uvec2 _155 = (packUint2x32(uvec2(exec_lo, exec_hi)));
    s28 = _155.y;
    s29 = _155.x;
```
while `unpackUint2x32` is expected, like:
```
uvec2 _155 = unpackUint2x32(packUint2x32(uvec2(exec_lo, exec_hi)));
```

Inori:
and also, both pack and unpack functionalities is missing if translated to hlsl

PR #1326, opened 2020-04-17
HLSL&GLSL: Support packUint2x32 and unpackUint2x32

Inori:
This add support for converting a `uint2` type to `uint64_t` type, and back from `uint64_t` to `uint2`.
This is corresponding to [GLSL's `packUint2x32` and `unpackUint2x32`](https://www.khronos.org/registry/OpenGL/extensions/NV/NV_gpu_shader5.txt)

And also, fix [an issue](https://github.com/KhronosGroup/SPIRV-Cross/issues/1325) I raised before.

### DIFF
--- a/spirv_glsl.cpp
+++ b/spirv_glsl.cpp
@@ -6683,6 +6683,8 @@ string CompilerGLSL::bitcast_glsl_op(const SPIRType &out_type, const SPIRType &i
 	// And finally, some even more special purpose casts.
 	if (out_type.basetype == SPIRType::UInt64 && in_type.basetype == SPIRType::UInt && in_type.vecsize == 2)
 		return "packUint2x32";
+	else if (out_type.basetype == SPIRType::UInt && in_type.basetype == SPIRType::UInt64 && out_type.vecsize == 2)
+		return "unpackUint2x32";
 	else if (out_type.basetype == SPIRType::Half && in_type.basetype == SPIRType::UInt && in_type.vecsize == 1)
 		return "unpackFloat2x16";
 	else if (out_type.basetype == SPIRType::UInt && in_type.basetype == SPIRType::Half && in_type.vecsize == 2)
--- a/spirv_hlsl.cpp
+++ b/spirv_hlsl.cpp
@@ -1499,6 +1499,24 @@ void CompilerHLSL::emit_resources()
 		statement("");
 	}
 
+	if (requires_uint2_packing)
+	{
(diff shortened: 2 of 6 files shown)


spirv_hlsl.cpp
```diff
@@ -5350,3 +5389,21 @@ bool CompilerHLSL::is_hlsl_resource_binding_used(ExecutionModel model, uint32_t
 	auto itr = resource_bindings.find(tuple);
 	return itr != end(resource_bindings) && itr->second.second;
 }
+
+CompilerHLSL::BitcastType CompilerHLSL::get_bitcast_type(
+    uint32_t result_type, uint32_t op0)
```
HansKristian-Work, line 5394: Formatting is off, there is a clang-format file you can use.

spirv_glsl.cpp
```diff
@@ -6683,6 +6683,8 @@ string CompilerGLSL::bitcast_glsl_op(const SPIRType &out_type, const SPIRType &i
 	// And finally, some even more special purpose casts.
 	if (out_type.basetype == SPIRType::UInt64 && in_type.basetype == SPIRType::UInt && in_type.vecsize == 2)
 		return "packUint2x32";
+	else if (out_type.basetype == SPIRType::UInt && in_type.basetype == SPIRType::UInt64 && out_type.vecsize == 2)
+		return "unpackUint2x32";
```
HansKristian-Work, line 6687: I saw GLSL was added here, you can use the same test shader for GLSL as well.

Inori:
Fixed

spirv_hlsl.cpp
```diff
@@ -1499,6 +1499,24 @@ void CompilerHLSL::emit_resources()
 		statement("");
 	}
 
+	if (requires_uint2_packing)
+	{
+		statement("uint64_t SPIRV_Cross_packUint2x32(uint2 value)");
+		begin_scope();
+		statement("return uint64_t(value.y) << 32 | uint64_t(value.x);");
```
HansKristian-Work, line 1506: Shouldn't the shift expression be inside parens here?

Inori:
@HansKristian-Work Sorry for the late reply.
I think I've fitured out the problem.

The reference hlsl shader can not compile because it uses `uint64_t` types which is only support above sm6.0, previously I modified the interface declaration manually to fie sm6.0, so there's no problem on compiling.

But currently spirv-cross only generate sm5.x hlsl shaders.
I'd like to hear what your opinion on this?

HansKristian-Work:
Merged in separate PR. Thanks!

Issue #1327, opened 2020-04-18
GLSL: Implement `GL_EXT_nonuniform_qualifier` (basically translate to `NV_gpu_program5`)


devshgraphicsprogramming:
Right now only `NV_gpu_program5` allows us to do what we can do with `GL_EXT_descriptor_indexing` which is index arrays of `sampler` divergently.

you should pretty much do the same thing you already do when you encounter `GL_EXT_shader_16bit_storage`

Right now even when I require `nonuniformEXT` I get a failure to cross compile to OpenGL semantic GLSL

devshgraphicsprogramming:
I've made a half-hearted attempt at solving this
https://github.com/KhronosGroup/SPIRV-Cross/pull/1328

PR #1328, opened 2020-04-18
Try and solve issue #1327

devshgraphicsprogramming:
actually for it to work, I'd have to remove (pass through) any `nonuniformEXT` calls in the generated GLSL shader

### DIFF
--- a/spirv_glsl.cpp
+++ b/spirv_glsl.cpp
@@ -501,6 +501,11 @@ void CompilerGLSL::find_static_extensions()
 		switch (cap)
 		{
 		case CapabilityShaderNonUniformEXT:
+			if (!options.vulkan_semantics)
+				require_extension_internal("GL_NV_gpu_shader5");
+			else
+				require_extension_internal("GL_EXT_nonuniform_qualifier");
+			break;
 		case CapabilityRuntimeDescriptorArrayEXT:
 			if (!options.vulkan_semantics)
 				SPIRV_CROSS_THROW("GL_EXT_nonuniform_qualifier is only supported in Vulkan GLSL.");
@@ -525,6 +530,11 @@ string CompilerGLSL::compile()
 {
 	if (options.vulkan_semantics)
 		backend.allow_precision_qualifiers = true;
+	else
+	{
+		// only NV_gpu_shader5 supports divergent indexing on OpenGL, and it does so without extra qualifiers
+		backend.nonuniform_qualifier = "";
+	}
 	backend.force_gl_in_out_block = true;
 	backend.supports_extensions = true;
 	backend.use_array_constructor = true;


spirv_glsl.cpp
```diff
@@ -501,6 +501,11 @@ void CompilerGLSL::find_static_extensions()
 		switch (cap)
 		{
 		case CapabilityShaderNonUniformEXT:
+			if (!options.vulkan_semantics)
+				require_extension_internal("GL_NV_gpu_program5");
```
HansKristian-Work, line 505: Should probably apply this to RuntimeDescriptorArrayEXT as well?

HansKristian-Work:
> actually for it to work, I'd have to remove (pass through) any `nonuniformEXT` calls in the generated GLSL shader

What do you mean by pass through here? If that EXT implies nonuniformEXT without you having to use it you can probably fiddle with https://github.com/KhronosGroup/SPIRV-Cross/blob/master/spirv_glsl.hpp#L424

spirv_glsl.cpp
devshgraphicsprogramming: well not really, since open ended arrays of descriptors (or any bindings) are not legal in OpenGL GLSL AFAIK.

basically this lets me divergently index SIZED arrays of descriptors on Nvidia GPUs, nothing more really

devshgraphicsprogramming:
> What do you mean by pass through here? If that EXT implies nonuniformEXT without you having to use it you can probably fiddle with https://github.com/KhronosGroup/SPIRV-Cross/blob/master/spirv_glsl.hpp#L424

On Nvidia GPUs the GLSL extension makes divergent adressing work transparently.

If you want to support this on other vendors then I have a NASTY little trick that works using `ARB_ballot`, in pseudocode
```glsl
T sampleTexture(samplerU tex, ...)
{
    T retval;
    uint outstandingSampleMask = ballotARB(true);
    // maybe unroll a few times manually
    while (outstandingSampleMask)
    {
        handle = readInvocationARB(tex,findLSB(outstandingSampleMask));
        bool canSample = handle==tex && (outstandingSampleMask&gl_SubGroupEqMaskARB)==gl_SubGroupEqMaskARB;
        outstandingSampleMask &= ~ballotARB(canSample);
        if (canSample)
            retval = texture(handle,...);
    }
    return retval;
}
```

devshgraphicsprogramming:
https://github.com/KhronosGroup/SPIRV-Cross/blob/master/spirv_glsl.hpp#L424

would have to be null/empty/"", but only when `!vulkanSemantics`

HansKristian-Work:
> If you want to support this on other vendors then I have a NASTY little trick that works using `ARB_ballot`, in pseudocode

No, I don't want that. Way too much complexity for such an obscure usecase. Supporting this by omitting nonuniformEXT is fine though.

devshgraphicsprogramming:
> 
> 
> https://github.com/KhronosGroup/SPIRV-Cross/blob/master/spirv_glsl.hpp#L424
> 
> would have to be null/empty/"", but only when `!vulkanSemantics`

^ no idea how to implement that ^

HansKristian-Work:
Override backend.nonuniform_literal here: https://github.com/KhronosGroup/SPIRV-Cross/blob/master/spirv_glsl.cpp#L530.

devshgraphicsprogramming:
> 
> 
> Override backend.nonuniform_literal here: https://github.com/KhronosGroup/SPIRV-Cross/blob/master/spirv_glsl.cpp#L530.

hmm actually it seems that my suggestion

> https://github.com/KhronosGroup/SPIRV-Cross/blob/master/spirv_glsl.hpp#L424
> would have to be null/empty/"", but only when !vulkanSemantics

would not solve the issue entirely.

devshgraphicsprogramming:
not sure how I'd make it just erase the `nonuniformEXT` function call generation all together.... should I just strip the decorations or something?

HansKristian-Work:
Setting backend.nonuniform_qualifier to "" should remove all of it.

spirv_glsl.cpp
devshgraphicsprogramming: are we resolved?

HansKristian-Work: yes

devshgraphicsprogramming:
done

PR #1337, opened 2020-04-21
Be a bit more careful what nonuniform state is propagated.

HansKristian-Work:


HansKristian-Work:
Trivial, just merging.

PR #1360, opened 2020-05-05
Fix #1359: MSL: If the packed type is scalar, don't emit "pack_" prefix.

### LINKED ISSUE #1359: MSL: bool array in struct is translated to packed_uint array which is not a valid type.
Hi,
I tried to translate this GLSL:
```
#version 450 core
struct Struct
{
    bool flags[1];
};

layout(set=0, binding=0, std140) uniform defaultUniformsVS
{
    Struct flags;
    vec2 uquad[4];
    mat4 umatrix;
};

layout (location = 0) in vec4 a_position;

void main()
{
    gl_Position = umatrix * vec4(uquad[gl_VertexIndex], a_position.z, a_position.w);
    if (flags.flags[0])
        gl_Position.z = 0.0;
}
```

And got this MSL:
```
#include <metal_stdlib>
#include <simd/simd.h>

using namespace metal;

struct Struct
{
    packed_uint flags[1];
};

struct defaultUniformsVS
{
    Struct flags;
    float4 uquad[4];
    float4x4 umatrix;
};

struct main0_out
{
};

struct main0_in
{
    float4 a_position [[attribute(0)]];
};

vertex main0_out main0(main0_in in [[stage_in]], constant defaultUniformsVS& _25 [[buffer(0)]], uint gl_VertexIndex [[vertex_id]])
{
    main0_out out = {};
    out.gl_Position = _25.umatrix * float4(_25.uquad[int(gl_VertexIndex)].x, _25.uquad[int(gl_VertexIndex)].y, in.a_position.z, in.a_position.w);
    if (uint(_25.flags.flags[0]) != 0u)
    {
        out.gl_Position.z = 0.0;
    }
    return out;
}
```

`bool` was translated to `packed_uint`. I don't think `packed_uint` is a valid type accepted by Metal compiler:
```
error: unknown type name 'packed_uint'
```
Only `packed_uint`2/3/4 are valid.

kakashidinho:
Scalar type is already packed in metal.

### DIFF
--- a/reference/shaders-msl/asm/vert/packed-bool-to-uint.asm.vert
+++ b/reference/shaders-msl/asm/vert/packed-bool-to-uint.asm.vert
@@ -1,0 +1,38 @@
+#include <metal_stdlib>
+#include <simd/simd.h>
+
+using namespace metal;
+
+struct Struct
+{
+    uint flags[1];
+};
+
+struct defaultUniformsVS
+{
+    Struct flags;
+    float4 uquad[4];
+    float4x4 umatrix;
+};
+
+struct main0_out
+{
+    float4 gl_Position [[position]];
+};
+
+struct main0_in
+{
+    float4 a_position [[attribute(0)]];
+};
+
+vertex main0_out main0(main0_in in [[stage_in]], constant defaultUniformsVS& _9 [[buffer(0)]], uint gl_VertexIndex [[vertex_id]])
+{
+    main0_out out = {};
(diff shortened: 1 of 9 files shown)


HansKristian-Work:
Seems like a simple fix. I'll need to analyze what is going on here and why the struct needs to be marked as packed to begin with.

kakashidinho:
I think this issue was introduced by 17ad62eea.
Previously validate_member_packing_rules_msl() could return false due to std140 array stride is different from msl array stride. Now this function will return true due to relax_array_stride.

HansKristian-Work:
Thanks! Yes, it was related to shenanigans around array stride. I'll also add some tweaks to not attempt to pack scalars in the first place.

PR #1364, opened 2020-05-16
Implement VK_KHR_ray_tracing in HLSL

maierfelix:
WIP:
 - Didn't tackle callable shaders yet
 - Tests - I hope someone can help on this

The trickiest part is that HLSL only accepts structs for hit attributes and payloads. When a primitive (such as float or vec4) is used, I fallback to temporary structs and copy the values around.

Also the ray payloads are passed into the entry function as *inout*, so are limited in their scope. To overcome this, I create a static global variable (to mimic GLSL) and copy the input and output back in the entry function.

Also I did some NV to KHR transitioning, hope that's fine, I can revert if not

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -345,18 +345,18 @@ static const char *execution_model_to_str(spv::ExecutionModel model)
 		return "fragment";
 	case ExecutionModelGLCompute:
 		return "compute";
-	case ExecutionModelRayGenerationNV:
-		return "raygenNV";
-	case ExecutionModelIntersectionNV:
-		return "intersectionNV";
-	case ExecutionModelCallableNV:
-		return "callableNV";
-	case ExecutionModelAnyHitNV:
-		return "anyhitNV";
-	case ExecutionModelClosestHitNV:
-		return "closesthitNV";
-	case ExecutionModelMissNV:
-		return "missNV";
+	case ExecutionModelRayGenerationKHR:
+		return "raygenKHR";
+	case ExecutionModelIntersectionKHR:
+		return "intersectionKHR";
+	case ExecutionModelCallableKHR:
+		return "callableKHR";
+	case ExecutionModelAnyHitKHR:
+		return "anyhitKHR";
+	case ExecutionModelClosestHitKHR:
+		return "closesthitKHR";
+	case ExecutionModelMissKHR:
+		return "missKHR";
 	default:
 		return "???";
 	}
--- a/spirv_hlsl.cpp
+++ b/spirv_hlsl.cpp
@@ -375,6 +375,9 @@ string CompilerHLSL::type_to_glsl(const SPIRType &type, uint32_t id)
 	case SPIRType::Sampler:
 		return comparison_ids.count(id) ? "SamplerComparisonState" : "SamplerState";
 
+	case SPIRType::AccelerationStructure:
+		return "RaytracingAccelerationStructure";
+
 	case SPIRType::Void:
 		return "void";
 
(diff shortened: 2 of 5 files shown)


main.cpp
```diff
@@ -345,18 +345,18 @@ static const char *execution_model_to_str(spv::ExecutionModel model)
...
-		return "intersectionNV";
-	case ExecutionModelCallableNV:
-		return "callableNV";
-	case ExecutionModelAnyHitNV:
-		return "anyhitNV";
-	case ExecutionModelClosestHitNV:
-		return "closesthitNV";
-	case ExecutionModelMissNV:
-		return "missNV";
+	case ExecutionModelRayGenerationKHR:
```
HansKristian-Work, line 348: KHR is still provisional, and we should not ship anything based on this extension until it is fully ratified. For now, please use NV. The enums are basically the same anyways.

spirv_hlsl.cpp
```diff
@@ -1196,6 +1357,52 @@ void CompilerHLSL::emit_resources()
 		emitted = true;
 	}
 
+	// HLSL requires ray payloads to be structs
+	if (execution.model == ExecutionModelClosestHitKHR || execution.model == ExecutionModelAnyHitKHR ||
+	    execution.model == ExecutionModelMissKHR)
+	{
+		auto *payload_var = get_ray_tracing_payload();
```
HansKristian-Work, line 1364: This seems a bit weird. There can be multiple output payloads in GLSL, each with their own location qualifier.

spirv_hlsl.cpp
```diff
@@ -3209,6 +3633,13 @@ void CompilerHLSL::emit_legacy_uniform(const SPIRVariable &var)
 
 void CompilerHLSL::emit_uniform(const SPIRVariable &var)
 {
+	// make rgen payload static
```
HansKristian-Work, line 3636: I think ray payloads could have it's own emit function, this feels a little clumsy.

main.cpp
maierfelix: Done

spirv_hlsl.cpp
maierfelix: get_ray_tracing_payload takes the incoming ray payload, I made this more obvious now

spirv_hlsl.cpp
maierfelix: Done

maierfelix:
I'd leave callable shaders out, someone with experience in using them should do the implementation.  

If you could create a reference test which I can base off, and also show how I can run the tests, that'd be great.

HansKristian-Work:
Ok, I'll try to write a more complete suite of tests for ray-tracing, the current test coverage is kind of lacking.

See https://github.com/KhronosGroup/SPIRV-Cross#testing for how to run the test suite.

HansKristian-Work:
Here: https://github.com/KhronosGroup/SPIRV-Cross/pull/1369. You should be able to copy them over to `shaders-hlsl` and use that as the basis for the HLSL implementation.

maierfelix:
Thanks!

> You place test shaders in shaders-hlsl/rchit/ (different folders for different shader types) and run the test scripts to create reference outputs, which are committed along with the PR.

Just to clear upfront, I'm using temporary variable identifiers [here](https://github.com/KhronosGroup/SPIRV-Cross/blob/8979b0bf4d72ee275d9cb22ee6445503634595c6/spirv_hlsl.cpp#L5679) for example. Isn't that becoming a problem when creating reference output files?

HansKristian-Work:
Why would that be a problem? RayDesc should be part of the "banned identifier list" though.

maierfelix:
I mean `ray_uid`

HansKristian-Work:
> I mean `ray_uid`

For temporaries like that you can reserve an ID for it with `uint32_t id = ir.increase_bound_by(1)` and use `to_name` on it to get IDs like `_1254`. These are guaranteed to not conflict with anything else.

HansKristian-Work:
Hi. It's been a while since last update. Should I expect further updates on this PR? If not, I might be able to continue on it.

maierfelix:
Hey sorry for not updating this, I've hit an unexpected chunk of work. It'd be great if you could continue on it

Issue #1368, opened 2020-05-19
HLSL: Support forcing readonly storage buffers to be UAV by ID


bbernhar:
I'm a Chromium developer working on WebGPU. We'd like to add functionality to extend `force_storage_buffer_as_uav` to work **per binding** instead of globally (all bindings). This ensures certain storage buffers that were declared _readonly_ in the shader but have storage usage (read/write) in the pipeline state still compile (ie. SRV(ti) + ByteAddressBuffer => UAV(ui) + RWByteAddressBuffer).

The change looks fairly straightforward - store a set of bindings which `set_hlsl_force_storage_buffer_as_uav(ID)` adds and then conditionally check it where `force_storage_buffer_as_uav` is used. The client would simply call `set_hlsl_force_storage_buffer_as_uav` before compiling.

Any concerns/comments of introducing this enhancement?

HansKristian-Work:
For raw buffers specifically, there isn't too much of a concern here. Fortunately, you cannot pass SSBOs around to functions, so munging with the type system is not a big deal. The only thing that should change is the declaration of the resource.

However, be aware that this cannot work for any type which *does* support being passed around to functions, like Buffer vs RWBuffer, Texture vs RWTexture, etc, since the function signature does not know anything about specific resources. I suspect you're going to go down this rabbit hole and figure out this approach is unimplementable. If you go that far, you'll need to add legalization passes yourself.

I earlier added readonly RWTexture -> Texture translation for WebGPU as well. Please see discussion here: https://github.com/KhronosGroup/SPIRV-Cross/issues/1286#issuecomment-594560682.

HansKristian-Work:
Before I commit to support this, please evaluate the implications above and verify that you're not going to design yourself into a corner. SPIRV-Cross will not support legalization passes or any complicated transforms to support this kind of thing. Replacing types is easy enough, but that's about as far as this approach can go.

bbernhar:
Thanks, @HansKristian-Work for your feedback. This would only need to support raw buffers (aka [RW]ByteAddressBuffer). I should have the PR for review soon.

Kangz:
Thanks @HansKristian-Work for the feedback. The situation for texture is indeed more complex and something we might run into in WebGPU later down the line. We might need to disallow such "implicit promotion" for storage textures.

HansKristian-Work:
Merged.

PR #1378, opened 2020-05-27
HLSL: Add option to treat certain SSBO bindings as UAV

bbernhar:
This change is the enhancement proposed in #1368. A new "per-binding" option flag can be specified and used when the global "all binding" option flag is unset. @HansKristian-Work PTAL, hopefully, this is sufficient.

### DIFF
--- a/spirv_hlsl.cpp
+++ b/spirv_hlsl.cpp
@@ -1966,7 +1966,7 @@ void CompilerHLSL::emit_buffer_block(const SPIRVariable &var)
 	if (is_uav)
 	{
 		Bitset flags = ir.get_buffer_block_flags(var);
-		bool is_readonly = flags.get(DecorationNonWritable) && !hlsl_options.force_storage_buffer_as_uav;
+		bool is_readonly = flags.get(DecorationNonWritable) && !is_hlsl_force_storage_buffer_as_uav(var.self);
 		bool is_coherent = flags.get(DecorationCoherent) && !is_readonly;
 		bool is_interlocked = interlocked_resources.count(var.self) > 0;
 		const char *type_name = "ByteAddressBuffer ";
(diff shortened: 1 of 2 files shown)


spirv_hlsl.cpp
```diff
@@ -5545,3 +5545,19 @@ CompilerHLSL::BitcastType CompilerHLSL::get_bitcast_type(uint32_t result_type, u
 
 	return BitcastType::TypeNormal;
 }
+
+bool CompilerHLSL::is_hlsl_force_storage_buffer_as_uav(ID binding) const
+{
+	if (hlsl_options.force_storage_buffer_as_uav)
+	{
+		return true;
+	};
```
HansKristian-Work, line 5554: Stray semicolon.

spirv_hlsl.hpp
```diff
@@ -114,6 +114,10 @@ class CompilerHLSL : public CompilerGLSL
 		// By default, a readonly storage buffer will be declared as ByteAddressBuffer (SRV) instead.
 		bool force_storage_buffer_as_uav = false;
 
+		// Similar to force_storage_buffer_as_uav but only forces set bindings.
+		// Use set_hlsl_force_storage_buffer_as_uav to specify the binding by ID.
+		bool force_storage_buffer_as_uav_by_id = false;
```
HansKristian-Work, line 119: I'm not sure why this needs to be a separate option. If you call set_hlsl_force_storage_buffer_as_uav that should be enough.

spirv_hlsl.hpp
```diff
@@ -186,6 +190,9 @@ class CompilerHLSL : public CompilerGLSL
 	void add_hlsl_resource_binding(const HLSLResourceBinding &resource);
 	bool is_hlsl_resource_binding_used(spv::ExecutionModel model, uint32_t set, uint32_t binding) const;
 
+	// Controls which storage buffer bindings will be forced to be declared as UAVs.
+	void set_hlsl_force_storage_buffer_as_uav(ID binding);
```
HansKristian-Work, line 194: Using ID here is a bit surprising, since everyone always wanted it to be based on set/binding instead. To get IDs you first need to perform reflection, where with set/binding you don't since you emit it straight from a descriptor set layout abstraction. I'd prefer if this API was based around set/binding.

bbernhar: Makes sense, SGTM.

PR #1388, opened 2020-06-08
GLSL: Support uint code for sparse residency query.

HansKristian-Work:


HansKristian-Work:
Trivial, tested locally.

PR #1389, opened 2020-06-08
GLSL: Handle the rest of GL_ARB_sparse_texture_clamp.

HansKristian-Work:
Missed these in initial sparse implementation.


HansKristian-Work:
Trivial, tested locally.

PR #1392, opened 2020-06-15
MSL: Fix up input variables' vector lengths in all stages.

cdavis5e:
Metal is picky about interface matching. If the types don't match
exactly, down to the number of vector components, Metal fails pipline
compilation. To support pipelines where the number of components
consumed by the fragment shader is less than that produced by the vertex
shader, we have to fix up the fragment shader to accept all the
components produced.

### DIFF
--- a/spirv_cross_c.cpp
+++ b/spirv_cross_c.cpp
@@ -1032,6 +1032,30 @@ spvc_result spvc_compiler_msl_add_vertex_attribute(spvc_compiler compiler, const
 #endif
 }
 
+spvc_result spvc_compiler_msl_add_shader_input(spvc_compiler compiler, const spvc_msl_shader_input *si)
+{
+#if SPIRV_CROSS_C_API_MSL
+	if (compiler->backend != SPVC_BACKEND_MSL)
+	{
+		compiler->context->report_error("MSL function used on a non-MSL backend.");
+		return SPVC_ERROR_INVALID_ARGUMENT;
+	}
+
+	auto &msl = *static_cast<CompilerMSL *>(compiler->compiler.get());
+	MSLShaderInput input;
+	input.location = si->location;
+	input.format = static_cast<MSLShaderInputFormat>(si->format);
+	input.builtin = static_cast<spv::BuiltIn>(si->builtin);
+	input.vecsize = si->vecsize;
+	msl.add_msl_shader_input(input);
+	return SPVC_SUCCESS;
+#else
+	(void)si;
+	compiler->context->report_error("MSL function used on a non-MSL backend.");
+	return SPVC_ERROR_INVALID_ARGUMENT;
+#endif
+}
+
 spvc_result spvc_compiler_msl_add_resource_binding(spvc_compiler compiler,
                                                    const spvc_msl_resource_binding *binding)
 {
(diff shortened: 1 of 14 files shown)


spirv_cross_c.cpp
```diff
@@ -1032,6 +1032,30 @@ spvc_result spvc_compiler_msl_add_vertex_attribute(spvc_compiler compiler, const
 #endif
 }
 
+spvc_result spvc_compiler_msl_add_shader_input(spvc_compiler compiler, const spvc_msl_shader_input *si)
```
HansKristian-Work, line 1035: Need to update API version as well.

spirv_cross_c.cpp
```diff
@@ -1032,6 +1032,30 @@ spvc_result spvc_compiler_msl_add_vertex_attribute(spvc_compiler compiler, const
...
+	{
+		compiler->context->report_error("MSL function used on a non-MSL backend.");
+		return SPVC_ERROR_INVALID_ARGUMENT;
+	}
+
+	auto &msl = *static_cast<CompilerMSL *>(compiler->compiler.get());
+	MSLShaderInput input;
+	input.location = si->location;
+	input.format = static_cast<MSLShaderInputFormat>(si->format);
+	input.builtin = static_cast<spv::BuiltIn>(si->builtin);
```
HansKristian-Work, line 1048: Builtin?

spirv_msl.hpp
```diff
@@ -456,8 +481,12 @@ class CompilerMSL : public CompilerGLSL
 	void set_argument_buffer_device_address_space(uint32_t desc_set, bool device_storage);
 
 	// Query after compilation is done. This allows you to check if a location or set/binding combination was used by the shader.
+	// Deprecated: use is_msl_shader_input_used() instead.
```
HansKristian-Work, line 484: There is a SPIRV_CROSS_DEPRECATED macro for this.

spirv_cross_c.cpp
HansKristian-Work: Oh, I guess it's there to replace vertex input API.

spirv_msl.cpp
```diff
@@ -93,7 +102,12 @@ void CompilerMSL::set_argument_buffer_device_address_space(uint32_t desc_set, bo
 
 bool CompilerMSL::is_msl_vertex_attribute_used(uint32_t location)
 {
-	return vtx_attrs_in_use.count(location) != 0;
+	return inputs_in_use.count(location) != 0;
```
HansKristian-Work, line 105: This should call is_msl_shader_input_used.

shaders-msl/frag/vecsize-mismtach.shader-inputs.frag
```diff
@@ -0,0 +1,17 @@
+#version 450
```
HansKristian-Work, line 1: Typo in file name.

HansKristian-Work:
I can tell that it's trying to remove components if vertex shader emits say 3 components and fragment shader only declares 2 components. What happens if you do this in reverse? E.g. vertex declares 2 components, but fragment reads 3? Is this case not handled because it is invalid SPIR-V? (Need to check if this is the case). I'd like this to be spelled out if that is the rationale behind not handling this case.

HansKristian-Work:
Found this in Vulkan spec:

```
Any input value to a shader stage is well-defined as long as the preceding stages writes to a
matching output, as described above.
Additionally, scalar and vector inputs are well-defined if there is a corresponding output satisfying
all of the following conditions:
• the input and output match exactly in decoration,
**• the output is a vector with the same basic type and has at least as many components as the
input, and**
• the common component type of the input and output is 16-bit integer or floating-point, or 32-bit
integer or floating-point (64-bit component types are excluded).
In this case, the components of the input will be taken from the first components of the output, and
any extra components of the output will be ignored.
```

spirv_cross_c.cpp
cdavis5e: I thought I did that... Fixed.

spirv_cross_c.cpp
cdavis5e: Yeah, this is a replacement for the vertex attribute API. I'm probably going to use it in #1294.

spirv_msl.hpp
cdavis5e: Done. I couldn't put it on the deprecated enums, because I don't think every compiler supports putting attributes on enumerants.

cdavis5e:
As you discovered, it *is* valid for a shader to consume fewer components than the previous shader produced, but *in*valid to consume more. In that case, it will likely break, or worse in the case of a tessellation shader, run but give broken results.

cdavis5e:
Hmm... The deprecation attributes I added caused Travis to fail. Do we need to build `spirv_cross_c.cpp` with `-Wno-deprecated-declarations`?

PR #1416, opened 2020-07-01
MSL: Workaround broken scalar access chain behavior in LLVM IR / AIR

### LINKED ISSUE #1415: GLSL compute shader that writes to same vec4 from different work items fails after translation to Metal Shading Language
Dear SPIRV-Cross-Team,

you might want to check this issue, where SPIRV-Cross converts a GLSL computer shader to a MSL shader that doesn't work:

https://github.com/KhronosGroup/MoltenVK/issues/942

It did not become clear who should fix the problem, that's why I report it here too, maybe you have a good solution.

Best regards,
Elmar

HansKristian-Work:
On Metal, vector types are declared as actual vectors, and LLVM simply refuses to access chain into a vector. Instead, if we attempt to do something like:

```
device float4 &v; v.x = 10.0;
```
we end up loading the entire vector, modifying it, and then storing everything back. This is not documented by MSL, but it is actual behavior observed.

This is broken in a concurrent environment where multiple threads can access the same vector, but different components. To mitigate this problem, we will instead cast such objects into pointer-to-scalar which enforces scalar access.

The alternative would be to emit packed vector types in these situations, but this is extremely annoying and would lead to extremely difficult edge cases in matrix packing and threadgroup storage. Casting pointers on-demand is by far the most robust and simple solution I can come up with.

Fix #1415.


HansKristian-Work:
@billhollings @cdavis5e I believe I found the best solution we can hope for here. I have limited capacity to test the implication of this change, so please review / test.

HansKristian-Work:
Ok, tested this a bit locally. Without this PR, I can indeed reproduce that scalar writes to vector are broken on Intel iGPU. With this PR, my ad-hoc testing works fine.

HansKristian-Work:
Going to merge this now, I don't anticipate any issues.

reference/shaders-msl-no-opt/comp/glsl.std450.comp
```diff
@@ -199,50 +199,50 @@ inline T spvFaceForward(T n, T i, T nref)
...
-    _19.res = exp2(_19.f32.x);
-    _19.res = log2(_19.f32.x);
-    _19.res = sqrt(_19.f32.x);
-    _19.res = rsqrt(_19.f32.x);
-    _19.res = abs(_19.f32.x);
-    _19.res = abs(_19.f32.x - _19.f32.y);
-    _19.res = sign(_19.f32.x);
-    _19.res = spvFaceForward(_19.f32.x, _19.f32.y, _19.f32.z);
-    _19.res = spvReflect(_19.f32.x, _19.f32.y);
-    _19.res = spvRefract(_19.f32.x, _19.f32.y, _19.f32.z);
```
kikaxa, line 239: @HansKristian-Work Hi!
Was looking through this PR and got to wonder if it can adversely affect the performance of such load constructs?
E.g.  a lot of per-component loads to the same vector, perhaps distributed in code. I'm concerned whether the compiler manages to coalesce these as well as before.

HansKristian-Work: It might be better since Apple's compiler refuses to consider scalar access chains at all. What could happen is that it ends up loading the full vector and only uses one component. There isn't much we can do about this, we're forced to emit this weird code due to buggy compilers.

HansKristian-Work: These patterns should already be dealt with at the SPIR-V level I think if the code is loading from the same data multiple times like this.

PR #1417, opened 2020-07-03
GLSL: Support multi-level struct flattening for I/O.

### LINKED ISSUE #351: GLSL to ES exception: Cannot flatten struct inside structs in I/O variables.
When cross-compiling the following shader to GLSL es 1.00, SPIRV-Cross throws an exception
which says: "Cannot flatten struct inside structs in I/O variables."

#version 450
struct VertexAttributes
{
	vec3 VertexColor ;
};

in vec3 VertexColor;

out vertexToFragment
{
	VertexAttributes attr;
} vertexOut;

in vertexToFragment
{
	VertexAttributes attr;	
} fragmentIn;

void entry_vert()
{
	vertexOut.attr.VertexColor = VertexColor;
}

out vec3 fragColor;
void entry_frag()
{
	fragColor = fragmentIn.attr.VertexColor;
}

Whereas cross-compiling it to desktop glsl, there is another issue: The shader has 2 interface blocks with exact same name but different storage qualifiers. Instead of output the names as is, SPIRV-Cross changed one of them, from "vertexToFragment" to "_24".

HansKristian-Work:
Fix #351.

### DIFF
--- a/reference/opt/shaders/legacy/fragment/multiple-struct-flattening.legacy.frag
+++ b/reference/opt/shaders/legacy/fragment/multiple-struct-flattening.legacy.frag
@@ -1,0 +1,36 @@
+#version 100
+precision mediump float;
+precision highp int;
+
+struct Foo
+{
+    highp vec4 a;
+    highp vec4 b;
+};
+
+struct Bar
+{
+    highp vec4 a;
+    highp vec4 b;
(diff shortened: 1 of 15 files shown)


HansKristian-Work:
Tested locally.

PR #1418, opened 2020-07-06
MSL: Fix mismatching input attachment indices

troughton:
When using e.g. input attachment index 1 without also using input attachment index 0 the incorrect `color(n)` value would be emitted for the input (i.e. `color(0)` instead of `color(1)`). Fix this by using the `InputAttachmentIndex` decoration to determine `n`.


HansKristian-Work:
Merged a more correct fix, thanks for bringing this to my attention.

PR #1419, opened 2020-07-06
MSL: Use input attachment index directly for resource index fallback.

HansKristian-Work:
Supersedes #1418.


HansKristian-Work:
Tested locally.

PR #1429, opened 2020-07-22
Enabling setting an additional fixed sampleMask in Metal fragment shaders.

ponitka:
In Metal render pipelines don't have an option to set a sampleMask
parameter, the only way to get that functionality is to set the
sample_mask output of the fragment shader to this value directly.
We also need to take care to combine the fixed sample mask with the
one that the shader might possibly output.

### DIFF
--- a/spirv_msl.cpp
+++ b/spirv_msl.cpp
@@ -156,9 +156,10 @@ void CompilerMSL::build_implicit_builtins()
 	    (active_input_builtins.get(BuiltInVertexId) || active_input_builtins.get(BuiltInVertexIndex) ||
 	     active_input_builtins.get(BuiltInBaseVertex) || active_input_builtins.get(BuiltInInstanceId) ||
 	     active_input_builtins.get(BuiltInInstanceIndex) || active_input_builtins.get(BuiltInBaseInstance));
+	bool need_sample_mask = msl_options.additional_fixed_sample_mask != 0xffffffff;
 	if (need_subpass_input || need_sample_pos || need_subgroup_mask || need_vertex_params || need_tesc_params ||
 	    need_multiview || need_dispatch_base || need_vertex_base_params || need_grid_params ||
-	    needs_subgroup_invocation_id)
+	    needs_subgroup_invocation_id || need_sample_mask)
 	{
 		bool has_frag_coord = false;
 		bool has_sample_id = false;
@@ -173,12 +174,27 @@ void CompilerMSL::build_implicit_builtins()
 		bool has_view_idx = false;
 		uint32_t workgroup_id_type = 0;
 
+		// FIXME: Investigate the fact that there are no checks for the entry point interface variables.
 		ir.for_each_typed_id<SPIRVariable>([&](uint32_t, SPIRVariable &var) {
-			if (var.storage != StorageClassInput || !ir.meta[var.self].decoration.builtin)
+			if (!ir.meta[var.self].decoration.builtin)
 				return;
 
 			// Use Metal's native frame-buffer fetch API for subpass inputs.
 			BuiltIn builtin = ir.meta[var.self].decoration.builtin_type;
+
+			if (var.storage == StorageClassOutput)
+			{
+				if (need_sample_mask && builtin == BuiltInSampleMask)
+				{
+					builtin_sample_mask_id = var.self;
+					mark_implicit_builtin(StorageClassOutput, BuiltInSampleMask, var.self);
+					does_shader_write_sample_mask = true;
+				}
+			}
+
+			if (var.storage != StorageClassInput)
+				return;
+
 			if (need_subpass_input && (!msl_options.is_ios() || !msl_options.ios_use_framebuffer_fetch_subpasses) &&
 			    builtin == BuiltInFragCoord)
 			{
@@ -560,6 +576,26 @@ void CompilerMSL::build_implicit_builtins()
 			}
 			set_name(var_id, "spvDispatchBase");
 			builtin_dispatch_base_id = var_id;
+		}
+
+		if (need_sample_mask && !does_shader_write_sample_mask)
+		{
+			uint32_t offset = ir.increase_bound_by(2);
+			uint32_t var_id = offset + 1;
+
+			// Create gl_SampleMask.
+			SPIRType uint_type_ptr_out;
+			uint_type_ptr_out = get_uint_type();
+			uint_type_ptr_out.pointer = true;
+			uint_type_ptr_out.parent_type = get_uint_type_id();
+			uint_type_ptr_out.storage = StorageClassOutput;
+			
+			auto &ptr_out_type = set<SPIRType>(offset, uint_type_ptr_out);
+			ptr_out_type.self = get_uint_type_id();
+			set<SPIRVariable>(var_id, offset, StorageClassOutput);
+			set_decoration(var_id, DecorationBuiltIn, BuiltInSampleMask);
+			builtin_sample_mask_id = var_id;
+			mark_implicit_builtin(StorageClassOutput, BuiltInSampleMask, var_id);
 		}
 	}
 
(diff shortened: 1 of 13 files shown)


spirv_msl.hpp
```diff
@@ -787,6 +792,8 @@ class CompilerMSL : public CompilerGLSL
 	uint32_t view_mask_buffer_id = 0;
 	uint32_t dynamic_offsets_buffer_id = 0;
 
+	bool does_shader_write_sample_mask = false;
```
HansKristian-Work, line 795: The API user shouldn't have to declare if the shader writes the sample mask. That's up to SPIRV-Cross to figure out.

spirv_msl.cpp
```diff
@@ -10062,6 +10105,37 @@ void CompilerMSL::fix_up_shader_inputs_outputs()
 				break;
 			}
 		}
+		else if (var.storage == StorageClassOutput && is_builtin_variable(var))
+		{
+			auto &entry_func = this->get<SPIRFunction>(ir.default_entry_point);
+			switch (bi_type)
+			{
+				case BuiltInSampleMask:
```
HansKristian-Work, line 10113: This switch should be a branch.

spirv_msl.cpp
```diff
@@ -149,8 +149,9 @@ void CompilerMSL::build_implicit_builtins()
 	bool need_dispatch_base =
 	    msl_options.dispatch_base && get_execution_model() == ExecutionModelGLCompute &&
 	    (active_input_builtins.get(BuiltInWorkgroupId) || active_input_builtins.get(BuiltInGlobalInvocationId));
+	bool need_sample_mask = msl_options.additional_fixed_sample_mask != 0xFFFFFFFF;
```
HansKristian-Work, line 152: Use lower-case for the mask.

spirv_msl.hpp
Kangz: I got tripped by the same thing: `does_shader_write_sample_mask ` isn't part of the options but is a protected member of` CompilerMSL` and is set to `true` when the SPIRV-Cross finds that the shader uses sampleMask already.

ponitka: What @Kangz said is true. We actually need to know whether the shader writes anything to the sample_mask at the moment when we run <CompilerMSL::fix_up_shader_inputs_outputs> and we can't just check if SampleMask is an active builtin then, because if we create it ourselves then we run mark_implicit_builtin for it too.

spirv_msl.cpp
ponitka: Done

spirv_msl.cpp
ponitka: Done

spirv_msl.cpp
```diff
@@ -561,6 +576,32 @@ void CompilerMSL::build_implicit_builtins()
...
 		}
+
+		if (need_sample_mask && !does_shader_write_sample_mask)
+		{
+			uint32_t offset = ir.increase_bound_by(3);
+			uint32_t type_id = offset;
+			uint32_t var_id = offset + 2;
+
+			// Create gl_SampleMask.
+			SPIRType uint_type;
```
HansKristian-Work, line 587: This should use the get_uint_type_id() stuff that was merged a few hours ago.

spirv_msl.cpp
```diff
@@ -166,11 +167,25 @@ void CompilerMSL::build_implicit_builtins()
...
 			// Use Metal's native frame-buffer fetch API for subpass inputs.
 			BuiltIn builtin = ir.meta[var.self].decoration.builtin_type;
+
+			if (var.storage == StorageClassOutput)
+			{
+				if (need_sample_mask && builtin == BuiltInSampleMask)
+				{
+					builtin_sample_mask_id = var.self;
+					mark_implicit_builtin(StorageClassOutput, BuiltInSampleMask, var.self);
+					does_shader_write_sample_mask = true;
```
HansKristian-Work, line 182: So I suppose this will be set to true if shader merely declares the SampleMask variable (even if it's unused)?

ponitka: Yes. I'm not sure what's the behaviour that we want in this case. If declaring the SampleMask as an output and not writing to it anything means that the output will be 0, then we need to assume that the shader writes to the sample mask.

HansKristian-Work: Perhaps emit an input fixup to force sample mask to 0xffffffff as well?

HansKristian-Work: In that case, the variable should be called does_shader_declare_sample_mask perhaps to avoid confusion.

ponitka: Emitting an input fixup might change the current behaviour of shaders if what I mentioned is true and declaring but not explicitly writing to the sample_mask means writing 0 to it. I don't know if this is the desired behaviour though.

HansKristian-Work: I think you can look at the entry point declaration to see if the SampleMask is used.

```
A variable in the Output storage class decorated with SampleMask is an array of integers forming a
bit array in a manner similar an input variable decorated with SampleMask, but where each bit
represents coverage as computed by the shader. Modifying the sample mask by writing zero to a
Chapter 14. Shader Interfaces | 795bit of SampleMask causes the sample to be considered uncovered. If this variable is also decorated
with OverrideCoverageNV, the fragment coverage is replaced with the sample mask bits set in the
shader otherwise the fragment coverage is ANDed with the bits of the sample mask. If the
fragment shader is being evaluated at any frequency other than per-fragment, bits of the sample
mask not corresponding to the current fragment shader invocation are ignored. This array must
be sized in the fragment shader either implicitly or explicitly, to be no larger than the
implementation-dependent maximum sample-mask (as an array of 32-bit elements), determined
by the maximum number of samples. If a fragment shader entry point’s interface includes an
output variable decorated with SampleMask, the sample mask will be undefined for any array
elements of any fragment shader invocations that fail to assign a value. If a fragment shader
entry point’s interface does not include an output variable decorated with SampleMask, the
sample mask has no effect on the processing of a fragment.
```

HansKristian-Work: So, an input fixup is not necessary. The application must assign something to it if it exists in the entry point. If it does not exist in the entry point, then we need to declare our own sample mask.

ponitka: Thanks, I think the current code is fine in this case.

HansKristian-Work: I don't see any check for the entry point interface variables though, this code will iterate over all variables in the module if I understand it correctly?

ponitka: It seems so, it's doing the same thing as for the other builtin variables. There are no entry point interface checks for BuiltInSampleID either.

HansKristian-Work: Hm, ok. Please just add a FIXME then so we don't forget that this might be something to revisit later.

ponitka: Added!

ponitka:
> * I would like CLI support for setting the sample mask, and some tests to go along with this PR. The tests would be:
> * Mask set in API, shader does not write sample mask.
> * Mask not set in API, shader writes sample mask.
> * Mask set in API and shader writes sample mask.

I added the CLI support. Can you give me a quick overview of where the tests should be?

HansKristian-Work:
Place shaders in `shaders-msl/frag/suitable-name.frag` and run test_shaders.sh. Reference shaders will appear in `reference/` and `reference/opt`. You can have the path include something like `.sample-mask.` and have `test_shaders.py` check that which would add a suitable argument to the CLI when running the test.

ponitka:
Thanks for explaining.

> * Mask set in API, shader does not write sample mask.
> * Mask set in API and shader writes sample mask.

Added those.

> * Mask not set in API, shader writes sample mask.

This was already covered by shaders-msl/frag/sample-mask.frag.

There's also a new test with the sample mask used both as an input and an output.

spirv_msl.cpp
ponitka: Done. I hope that's the correct way to merge stuff.

HansKristian-Work: I think you need a rebase.

HansKristian-Work: Also, there was an update of glslang/spirv-tools references in the interim as well, so beware that you probably need to rebuild that locally to create the reference test files.

ponitka: Should be fine now.

Kangz:
@austinEng FYI

PR #1434, opened 2020-07-28
GLSL: Add option to force flattening IO blocks.

### LINKED ISSUE #1427: GLSL: Opt-in to flatten I/O blocks
I think it's better to always (or by flag in options) flatten I/O blocks if they are not directly supported by specified glsl version than to rely on the GL_EXT_shader_io_blocks extension. Currently I/O blocks may not be supported on es 310 (if extension is not presented), but are supported on older versions through flattening. It's a bit strange.

HansKristian-Work:
It is not always desirable to use actual blocks.
A prime example in the case where EXT_shader_io_blocks is not supported
on the target implementation.

Fix #1427.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -323,7 +323,7 @@ if (SPIRV_CROSS_STATIC)
 endif()
 
 set(spirv-cross-abi-major 0)
-set(spirv-cross-abi-minor 36)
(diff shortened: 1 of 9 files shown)


HansKristian-Work:
Benign failure on AppVeyor, consider it passing.

PR #1436, opened 2020-07-29
GLSL/HLSL: Be more exhaustive with aliasing struct type declarations.

### LINKED ISSUE #1408: A feature for flattening inside interface blocks
Is it possible to introduce a feature for flattening all members that are of a structure type inside interface blocks. It will simplify a lot name mangling (unification) task in glsl where linking by name. Name mangling is necessary if the debug info was stripped and if you want the linker not to rely on names.
Below the example of flattening:
```
struct ExampleInnerStruct
{
	vec4 a;
};

struct ExampleStruct
{
	vec4 a;
	ExampleInnerStruct b;
};

layout(location = 0) in InputBlock
{
	ExampleStruct a;
	vec4 b;
};

layout(binding = 0) uniform UniformBlock
{
	TestStruct a;
	vec4 b;
	vec4 c;
};
```
transform to:
```
layout(location = 0) in InputBlock
{
	vec4 a_a;
	vec4 a_b_a;
	vec4 b;
};

layout(binding = 0) uniform UniformBlock
{
	vec4 a_a;
	vec4 a_b_a;
	vec4 b;
	vec4 c;
};
```

HansKristian-Work:
Fixes some issues which now allows the same struct type to be declared as just one struct in all cases, not just buffer block / value type aliasing cases. This makes it far easier to link GLSL shaders by name, since type names and member names need to match up exactly.

Fix #1408.

### DIFF
--- a/reference/opt/shaders-hlsl/asm/comp/block-name-alias-global.asm.comp
+++ b/reference/opt/shaders-hlsl/asm/comp/block-name-alias-global.asm.comp
@@ -4,22 +4,16 @@ struct A
     int b;
 };
 
-struct A_1
-{
-    int a;
-    int b;
-};
-
 RWByteAddressBuffer C1 : register(u1);
 cbuffer C2 : register(b2)
 {
-    A_1 C2_1_Data[1024] : packoffset(c0);
+    A C2_1_Data[1024] : packoffset(c0);
 };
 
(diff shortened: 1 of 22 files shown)


HansKristian-Work:
Benign failure on AppVeyor, passed in other run.

PR #1446, opened 2020-08-13
Fix #1445: MSL: Enclose args when convert distance(a,b) to abs(a-b)

### LINKED ISSUE #1445: MSL: expressions are not enclosed when converting distance(a,b) -> abs(a-b)
This is the source GLSL:
```
#version 450 core
layout(location = 0) in vec4 aPosition;
layout(location = 0) out float vDist;
void main()
{
   gl_Position = aPosition;
   vDist = distance(aPosition.x - 1.0, aPosition.y - 1.0);
}

```

This is converted MSL:
```
#include <metal_stdlib>
#include <simd/simd.h>

using namespace metal;

struct main0_out
{
    float vDist [[user(locn0)]];
};

struct main0_in
{
    float4 aPosition [[attribute(0)]];
};

vertex main0_out main0(main0_in in [[stage_in]])
{
    main0_out out = {};
    out.gl_Position = in.aPosition;
    out.vDist = abs(in.aPosition.x - 1.0 - in.aPosition.y - 1.0);
    return out;
}
```

Note `distance(aPosition.x - 1.0, aPosition.y - 1.0)` is translated to
         `abs(in.aPosition.x - 1.0 - in.aPosition.y - 1.0)` which is wrong.
It should be:
- `abs(in.aPosition.x - 1.0 - in.aPosition.y + 1.0)`
- or `abs((in.aPosition.x - 1.0) - (in.aPosition.y - 1.0))`

kakashidinho:


kakashidinho:
What is the difference between `reference` and `reference/opt`

cdavis5e:
> What is the difference between `reference` and `reference/opt`?

The second set of reference results has `spirv-opt` run on the SPIR-V before running SPIRV-Cross.

PR #1450, opened 2020-08-21
Overhaul how we deal with reserved identifiers.

### LINKED ISSUE #1440: MSL: Why do names starting with '_' followed by a digit e.g. "_1PixelColor" get changed to _m<num>?
Hi,
Do you reserve names starting with '_' followed by a digit for something? from what I've seen these names are valid MSL, this is kind of a hidden gotcha, spent today debugging why our UI is misbehaving, only to find out that calling:
`spvc_compiler_get_member_name(spvCompiler, bufferResource.base_type_id, j);` on a variable named  `_1PixelClipInvView`returns an empty string. It seems that the names are changed to "_m" followed by their index int the UBO, so it looks intentional, however haven't found this documented anywhere.
The reflection result seems off to me, but I might not know the full context.
(Edit: accidentally submitted the issue a bit early so adding a repro on shader playground http://shader-playground.timjones.io/9734fd726901260175252c5c3a9bc812, can spend some time on a better reproduction shader if necessary)

HansKristian-Work:
- Do not silently drop reserved identifiers in the parser. This makes it
  possible to reflect identifiers which are reserved by the
  cross-compiler module.
- Instead of dropping the name, emit _RESERVED_IDENTIFIER_FIXUP in the
  source to make it clear that a name has been rewritten.
- Document what is reserved and not.

Fix #1440


eddyb:
I believe this may have fixed a bug where MSL where this was replacing the sanitized entry point name with the original one: https://github.com/KhronosGroup/SPIRV-Cross/blob/e50f7d1ce8e162d0c826e84168cfa234e4de4ec9/spirv_msl.cpp#L12188-L12189

(In Rust-GPU we ended up with entry-point names like `foo::bar`, from module nesting - like C++ namespaces)

But now that the sanitizing happens late, as opposed to on `set_name`, MSL's replacement shouldn't have any effects anymore.
Haven't confirmed it's a fix tho tho, only just tracked it down from the interaction of MoltenVK and Rust-GPU (which outputs SPIR-V).

PR #1454, opened 2020-09-01
MSL: Don't set the layer for multiview if the device doesn't support it.

cdavis5e:
Some older iOS devices don't support layered rendering. In that case,
don't set `[[render_target_array_index]]`, because the compiler will
reject the shader in that case. The client will then have to unroll the
render pass manually.

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -549,6 +549,7 @@ struct CLIArguments
 	bool msl_invariant_float_math = false;
 	bool msl_emulate_cube_array = false;
 	bool msl_multiview = false;
+	bool msl_multiview_layered_rendering = true;
 	bool msl_view_index_from_device_index = false;
 	bool msl_dispatch_base = false;
 	bool msl_decoration_binding = false;
@@ -732,6 +733,8 @@ static void print_help_msl()
 	                "\t[--msl-device-argument-buffer <descriptor set index>]:\n\t\tUse device address space to hold indirect argument buffers instead of constant.\n"
 	                "\t\tComes up when trying to support argument buffers which are larger than 64 KiB.\n"
 	                "\t[--msl-multiview]:\n\t\tEnable SPV_KHR_multiview emulation.\n"
+	                "\t[--msl-multiview-no-layered-rendering]:\n\t\tDon't set [[render_target_array_index]] in multiview shaders.\n"
+	                "\t\tUseful for devices which don't support layered rendering. Only effective when --msl-multiview is enabled.\n"
 	                "\t[--msl-view-index-from-device-index]:\n\t\tTreat the view index as the device index instead.\n"
 	                "\t\tFor multi-GPU rendering.\n"
 	                "\t[--msl-dispatch-base]:\n\t\tAdd support for vkCmdDispatchBase() or similar APIs.\n"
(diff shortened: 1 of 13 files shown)


cdavis5e:
I just noticed that the reference output is as though the option weren't used. Something isn't right here...

main.cpp
```diff
@@ -732,6 +733,8 @@ static void print_help_msl()
 	                "\t[--msl-device-argument-buffer <descriptor set index>]:\n\t\tUse device address space to hold indirect argument buffers instead of constant.\n"
 	                "\t\tComes up when trying to support argument buffers which are larger than 64 KiB.\n"
 	                "\t[--msl-multiview]:\n\t\tEnable SPV_KHR_multiview emulation.\n"
+	                "\t[--msl-disable-layered-rendering]:\n\t\tDon't set [[render_target_array_index]] in multiview shaders.\n"
+					"\t\tUseful for devices which don't support layered rendering. Only effective when --msl-multiview is enabled.\n"
```
HansKristian-Work, line 737: Formatting. Need to do it manually since the block is marked format off.

HansKristian-Work: Probably call this something else to make it clear this is about multiview. I didn't understand it before I read this description.

HansKristian-Work (changes requested):
Seems fine other than the nit comments.

HansKristian-Work:
> I just noticed that the reference output is as though the option weren't used. Something isn't right here...

Is this still true? Seemed fine to me ...

cdavis5e:
> > I just noticed that the reference output is as though the option weren't used. Something isn't right here...
> 
> Is this still true? Seemed fine to me ...

I wrote that comment before fixing it and force-pushing it. You didn't see it before, when it was broken.

test_shaders.py
```diff
@@ -232,6 +232,8 @@ def cross_compile_msl(shader, spirv, opt, iterations, paths):
         msl_args.append('--emit-line-directives')
     if '.multiview.' in shader:
         msl_args.append('--msl-multiview')
+    if '.no-layered.' in shader:
+        msl_args.append('--msl-disable-layered-rendering')
```
HansKristian-Work, line 236: Sure?

HansKristian-Work (commented):
Sure this passes?

test_shaders.py
HansKristian-Work: Shouldn't that be --msl-multiview-disable-layered-rendering?

cdavis5e: Oops. Fixed.

cdavis5e:
How come it's adding `gl_InstanceIndex` and `gl_BaseInstance` here, but not on my machine?

cdavis5e:
> How come it's adding `gl_InstanceIndex` and `gl_BaseInstance` on Travis, but not on my machine?

Oh I know why. It's because this branch was made without #1453.

cdavis5e:
> > How come it's adding `gl_InstanceIndex` and `gl_BaseInstance` on Travis, but not on my machine?
> 
> Oh I know why. It's because this branch was made without #1453.

More correctly, because I forgot to have SPIRV-Cross not emit those builtins in no-layered mode, and the reference outputs on my machine were generated without #1453. I fixed it not to emit those unneeded builtins now.

PR #1459, opened 2020-09-02
MSL: Also AND input sample mask with the additional fixed mask.

cdavis5e:


cdavis5e:
I'm not even sure this change is needed. It doesn't make any difference in the CTS.

PR #1461, opened 2020-09-04
Fix -Wduplicate-enum and -Wrange-for-analysis.

Kangz:
PTAL @HansKristian-Work

### DIFF
--- a/spirv_reflect.cpp
+++ b/spirv_reflect.cpp
@@ -651,7 +651,7 @@ void CompilerReflection::emit_specialization_constants()
 		return;
 
 	json_stream->emit_json_key_array("specialization_constants");
-	for (const auto spec_const : specialization_constants)
+	for (const auto &spec_const : specialization_constants)
 	{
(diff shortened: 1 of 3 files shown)


spirv_reflect.cpp
```diff
@@ -651,7 +651,7 @@ void CompilerReflection::emit_specialization_constants()
 		return;
 
 	json_stream->emit_json_key_array("specialization_constants");
-	for (const auto spec_const : specialization_constants)
+	for (const auto& spec_const : specialization_constants)
```
HansKristian-Work, line 654: Why? Also, formatting, & binds to right.

HansKristian-Work (changes requested):
Minor nit.

spirv_reflect.cpp
Kangz: `-Wrange-for-analysis` complains because `decltype(spec_const)` is a structure with more than one member, so it is more efficient to use `const &`

HansKristian-Work (approved):

PR #1463, opened 2020-09-14
Implement a simple evaluator of specialization constants.

### LINKED ISSUE #1462: [GLSL] Exception inside get_active_buffer_ranges on compute shader
"Bad cast" exception is thrown inside get_active_buffer_ranges function for the following shader:
https://github.com/SaschaWillems/Vulkan/blob/master/data/shaders/glsl/computecullandlod/cull.comp
https://github.com/SaschaWillems/Vulkan/blob/master/data/shaders/glsl/computecullandlod/cull.comp.spv

The ID of the buffer provided as argument to get_active_buffer_ranges is 0x54.
resource.id = 0x54
resource.type_id = 0x53
resource.base_type_id = 0x52
resource,name = "UBOOut"

HansKristian-Work:
In some cases, we need to get a literal value from a spec constant op.
Mostly relevant when emitting buffers, so implement a 32-bit integer
scalar subset of the evaluator. Can be extended as needed to support
evaluating any specialization constant operation.

Fix #1462.

### DIFF
--- a/reference/shaders-no-opt/comp/specialization-constant-evaluation.comp
+++ b/reference/shaders-no-opt/comp/specialization-constant-evaluation.comp
@@ -1,0 +1,321 @@
+#version 450
+layout(local_size_x = 1, local_size_y = 1, local_size_z = 1) in;
+
+#ifndef SPIRV_CROSS_CONSTANT_ID_2
+#define SPIRV_CROSS_CONSTANT_ID_2 1
+#endif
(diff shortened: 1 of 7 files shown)


shangjiaxuan:
I am making a pr at https://github.com/KhronosGroup/SPIRV-Reflect/pull/154, and would like to use test glsl code from here. Is this okay? Both repos are under khronos, so I assumed it was okay. But on second thought, it should be better to ammend the commit to add co-author.

Issue #1352, opened 2020-04-28
GLSL: Support `GL_KHR_shader_subgroup_ballot`


devshgraphicsprogramming:
`GL_KHR_shader_subgroup_ballot` is 1:1 match with OpenGL's `GL_ARB_shader_ballot` except that the masks and bitfileds are implemented as `uint64_t` not `uvec4`.

This makes it require `GL_ARB_gpu_shader_int64`, and not any of  `GL_AMD_gpu_shader_int64` or `GL_NV_gpu_shader5` because the `GL_ARB_shader_ballot` extension spec explicitly states the requirement for `GL_ARB_gpu_shader_int64`.

Some of functionality can also be implemented via `GL_NV_shader_thread_group` which if available should be preferred because it uses the `uint` type and all bitmask operations are 2x less complex.

The masks are easily defined as well as the ballot operations
```
#extension ARB_gpu_shader5 : require
#extension GL_NV_shader_thread_group : enable
#extension GL_NV_shader_thread_shuffle : enable
#ifdef GL_NV_shader_thread_group 
    gl_SubgroupEqMask = uvec4(gl_ThreadEqMaskNV,0u,0u,0u);
    gl_SubgroupGeMask = uvec4(gl_ThreadGeMaskNV,0u,0u,0u);
    gl_SubgroupGtMask = uvec4(gl_ThreadGtMaskNV,0u,0u,0u);
    gl_SubgroupLeMask = uvec4(gl_ThreadLeMaskNV,0u,0u,0u);
    gl_SubgroupLtMask = uvec4(gl_ThreadLtMaskNV,0u,0u,0u);

    subgroupBallot = uvec4(ballotThreadNV,0u,0u,0u)
#else
    #extension GL_ARB_shader_ballot : require
    #extension GL_ARB_gpu_shader_int64 : require

    gl_SubgroupEqMask = uvec4(unpackUint2x32(gl_SubGroupEqMaskARB),0u,0u);
    gl_SubgroupGeMask = uvec4(unpackUint2x32(gl_SubGroupGeMaskARB),0u,0u);
    gl_SubgroupGtMask = uvec4(unpackUint2x32(gl_SubGroupGtMaskARB),0u,0u);
    gl_SubgroupLeMask = uvec4(unpackUint2x32(gl_SubGroupLeMaskARB),0u,0u);
[...]

Issue #1351, opened 2020-04-28
GLSL: Support `GL_KHR_shader_subgroup_basic`


devshgraphicsprogramming:
Equivalents to `GL_KHR_shader_subgroup_basic`, are `GL_ARB_shader_ballot` or `GL_NV_shader_thread_group`. Somewhat involved can be `AMD_gcn_shader` but it cannot save us from requiring  `GL_ARB_shader_ballot`.

This extension is a bit complex to support, it depends on whether:
- We're only using the subgroup masks and sizes
- We are using `subgroupElect`
- We require memory barriers

## Subgroup Masks

Are relatively trivial to implement:

Pardon the abuse of notation that I use `vk_` prefix for vulkan semantic GLSL built-ins
```glsl
#extension GL_NV_shader_thread_group : enable

// prefer `GL_NV_shader_thread_group` cause these variables are constants
#ifdef GL_NV_shader_thread_group 
    vk_SubgroupSize = gl_WarpSizeNV
    vk_SubgroupInvocationID = gl_ThreadInWarpNV
#else 
    #extension GL_ARB_shader_ballot : require

    #extension GL_AMD_gcn_shader : enable
    // prefer `GL_AMD_gcn_shader` because gl_SIMDGroupSizeAMD is a constant
    #ifdef GL_AMD_gcn_shader
        vk_SubgroupSize = uint(gl_SIMDGroupSizeAMD);
    #else 
        vk_SubgroupSize = gl_SubGroupSizeARB;
    #endif 
    vk_SubgroupInvocationID = gl_SubGroupInvocationARB;
#endif 
// `gl_WarpIDNV` not appropriate because its per-SM not per-WorkGroup
vk_SubgroupID = vk_LocalInvocationIndex/vk_SubgroupSize ;

vk_NumSubgroups = vk_WorkGroupSize.x*vk_WorkGroupSize.y*vk_WorkGroupSize.z/vk_SubgroupSize; 
#ifdef GL_NV_shader_thread_group
    // you can probably turn that into a compile constant if no runtime decoration on `vk_WorkGroupSize` and `GL_NV_shader_thread_group`
#endif 
```

## Subgroup Elect

```glsl
bool subgroupElect()
{
#ifdef GL_NV_shader_thread_group 
    int firstLive = findLSB(ballotThreadNV(true));
    return gl_ThreadInWarpNV==uint(firstLive);
#else 
[...]

PR #1476, opened 2020-10-01
WIP: Partial Subgroup Intrinsics Support

### LINKED ISSUE #1350: GLSL: Support `GL_KHR_shader_subgroup_vote`
`GL_KHR_shader_subgroup_vote` is 1:1 match with OpenGL's `GL_ARB_shader_group_vote`  which is also core in 4.6, the implementation would be trivial (replace function names), except for `subgroupAllEqual` on types other than `bool`.

However it can also be implemented through `GL_NV_gpu_shader5` or `GL_AMD_gcn_shader`
```glsl
#extension GL_ARB_shader_group_vote : enable

#ifdef GL_ARB_shader_group_vote
    subgroupAll = allInvocationsARB
    subgroupAny = allInvocationsARB
    bool subgroupAllEqual(genType value) {return allInvocationsEqualARB(equal(value));}
#else 
    #extension GL_NV_gpu_shader5 : enable

    #ifdef GL_NV_gpu_shader5
        subgroupAll = allThreadsNV
        subgroupAny = anyThreadNV
        bool subgroupAllEqual(bool value) {return allThreadsEqualNV(value);}
    #else 
        #extension GL_AMD_gcn_shader : require
        #extension AMD_gpu_shader_int64 : enable
        #extension NV_gpu_shader5 : enable

        bool subgroupAll(bool value) {return ballotAMD(value)==ballotAMD(true);}
        bool subgroupAny(bool value) {return ballotAMD(value)!=0uLL;}
        bool subgroupAllEqual(bool value)
        {
            uint64_t b = ballotAMD(value);
            return b==0uLL || b==ballot(true);
        }
    #endif
#endif 
```

P.S. To implement `subgroupAllEqual` on other types than `bool` we'd need to make use of `GL_KHR_shader_subgroup_ballot` and `GL_KHR_shader_subgroup_basic`

```glsl
#extension GL_KHR_shader_subgroup_basic : require
#extension GL_KHR_shader_subgroup_ballot : require

bool subgroupAllEqual(NotBoolType value)
{
    return subgroupAllEqual(subgroupBroadcast(value,subgroupElect(void))==value);
}
```

devshgraphicsprogramming:
Fixes #1350 #1351 and #1352

### DIFF
--- a/reference/shaders-no-opt/comp/subgroups_basicvoteballot.vk.comp
+++ b/reference/shaders-no-opt/comp/subgroups_basicvoteballot.vk.comp
@@ -1,0 +1,371 @@
+#version 450
+#if defined(GL_KHR_shader_subgroup_ballot)
+#extension GL_KHR_shader_subgroup_ballot : require
+#elif defined(GL_NV_shader_thread_group)
+#extension GL_NV_shader_thread_group : require
+#elif defined(GL_ARB_shader_ballot) && defined(GL_ARB_shader_int64)
+#extension GL_ARB_shader_int64 : enable
+#extension GL_ARB_shader_ballot : require
+#else
+#error No extensions available to emulate requested subgroup feature.
+#endif
+#if defined(GL_KHR_shader_subgroup_basic)
+#extension GL_KHR_shader_subgroup_basic : require
+#elif defined(GL_NV_shader_thread_group)
+#extension GL_NV_shader_thread_group : require
+#elif defined(GL_ARB_shader_ballot) && defined(GL_ARB_shader_int64)
+#extension GL_ARB_shader_int64 : enable
+#extension GL_ARB_shader_ballot : require
+#elif defined(GL_AMD_gcn_shader) && (defined(GL_AMD_gpu_shader_int64) || defined(GL_NV_gpu_shader5))
+#extension GL_AMD_gpu_shader_int64 : enable
+#extension GL_NV_gpu_shader5 : enable
+#extension GL_AMD_gcn_shader : require
+#else
+#error No extensions available to emulate requested subgroup feature.
+#endif
+#if defined(GL_KHR_shader_subgroup_basic)
+#extension GL_KHR_shader_subgroup_basic : require
+#elif defined(GL_NV_shader_thread_group)
+#extension GL_NV_shader_thread_group : require
+#elif defined(GL_ARB_shader_ballot) && defined(GL_ARB_shader_int64)
+#extension GL_ARB_shader_int64 : enable
+#extension GL_ARB_shader_ballot : require
+#else
+#error No extensions available to emulate requested subgroup feature.
+#endif
+#if defined(GL_KHR_shader_subgroup_basic)
+#extension GL_KHR_shader_subgroup_basic : require
+#elif defined(GL_NV_shader_thread_group)
+#extension GL_NV_shader_thread_group : require
+#else
+#error No extensions available to emulate requested subgroup feature.
+#endif
+#if defined(GL_KHR_shader_subgroup_basic)
+#extension GL_KHR_shader_subgroup_basic : require
+#elif defined(GL_NV_shader_thread_group)
+#extension GL_NV_shader_thread_group : require
+#else
+#error No extensions available to emulate requested subgroup feature.
+#endif
+#if defined(GL_KHR_shader_subgroup_ballot)
+#extension GL_KHR_shader_subgroup_ballot : require
+#elif defined(GL_ARB_shader_ballot) && defined(GL_ARB_shader_int64)
+#extension GL_ARB_shader_int64 : enable
+#extension GL_ARB_shader_ballot : require
+#elif defined(GL_NV_shader_thread_shuffle)
+#extension GL_NV_shader_thread_shuffle : require
+#else
+#error No extensions available to emulate requested subgroup feature.
+#endif
+#if defined(GL_KHR_shader_subgroup_ballot)
+#extension GL_KHR_shader_subgroup_ballot : require
+#elif defined(GL_NV_shader_thread_group)
+#extension GL_NV_shader_thread_group : require
+#endif
+#if defined(GL_KHR_shader_subgroup_vote)
+#extension GL_KHR_shader_subgroup_vote : require
+#elif defined(GL_AMD_gcn_shader) && (defined(GL_AMD_gpu_shader_int64) || defined(GL_NV_gpu_shader5))
+#extension GL_AMD_gpu_shader_int64 : enable
+#extension GL_NV_gpu_shader5 : enable
+#extension GL_AMD_gcn_shader : require
+#elif defined(GL_NV_gpu_shader_5)
+#extension GL_NV_gpu_shader_5 : require
+#elif defined(GL_ARB_shader_group_vote)
+#extension GL_ARB_shader_group_vote : require
+#else
+#error No extensions available to emulate requested subgroup feature.
+#endif
+#if defined(GL_KHR_shader_subgroup_basic)
+#extension GL_KHR_shader_subgroup_basic : require
+#elif defined(GL_NV_shader_thread_group)
+#extension GL_NV_shader_thread_group : require
+#elif defined(GL_ARB_shader_ballot) && defined(GL_ARB_shader_int64)
+#extension GL_ARB_shader_int64 : enable
+#extension GL_ARB_shader_ballot : require
+#elif defined(GL_AMD_gcn_shader) && (defined(GL_AMD_gpu_shader_int64) || defined(GL_NV_gpu_shader5))
+#extension GL_AMD_gpu_shader_int64 : enable
+#extension GL_NV_gpu_shader5 : enable
+#extension GL_AMD_gcn_shader : require
+#else
+#error No extensions available to emulate requested subgroup feature.
+#endif
+#if defined(GL_KHR_shader_subgroup_basic)
+#extension GL_KHR_shader_subgroup_basic : require
+#endif
+#if defined(GL_KHR_shader_subgroup_ballot)
+#extension GL_KHR_shader_subgroup_ballot : require
+#elif defined(GL_NV_shader_thread_group)
+#extension GL_NV_shader_thread_group : require
+#elif defined(GL_ARB_shader_ballot) && defined(GL_ARB_shader_int64)
+#extension GL_ARB_shader_int64 : enable
+#extension GL_ARB_shader_ballot : require
+#else
+#error No extensions available to emulate requested subgroup feature.
+#endif
+#if defined(GL_NV_shader_thread_group)
+#extension GL_NV_shader_thread_group : require
+#endif
+layout(local_size_x = 1, local_size_y = 1, local_size_z = 1) in;
+
+layout(binding = 0, std430) buffer SSBO
+{
+    float FragColor;
+} _9;
+
+#if defined(GL_KHR_shader_subgroup_ballot)
+#elif defined(GL_NV_shader_thread_group)
+#define gl_SubgroupEqMask uvec4(gl_ThreadEqMaskNV, 0u, 0u, 0u)
+#define gl_SubgroupGeMask uvec4(gl_ThreadGeMaskNV, 0u, 0u, 0u)
+#define gl_SubgroupGtMask uvec4(gl_ThreadGtMaskNV, 0u, 0u, 0u)
+#define gl_SubgroupLeMask uvec4(gl_ThreadLeMaskNV, 0u, 0u, 0u)
+#define gl_SubgroupLtMask uvec4(gl_ThreadLtMaskNV, 0u, 0u, 0u)
+#elif defined(GL_ARB_shader_ballot)
+#define gl_SubgroupEqMask uvec4(unpackUint2x32(gl_SubGroupEqMaskARB), 0u, 0u)
+#define gl_SubgroupGeMask uvec4(unpackUint2x32(gl_SubGroupGeMaskARB), 0u, 0u)
+#define gl_SubgroupGtMask uvec4(unpackUint2x32(gl_SubGroupGtMaskARB), 0u, 0u)
+#define gl_SubgroupLeMask uvec4(unpackUint2x32(gl_SubGroupLeMaskARB), 0u, 0u)
+#define gl_SubgroupLtMask uvec4(unpackUint2x32(gl_SubGroupLtMaskARB), 0u, 0u)
+#endif
+#if defined(GL_KHR_shader_subgroup_basic)
+#elif defined(GL_NV_shader_thread_group)
+#define gl_SubgroupSize gl_WarpSizeNV
+#elif defined(GL_ARB_shader_ballot)
+#define gl_SubgroupSize gl_SubGroupSizeARB
+#elif defined(GL_AMD_gcn_shader)
+#define gl_SubgroupSize uint(gl_SIMDGroupSizeAMD)
+#endif
+#if defined(GL_KHR_shader_subgroup_basic)
+#elif defined(GL_NV_shader_thread_group)
+#define gl_SubgroupInvocationID gl_ThreadInWarpNV
+#elif defined(GL_ARB_shader_ballot)
+#define gl_SubgroupInvocationID gl_SubGroupInvocationARB
+#endif
+#if defined(GL_KHR_shader_subgroup_basic)
+#elif defined(GL_NV_shader_thread_group)
+#define gl_SubgroupID gl_WarpIDNV
+#endif
+#if defined(GL_KHR_shader_subgroup_basic)
+#elif defined(GL_NV_shader_thread_group)
+#define gl_NumSubgroups gl_WarpsPerSMNV
+#endif
+#if defined(GL_KHR_shader_subgroup_ballot)
+#elif defined(GL_ARB_shader_ballot)
+int subgroupBroadcastFirst(int value) { return readFirstInvocationARB(value); }
+ivec2 subgroupBroadcastFirst(ivec2 value) { return readFirstInvocationARB(value); }
+ivec3 subgroupBroadcastFirst(ivec3 value) { return readFirstInvocationARB(value); }
+ivec4 subgroupBroadcastFirst(ivec4 value) { return readFirstInvocationARB(value); }
+uint subgroupBroadcastFirst(uint value) { return readFirstInvocationARB(value); }
+uvec2 subgroupBroadcastFirst(uvec2 value) { return readFirstInvocationARB(value); }
+uvec3 subgroupBroadcastFirst(uvec3 value) { return readFirstInvocationARB(value); }
+uvec4 subgroupBroadcastFirst(uvec4 value) { return readFirstInvocationARB(value); }
+float subgroupBroadcastFirst(float value) { return readFirstInvocationARB(value); }
+vec2 subgroupBroadcastFirst(vec2 value) { return readFirstInvocationARB(value); }
+vec3 subgroupBroadcastFirst(vec3 value) { return readFirstInvocationARB(value); }
+vec4 subgroupBroadcastFirst(vec4 value) { return readFirstInvocationARB(value); }
+double subgroupBroadcastFirst(double value) { return readFirstInvocationARB(value); }
+dvec2 subgroupBroadcastFirst(dvec2 value) { return readFirstInvocationARB(value); }
+dvec3 subgroupBroadcastFirst(dvec3 value) { return readFirstInvocationARB(value); }
+dvec4 subgroupBroadcastFirst(dvec4 value) { return readFirstInvocationARB(value); }
+int subgroupBroadcast(int value, uint id) { return readInvocationARB(value, id); }
+ivec2 subgroupBroadcast(ivec2 value, uint id) { return readInvocationARB(value, id); }
+ivec3 subgroupBroadcast(ivec3 value, uint id) { return readInvocationARB(value, id); }
+ivec4 subgroupBroadcast(ivec4 value, uint id) { return readInvocationARB(value, id); }
+uint subgroupBroadcast(uint value, uint id) { return readInvocationARB(value, id); }
+uvec2 subgroupBroadcast(uvec2 value, uint id) { return readInvocationARB(value, id); }
+uvec3 subgroupBroadcast(uvec3 value, uint id) { return readInvocationARB(value, id); }
+uvec4 subgroupBroadcast(uvec4 value, uint id) { return readInvocationARB(value, id); }
+float subgroupBroadcast(float value, uint id) { return readInvocationARB(value, id); }
+vec2 subgroupBroadcast(vec2 value, uint id) { return readInvocationARB(value, id); }
+vec3 subgroupBroadcast(vec3 value, uint id) { return readInvocationARB(value, id); }
+vec4 subgroupBroadcast(vec4 value, uint id) { return readInvocationARB(value, id); }
+double subgroupBroadcast(double value, uint id) { return readInvocationARB(value, id); }
+dvec2 subgroupBroadcast(dvec2 value, uint id) { return readInvocationARB(value, id); }
+dvec3 subgroupBroadcast(dvec3 value, uint id) { return readInvocationARB(value, id); }
+dvec4 subgroupBroadcast(dvec4 value, uint id) { return readInvocationARB(value, id); }
+#elif defined(GL_NV_shader_thread_shuffle)
+int subgroupBroadcastFirst(int value) { return shuffleNV(value, findLSB(ballotThreadNV(true)), gl_WarpSizeNV); }
+ivec2 subgroupBroadcastFirst(ivec2 value) { return shuffleNV(value, findLSB(ballotThreadNV(true)), gl_WarpSizeNV); }
+ivec3 subgroupBroadcastFirst(ivec3 value) { return shuffleNV(value, findLSB(ballotThreadNV(true)), gl_WarpSizeNV); }
+ivec4 subgroupBroadcastFirst(ivec4 value) { return shuffleNV(value, findLSB(ballotThreadNV(true)), gl_WarpSizeNV); }
+uint subgroupBroadcastFirst(uint value) { return shuffleNV(value, findLSB(ballotThreadNV(true)), gl_WarpSizeNV); }
+uvec2 subgroupBroadcastFirst(uvec2 value) { return shuffleNV(value, findLSB(ballotThreadNV(true)), gl_WarpSizeNV); }
+uvec3 subgroupBroadcastFirst(uvec3 value) { return shuffleNV(value, findLSB(ballotThreadNV(true)), gl_WarpSizeNV); }
+uvec4 subgroupBroadcastFirst(uvec4 value) { return shuffleNV(value, findLSB(ballotThreadNV(true)), gl_WarpSizeNV); }
+float subgroupBroadcastFirst(float value) { return shuffleNV(value, findLSB(ballotThreadNV(true)), gl_WarpSizeNV); }
+vec2 subgroupBroadcastFirst(vec2 value) { return shuffleNV(value, findLSB(ballotThreadNV(true)), gl_WarpSizeNV); }
+vec3 subgroupBroadcastFirst(vec3 value) { return shuffleNV(value, findLSB(ballotThreadNV(true)), gl_WarpSizeNV); }
+vec4 subgroupBroadcastFirst(vec4 value) { return shuffleNV(value, findLSB(ballotThreadNV(true)), gl_WarpSizeNV); }
+double subgroupBroadcastFirst(double value) { return shuffleNV(value, findLSB(ballotThreadNV(true)), gl_WarpSizeNV); }
+dvec2 subgroupBroadcastFirst(dvec2 value) { return shuffleNV(value, findLSB(ballotThreadNV(true)), gl_WarpSizeNV); }
+dvec3 subgroupBroadcastFirst(dvec3 value) { return shuffleNV(value, findLSB(ballotThreadNV(true)), gl_WarpSizeNV); }
+dvec4 subgroupBroadcastFirst(dvec4 value) { return shuffleNV(value, findLSB(ballotThreadNV(true)), gl_WarpSizeNV); }
+int subgroupBroadcast(int value, uint id) { return shuffleNV(value, id, gl_WarpSizeNV); }
+ivec2 subgroupBroadcast(ivec2 value, uint id) { return shuffleNV(value, id, gl_WarpSizeNV); }
+ivec3 subgroupBroadcast(ivec3 value, uint id) { return shuffleNV(value, id, gl_WarpSizeNV); }
+ivec4 subgroupBroadcast(ivec4 value, uint id) { return shuffleNV(value, id, gl_WarpSizeNV); }
+uint subgroupBroadcast(uint value, uint id) { return shuffleNV(value, id, gl_WarpSizeNV); }
+uvec2 subgroupBroadcast(uvec2 value, uint id) { return shuffleNV(value, id, gl_WarpSizeNV); }
+uvec3 subgroupBroadcast(uvec3 value, uint id) { return shuffleNV(value, id, gl_WarpSizeNV); }
+uvec4 subgroupBroadcast(uvec4 value, uint id) { return shuffleNV(value, id, gl_WarpSizeNV); }
+float subgroupBroadcast(float value, uint id) { return shuffleNV(value, id, gl_WarpSizeNV); }
+vec2 subgroupBroadcast(vec2 value, uint id) { return shuffleNV(value, id, gl_WarpSizeNV); }
+vec3 subgroupBroadcast(vec3 value, uint id) { return shuffleNV(value, id, gl_WarpSizeNV); }
+vec4 subgroupBroadcast(vec4 value, uint id) { return shuffleNV(value, id, gl_WarpSizeNV); }
+double subgroupBroadcast(double value, uint id) { return shuffleNV(value, id, gl_WarpSizeNV); }
+dvec2 subgroupBroadcast(dvec2 value, uint id) { return shuffleNV(value, id, gl_WarpSizeNV); }
+dvec3 subgroupBroadcast(dvec3 value, uint id) { return shuffleNV(value, id, gl_WarpSizeNV); }
+dvec4 subgroupBroadcast(dvec4 value, uint id) { return shuffleNV(value, id, gl_WarpSizeNV); }
+#endif
+#if defined(GL_KHR_shader_subgroup_ballot)
+#elif defined(GL_NV_shader_thread_group)
+uint subgroupBallotFindLSB(uvec4 value) { return findLSB(value.x); }
+uint subgroupBallotFindMSB(uvec4 value) { return findMSB(value.x); }
+#else
+uint subgroupBallotFindLSB(uvec4 value) {
+int firstLive = findLSB(value.x);
+return uint(firstLive != -1 ? firstLive : (findLSB(value.y) + 32));
+}
+uint subgroupBallotFindMSB(uvec4 value) {
+int firstLive = findMSB(value.y);
+return uint(firstLive != -1 ? (firstLive + 32) : findMSB(value.x));
+}
+#endif
+#if defined(GL_KHR_shader_subgroup_vote)
+#elif defined(GL_AMD_gcn_shader)
+bool subgroupAll(bool value) { return ballotAMD(value)==ballotAMD(true); }
+bool subgroupAny(bool value) { return ballotAMD(value)!=0ull; }
+bool subgroupAllEqual(bool value) { uint64_t b=ballotAMD(value); return b==0uLL || b==ballotAMD(true); }
+#elif defined(GL_NV_gpu_shader_5)
+bool subgroupAll(bool value) { return allThreadsNV(value); }
+bool subgroupAny(bool value) { return anyThreadNV(value); }
+bool subgroupAllEqual(bool value) { return allThreadsEqualNV(value); }
+#elif defined(GL_ARB_shader_group_vote)
+bool subgroupAll(bool v) { return allInvocationsARB(v); }
+bool subgroupAny(bool v) { return anyInvocationARB(v); }
+bool subgroupAllEqual(bool v) { return allInvocationsEqualARB(v); }
+#endif
+#ifndef GL_KHR_shader_subgroup_vote
+#define _SPIRV_CROSS_SUBGROUP_ALL_EQUAL_WORKAROUND(type) bool subgroupAllEqual(type value) { return subgroupAllEqual(subgroupBroadcastFirst(value)==value); }
+_SPIRV_CROSS_SUBGROUP_ALL_EQUAL_WORKAROUND(int)
+_SPIRV_CROSS_SUBGROUP_ALL_EQUAL_WORKAROUND(ivec2)
+_SPIRV_CROSS_SUBGROUP_ALL_EQUAL_WORKAROUND(ivec3)
+_SPIRV_CROSS_SUBGROUP_ALL_EQUAL_WORKAROUND(ivec4)
+_SPIRV_CROSS_SUBGROUP_ALL_EQUAL_WORKAROUND(uint)
+_SPIRV_CROSS_SUBGROUP_ALL_EQUAL_WORKAROUND(uvec2)
+_SPIRV_CROSS_SUBGROUP_ALL_EQUAL_WORKAROUND(uvec3)
+_SPIRV_CROSS_SUBGROUP_ALL_EQUAL_WORKAROUND(uvec4)
+_SPIRV_CROSS_SUBGROUP_ALL_EQUAL_WORKAROUND(float)
+_SPIRV_CROSS_SUBGROUP_ALL_EQUAL_WORKAROUND(vec2)
+_SPIRV_CROSS_SUBGROUP_ALL_EQUAL_WORKAROUND(vec3)
+_SPIRV_CROSS_SUBGROUP_ALL_EQUAL_WORKAROUND(vec4)
+_SPIRV_CROSS_SUBGROUP_ALL_EQUAL_WORKAROUND(double)
+_SPIRV_CROSS_SUBGROUP_ALL_EQUAL_WORKAROUND(dvec2)
+_SPIRV_CROSS_SUBGROUP_ALL_EQUAL_WORKAROUND(dvec3)
+_SPIRV_CROSS_SUBGROUP_ALL_EQUAL_WORKAROUND(dvec4)
+#undef _SPIRV_CROSS_SUBGROUP_ALL_EQUAL_WORKAROUND
+#endif
+#if defined(GL_KHR_shader_subgroup_ballot)
+#elif defined(GL_NV_shader_thread_group)
+uvec4 subgroupBallot(bool v) { return uvec4(ballotThreadNV(v), 0u, 0u, 0u); }
+uvec4 subgroupBallot(bool v) { return uvec4(unpackUint2x32(ballotARB(v)), 0u, 0u); }
+#elif defined(GL_ARB_shader_ballot)
+uvec4 subgroupBallot(bool v) { return uvec4(unpackUint2x32(ballotARB(v)), 0u, 0u); }
+#endif
+#ifndef GL_KHR_shader_subgroup_basic
+bool subgroupElect() {
+uvec4 activeMask = subgroupBallot(true);
+uint firstLive = subgroupBallotFindLSB(activeMask);
+return gl_SubgroupInvocationID == firstLive;
+}
+#endif
+#ifndef GL_KHR_shader_subgroup_basic
+void subgroupBarrier() { /*NOOP*/ }
+#endif
+#ifndef GL_KHR_shader_subgroup_basic
+void subgroupMemoryBarrier() { groupMemoryBarrier(); }
+void subgroupMemoryBarrierBuffer() { groupMemoryBarrier(); }
+void subgroupMemoryBarrierShared() { groupMemoryBarrier(); }
+void subgroupMemoryBarrierImage() { groupMemoryBarrier(); }
+#endif
+#ifndef GL_KHR_shader_subgroup_ballot
+bool subgroupInverseBallot(uvec4 value, uint index)
+{
+return any(notEqual(value.xy & gl_SubgroupEqMask.xy, uvec2(0u)));
+}
+uint subgroupBallotInclusiveBitCount(uvec4 value)
+{
+uvec2 v = value.xy & gl_SubgroupLeMask.xy;
+ivec2 c = bitCount(v);
+#ifdef GL_NV_shader_thread_group
+return uint(c.x);
+#else
+return uint(c.x + c.y);
+#endif
+}
+uint subgroupBallotExclusiveBitCount(uvec4 value)
+{
+uvec2 v = value.xy & gl_SubgroupLtMask.xy;
+ivec2 c = bitCount(v);
+#ifdef GL_NV_shader_thread_group
+return uint(c.x);
+#else
+return uint(c.x + c.y);
+#endif
+}
+#endif
+#ifndef GL_KHR_shader_subgroup_ballot
+uint subgroupBallotBitCount(uvec4 value)
+{
+ivec2 c = bitCount(value.xy);
+#ifdef GL_NV_shader_thread_group
+return uint(c.x);
+#else
+return uint(c.x + c.y);
+#endif
+}
+#endif
+#ifndef GL_KHR_shader_subgroup_ballot
+bool subgroupBallotBitExtract(uvec4 value, uint index)
+{
+#ifdef GL_NV_shader_thread_group
+uint shifted = value.x >> index;
+#else
+uint shifted = value[index >> 5u] >> (index & 0x1fu);
+#endif
+return (shifted & 1u) != 0u;
+}
+#endif
+void main()
+{
+    _9.FragColor = float(gl_NumSubgroups);
+    _9.FragColor = float(gl_SubgroupID);
+    _9.FragColor = float(gl_SubgroupSize);
+    _9.FragColor = float(gl_SubgroupInvocationID);
+    subgroupMemoryBarrier();
+    subgroupBarrier();
+    subgroupMemoryBarrier();
+    subgroupMemoryBarrierBuffer();
+    subgroupMemoryBarrierShared();
+    subgroupMemoryBarrierImage();
+    bool elected = subgroupElect();
+    _9.FragColor = vec4(gl_SubgroupEqMask).x;
+    _9.FragColor = vec4(gl_SubgroupGeMask).x;
+    _9.FragColor = vec4(gl_SubgroupGtMask).x;
+    _9.FragColor = vec4(gl_SubgroupLeMask).x;
+    _9.FragColor = vec4(gl_SubgroupLtMask).x;
+    vec4 broadcasted = subgroupBroadcast(vec4(10.0), 8u);
+    vec3 first = subgroupBroadcastFirst(vec3(20.0));
+    uvec4 ballot_value = subgroupBallot(true);
+    bool inverse_ballot_value = subgroupInverseBallot(ballot_value);
+    bool bit_extracted = subgroupBallotBitExtract(uvec4(10u), 8u);
+    uint bit_count = subgroupBallotBitCount(ballot_value);
+    uint inclusive_bit_count = subgroupBallotInclusiveBitCount(ballot_value);
+    uint exclusive_bit_count = subgroupBallotExclusiveBitCount(ballot_value);
+    uint lsb = subgroupBallotFindLSB(ballot_value);
+    uint msb = subgroupBallotFindMSB(ballot_value);
+    bool has_all = subgroupAll(true);
+    bool has_any = subgroupAny(true);
+    bool has_equal_bool = subgroupAllEqual(true);
+    bool has_equal_T = subgroupAllEqual(uvec3(5u));
+}
+
--- a/spirv_cross_containers.hpp
+++ b/spirv_cross_containers.hpp
@@ -63,8 +63,7 @@ public:
 private:
 #if defined(_MSC_VER) && _MSC_VER < 1900
 	// MSVC 2013 workarounds, sigh ...
-	union
-	{
+	union {
 		char aligned_char[sizeof(T) * N];
 		double dummy_aligner;
 	} u;
@@ -210,6 +209,10 @@ public:
 		for (size_t i = 0; i < count; i++, arg_list_begin++)
 			new (&this->ptr[i]) T(*arg_list_begin);
 		this->buffer_size = count;
+	}
+
+	SmallVector(std::initializer_list<T> init) SPIRV_CROSS_NOEXCEPT : SmallVector(init.begin(), init.end())
+	{
 	}
 
 	SmallVector(SmallVector &&other) SPIRV_CROSS_NOEXCEPT : SmallVector()
--- a/spirv_glsl.cpp
+++ b/spirv_glsl.cpp
@@ -548,6 +548,7 @@ string CompilerGLSL::compile()
 
 		emit_header();
 		emit_resources();
+		emit_extension_workarounds(ir.entry_points[ir.default_entry_point].model);
 
 		emit_function(get<SPIRFunction>(ir.default_entry_point), Bitset());
 
@@ -618,6 +619,22 @@ void CompilerGLSL::build_workgroup_size(SmallVector<string> &arguments, const Sp
 	}
 	else
 		arguments.push_back(join("local_size_z = ", execution.workgroup_size.z));
+}
+
+void CompilerGLSL::request_subgroup_feature(ShaderSubgroupSupportHelper::Feature feature)
+{
+	if (options.vulkan_semantics)
+	{
+		const ShaderSubgroupSupportHelper::Candidate khrExt =
+		    shader_subgroup_supporter.get_KHR_extension_for_feature(feature);
+		require_extension_internal(shader_subgroup_supporter.get_extension_name(khrExt));
+	}
+	else
+	{
+		if (!shader_subgroup_supporter.is_feature_requested(feature))
+			force_recompile();
+		shader_subgroup_supporter.request_feature(feature);
+	}
 }
 
 void CompilerGLSL::emit_header()
@@ -720,6 +737,45 @@ void CompilerGLSL::emit_header()
 		}
 		else
 			statement("#extension ", ext, " : require");
+	}
+
+	if (!options.vulkan_semantics)
+	{
+		using Supp = ShaderSubgroupSupportHelper;
+
+		Supp::Result result = shader_subgroup_supporter.resolve();
+		for (uint32_t ft = 0u; ft < Supp::FeatureCount; ++ft)
+		{
+			Supp::Feature feature = static_cast<Supp::Feature>(ft);
+			if (!shader_subgroup_supporter.is_feature_requested(feature))
+				continue;
+
+			auto exts = shader_subgroup_supporter.get_candidates_for_feature(feature, result);
+			if (exts.empty())
+				continue;
+
+			for (auto it = exts.begin(); it != exts.end(); ++it)
+			{
+				const Supp::Candidate ext = *it;
+
+				std::string name = Supp::get_extension_name(ext);
+				std::string extraPredicate = Supp::get_extra_required_extension_predicate(ext);
+				auto extraNames = Supp::get_extra_required_extension_names(ext);
+				statement(it != exts.begin() ? "#elif" : "#if", " defined(", name, ")",
+				          extraPredicate.empty() ? "" : (" && " + extraPredicate));
+				for (const auto &e : extraNames)
+					statement("#extension ", e, " : enable");
+				statement("#extension ", name, " : require");
+			}
+
+			if (!Supp::can_feature_be_implemented_wo_extensions(feature))
+			{
+				statement("#else");
+				statement("#error No extensions available to emulate requested subgroup feature.");
+			}
+
+			statement("#endif");
+		}
 	}
 
 	for (auto &header : header_lines)
@@ -2890,8 +2946,7 @@ void CompilerGLSL::declare_undefined_values()
 		if (options.force_zero_initialized_variables && type_can_zero_initialize(type))
 			initializer = join(" = ", to_zero_initialized_expression(undef.basetype));
 
-		statement(variable_decl(type, to_name(undef.self), undef.self), initializer,
-		          ";");
+		statement(variable_decl(type, to_name(undef.self), undef.self), initializer, ";");
 		emitted = true;
 	});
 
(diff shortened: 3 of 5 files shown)


spirv_glsl.hpp
```diff
@@ -399,6 +399,8 @@ class CompilerGLSL : public Compiler
 	void end_scope_decl(const std::string &decl);
 
 	Options options;
+	bool declaredKHR_shader_subgroup_basic = false;
+	bool declaredKHR_shader_subgroup_vote = false;
```
devshgraphicsprogramming, line 403: ballot is missing.

devshgraphicsprogramming:
A few recaps from my conversation with @HansKristian-Work 

> CompilerGLSL::emit_subgroup_op implements subgroup ops

> Individual opcodes are handled in CompilerGLSL::emit_instruction

> for test files, .nocompat. means don't emit desktop GLSL
.vk. means it's vulkan GLSL and should normally emit Vulkan GLSL and desktop GLSL variants
those paths are checked for in test_shaders.py
so to add desktop GLSL support, nocompat should be removed
and it will generate non-vulkan reference output as well

> `.invalid` means don't run through spirv-val
usually because some version had a bug that blocked test suite
generally spir-v is validated


@Crisspl the problem is that test requires all extension at once
We will only implement 3, basic, vote and ballot. 
We have been give the go ahead to splice out that one test into two separate tests.
Our new one should be called `shaders-no-opt/comp/subgroups_basicvoteballot.vk.comp`


Instructions to run:
> just plop in file
and run test_shaders.sh
new file will show up in reference/
or test_shaders.py shaders-no-opt --update
to not test everything
full suite takes a while

spirv_glsl.hpp
devshgraphicsprogramming: I'd rather track the following (instead of KHR extensions):
- needs group voting
- needs subgroup masks
- needs subgroup elect
- needs subgroup ballots

@HansKristian-Work whats the ordering between `emit_header` and `emit_subgroup_op`, the logical one?

spirv_glsl.cpp
```diff
@@ -718,6 +719,54 @@ void CompilerGLSL::emit_header()
...
+				statement("#extension GL_KHR_shader_subgroup_vote : require");
+			else
+			{
+				statement("#if defined(GL_ARB_shader_group_vote)");
+				statement("#extension GL_ARB_shader_group_vote : require");
+				statement("#elif defined(GL_NV_gpu_shader5)");
+				statement("#extension GL_NV_gpu_shader5 : require");
+				statement("#elif defined(GL_AMD_gcn_shader)&&defined(GL_AMD_gpu_shader_int64)");
+				statement("#extension GL_AMD_gcn_shader : require");
+				statement("#extension GL_AMD_gpu_shader_int64 : enable"); //needed?
```
devshgraphicsprogramming, line 735: https://www.khronos.org/registry/OpenGL/extensions/AMD/AMD_gcn_shader.txt

> AMD_gpu_shader_int64 or NV_gpu_shader5 is required.

you need to soft enable both, since spec says "required" and not "implicitly enabled"

spirv_glsl.cpp
```diff
@@ -718,6 +719,54 @@ void CompilerGLSL::emit_header()
...
+		else if (ext == "GL_KHR_shader_subgroup_basic")
+		{
+			declaredKHR_shader_subgroup_basic = true;
+			if (options.vulkan_semantics)
+				statement("#extension GL_KHR_shader_subgroup_basic : require");
+			else
+			{
+				statement("#if defined(GL_NV_shader_thread_group)");
+
+				statement("#define GL_NV_shader_thread_group");
```
devshgraphicsprogramming, line 750: typo?

spirv_glsl.cpp
```diff
@@ -718,6 +719,54 @@ void CompilerGLSL::emit_header()
...
+				statement("#elif defined(GL_ARB_shader_ballot)");
+
+				statement("#extension GL_ARB_shader_ballot : require");
+				statement("#extension GL_AMD_gcn_shader : enable");
+				statement("#ifdef GL_AMD_gcn_shader");
+				statement("#define gl_SubgroupSize uint(gl_SIMDGroupSizeAMD)");
+				statement("#else defined(GL_ARB_shader_ballot)");
+				statement("#define gl_SubgroupSize gl_SubGroupSizeARB");
+				statement("#endif");
+				statement("#define gl_SubgroupInvocationID gl_SubGroupInvocationARB");
```
devshgraphicsprogramming, line 761: @HansKristian-Work is that the right place to put these?

Also should we `#define` or should we declare a `const uint` ?

Furthermore @Crisspl anything starting with `gl_` in GLSL is reserved, so I guess we'll need to use a different prefix like SPIRV-Cross uses for its `gl_InstanceIndex` replacement.

spirv_glsl.cpp
```diff
@@ -718,6 +719,54 @@ void CompilerGLSL::emit_header()
...
+				statement("#extension GL_NV_gpu_shader5 : require");
+				statement("#elif defined(GL_AMD_gcn_shader)&&defined(GL_AMD_gpu_shader_int64)");
+				statement("#extension GL_AMD_gcn_shader : require");
+				statement("#extension GL_AMD_gpu_shader_int64 : enable"); //needed?
+				statement("#else");
+				statement("#error No extensions available for subgroupAllEqual()");
+				statement("#endif");
+			}
+		}
+		else if (ext == "GL_KHR_shader_subgroup_basic")
```
devshgraphicsprogramming, line 741: any `GL_KHR_shader_subgroup_*` extension implies `GL_KHR_shader_subgroup_basic` implicitly.

spirv_glsl.cpp
devshgraphicsprogramming: > If you emit a constant, you can call it SPIRV_Cross_gl_something, and change the name where builtin name is queried depending on vulkan_semantic. You can emit these constants in emit_resources(). You should be able to detect if the capability is enabled and only emit builtin constants based on that.

https://github.com/KhronosGroup/SPIRV-Cross/issues/1352#issuecomment-621796811

spirv_glsl.cpp
```diff
@@ -3258,6 +3306,78 @@ void CompilerGLSL::emit_resources()
...
+	/* TODO?
+	https://github.com/KhronosGroup/SPIRV-Cross/issues/1351
+	Neither GL_ARB_shader_ballot, GL_NV_shader_thread_group or AMD_gcn_shader support subgroup barriers (execution or memory) explicitly.
+	However the wording of GL_NV_shader_thread_group and ARB_shader_ballot explicitly stat that invocations run in lock-step.
+	So this should mean that Turing and Volta (or any other GPU architecture which supports per-invocation Program Counters)
+	should either not report these extensions or insert synchronization implicitly in the shaders using those extensions.
+	**This could allow us to turn any subgroupBarrier into a NO-OP.**
+	However the extensions do not guarantee anything about memory dependencies.
+	So the presence of subgroupMemoryBarrier* function call in the SPV should produce an error.
+	*/
```
devshgraphicsprogramming, line 3362: yep, error out if `subgroupMemoryBarrier*` SPIR-V op is attempted to be emitted.

But turn the `subgroupBarrier` into a NOOP (maybe emit a comment for clarity).

spirv_glsl.cpp
```diff
@@ -3258,6 +3306,78 @@ void CompilerGLSL::emit_resources()
...
 
+void CompilerGLSL::emit_extension_workarounds()
+{
+	//GL_KHR_shader_subgroup_vote
+	if (!options.vulkan_semantics && declaredKHR_shader_subgroup_vote)
+	{
+		statement("#ifdef GL_ARB_shader_group_vote");
+		statement("bool subgroupAll(in bool v) { return allInvocationsARB(v); }");
+		statement("bool subgroupAny(in bool v) { return anyInvocationARB(v); }");
+		statement("#define SUBGROUP_ALL_EQUAL_WORKAROUND(type) bool subgroupAllEqual(type value) { return "
```
devshgraphicsprogramming, line 3317: `bool subgroupAllEqual(bool value) {return allInvocationsEqualARB(value);}` for the basic `bool` version, then we can play with the macro for the other types

spirv_glsl.cpp
```diff
@@ -3258,6 +3306,78 @@ void CompilerGLSL::emit_resources()
...
+		statement("SUBGROUP_ALL_EQUAL_WORKAROUND(uvec4)");
+		statement("SUBGROUP_ALL_EQUAL_WORKAROUND(float)");
+		statement("SUBGROUP_ALL_EQUAL_WORKAROUND(vec2)");
+		statement("SUBGROUP_ALL_EQUAL_WORKAROUND(vec3)");
+		statement("SUBGROUP_ALL_EQUAL_WORKAROUND(vec4)");
+		statement("SUBGROUP_ALL_EQUAL_WORKAROUND(double)");
+		statement("SUBGROUP_ALL_EQUAL_WORKAROUND(dvec2)");
+		statement("SUBGROUP_ALL_EQUAL_WORKAROUND(dvec3)");
+		statement("SUBGROUP_ALL_EQUAL_WORKAROUND(dvec4)");
+		statement("#undef SUBGROUP_ALL_EQUAL_WORKAROUND");
```
devshgraphicsprogramming, line 3335: To be able to service these versions of `subgroupAllEqual`, you'd need to implicitly enable `GL_KHR_shader_subgroup_ballot` (which you should only do if you detect the SPIR-V Opcodes for `subgroupAllEqual` present in the source)

devshgraphicsprogramming:
```glsl
bool subgroupAllEqual(NotBoolType value)
{
    return subgroupAllEqual(subgroupBroadcast(value,subgroupElect(void))==value);
}
```
That's what the macro should look like.

devshgraphicsprogramming: Also define and use the macro later, its the same workaround, regardless of the ARB/NV/AMD extension used to emulate.

spirv_glsl.cpp
```diff
@@ -718,6 +719,54 @@ void CompilerGLSL::emit_header()
...
+			declaredKHR_shader_subgroup_vote = true;
+			if (options.vulkan_semantics)
+				statement("#extension GL_KHR_shader_subgroup_vote : require");
+			else
+			{
+				statement("#if defined(GL_ARB_shader_group_vote)");
+				statement("#extension GL_ARB_shader_group_vote : require");
+				statement("#elif defined(GL_NV_gpu_shader5)");
+				statement("#extension GL_NV_gpu_shader5 : require");
+				statement("#elif defined(GL_AMD_gcn_shader)&&defined(GL_AMD_gpu_shader_int64)");
```
HansKristian-Work, line 733: Style nit, space between `defined() && defined()`

spirv_glsl.cpp
HansKristian-Work: SubgroupSize is not a constant necessarily, so `#define` overriding it is fine.

spirv_glsl.cpp
```diff
@@ -718,6 +719,54 @@ void CompilerGLSL::emit_header()
...
+				statement("#ifdef GL_AMD_gcn_shader");
+				statement("#define gl_SubgroupSize uint(gl_SIMDGroupSizeAMD)");
+				statement("#else defined(GL_ARB_shader_ballot)");
+				statement("#define gl_SubgroupSize gl_SubGroupSizeARB");
+				statement("#endif");
+				statement("#define gl_SubgroupInvocationID gl_SubGroupInvocationARB");
+
+				statement("#endif");
+
+				statement("#define gl_SubgroupID (gl_LocalInvocationIndex/gl_SubgroupSize)");
```
HansKristian-Work, line 765: Space between divider.

spirv_glsl.cpp
```diff
@@ -718,6 +719,54 @@ void CompilerGLSL::emit_header()
...
+				statement("#define gl_NumSubgroups "
+				          "(gl_WorkGroupSize.x*gl_WorkGroupSize.y*gl_WorkGroupSize.z/gl_SubgroupSize)");
```
HansKristian-Work, line 767: This gets a bit sketchy. Workgroup size might not align with the subgroup size, and the rounding behavior is not exactly known to the application. E.g. a workgroup size of (5, 3, 7) is hard to know for sure what happens.

spirv_glsl.cpp
Crisspl: ?

spirv_glsl.hpp
```diff
@@ -399,6 +399,8 @@ class CompilerGLSL : public Compiler
 	void end_scope_decl(const std::string &decl);
 
 	Options options;
+	bool declaredKHR_shader_subgroup_basic = false;
```
HansKristian-Work, line 402: Should be `declared_KHR_shader_subgroup_basic`, not camel case. Since there is an entry point for emitting extension workarounds, probably best to hoist all these members into an anonymous struct.

spirv_glsl.hpp
HansKristian-Work: Not sure I understand the question.

spirv_glsl.cpp
```diff
@@ -3258,6 +3306,78 @@ void CompilerGLSL::emit_resources()
...
+	Neither GL_ARB_shader_ballot, GL_NV_shader_thread_group or AMD_gcn_shader support subgroup barriers (execution or memory) explicitly.
+	However the wording of GL_NV_shader_thread_group and ARB_shader_ballot explicitly stat that invocations run in lock-step.
+	So this should mean that Turing and Volta (or any other GPU architecture which supports per-invocation Program Counters)
+	should either not report these extensions or insert synchronization implicitly in the shaders using those extensions.
+	**This could allow us to turn any subgroupBarrier into a NO-OP.**
+	However the extensions do not guarantee anything about memory dependencies.
+	So the presence of subgroupMemoryBarrier* function call in the SPV should produce an error.
+	*/
+		const std::string workaround = //multi-line "statements" are ok?
+R"(bool subgroupElect()
```
HansKristian-Work, line 3364: Use multiple statements.

spirv_glsl.cpp
```diff
@@ -3258,6 +3306,78 @@ void CompilerGLSL::emit_resources()
...
+	//GL_KHR_shader_subgroup_vote
+	if (!options.vulkan_semantics && declaredKHR_shader_subgroup_vote)
+	{
+		statement("#ifdef GL_ARB_shader_group_vote");
+		statement("bool subgroupAll(in bool v) { return allInvocationsARB(v); }");
+		statement("bool subgroupAny(in bool v) { return anyInvocationARB(v); }");
+		statement("#define SUBGROUP_ALL_EQUAL_WORKAROUND(type) bool subgroupAllEqual(type value) { return "
+		          "allInvocationsEqualARB(equal(value)); }");
+		statement("SUBGROUP_ALL_EQUAL_WORKAROUND(int)");
+		statement("SUBGROUP_ALL_EQUAL_WORKAROUND(ivec2)");
```
HansKristian-Work, line 3320: Can have an array of the type names and just emit statement in a loop to keep the LOC down.

spirv_glsl.cpp
```diff
@@ -3258,6 +3306,78 @@ void CompilerGLSL::emit_resources()
 	declare_undefined_values();
 }
 
+void CompilerGLSL::emit_extension_workarounds()
+{
+	//GL_KHR_shader_subgroup_vote
```
HansKristian-Work, line 3311: Space between `//` and `GL`

spirv_glsl.cpp
```diff
@@ -9852,7 +9972,7 @@ void CompilerGLSL::emit_instruction(const Instruction &instruction)
 	{
 		auto &type = get<SPIRType>(ops[0]);
 		if (type.vecsize > 1)
-			GLSL_UFOP(not );
+			GLSL_UFOP(not);
```
HansKristian-Work, line 9975: Fallout from running formatting script?

spirv_glsl.cpp
```diff
@@ -2890,8 +2939,7 @@ void CompilerGLSL::declare_undefined_values()
 		if (options.force_zero_initialized_variables && type_can_zero_initialize(type))
 			initializer = join(" = ", to_zero_initialized_expression(undef.basetype));
 
-		statement(variable_decl(type, to_name(undef.self), undef.self), initializer,
-		          ";");
+		statement(variable_decl(type, to_name(undef.self), undef.self), initializer, ";");
```
HansKristian-Work, line 2942: Fallout from using formatting script?

spirv_glsl.cpp
HansKristian-Work: `subgroupBroadcast(value, subgroupElect(void))`, shouldn't that be `subgroupBroadcastFirst(value) == value`?

spirv_glsl.hpp
devshgraphicsprogramming: will `emit_header` be called before `emit_subgroup_op` ?

spirv_glsl.cpp
devshgraphicsprogramming: why are you defining an extension?

spirv_glsl.cpp
devshgraphicsprogramming: we don't mean spec constant or a uniform, we could just declare it as a `const uint SugroupSize = ;`

Should we prefix with `SPIRV_Cross_gl` instead of `gl`?

Crisspl: > anything starting with gl_ in GLSL is reserved, so I guess we'll need to use a different prefix

I thought it applies to GLSL only, without preprocessor?

> If you emit a constant, you can call it SPIRV_Cross_gl_something, and change the name where builtin name is queried depending on vulkan_semantic. 

Change the name?

spirv_glsl.cpp
Crisspl: > But turn the subgroupBarrier into a NOOP (maybe emit a comment for clarity).

Doesnt matter in case when error is generated anyway

spirv_glsl.cpp
Crisspl: ah, thanks i forgot about bool overload

spirv_glsl.cpp
devshgraphicsprogramming: yeah you're right about the  `subgroupBroadcastFirst`, edited.

spirv_glsl.hpp
devshgraphicsprogramming: @HansKristian-Work I dont like that approach anyway, because we should analyze the ops used and not the extensions declared because they dont map 1:1 with desktop GL

spirv_glsl.cpp
devshgraphicsprogramming: probably.

spirv_glsl.cpp
```diff
@@ -3258,6 +3306,78 @@ void CompilerGLSL::emit_resources()
 	declare_undefined_values();
 }
 
+void CompilerGLSL::emit_extension_workarounds()
+{
+	//GL_KHR_shader_subgroup_vote
+	if (!options.vulkan_semantics && declaredKHR_shader_subgroup_vote)
+	{
+		statement("#ifdef GL_ARB_shader_group_vote");
+		statement("bool subgroupAll(in bool v) { return allInvocationsARB(v); }");
```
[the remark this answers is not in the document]
devshgraphicsprogramming: wdym?

spirv_glsl.cpp
devshgraphicsprogramming: having a `subgroupBarrier` shouldn't be an error, its an execution barrier which with the ARB, AMD and NV extensions we're using turns into a NOOP (because the extensions state that subgroups execute in lockstep so this barrier is implicit).

spirv_glsl.cpp
devshgraphicsprogramming: Well for `GL_NV_shader_thread_group` we can define the following (@Crisspl the NV extension should always be our "primary"/preferred implementation)
```glsl
gl_SubgroupSize = gl_WarpSizeNV
gl_SubgroupInvocationID = gl_ThreadInWarpNV 
gl_SubgroupID = gl_WarpIDNV
gl_NumSubgroups = gl_WarpsPerSMNV
```

devshgraphicsprogramming: on AMD/ARB the `gl_SubgroupSize` can be obtained from `gl_SIMDGroupSizeAMD` and `gl_SubGroupSizeARB` respectively.

Non-NV requires `ARB_shader_ballot` to obtain `gl_SubgroupInvocationID` which is `gl_SubGroupInvocationARB`

devshgraphicsprogramming: You're right that `gl_NumSubgroup` and `gl_NumSubgroups` are impossible to obtain on non-NVidia systems, @HansKristian-Work

spirv_glsl.hpp
devshgraphicsprogramming: > I'd rather track the following (instead of KHR extensions):
> 
> * needs subgroup voting on booleans
> * needs subgroup voting on non-booleans
> * needs subgroup masks
> * needs subgroup elect
> * needs subgroup ballots
> 

I'd also track:
- needs the input `gl_SubgroupID` or `gl_NumSubgroups` (requires NV)
- needs the input `gl_SubgroupSize` requires (requires NV, AMD or ARB_shader_ballot )
- needs the input `gl_SubgroupInvocationID` (requires NV or ARB_shader_ballot )

Basically we need to analyse the SPIR-V opcodes and not the declared extension usage, because outside the NV ecosystem the support for the Vulkan extension is "partial".

Then we can establish which combos of extensions can potentially implement the shader.

devshgraphicsprogramming:
@Crisspl apparently OpenGL has the extensions now in new drivers https://www.khronos.org/registry/OpenGL/extensions/KHR/KHR_shader_subgroup.txt

So I guess our fallback chain is now:
1) `KHR_shader_subgroup`
2) `NV`
3) `AMD`
4) the old `ARB` extensions

spirv_glsl.hpp
Crisspl: Spirv-cross kind of already does it by requiring specific extensions when found specific feature (builtin input/opcode/etc..), so I think it's better to use exisiting code and leave those `declared_<extension_name>` flags

for example see https://github.com/KhronosGroup/SPIRV-Cross/blob/master/spirv_glsl.cpp#L7356 or https://github.com/KhronosGroup/SPIRV-Cross/blob/master/spirv_glsl.cpp#L10906

devshgraphicsprogramming: yes, but...

these are part of `builtin_to_glsl` and `emit_instruction` which get called after the header is already emitted?

Anyway @HansKristian-Work suggested we use `traverse_all_reachable_opcodes`

Crisspl: hm, yea, you're right... No idea why spirv-cross does it then..

Crisspl: Ah, because it then forces recompilation. So actually it will do

spirv_glsl.cpp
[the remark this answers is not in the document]
Crisspl: that `in` is default qualifier and so theres no point in having it in code, i suppose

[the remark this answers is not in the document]
devshgraphicsprogramming: ok

spirv_glsl.cpp
```diff
@@ -666,6 +666,13 @@ void CompilerGLSL::emit_header()
 
 	for (auto &ext : forced_extensions)
 	{
+		if (ext == "GL_KHR_shader_subgroup")
+		{
+			declared_KHR_shader_subgroup_ballot = true;
+			declared_KHR_shader_subgroup_basic = true;
+			declared_KHR_shader_subgroup_vote = true;
```
devshgraphicsprogramming, line 673: explain?

spirv_glsl.cpp
```diff
@@ -6832,6 +6830,9 @@ void CompilerGLSL::emit_subgroup_op(const Instruction &i)
 	case OpGroupNonUniformAny:
 	case OpGroupNonUniformAllEqual:
 		require_extension_internal("GL_KHR_shader_subgroup_vote");
+		// to be able to emulate subgroupAllEqual(T) where T is not bool
+		if (!options.vulkan_semantics)
+			require_extension_internal("GL_KHR_shader_subgroup_ballot");
```
devshgraphicsprogramming, line 6835: you need to track more granular.

spirv_glsl.cpp
```diff
@@ -10937,8 +10938,14 @@ void CompilerGLSL::emit_instruction(const Instruction &instruction)
...
 		if (execution_scope == ScopeSubgroup || memory == ScopeSubgroup)
 		{
-			if (!options.vulkan_semantics)
-				SPIRV_CROSS_THROW("Can only use subgroup operations in Vulkan semantics.");
+			// OpControlBarrier with ScopeSubgroup is subgroupBarrier()
+			if (!options.vulkan_semantics && opcode != OpControlBarrier)
+			{
+				// Subgroup memory barriers are replaced by workgroup memory barriers if not vulkan_semantics
+				memory = ScopeWorkgroup;
+				execution_scope = ScopeWorkgroup;
```
devshgraphicsprogramming, line 10946: not sure what changing these two variables will do.

Crisspl: workgroup mem barrier and subgroup mem barrier have same SPIRV opcode but differs in Memory field. You have to read code below

spirv_glsl.cpp
Crisspl: why?

spirv_glsl.cpp
Crisspl: GL_KHR_shader_subgroup implies all of them

devshgraphicsprogramming: but there's no point tracking which were declared, since we need to track the reachable opcodes (parts of extensions used) because the problems with replicating functionality

spirv_glsl.cpp
devshgraphicsprogramming: yeah, but you can't emit a barrier in a non compute/non eval shader, so its best to just emit the statement you need and `break` like the case below for TES

spirv_glsl.cpp
Crisspl: we discussed it here https://github.com/KhronosGroup/SPIRV-Cross/pull/1476#discussion_r498804887
spirv-cross already does it

spirv_glsl.hpp
HansKristian-Work: > will `emit_header` be called before `emit_subgroup_op` ?

emit_header will be called before any functions. emit_header and emit_resources are the first things to be emitted.

Crisspl: i figured `require_extension_internal` forces recompilation so it doesnt really matter

spirv_glsl.cpp
Crisspl: fixed

spirv_glsl.hpp
devshgraphicsprogramming: > i figured `require_extension_internal` forces recompilation so it doesnt really matter

sounds like compile times will explode, its best to run poor-man's static analysis with `traverse_all_reachable_opcodes` before `emit_header` and tag all the opcodes groups and builtin groups that are of interest.

spirv_glsl.cpp
devshgraphicsprogramming: because you often get a situation where one extension defines multiple functionalities that need different subsets of desktop OpenGL extensions to express, the subset of extensions could be bigger if you actually analyze what functionality is actually being used.

For example:

# `GL_KHR_shader_subgroup_basic` (which is implicitly enabled by all the other ones)

If `gl_NumSubgroups` and `gl_SubgroupID` are used at all, then only the Nvidia extension can be used.

Do you want to implicitly enable all of the functionality that `GL_KHR_shader_subgroup_basic` provides and have SPIRV-Cross fail to produce dekstop OpenGL GLSL for other vendors, even when these two special builtins are not being used?

Do you wish to fail to compile a shader that does not use the above and does not use `gl_SubgroupInvocationID` when only `AMD_gcn_shader` is available?

spirv_glsl.cpp
```diff
@@ -718,6 +719,92 @@ void CompilerGLSL::emit_header()
...
 			statement("#endif");
 		}
+		else if (ext == "GL_KHR_shader_subgroup_ballot")
+		{
+			statement("#ifdef GL_KHR_shader_subgroup");
+			statement("#extension GL_KHR_shader_subgroup : require");
+
+			statement("#elif defined(GL_NV_shader_thread_group) && defined(GL_NV_shader_thread_shuffle)");
+			statement("#extension GL_NV_shader_thread_group : require");
+			statement("#extension GL_NV_shader_thread_shuffle : require");
```
devshgraphicsprogramming, line 729: why do you need shuffle for ballot?

spirv_glsl.cpp
```diff
@@ -3258,6 +3344,107 @@ void CompilerGLSL::emit_resources()
...
+			statement(std::string("SUBGROUP_ALL_EQUAL_WORKAROUND(") + t + ")");
+		}
+		statement("#undef SUBGROUP_ALL_EQUAL_WORKAROUND");
+
+		statement("#elif defined(GL_NV_gpu_shader5)");
+		statement("bool subgroupAll(bool value) { return allThreadsNV(value); }");
+		statement("bool subgroupAny(bool value) { return anyThreadNV(value); }");
+		statement("bool subgroupAllEqual(bool value) { return allThreadsEqualNV(value); }");
+
+		statement("#elif defined(GL_AMD_gcn_shader)&&defined(GL_AMD_gpu_shader_int64)");
```
devshgraphicsprogramming, line 3419: `gcn_shader` requires ARB extension 64-bit integers, not AMD one, IIRC

spirv_glsl.cpp
devshgraphicsprogramming: Basically I see two sane methods to approach this, either:

A) split the subgroup extensions into more "meta-extensions" and require these when finding an opcode/builtin

B) Keep a list of states for `GL_NV_shader_thread_group, GL_NV_shader_thread_shuffle, GL_ARB_shader_group_vote, GL_NV_gpu_shader5, GL_AMD_gcn_shader, GL_ARB_shader_ballot` and flag them as:
- "not needed" (no opcode or builtin has been encountered which could be expressed in a way requiring this extension)
- "potentially needed" (logical inverse of the above)
- "required" (only way to express a functionality, i.e. `gl_NumSubgroups` forces NV extension)
- "redundant" (another extension which is "required" and the extensions it requires covers this extension 100% in features which will be used)

spirv_glsl.hpp
devshgraphicsprogramming: https://github.com/KhronosGroup/SPIRV-Cross/pull/1476/files#r498956448

spirv_glsl.cpp
devshgraphicsprogramming: I guess you need to test some driver compilers in OpenGL if they let you feed them GLSL core profile shaders with `#define` which define things which start with `gl_` and then use them in the code

spirv_glsl.cpp
Crisspl: https://www.khronos.org/registry/OpenGL/extensions/AMD/AMD_gcn_shader.txt

> AMD_gpu_shader_int64 or NV_gpu_shader5 is required.

devshgraphicsprogramming: ah crap, thanks.

spirv_glsl.hpp
```diff
@@ -64,6 +65,149 @@ typedef uint32_t AccessChainFlags;
...
+			NV_gpu_shader_5,
+			NV_shader_thread_group,
+			NV_shader_thread_shuffle,
+			ARB_shader_ballot,
+			ARB_shader_group_vote,
+			AMD_gcn_shader,
+
+			CandidateCount
+		};
+		static std::string getExtensionName(Candidate c)
```
devshgraphicsprogramming, line 82: @HansKristian-Work do you prefer a static constexpr array?

spirv_glsl.hpp
```diff
@@ -64,6 +65,149 @@ typedef uint32_t AccessChainFlags;
...
+			}
+		}
+		static SmallVector<std::string> getExtraRequiredExtensionNames(Candidate c)
+		{
+			switch (c)
+			{
+			case ARB_shader_ballot:
+				return { "GL_ARB_shader_int64" };
+			case AMD_gcn_shader:
+				return { "GL_AMD_gpu_shader_int64" };
```
devshgraphicsprogramming, line 109: but it can be either that or NV_gpu_shader5

spirv_glsl.hpp
```diff
@@ -64,6 +65,149 @@ typedef uint32_t AccessChainFlags;
...
+			SubgroupInvocationID,
+			SubgroupID,
+			NumSubgroups,
+			SubgroupBrodcast_First,
+			SubgroupBallotFindLSB_MSB,
+			SubgroupAll_Any_AllEqual,
+			SubgroupElect,
+			SubgroupBarrier,
+			SubgroupMemBarrier,
+			SubgroupBallot,
```
devshgraphicsprogramming, line 128: a little bit of documentation for the people who come after us would be useful.

spirv_glsl.hpp
```diff
@@ -64,6 +65,149 @@ typedef uint32_t AccessChainFlags;
...
+			case SubgroupInvocationID:
+				return { NV_shader_thread_group, ARB_shader_ballot };
+			case SubgroupID:
+				return { NV_shader_thread_group };
+			case NumSubgroups:
+				return { NV_shader_thread_group };
+			case SubgroupBrodcast_First:
+				return { NV_shader_thread_shuffle, ARB_shader_ballot };
+			case SubgroupBallotFindLSB_MSB:
+				return { NV_shader_thread_group };
```
devshgraphicsprogramming, line 192: why? LSB and MSB of the ballot can be done with anything.

spirv_glsl.hpp
HansKristian-Work: `static const char *[]` seems appropriate here.

spirv_glsl.hpp
```diff
@@ -64,6 +65,149 @@ typedef uint32_t AccessChainFlags;
...
+			case SubgroupAll_Any_AllEqual:
+				return { NV_gpu_shader_5, ARB_shader_group_vote, AMD_gcn_shader };
+			case SubgroupElect:
+				return { NV_shader_thread_group };
```
devshgraphicsprogramming, line 196: we can do our own subgroup elect workaround.

spirv_glsl.hpp
```diff
@@ -64,6 +65,149 @@ typedef uint32_t AccessChainFlags;
 
 class CompilerGLSL : public Compiler
 {
+	struct KHR_shader_subgroupSupportHelper
```
HansKristian-Work, line 68: Move this below `private:`. Also, in the public header, avoid the inlined definitions to keep the code more compact.

HansKristian-Work: The class names should be pascal case, i.e. ShaderSubgroupSupportHelper. Don't need the KHR.

spirv_glsl.hpp
```diff
@@ -64,6 +65,149 @@ typedef uint32_t AccessChainFlags;
...
+				return "GL_ARB_shader_ballot";
+			case ARB_shader_group_vote:
+				return "GL_ARB_shader_group_vote";
+			case AMD_gcn_shader:
+				return "GL_AMD_gcn_shader";
+			default:
+				return "";
+			}
+		}
+		static SmallVector<std::string> getExtraRequiredExtensionNames(Candidate c)
```
HansKristian-Work, line 102: No camel case, get_extra_required_extension_names.

spirv_cross_containers.hpp
```diff
@@ -212,6 +211,10 @@ class SmallVector : public VectorView<T>
 		this->buffer_size = count;
 	}
 
+	SmallVector(std::initializer_list<T> init) SPIRV_CROSS_NOEXCEPT : SmallVector(init.begin(), init.end())
```
HansKristian-Work, line 214: What is this used for?

spirv_glsl.hpp
```diff
@@ -19,6 +19,7 @@
 
 #include "GLSL.std.450.h"
 #include "spirv_cross.hpp"
+#include <array>
```
HansKristian-Work, line 22: Is this needed?

spirv_glsl.cpp
```diff
@@ -3258,6 +3301,292 @@ void CompilerGLSL::emit_resources()
 	declare_undefined_values();
 }
 
+void CompilerGLSL::emit_extension_workarounds()
+{
+	const auto TYPES = { "int",   "ivec2", "ivec3", "ivec4", "uint",   "uvec2", "uvec3", "uvec4",
```
HansKristian-Work, line 3306: It can be awkward to deduce what type this is. I suggest `static const char *[]` here.

spirv_glsl.hpp
Crisspl: yes, used in Supporter class ::Result

spirv_cross_containers.hpp
Crisspl: to easily create vector from list of elements. See `KHR_shader_subgroupSupportHelper::getCandidatesForFeature`

devshgraphicsprogramming: so you can construct from a C++11 intializer list I guess?

spirv_glsl.hpp
devshgraphicsprogramming: return type of the subgroup helper (ordered list of extensions to use)

spirv_glsl.cpp
```diff
@@ -719,94 +719,51 @@ void CompilerGLSL::emit_header()
...
-			if (options.vulkan_semantics)
-				statement("#extension GL_KHR_shader_subgroup_vote : require");
-			else
+			Supp::Feature feature = static_cast<Supp::Feature>(ft);
+			if (!shader_subgroup_supporter.isFeatureRequested(feature))
+				continue;
+
+			auto exts = shader_subgroup_supporter.getCandidatesForFeature(feature, result);
+			if (exts.empty())
+				continue;
```
devshgraphicsprogramming, line 739: wouldn't that require you to emit an `#error` statement?

spirv_glsl.cpp
```diff
@@ -719,94 +719,51 @@ void CompilerGLSL::emit_header()
...
+
+			statement("#ifdef GL_KHR_shader_subgroup");
+			statement("#extension GL_KHR_shader_subgroup : require");
```
devshgraphicsprogramming, line 742: doesn't this require the whole set of 6 extensions?

spirv_glsl.cpp
```diff
@@ -719,94 +719,51 @@ void CompilerGLSL::emit_header()
...
+				else
+				{
+					std::string extra;
+					for (const auto &e : extraNames)
+						extra += " && defined(" + e + ")";
+					statement("#elif defined(", name, ")", extra);
+				}
+				for (const auto &e : extraNames)
+					statement("#extension ", e, " : required");
+				statement("#extension ", name, " : required");
```
devshgraphicsprogramming, line 758: how does this work, what is it supposed to emit?

spirv_glsl.cpp
```diff
@@ -3348,100 +3305,285 @@ void CompilerGLSL::emit_extension_workarounds()
...
+			{
+				const Supp::Candidate e = exts[i];
+				const std::string name = Supp::getExtensionName(e);
 
-		statement("#elif defined(GL_ARB_shader_ballot)");
+				statement(i == 0u ? "#if" : "#elif", " defined(", name, ")");
+				switch (e)
+				{
+				case Supp::NV_shader_thread_group:
+					statement("#define gl_SubgroupEqMask uvec4(gl_ThreadEqMaskNV, 0u, 0u, 0u)");
```
devshgraphicsprogramming, line 3328: I see a problem here, one that can be easily solved.

If `KHR_shader_subgroup_X` is defined (available) then you shouldn't emit any workarounds for `KHR_shader_subgroup_X`

This you could solve by actually having the `KHR` extensions in the candidate set, and have them take precedence even over the NV extensions.

spirv_glsl.cpp
```diff
@@ -3348,100 +3305,285 @@ void CompilerGLSL::emit_extension_workarounds()
...
+						statement(string(t) + " subgroupBroadcastFirst(" + t +
+						          " value) { return shuffleNV(value, 0, gl_WarpSizeNV); }");
+					for (const char *t : TYPES)
+						statement(string(t) + " subgroupBroadcast(" + t +
+						          " value, uint id) { return shuffleNV(value, id, gl_WarpSizeNV); }");
+					break;
+				case Supp::ARB_shader_ballot:
+					for (const char *t : TYPES)
+						statement(string(t) + " subgroupBroadcastFirst(" + t +
+						          " value) { return readInvocationARB(value); }");
```
devshgraphicsprogramming, line 3453: readFirstInvocationARB

spirv_glsl.cpp
```diff
@@ -3348,100 +3305,285 @@ void CompilerGLSL::emit_extension_workarounds()
...
+				const Supp::Candidate e = exts[i];
+				const std::string name = Supp::getExtensionName(e);
+
+				statement(i == 0u ? "#if" : "#elif", " defined(", name, ")");
+				switch (e)
+				{
+				case Supp::NV_shader_thread_shuffle:
+					for (const char *t : TYPES)
+						statement(string(t) + " subgroupBroadcastFirst(" + t +
+						          " value) { return shuffleNV(value, 0, gl_WarpSizeNV); }");
```
devshgraphicsprogramming, line 3445: `shuffleNV` has slightly different semantics to `subgroupBroadcastFirst` and `readFirstInvocationARB`, you'd need to use ` findLSB(ballotThreadNV(true))` instead of 0 here.

spirv_glsl.cpp
```diff
@@ -3348,100 +3305,285 @@ void CompilerGLSL::emit_extension_workarounds()
...
+			statement("int firstLive = findLSB(value.x);");
+			statement("return uint(firstLive != -1 ? firstLive : (findLSB(value.y) + 32));");
+			statement("}");
+			statement("uint subgroupBallotFindMSB(uvec4 value) {");
+			statement("int firstLive = findMSB(value.y);");
+			statement("return uint(firstLive != -1 ? (firstLive + 32) : findMSB(value.x));");
+			statement("}");
+			statement("#endif");
+		}
+		else if (shader_subgroup_supporter.isFeatureRequested(Supp::SubgroupAll_Any_AllEqual))
```
devshgraphicsprogramming, line 3491: split this feature into 2, one for bool and one for other types.

The other type feature will need to imply the usage of bool feature  + subgroup broadcast first)

spirv_glsl.cpp
Crisspl: not relevant any more

spirv_glsl.cpp
Crisspl:
```
#elif defined(ext1) && defined(extra1)
#extension ext1 : required
#extension extra1 : required
#elif defined(ext2) && defined(extra2)
#extension ext2 : required
#extension extra2 : required
...
...
...
#else
#error
#endif
```

devshgraphicsprogramming: not sure the `extra` extension thing is a good design idea. I'd rather have the candidate extension have its own array of statements for dependancy/ladder extensions.

I.e for `AMD_gcn_shader` this could be
```cpp
{
"(defined(GL_AMD_gpu_int64_t) || defined(GL_NV_gpu_shader5))",
"#extension GL_AMD_gpu_int64_t : enable",
"#extension GL_NV_gpu_shader5 : enable"
}
```

Because there's a lot of "either/or" relations and possibly extensions depending on extensions.

Crisspl: yeah, i noticed before and already fixed

spirv_glsl.cpp
Crisspl: fixed

spirv_glsl.cpp
Crisspl: no, empty candidates list means no extension is needed to emulate

devshgraphicsprogramming: ah ok, better document it.

spirv_glsl.cpp
```diff
@@ -725,44 +725,39 @@ void CompilerGLSL::emit_header()
...
-					std::string extra;
-					for (const auto &e : extraNames)
-						extra += " && defined(" + e + ")";
-					statement("#elif defined(", name, ")", extra);
-				}
+				const Supp::Candidate ext = *it;
+
+				std::string name = Supp::get_extension_name(ext);
+				std::string extra = Supp::get_extra_required_extension_checks(ext);
+				auto extraNames = Supp::get_extra_required_extension_names(ext);
```
devshgraphicsprogramming, line 747: I'd get the `extraPredicate` since right now you'll be requiring both AMD and NV extension.

its much easier to have
```glsl
#if defined(GL_VENDOR_extension_name) && ( predicate )
``` 
```

File: spirv_glsl.cpp [hunk 42]
```diff
@@ -725,44 +725,39 @@ void CompilerGLSL::emit_header()
...
+				statement(it != exts.begin() ? "#elif" : "#if", " defined(", name, ")",
+				          extra.empty() ? "" : (" && " + extra));
 				for (const auto &e : extraNames)
-					statement("#extension ", e, " : required");
+					statement("#extension ", e, " : enable");
 				statement("#extension ", name, " : required");
 			}
-			if (!exts.empty()) // if getCandidatesForFeature() returns empty list, it means no extensions are needed
+			// if get_candidates_for_feature() returns empty list, it means no extensions are needed
+			if (!exts.empty() || Supp::can_feature_be_implemented_wo_extensions(feature))
```
devshgraphicsprogramming, line 755: btw right now it looks like you require NV extension for LSB and MSB of the ballot,

spirv_glsl.cpp
```diff
@@ -3584,6 +3581,15 @@ void CompilerGLSL::emit_extension_workarounds()
 			}
 			statement("#endif");
 		}
+		else if (shader_subgroup_supporter.is_feature_requested(Supp::SubgroupMemBarrier))
+		{
+			statement("#ifndef GL_KHR_shader_subgroup_basic");
```
devshgraphicsprogramming, line 3586: I'd issue more statements to obtain a 
```glsl
#ifdef GL_KHR_shader_subgroup_basic
#else
// the stuff
#endif
```

devshgraphicsprogramming: Actually you'd need to discriminate if you're compiling a Compute / Tessellation Eval Shader or the rest.

Because if its a shader with workgroups (Compute/TES) then you can replace all the barriers by `groupMemoryBarrier` because 

> groupMemoryBarrier waits on the completion of all memory accesses performed by an invocation of a compute shader relative to the same access performed by other invocations in the same work group and then returns with no other effect.

spirv_glsl.hpp
```diff
@@ -130,16 +138,36 @@ class CompilerGLSL : public Compiler
...
+			const static bool retval[FeatureCount]{
+				false,
+				false,
+				false,
+				false,
+				false,
+				false,
+				true, // SubgroupBalloFindLSB_MSB
+				false,
+				true, // SubgroupElect
```
devshgraphicsprogramming, line 152: yes, but using these features implies using others.

You need another array of what feature implies other features.

spirv_glsl.cpp
devshgraphicsprogramming: @Crisspl did you check if Vendor OpenGL runtime compilers will compile GLSL with a `#define` where the symbol starts with `gl_` ?

spirv_glsl.cpp
devshgraphicsprogramming: ok now seeing the code, I can see that you've already done that

would be useful to call `extra` `extraPredicate` instead.

spirv_glsl.cpp
Crisspl: yes, because in case of NV you have to perform findLSB/MSB on just one uint instead of two. It's your idea from https://github.com/KhronosGroup/SPIRV-Cross/issues/1352

devshgraphicsprogramming: yeah, but if NV extension is not available, we should sitll have working subgroup ballot LSB/MSB

spirv_glsl.cpp
```diff
@@ -3583,12 +3583,22 @@ void CompilerGLSL::emit_extension_workarounds()
...
+				statement("#ifndef GL_KHR_shader_subgroup_basic");
+				statement("void subgroupMemoryBarrier() { memoryBarrier(); }");
+				statement("void subgroupMemoryBarrierBuffer() { memoryBarrierBuffer(); }");
+				statement("void subgroupMemoryBarrierImage() { memoryBarrierImage(); }");
+				statement("#endif");
+			}
+			else
+			{
+				statement("#error Subgroup memory barriers can be emulated only in tesselation control and compute "
+				          "execution models.");
```
devshgraphicsprogramming, line 3597: this is not what I meant.

spirv_glsl.cpp
```diff
@@ -3583,12 +3583,22 @@ void CompilerGLSL::emit_extension_workarounds()
...
+			}
+
+			if (model == spv::ExecutionModelGLCompute)
+				statement("void subgroupMemoryBarrierShared() { memoryBarrierShared(); }");
```
devshgraphicsprogramming, line 3601: tessellation can use shared memory, so you can hoist it back.

spirv_glsl.cpp
devshgraphicsprogramming: I meant

### For Compute or Evaluation
```cpp
			statement("void subgroupMemoryBarrier() { groupMemoryBarrier(); }");
			statement("void subgroupMemoryBarrierBuffer() { groupMemoryBarrier(); }");
			statement("void subgroupMemoryBarrierShared() { groupMemoryBarrier(); }");
			statement("void subgroupMemoryBarrierImage() { groupMemoryBarrier(); }");
```
### For Everything Else
```cpp
			statement("void subgroupMemoryBarrier() { memoryBarrier(); }");
			statement("void subgroupMemoryBarrierBuffer() { memoryBarrierBuffer(); }");
			statement("void subgroupMemoryBarrierImage() { memoryBarrierImage(); }");
```

spirv_glsl.hpp
Crisspl: > yes, but using these features implies using others.

how? no workaround depends on another one

spirv_glsl.cpp
Crisspl: I've read GLSL spec and `memoryBarrierShared` is for compute only

Crisspl: same for `groupMemoryBarrier`

spirv_glsl.cpp
devshgraphicsprogramming: blergh I made a mistake, its the Control shader not the Evaluation shader that's like compute

spirv_glsl.hpp
devshgraphicsprogramming: Let me give the only example, assuming you will split `SubgroupAll_Any_AllEqual` into `SubgroupAll_Any_AllBool` and `SubgroupAllEqualT`:

## Feature::SubgroupAllEqualT

Requires both:
- SubgroupBrodcast_First
- SubgroupAll_Any_AllBool

devshgraphicsprogramming: I foresee that when we implement the rest of the intrinsics (scans and shuffles), we will often need to piggyback off other features for workarounds.

spirv_glsl.hpp
devshgraphicsprogramming: @crisspl one more example of something that requires another feature

subgroupElect can be achieved via `LSB` of a ballot, so needs ballot feature.

spirv_glsl.hpp
devshgraphicsprogramming: Also just got reminded of subgroup elect https://github.com/KhronosGroup/SPIRV-Cross/pull/1476#discussion_r500395226

spirv_glsl.cpp
devshgraphicsprogramming: and forgot that TCS can only use `barrier` but not shared mem and doesnt have explicit workgroups

spirv_glsl.cpp
Crisspl: thats why there is `can_feature_be_implemented_wo_extensions` function ;)

spirv_glsl.cpp
```diff
@@ -3535,70 +3559,53 @@ void CompilerGLSL::emit_extension_workarounds(spv::ExecutionModel model)
...
+			statement("#ifndef GL_KHR_shader_subgroup_basic");
 			statement("bool subgroupElect() {");
-			statement("uvec2 activeMask = unpackUint2x32(ballotARB(true));");
-			statement("int firstLive = findLSB(activeMask.x);");
-			statement("if (firstLive == -1)");
-			statement("firstLive = findLSB(activeMask.y)+32;");
-			statement("return gl_SubgroupInvocationID == uint(firstLive);");
+			statement("uvec4 activeMask = subgroupBallot(true);");
+			statement("uint firstLive = subgroupBallotFindLSB(activeMask);");
+			statement("return gl_SubgroupInvocationID == firstLive;");
```
devshgraphicsprogramming, line 3578: does NV/AMD not have specific instructions for electing an invocation?

spirv_glsl.hpp
```diff
@@ -138,34 +139,48 @@ class CompilerGLSL : public Compiler
...
+			}
+		}
+		static FeatureMask get_feature_dependency_mask(Feature feature)
+		{
+			return build_mask(get_feature_dependencies(feature));
+		}
+
 		static bool can_feature_be_implemented_wo_extensions(Feature feature)
 		{
 			const static bool retval[FeatureCount]{
```
devshgraphicsprogramming, line 164: I think this should be a switch, or get constructed explicitly via
`retval[feature] = value`
so this mapping is not vulnerable to enumerations changing or getting inserted inbetween.

spirv_glsl.cpp
devshgraphicsprogramming: ok I checked on NVidia and Intel, `#define` with a `gl_` prefix (LOWER CASE) is fine, upper case is not... so we're in the clear

spec is also worded so.

devshgraphicsprogramming: @HansKristian-Work you're fine with this `#define gl_` ?

spirv_glsl.hpp
devshgraphicsprogramming: @Crisspl ?

Crisspl:
@HansKristian-Work I wasnt able to actually run test suite (python gives me this: `Error: [WinError 2] The system cannot find the file specified`) and so it didnt put new reference file. I've put one manually, but I'm not sure it is like it's supposed to be. Please check before merge.

reference/shaders-no-opt/comp/subgroups_basicvoteballot.vk.comp
```diff
@@ -284,6 +284,51 @@ void subgroupMemoryBarrierBuffer() { groupMemoryBarrier(); }
...
+#endif
+}
+uint subgroupBallotExclusiveBitCount(uvec4 value)
+{
+uvec4 v = value & gl_SubgroupLtMask;
+ivec4 c = bitCount(v);
+#ifdef GL_NV_shader_thread_group
+return uint(c.x);
+#else
+return uint(c.x + c.y + c.z + c.w);
```
devshgraphicsprogramming, line 309: actually you only need to do `c.x+c.y` because nobody specifies subgroups > 64 in GL extensions

reference/shaders-no-opt/comp/subgroups_basicvoteballot.vk.comp
```diff
@@ -284,6 +284,51 @@ void subgroupMemoryBarrierBuffer() { groupMemoryBarrier(); }
...
+#endif
+}
+uint subgroupBallotExclusiveBitCount(uvec4 value)
+{
+uvec4 v = value & gl_SubgroupLtMask;
+ivec4 c = bitCount(v);
```
devshgraphicsprogramming, line 305: only a `bitcount(v.xy)` needed

reference/shaders-no-opt/comp/subgroups_basicvoteballot.vk.comp
```diff
@@ -284,6 +284,51 @@ void subgroupMemoryBarrierBuffer() { groupMemoryBarrier(); }
 void subgroupMemoryBarrierShared() { groupMemoryBarrier(); }
 void subgroupMemoryBarrierImage() { groupMemoryBarrier(); }
 #endif
+#ifndef GL_KHR_shader_subgroup_ballot
+bool subgroupInverseBallot(uvec4 value, uint index)
+{
+return any(notEqual(value & gl_SubgroupEqMask, uvec4(0u)));
```
devshgraphicsprogramming, line 290: use only the `xy` components and `uvec4`

devshgraphicsprogramming: basically all my issues are that you should only use `xy` parts

reference/shaders-no-opt/comp/subgroups_basicvoteballot.vk.comp
devshgraphicsprogramming: inclusive and exclusive bitcounts could be defined in terms of `subgroupBallotBitCount`

reference/shaders-no-opt/comp/subgroups_basicvoteballot.vk.comp
```diff
@@ -284,6 +284,51 @@ void subgroupMemoryBarrierBuffer() { groupMemoryBarrier(); }
...
+return uint(c.x);
+#else
+return uint(c.x + c.y + c.z + c.w);
+#endif
+}
+#endif
+#ifndef GL_KHR_shader_subgroup_ballot
+uint subgroupBallotBitCount(uvec4 value)
+{
+uvec4 v = value & subgroupBallot(true);
```
devshgraphicsprogramming, line 316: this is a tautology, `value` must have come from `subgroupBallot` so its implied that inactive invocations will set the bit to 0

reference/shaders-no-opt/comp/subgroups_basicvoteballot.vk.comp
```diff
@@ -284,6 +284,51 @@ void subgroupMemoryBarrierBuffer() { groupMemoryBarrier(); }
...
+return uint(c.x);
+#else
+return uint(c.x + c.y + c.z + c.w);
+#endif
+}
+#endif
+#ifndef GL_KHR_shader_subgroup_ballot
+bool subgroupBallotBitExtract(uvec4 value, uint index)
+{
+uvec4 mask = subgroupBallot(gl_SubgroupInvocationID == index);
```
devshgraphicsprogramming, line 328: you can extract the bit straight from `value`

like this
```glsl
#ifdef GL_NV_shader_thread_group
   uint shifted = value.x >> index;
#else
   uint shifted = value[index >> 5u] >> (index & 0x1fu);
#endif
   return (shifted & 0x1u) != 0u;
```

spirv_glsl.hpp
Crisspl: done

reference/shaders-no-opt/comp/subgroups_basicvoteballot.vk.comp
Crisspl: done

HansKristian-Work (changes requested):
First, please squash the PR.

Then, there is a compilation issue with GCC 4.8.

https://travis-ci.org/github/KhronosGroup/SPIRV-Cross/jobs/733747288#L1178

To declare `std::unordered_set<enum>` you have to also define a hash function for that enum by declaring a specialized `std::hash` manually. That will get very messy, so just use std::unordered_set of uint32_t here instead.

There's also some scary warnings that must be fixed. Returning a std::initializer_list does not work.

```
In file included from spirv_glsl.cpp:17:
spirv_glsl.hpp: In static member function ‘static std::initializer_list<spirv_cross::CompilerGLSL::ShaderSubgroupSupportHelper::Feature> spirv_cross::CompilerGLSL::ShaderSubgroupSupportHelper::get_feature_dependencies(spirv_cross::CompilerGLSL::ShaderSubgroupSupportHelper::Feature)’:
spirv_glsl.hpp:154:67: warning: returning temporary ‘initializer_list’ does not extend the lifetime of the underlying array [-Winit-list-lifetime]
  154 |     return { SubgroupBrodcast_First, SubgroupAll_Any_AllEqualBool };
      |                                                                   ^
spirv_glsl.hpp:156:78: warning: returning temporary ‘initializer_list’ does not extend the lifetime of the underlying array [-Winit-list-lifetime]
  156 |     return { SubgroupBallotFindLSB_MSB, SubgroupBallot, SubgroupInvocationID };
      |                                                                              ^
spirv_glsl.hpp:158:27: warning: returning temporary ‘initializer_list’ does not extend the lifetime of the underlying array [-Winit-list-lifetime]
  158 |     return { SubgroupMask };
      |                           ^
spirv_glsl.hpp:160:29: warning: returning temporary ‘initializer_list’ does not extend the lifetime of the underlying array [-Winit-list-lifetime]
  160 |     return { SubgroupBallot };
      |                             ^
spirv_glsl.hpp: In member function ‘spirv_cross::CompilerGLSL::ShaderSubgroupSupportHelper::Result spirv_cross::CompilerGLSL::ShaderSubgroupSupportHelper::resolve() const’:
spirv_glsl.hpp:247:19: warning: unused variable ‘d’ [-Wunused-variable]
  247 |      for (Feature d : deps)
      |                   ^
spirv_glsl.cpp: In member function ‘void spirv_cross::CompilerGLSL::emit_extension_workarounds(spv::ExecutionModel)’:
spirv_glsl.cpp:3662:15: warning: this statement may fall through [-Wimplicit-fallthrough=]
 3662 |      statement("uvec4 subgroupBallot(bool v) { return uvec4(ballotThreadNV(v), 0u, 0u, 0u); }");
      |      ~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
spirv_glsl.cpp:3663:5: note: here
 3663 |     case Supp::ARB_shader_ballot:
      |     ^~~~
```

reference/shaders-no-opt/frag/fp16.invalid.desktop.frag
```diff
@@ -50,7 +50,7 @@ void test_conversions()
     float16_t one = test_result();
     int a = int(one);
     uint b = uint(one);
-    bool c = !(one == float16_t(0.0));
+    bool c = one != float16_t(0.0);
```
HansKristian-Work, line 53: This file should not change. It failed local testing.

HansKristian-Work: Was this commited by mistake?

spirv_glsl.cpp
devshgraphicsprogramming: @Crisspl ?

spirv_glsl.cpp
devshgraphicsprogramming: @HansKristian-Work want anything done about it?

spirv_glsl.cpp
devshgraphicsprogramming: @HansKristian-Work want anything done about it?

spirv_glsl.hpp
```diff
@@ -241,6 +501,9 @@ class CompilerGLSL : public Compiler
 	// - Images which are statically used at least once with Dref opcodes.
 	bool variable_is_depth_or_compare(VariableID id) const;
 
+	// TODO remove this function when all subgroup ops are supported (or make it always return true)
+	static bool is_supported_subgroup_op(spv::Op op);
```
HansKristian-Work, line 505: Does this have to be public? It does not seem to belong to the public API.

spirv_glsl.hpp
```diff
@@ -64,6 +65,265 @@ typedef uint32_t AccessChainFlags;
 
 class CompilerGLSL : public Compiler
 {
+protected:
+	struct ShaderSubgroupSupportHelper
```
HansKristian-Work, line 69: As mentioned earlier, please move this class to the bottom of CompilerGLSL (below protected).
Also, there shouldn't be any inline code for this. The implementation should be done in spirv_glsl.cpp.

spirv_glsl.cpp
HansKristian-Work: Ideally, unrelated formatting should not be part of this PR. I'll probably need to do a "format_all.sh" commit soon.

spirv_glsl.cpp
```diff
@@ -3258,6 +3313,378 @@ void CompilerGLSL::emit_resources()
...
+				statement("void subgroupMemoryBarrier() { memoryBarrier(); }");
+				statement("void subgroupMemoryBarrierBuffer() { memoryBarrierBuffer(); }");
+				statement("void subgroupMemoryBarrierImage() { memoryBarrierImage(); }");
+				statement("#endif");
+			}
+		}
+		if (shader_subgroup_supporter.is_feature_requested(Supp::SubgroupInverseBallot_InclBitCount_ExclBitCout))
+		{
+			statement("#ifndef GL_KHR_shader_subgroup_ballot");
+			statement("bool subgroupInverseBallot(uvec4 value, uint index)");
```
HansKristian-Work, line 3629: InverseBallot does not take an index.

spirv_glsl.hpp
Crisspl: Thought it might be useful for users. I can make it protected..

Crisspl:
> To declare std::unordered_set<enum> you have to also define a hash function for that enum by declaring a specialized std::hash manually. That will get very messy, so just use std::unordered_set of uint32_t here instead.

Oh, thats why unixes fail. Actually it's compiler's fault, c++ standard states that std::hash should be provided for all enum types. Just a fun fact.

HansKristian-Work:
> > To declare std::unordered_set you have to also define a hash function for that enum by declaring a specialized std::hash manually. That will get very messy, so just use std::unordered_set of uint32_t here instead.
> 
> Oh, thats why unixes fail. Actually it's compiler's fault, c++ standard states that std::hash should be provided for all enum types. Just a fun fact.

Yes, modern GCC does not fail, but GCC 4.8 is min-spec for SPIRV-Cross. Compiler workarounds sometimes happen. There's a million workarounds for MSVC 2013/2015 in here :(

reference/shaders-no-opt/frag/fp16.invalid.desktop.frag
Crisspl: yes, sorry, test suite `--update` did that and i didnt notice

Crisspl:
Ok, everything fixed as you wish

HansKristian-Work:
Ok, next step is squashing the history to one commit, as this is one logical change to me. Multiple commits is fine too if each commit accomplishes something meaningful on their own.

HansKristian-Work:
After that, it's mostly stylistic considerations. The implementation strategy seems to be well thought out, and I don't have experience using subgroups on GL, so I'll trust your implementation. We also have a framework for future subgroup work, so the next PR should be much lighter.

For stylistic considerations, let me know if you prefer me to review all of them in detail, or I can merge it locally, and add fixup nits in a different commit and push that. The time spent for me should be similar anyways.

Crisspl:
Squashed all commits into 1.

> or I can merge it locally, and add fixup nits in a different commit and push that. 

If you could, it would be great.

spirv_glsl.cpp
Crisspl: Looked for in https://www.khronos.org/registry/OpenGL/extensions/ , havent found anything directly corresponding

HansKristian-Work:
Merged on master. Thanks for your contribution!

PR #1485, opened 2020-10-13
MSL: Support SPV_EXT_demote_to_helper_invocation for MSL 2.3.

cdavis5e:
MSL 2.3 has everything needed to support this extension on all
platforms. The existing `discard_fragment()` function was given demote
semantics, similar to Direct3D, and the `simd_is_helper_thread()`
function was finally added to iOS.

I've left the old test alone. Should I remove it in favor of these?


HansKristian-Work (approved):
LGTM!

PR #1489, opened 2020-10-13
MSL: Fix calculation of atomic image buffer address.

cdavis5e:
Fix reversed coordinates: `y` should be used to calculate the row
address. Align row address to the row stride.

Honestly, I don't know how this worked at all for Epic. It definitely
didn't work in the CTS prior to this.

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -563,6 +563,8 @@ struct CLIArguments
 	bool msl_vertex_for_tessellation = false;
 	uint32_t msl_additional_fixed_sample_mask = 0xffffffff;
 	bool msl_arrayed_subpass_input = false;
+	uint32_t msl_r32ui_linear_texture_alignment = 4;
+	uint32_t msl_r32ui_alignment_constant_id = 65535;
 	bool glsl_emit_push_constant_as_ubo = false;
 	bool glsl_emit_ubo_as_plain_uniforms = false;
 	bool glsl_force_flattened_io_blocks = false;
@@ -768,7 +770,11 @@ static void print_help_msl()
 	                "\t[--msl-additional-fixed-sample-mask <mask>]:\n"
 	                "\t\tSet an additional fixed sample mask. If the shader outputs a sample mask, then the final sample mask will be a bitwise AND of the two.\n"
 	                "\t[--msl-arrayed-subpass-input]:\n\t\tAssume that images of dimension SubpassData have multiple layers. Layered input attachments are accessed relative to BuiltInLayer.\n"
-	                "\t\tThis option has no effect if multiview is also enabled.\n");
+	                "\t\tThis option has no effect if multiview is also enabled.\n"
+	                "\t[--msl-r32ui-linear-texture-align <alignment>]:\n\t\tThe required alignment of linear textures of format MTLPixelFormatR32Uint.\n"
+	                "\t\tThis is used to align the row stride for atomic accesses to such images.\n"
+	                "\t[--msl-r32ui-linear-texture-align-constant-id <id>]:\n\t\tThe function constant ID to use for the linear texture alignment.\n"
+	                "\t\tOn MSL 1.2 or later, you can override the alignment by setting this function constant.\n");
 	// clang-format on
 }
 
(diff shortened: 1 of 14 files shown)


main.cpp
```diff
@@ -768,7 +769,9 @@ static void print_help_msl()
 	                "\t[--msl-additional-fixed-sample-mask <mask>]:\n"
 	                "\t\tSet an additional fixed sample mask. If the shader outputs a sample mask, then the final sample mask will be a bitwise AND of the two.\n"
 	                "\t[--msl-arrayed-subpass-input]:\n\t\tAssume that images of dimension SubpassData have multiple layers. Layered input attachments are accessed relative to BuiltInLayer.\n"
-	                "\t\tThis option has no effect if multiview is also enabled.\n");
+	                "\t\tThis option has no effect if multiview is also enabled.\n"
+	                "\t[--msl-r32ui-linear-texture-align <alignment>]:\n\t\tThe required alignment of linear textures of format MTLPixelFormatR32Uint.\n"
+					"\t\tThis is used to align the row stride for atomic accesses to such images.\n");
```
HansKristian-Work, line 774: Formatting.

main.cpp
```diff
@@ -1427,6 +1431,8 @@ static int main_inner(int argc, char *argv[])
 	cbs.add("--msl-additional-fixed-sample-mask",
 	        [&args](CLIParser &parser) { args.msl_additional_fixed_sample_mask = parser.next_hex_uint(); });
 	cbs.add("--msl-arrayed-subpass-input", [&args](CLIParser &) { args.msl_arrayed_subpass_input = true; });
+	cbs.add("--msl-r32ui-linear-texture-align",
+			[&args](CLIParser &parser) { args.msl_r32ui_linear_texture_alignment = parser.next_uint(); });
```
HansKristian-Work, line 1435: Formatting.

spirv_msl.cpp
```diff
@@ -4195,7 +4195,12 @@ void CompilerMSL::emit_custom_functions()
 		case SPVFuncImplImage2DAtomicCoords:
 		{
 			statement("// Returns buffer coords corresponding to 2D texture coords for emulating 2D texture atomics");
-			statement("#define spvImage2DAtomicCoord(tc, tex) (((tex).get_width() * (tc).x) + (tc).y)");
+			if (msl_options.r32ui_linear_texture_alignment <= 4)
+				statement("#define spvImage2DAtomicCoord(tc, tex) (((tex).get_width() * (tc).y) + (tc).x)");
+			else
+				statement("#define spvImage2DAtomicCoord(tc, tex) (((((tex).get_width() + ",
```
HansKristian-Work, line 4201: Wrap both cases in braces.

main.cpp
cdavis5e: I caught that like right before you commented, but forgot to push.

HansKristian-Work (changes requested):
Just nits, quite curious indeed. I guess they only used 1D images perhaps. Ping @LukasBanana.

HansKristian-Work (changes requested):
r32ui_linear_texture_alignment should probably be emitted as a function constant instead, and the option should be the ID. It seems useful for applications that want to ship converted MSL to supply the alignment after converting MSL.

cdavis5e:
> r32ui_linear_texture_alignment should probably be emitted as a function constant instead, and the option should be the ID. It seems useful for applications that want to ship converted MSL to supply the alignment after converting MSL.

Why not both? Function constants don't work on MSL 1.1 anyway.

spirv_msl.cpp
```diff
@@ -4195,7 +4195,16 @@ void CompilerMSL::emit_custom_functions()
...
-			statement("#define spvImage2DAtomicCoord(tc, tex) (((tex).get_width() * (tc).x) + (tc).y)");
+			if (msl_options.r32ui_linear_texture_alignment <= 4)
+			{
+				statement("#define spvImage2DAtomicCoord(tc, tex) (((tex).get_width() * (tc).y) + (tc).x)");
+			}
+			else
+			{
+				statement("#define spvImage2DAtomicCoord(tc, tex) (((((tex).get_width() + ",
+				          msl_options.r32ui_linear_texture_alignment / 4, ") & ~",
+				          msl_options.r32ui_linear_texture_alignment / 4, ") * (tc).y) + (tc).x)");
```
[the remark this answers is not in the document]
cdavis5e: Oops. Fixed.

HansKristian-Work:
> > r32ui_linear_texture_alignment should probably be emitted as a function constant instead, and the option should be the ID. It seems useful for applications that want to ship converted MSL to supply the alignment after converting MSL.
> 
> Why not both? Function constants don't work on MSL 1.1 anyway.

Yes, both is fine I guess. I guess when you're compiling to MSL in runtime you don't really need function constants either.

HansKristian-Work (approved):

LukasBanana:
> Just nits, quite curious indeed. I guess they only used 1D images perhaps. Ping @LukasBanana.

Sorry for the late reply. I can confirm the swapped `x` and `y` coordinates was a bug. We have it fixed in our P4V repository but not with the addition of variable alignment.

PR #1494, opened 2020-10-14
MSL: Don't use a bitcast for tessellation levels in tesc shaders.

cdavis5e:
`half` cannot be bitcasted to `float`, because the two types are not the
same size. Use an expanding cast instead.

### DIFF
--- a/spirv_msl.cpp
+++ b/spirv_msl.cpp
@@ -13299,7 +13299,7 @@ void CompilerMSL::remap_constexpr_sampler_by_binding(uint32_t desc_set, uint32_t
 	constexpr_samplers_by_binding[{ desc_set, binding }] = sampler;
 }
 
-void CompilerMSL::bitcast_from_builtin_load(uint32_t source_id, std::string &expr, const SPIRType &expr_type)
+void CompilerMSL::cast_from_builtin_load(uint32_t source_id, std::string &expr, const SPIRType &expr_type)
 {
 	auto *var = maybe_get_backing_variable(source_id);
 	if (var)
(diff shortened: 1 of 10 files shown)


spirv_msl.cpp
```diff
@@ -13344,7 +13344,17 @@ void CompilerMSL::bitcast_from_builtin_load(uint32_t source_id, std::string &exp
 	}
 
 	if (expected_type != expr_type.basetype)
-		expr = bitcast_expression(expr_type, expected_type, expr);
+	{
+		if (expected_type == SPIRType::Half)
```
HansKristian-Work, line 13348: The more appropriate thing to do here is to detect if there is a size mismatch and do value cast in that case. Should probably rename the function to reflect this.

HansKristian-Work: bitcast_to_builtin_store does an explicit check, can probably refactor that too while you're at it. Alternatively just do half -> float check.

HansKristian-Work (changes requested):
Minor nit.

spirv_msl.cpp
cdavis5e: Refactored both of them.

spirv_msl.cpp
```diff
@@ -13331,20 +13332,34 @@ void CompilerMSL::bitcast_from_builtin_load(uint32_t source_id, std::string &exp
 	case BuiltInBaseInstance:
 	case BuiltInBaseVertex:
 		expected_type = SPIRType::UInt;
+		expected_width = 4;
```
HansKristian-Work, line 13335: Expression width in in bits, not bytes.

cdavis5e: And yet, miraculously, it didn't produce any changes in the tests...

HansKristian-Work (approved):

spirv_msl.cpp
HansKristian-Work: Ye, I guess it's because bitcasts tend to be implemented as value casts when the bit width matches. Changing sign is the only real use case for the builtin cast ...

PR #1495, opened 2020-10-15
MSL: Handle Offset and Grad operands for 1D-as-2D textures.

cdavis5e:


cdavis5e:
GCC isn't smart enough to realize that `offset_type` will always be initialized whenever `offset_expr` is non-empty. This is why a lot of people don't like `-Wmaybe-uninitialized`...

HansKristian-Work (approved):
LGTM.

PR #1499, opened 2020-10-22
Subgroup fixes

cdavis5e:
Assorted fixes to subgroup support. This gets most of the subgroup tests passing for compute shaders. The remaining subgroup tests fail either because the graphics tests use isoline tessellation, or because of bugs in Metal.


spirv_msl.cpp
```diff
@@ -4610,6 +4610,89 @@ void CompilerMSL::emit_custom_functions()
...
+			statement("inline T spvSubgroupBroadcast(T value, ushort lane)");
+			begin_scope();
+			if (msl_options.is_ios())
+				statement("return quad_broadcast(value, lane);");
+			else
+				statement("return simd_broadcast(value, lane);");
+			end_scope();
+			statement("");
+			statement("template<>");
+			statement("inline bool spvSubgroupBroadcast(bool value, ushort lane)");
```
HansKristian-Work, line 4626: Can we make this more compact? Either specialized template on boolean types, or emitting code in a loop.

HansKristian-Work (changes requested):
Nice! Yes, I figured we'd need something like CTS to properly test subgroups ...

My only real concern is the intense code bloat to stamp out variants. Please try to find a way to make the MSL codegen less repetitive.

spirv_msl.cpp
cdavis5e: I didn't even think TP inference with a non-type parameter would work. But it worked pretty well, in fact.

cdavis5e:
I simplified the `SubgroupG[et]Mask` changes a bit with `insert_bits()`. This should make it a bit faster and easier to comprehend.

HansKristian-Work (approved):
LGTM.

PR #1501, opened 2020-10-23
added metal keyworld: "level"

atyuwen:
To avoid name conflict between a variable named `level` and the `level` constructor in Metal.


cdavis5e:
Should probably add `gradient2d`, `gradientcube`, `gradient3d`, and `min_lod_clamp`, too.

HansKristian-Work:
Need up update reference files as well, level changed to level0 in some of the shaders.

PR #1505, opened 2020-10-26
MSL: For 2.1+, don't disable rasterization for vertex writes.

cdavis5e:
Writing to buffers actually works starting in MSL 2.1 (macOS 10.14,
iOS 12). Writing to textures works starting in MSL 2.2 (macOS 10.15, iOS
13).

No tests unfortunately, because the MSL 2.2 compiler and above produce a
warning that cannot be disabled, because it has no associated option.


HansKristian-Work (approved):
LGTM.

PR #1506, opened 2020-10-26
Fixing HLSL compiler resource binding remap bug

jojon5:


HansKristian-Work:
Are you sure this is a bug? The check is supposed to be inverted.

HansKristian-Work (changes requested):
I don't think this change is correct. The check currently is essentially `if (autobinding == 0) { use_explicit_binding() } else { let_hlsl_compiler_decide() }`. It might be a bit confusing since the check is inverted, so a clarifying comment instead would be appropriate.

jojon5:
Not sure, I was hoping to get clarification on the bug I was seeing in my app. This change solved my problem. But I suppose it could possibly be my application incorrectly configuring the compiler object (or the bug is somewere else)

HansKristian-Work:
If you want to let the HLSL compiler decide on bindings for you, i.e., you reflect the resulting DXBC yourself, then you need to add the AUTO flags to the compiler first, otherwise, the bindings will come directly from the SPIR-V.

HansKristian-Work:
Closing this. If you have further issues on HLSL autobinding, you can open an issue.

PR #1509, opened 2020-10-28
MSL: Allow post-depth coverage on Mac in MSL 2.3.

cdavis5e:
It's still only supported on Apple GPUs, but Macs will have those soon.


HansKristian-Work (approved):

cdavis5e:
> Will merge when Travis passes.

What's up with Travis lately? It's been almost a full day since I pushed this, but Travis still hasn't started building this.

HansKristian-Work:
Yes, it's rather curious. Travis has been exceptionally slow lately. I'll just go ahead and merge, since this only affects macOS, is a small patch and I assume you ran test suite.

PR #1516, opened 2020-10-30
MSL: Support run-time sized image and sampler arrays

billhollings:
MSL: Support run-time sized image and sampler arrays
(GL_EXT_nonuniform_qualifier/SPV_EXT_descriptor_indexing).

MSL does not natively support unsized arrays.
MSLResourceBinding includes array size through API, and substitutes
in that size if the image or sampler array is not explicitly sized.

Also includes fix to support SPIRCombinedImageSampler type in OpCopyObject in MSL.

I couldn't come up with any simple automated tests to add for these changes.

### DIFF
--- a/spirv_glsl.cpp
+++ b/spirv_glsl.cpp
@@ -10020,10 +10020,20 @@ void CompilerGLSL::emit_instruction(const Instruction &instruction)
 		bool pointer = get<SPIRType>(result_type).pointer;
 
 		auto *chain = maybe_get<SPIRAccessChain>(rhs);
+		auto *imgsamp = maybe_get<SPIRCombinedImageSampler>(rhs);
 		if (chain)
 		{
 			// Cannot lower to a SPIRExpression, just copy the object.
 			auto &e = set<SPIRAccessChain>(id, *chain);
+			e.self = id;
+		}
+		else if (imgsamp)
+		{
+			// Cannot lower to a SPIRExpression, just copy the object.
+			// GLSL does not currently use this type and will never get here, but MSL does.
+			// Handled here instead of CopilerMSL for better integration and general handling,
+			// and in case GLSL or other subclasses require it in the future.
+			auto &e = set<SPIRCombinedImageSampler>(id, *imgsamp);
 			e.self = id;
 		}
 		else if (expression_is_lvalue(rhs) && !pointer)
--- a/spirv_glsl.cpp
+++ b/spirv_glsl.cpp
@@ -10031,7 +10031,7 @@ void CompilerGLSL::emit_instruction(const Instruction &instruction)
 		{
 			// Cannot lower to a SPIRExpression, just copy the object.
 			// GLSL does not currently use this type and will never get here, but MSL does.
-			// Handled here instead of CopilerMSL for better integration and general handling,
+			// Handled here instead of CompilerMSL for better integration and general handling,
 			// and in case GLSL or other subclasses require it in the future.
 			auto &e = set<SPIRCombinedImageSampler>(id, *imgsamp);
 			e.self = id;
--- a/spirv_msl.cpp
+++ b/spirv_msl.cpp
@@ -105,9 +105,8 @@ bool CompilerMSL::is_msl_resource_binding_used(ExecutionModel model, uint32_t de
 
 uint32_t CompilerMSL::get_resource_array_size(uint32_t id) const
 {
-	StageSetBinding tuple = { get_entry_point().model,
-							  get_decoration(id, DecorationDescriptorSet),
-							  get_decoration(id, DecorationBinding) };
+	StageSetBinding tuple = { get_entry_point().model, get_decoration(id, DecorationDescriptorSet),
+		                      get_decoration(id, DecorationBinding) };
 	auto itr = resource_bindings.find(tuple);
 	return itr != end(resource_bindings) ? itr->second.first.size : 0;
 }
--- a/spirv_msl.cpp
+++ b/spirv_msl.cpp
@@ -103,12 +103,14 @@ bool CompilerMSL::is_msl_resource_binding_used(ExecutionModel model, uint32_t de
 	return itr != end(resource_bindings) && itr->second.second;
 }
 
+// Returns the size of the array of resources used by the variable with the specified id.
+// The returned value is retrieved from the resource binding added using add_msl_resource_binding().
 uint32_t CompilerMSL::get_resource_array_size(uint32_t id) const
 {
 	StageSetBinding tuple = { get_entry_point().model, get_decoration(id, DecorationDescriptorSet),
 		                      get_decoration(id, DecorationBinding) };
 	auto itr = resource_bindings.find(tuple);
-	return itr != end(resource_bindings) ? itr->second.first.size : 0;
+	return itr != end(resource_bindings) ? itr->second.first.count : 0;
 }
 
 uint32_t CompilerMSL::get_automatic_msl_resource_binding(uint32_t id) const
--- a/spirv_msl.cpp
+++ b/spirv_msl.cpp
@@ -103,6 +103,15 @@ bool CompilerMSL::is_msl_resource_binding_used(ExecutionModel model, uint32_t de
 	return itr != end(resource_bindings) && itr->second.second;
 }
 
+uint32_t CompilerMSL::get_resource_array_size(uint32_t id) const
+{
+	StageSetBinding tuple = { get_entry_point().model,
+							  get_decoration(id, DecorationDescriptorSet),
+							  get_decoration(id, DecorationBinding) };
+	auto itr = resource_bindings.find(tuple);
+	return itr != end(resource_bindings) ? itr->second.first.size : 0;
+}
+
 uint32_t CompilerMSL::get_automatic_msl_resource_binding(uint32_t id) const
 {
 	return get_extended_decoration(id, SPIRVCrossDecorationResourceIndexPrimary);
@@ -8136,7 +8145,7 @@ void CompilerMSL::emit_function_prototype(SPIRFunction &func, const Bitset &)
 
 			// Manufacture automatic sampler arg for SampledImage texture
 			if (arg_type.image.dim != DimBuffer)
-				decl += join(", thread const ", sampler_type(arg_type), " ", to_sampler_expression(arg.id));
+				decl += join(", thread const ", sampler_type(arg_type, arg.id), " ", to_sampler_expression(arg.id));
 		}
 
 		// Manufacture automatic swizzle arg.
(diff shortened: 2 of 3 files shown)


spirv_msl.hpp
```diff
@@ -60,7 +60,10 @@ struct MSLShaderInput
 
 // Matches the binding index of a MSL resource for a binding within a descriptor set.
 // Taken together, the stage, desc_set and binding combine to form a reference to a resource
-// descriptor used in a particular shading stage.
+// descriptor used in a particular shading stage. The size element indicates the number of
+// resources consumed by this binding, if the binding represents an array of resources.
+// If the resource array is a run-time-sized array, which are legal in GLSL or SPIR-V, this value
+// will be used to declare the array size in MSL, which does not support run-time-sized arrays.
```
HansKristian-Work, line 66: How big can this array be? Vulkan min-spec here is 512k FWIW, and D3D12 is 1000000. Is it possible to use the trick of `[1]` array size for unsized arrays? Do you expect that we will need to emit the "true" unsized descriptor array in Metal, which is `const device texture2d<float> *textures` for example (Tier 2 IAB)?

spirv_msl.cpp
```diff
@@ -103,6 +103,15 @@ bool CompilerMSL::is_msl_resource_binding_used(ExecutionModel model, uint32_t de
 	return itr != end(resource_bindings) && itr->second.second;
 }
 
+uint32_t CompilerMSL::get_resource_array_size(uint32_t id) const
+{
+	StageSetBinding tuple = { get_entry_point().model,
+							  get_decoration(id, DecorationDescriptorSet),
```
HansKristian-Work, line 109: Formatting.

spirv_glsl.cpp
```diff
@@ -10020,12 +10020,22 @@ void CompilerGLSL::emit_instruction(const Instruction &instruction)
...
 		{
 			// Cannot lower to a SPIRExpression, just copy the object.
 			auto &e = set<SPIRAccessChain>(id, *chain);
 			e.self = id;
 		}
+		else if (imgsamp)
+		{
+			// Cannot lower to a SPIRExpression, just copy the object.
+			// GLSL does not currently use this type and will never get here, but MSL does.
+			// Handled here instead of CopilerMSL for better integration and general handling,
```
HansKristian-Work, line 10034: CopilerMSL typo

spirv_glsl.cpp
```diff
@@ -10020,12 +10020,22 @@ void CompilerGLSL::emit_instruction(const Instruction &instruction)
...
 
 		auto *chain = maybe_get<SPIRAccessChain>(rhs);
+		auto *imgsamp = maybe_get<SPIRCombinedImageSampler>(rhs);
 		if (chain)
 		{
 			// Cannot lower to a SPIRExpression, just copy the object.
 			auto &e = set<SPIRAccessChain>(id, *chain);
 			e.self = id;
 		}
+		else if (imgsamp)
```
HansKristian-Work, line 10030: When does this come up in practice when you say that MSL can hit it, but not GLSL?

HansKristian-Work (commented):
See comments, I'd like to make sure that will actually be sufficient for implementing descriptor indexing.

spirv_msl.hpp
cdavis5e: > Is it possible to use the trick of `[1]` array size for unsized arrays?

It is. I just tested it.

spirv_glsl.cpp
billhollings: One of the Vulkan CTS descriptor set indexing tests includes this SPIR-V:

         %31 = OpTypeSampledImage %10
         %32 = OpSampledImage %31 %20 %30
         %33 = OpCopyObject %31 %32


Without this conversion, the `OpCopyObject` processing code here falls into `to_expression()`, which then hits `case TypeCombinedImageSampler:` and throws an error/assertion.

The code comments related to that throw indicate that GLSL does not use `TypeCombinedImageSampler` because `OpSampledImage` converts it to an expression, but MSL certainly does use `TypeCombinedImageSampler`.

spirv_msl.hpp
billhollings: > How big can this array be?

MoltenVK doesn't support Argument Buffers yet...but Metal Argument Buffers can indeed support 500K textures on recent systems.

> > Is it possible to use the trick of `[1]` array size for unsized arrays?
> 
> It is. I just tested it.

Hmmm...when running CTS `dEQP-VK.descriptor_indexing.sampled_image`, the shader created for that test:

```
fragment main0_out main0(main0_in in [[stage_in]], array<texture2d<float>, 128> _13 [[texture(0)]], array<sampler, 1> _26 [[sampler(0)]])
{
    main0_out out = {};
    out.m_9 = _13[in.m_16].sample(_26[0], float2(0.0));
    return out;
}
```

works (the `128` comes from the binding layout in the test itself..but I think it only creates 30 or so actual textures).

But attempting to trick it with a length of `1`:

    fragment main0_out main0(main0_in in [[stage_in]], array<texture2d<float>, 1> _13 [[texture(0)]], array<sampler, 1> _26 [[sampler(0)]])

results in runtime errors:

```
2020-10-30 14:23:53.680356-0400 deqp-vk[95174:2468567] -[_MTLCommandBuffer didCompleteWithStartTime:endTime:error:], line 645: error 'Execution of the command buffer was aborted due to an error during execution. Internal Error (IOAF code 2067)'
2020-10-30 14:23:53.680453-0400 deqp-vk[95174:2468567] Execution of the command buffer was aborted due to an error during execution. Internal Error (IOAF code 2067)
```

_**and**_ messes with the responsiveness of my machine. It looks like there must be something like buffer overflow happening somewhere in the flow.

There are no Metal validation assertions, so it looks like Apple just isn't paying attention to this use case...which leads me to think that Apple likely _**will**_ put validation tests in when they realize that it causes serious issues. Either way, it doesn't look like it will work for us.

cdavis5e: Oh. I meant it works with argument buffers. Since `VK_EXT_descriptor_indexing` has a minimum of 500k `UpdateAfterBind` resources, argument buffers are quite necessary to support this properly.

billhollings: >Oh. I meant it works with argument buffers. 

Ah! You tested this outside of MoltenVK then?

>Since VK_EXT_descriptor_indexing has a minimum of 500k UpdateAfterBind resources, argument buffers are quite necessary to support this properly.

Definitely! That expansion was one of the primary motivations for `VK_EXT_descriptor_indexing`, and Argument Buffers need to be part of this.

But I think we will need to support both options for the near future. Argument Buffers are only a couple of years old, and Tier 1 (iOS before A13 and macOS Integrated GPU's) doesn't offer any larger numbers than discrete argument tables anyway.

I'm well into an initial `VK_EXT_descriptor_indexing` implementation without them. As soon as that's done, I'll go back and rework descriptor sets in general to use either argument buffers or not, depending on their availability.

With all this in mind...I think including `MSLResourceBinding::size` here is a relatively innocuous and generally useful item, and for the time being we can use it for discrete resources. If it doesn't need to be used when Argument Buffers are in play, then we can ignore it on those code paths, and just plug in a `1`.

Also...and BTW...this particular change needn't be inherently tied to descriptor indexing only. A GLSL shader that makes use of `GL_EXT_nonuniform_qualifier` can pass unsized arrays which make it all the way through SPIR-V to MSL.

spirv_msl.cpp
billhollings: Done.

spirv_glsl.cpp
billhollings: Done.

spirv_glsl.cpp
HansKristian-Work: I see.

spirv_msl.cpp
```diff
@@ -103,6 +103,14 @@ bool CompilerMSL::is_msl_resource_binding_used(ExecutionModel model, uint32_t de
 	return itr != end(resource_bindings) && itr->second.second;
 }
 
+uint32_t CompilerMSL::get_resource_array_size(uint32_t id) const
+{
+	StageSetBinding tuple = { get_entry_point().model, get_decoration(id, DecorationDescriptorSet),
+		                      get_decoration(id, DecorationBinding) };
```
HansKristian-Work, line 109: Formatting.

spirv_msl.hpp
```diff
@@ -517,6 +521,10 @@ class CompilerMSL : public CompilerGLSL
 	// by remap_constexpr_sampler(_by_binding).
 	bool is_msl_resource_binding_used(spv::ExecutionModel model, uint32_t set, uint32_t binding) const;
 
+	// Returns the size of the array of resources used by the variable with the specified id.
+	// The returned value is retrieved from the resource binding added using add_msl_resource_binding().
+	uint32_t get_resource_array_size(uint32_t id) const;
```
HansKristian-Work, line 526: Why should this be public?

HansKristian-Work (changes requested):
Minor nit comments, otherwise, seems fine.

spirv_msl.hpp
```diff
@@ -71,6 +74,7 @@ struct MSLResourceBinding
 	spv::ExecutionModel stage = spv::ExecutionModelMax;
 	uint32_t desc_set = 0;
 	uint32_t binding = 0;
+	uint32_t size = 0;
```
HansKristian-Work, line 77: This should have a comment saying what it's for (resolving unsized arrays), so that users don't feel compelled to specify size everywhere. Could perhaps be called something else that would clarify intent as well. size would mean bit width as well for example ...

spirv_msl.hpp
cdavis5e: > > Oh. I meant it works with argument buffers.
> 
> Ah! You tested this outside of MoltenVK then?

Yes.

> With all this in mind...I think including `MSLResourceBinding::size` here is a relatively innocuous and generally useful item, and for the time being we can use it for discrete resources. If it doesn't need to be used when Argument Buffers are in play, then we can ignore it on those code paths, and just plug in a `1`.
> 
> Also...and BTW...this particular change needn't be inherently tied to descriptor indexing only. A GLSL shader that makes use of `GL_EXT_nonuniform_qualifier` can pass unsized arrays which make it all the way through SPIR-V to MSL.

Very well, then.

spirv_msl.cpp
billhollings: Hmmm...this is what `clang-format -style=file -i spirv_msl.cpp` produces on my Mac.

spirv_msl.hpp
billhollings: Moved it to protected.

spirv_msl.hpp
billhollings: There was already an explanation of the field in the comments. I've added an additional line clarifying that the field is not required for resources that are not held in a run-time-sized array in the shader.

I used `size` because it represents an array size. I've changed the name to `count`, which aligns with the corresponding name `descriptorCount` in Vulkan. I don't want to explicitly reference run-time-sizing in the name because having the resource count available may find other general uses later on.

HansKristian-Work (approved):

PR #1520, opened 2020-10-31
HLSL: Support depth comparison texture sampling in SM 2/3.

rdb:
In DX9, all samples of a depth texture result in a depth comparison (in fact, disabling this requires use of the special `INTZ` format, only available on DX10+ hardware).  This works with all 4-component texture sample functions (tex2Dlod, tex2Dbias, tex2Dproj).

This PR provides access to this functionality by enabling use of the Dref texture sample instruction and passing in the depth compare value into the appropriate register.  It assumes that there is indeed a depth texture bound to the appropriate register (which is the application's responsibility, and not something we can check).

Even though the regular `texld` instruction does support this as well, HLSL does not (to my knowledge) allow access to this variant since `tex2D` only accepts 2 coordinates.  To work around this, this PR turns an OpImageSampleDrefImplicitLod into a `tex2Dproj` with a w coordinate of 1.0, which is effectively equivalent.

Thanks for your consideration!

### DIFF
--- a/reference/shaders-hlsl/frag/tex-sampling.sm30.frag
+++ b/reference/shaders-hlsl/frag/tex-sampling.sm30.frag
@@ -1,0 +1,83 @@
+uniform sampler1D tex1d;
+uniform sampler2D tex2d;
+uniform sampler3D tex3d;
+uniform samplerCUBE texCube;
+uniform sampler1D tex1dShadow;
+uniform sampler2D tex2dShadow;
+
+static float texCoord1d;
+static float2 texCoord2d;
+static float3 texCoord3d;
+static float4 FragColor;
+static float4 texCoord4d;
+
+struct SPIRV_Cross_Input
+{
+    float texCoord1d : TEXCOORD0;
+    float2 texCoord2d : TEXCOORD1;
+    float3 texCoord3d : TEXCOORD2;
+    float4 texCoord4d : TEXCOORD3;
+};
+
+struct SPIRV_Cross_Output
+{
+    float4 FragColor : COLOR0;
+};
+
+void frag_main()
+{
+    float4 texcolor = tex1D(tex1d, texCoord1d);
+    texcolor += tex1Dlod(tex1d, float4(texCoord1d, 0.0, 0.0, 2.0f));
+    texcolor += tex1Dgrad(tex1d, texCoord1d, 1.0f, 2.0f);
+    float2 _34 = float2(texCoord1d, 2.0f);
+    texcolor += tex1Dproj(tex1d, float4(_34.x, 0.0, 0.0, _34.y));
+    texcolor += tex1Dbias(tex1d, float4(texCoord1d, 0.0, 0.0, 1.0f));
+    texcolor += tex2D(tex2d, texCoord2d);
+    texcolor += tex2Dlod(tex2d, float4(texCoord2d, 0.0, 2.0f));
+    texcolor += tex2Dgrad(tex2d, texCoord2d, float2(1.0f, 2.0f), float2(3.0f, 4.0f));
+    float3 _73 = float3(texCoord2d, 2.0f);
+    texcolor += tex2Dproj(tex2d, float4(_73.xy, 0.0, _73.z));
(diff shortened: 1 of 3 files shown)


HansKristian-Work (changes requested):
I'd like to see a test here to see what is going on.

rdb:
I added a test for all the samplers supported by SM 3.0.

As you can see it's just a matter of passing the dref into the third coordinate.  I verified that the resulting shader works correctly for shadow mapping on both Intel and NVIDIA hardware, and furthermore, this is also what the Cg compiler is producing when using a shadow sampler and targeting SM 3.0 (aside from the fact that it can directly generate a `texld` instead of `texldp`, but that has no effect on the result).

Thanks!

reference/shaders-hlsl/frag/tex-sampling.sm30.frag
```diff
@@ -0,0 +1,87 @@
...
+    float _180 = _179.z;
+    texcolor.w += tex1Dbias(tex1dShadow, float4(_179.x, _180, _180, 1.0f));
+    float3 _194 = float3(texCoord2d, 0.0f);
+    texcolor.w += tex2Dproj(tex2dShadow, float4(_194.xy, _194.z, 1.0));
+    float3 _205 = float3(texCoord2d, 0.0f);
+    texcolor.w += tex2Dlod(tex2dShadow, float4(_205.xy, _205.z, 2.0f));
+    float4 _216 = float4(texCoord2d, 0.0f, 2.0f);
+    float4 _219 = _216;
+    _219.z = _216.w;
+    texcolor.w += tex2Dproj(tex2dShadow, float4(_219.xy, _216.z, _219.z));
```
HansKristian-Work, line 71: This looks weird. What is the return type of tex2Dproj here? I would guess float4, in which case, we need to emit a `.x` swizzle after the expression since Dref is expected to be scalar in SPIR-V.

spirv_hlsl.cpp
```diff
@@ -2994,35 +2993,45 @@ void CompilerHLSL::emit_texture_op(const Instruction &i, bool sparse)
...
+
+			if (grad_x || grad_y)
+				SPIRV_CROSS_THROW("Depth comparison is not supported for grad sampling in HLSL shader model 2/3.");
+
+			forward = forward && should_forward(dref);
 			for (uint32_t size = coord_components; size < 3; ++size)
-				coord_filler += ", 0.0";
-			coord_expr = "float4(" + coord_expr + coord_filler + ", " + to_expression(lod) + ")";
-			modifier_count++;
+				coord_expr += ", " + to_expression(dref);
```
HansKristian-Work, line 3006: Should dref be duplicated twice out to 3 components when using tex1D?

HansKristian-Work: I guess it doesn't hurt though, but ...

reference/shaders-hlsl/frag/tex-sampling.sm30.frag
rdb: Good catch, I'll change that.

HansKristian-Work (changes requested):
Also missing `reference/opt/shaders-hlsl/frag/tex-sampling.sm30.frag` to be committed.

spirv_hlsl.cpp
rdb: Not necessary, just matching the behaviour of the NV Cg compiler, which put the depth in both y and z.  My guess was that it might have been one less instruction to just swizzle than having to add in a 0?  No idea.  Should I change it?

HansKristian-Work: I would normally expect the second coordinate to be 0 for 1D, but if Cg does it that way I don't mind it being there as-is. Just add a comment I guess to not confuse future readers.

rdb:
I just went ahead and just changed the second coordinate to 0.0, because it's clearer.  I verified that the resulting shader still works correctly with a 1D depth map.

I also added a return swizzle and added the optimized reference file as requested.

HansKristian-Work (approved):
LGTM, thanks.

PR #1521, opened 2020-10-31
Fix some bad assumptions about emulating `subgroupBarrier`

devshgraphicsprogramming:
It turns out a shared memory barrier is needed even when threads are guaranteed to be in lockstep.

I have some code that breaks in OpenGL on a Nvidia 1050 Mobile, if a `memoryBarrierShared` is not present (even with the NV_shader_thread_group extension enabled that guarantees lockstep execution within a warp).


devshgraphicsprogramming:
@Crisspl could you update the test suite reference shaders on Monday?

HansKristian-Work (approved):

PR #1525, opened 2020-11-03
MSL: Support pull-model interpolation on MSL 2.3+.

cdavis5e:
New in MSL 2.3 is a template that can be used in the place of a scalar
type in a stage-in struct. This template has methods which interpolate
the varying at the given points. Curiously, you can't set interpolation
attributes on such a varying; perspective-correctness is encoded in the
type, while interpolation must be done using one of the methods. This
makes using this somewhat awkward from SPIRV-Cross, requiring us to jump
through a bunch of hoops to make this all work.

Using varyings from functions in particular is a pain point, requiring
us to pass the stage-in struct itself around. An alternative is to pass
references to the interpolants; except this will fall over badly with
composite types, which naturally must be flattened.  As with
tessellation, dynamic indexing isn't supported with pull-model
interpolation. This is because of the need to reference the original
struct member in order to call one of the pull-model interpolation
methods on it. Also, this is done at the variable level; this means that
if one varying in a struct is used with the pull-model functions, then
the entire struct is emitted as pull-model interpolants.

For some reason, this was not documented in the MSL spec, though there
is a property on `MTLDevice`, `supportsPullModelInterpolation`,
indicating support for this, which *is* documented. This does not appear
to be implemented yet for AMD: it returns `NO` from
`supportsPullModelInterpolation`, and pipelines with shaders using the
templates fail to compile. It *is* implemeted for Intel. It's probably
also implemented for Apple GPUs: on Apple Silicon, OpenGL calls down to
Metal, and it wouldn't be possible to use the interpolation functions
without this implemented in Metal.

Based on my testing, where SPIR-V and GLSL have the offset relative to
the pixel center, in Metal it appears to be relative to the pixel's
upper-left corner, as in HLSL. Therefore, I've added an offset 0.4375,
i.e. one half minus one sixteenth, to all arguments to
`interpolate_at_offset()`.

This also fixes a long-standing bug: if a pull-model interpolation
function is used on a varying, make sure that varying is declared. We
were already doing this only for the AMD pull-model function,
`interpolateAtVertexAMD()`; for reasons which are completely beyond me,
we weren't doing this for the base interpolation functions. I also note
that there are no tests for the interpolation functions for GLSL or
HLSL.

### DIFF
--- a/spirv_cross.cpp
+++ b/spirv_cross.cpp
@@ -699,8 +699,31 @@ bool Compiler::InterfaceVariableAccessHandler::handle(Op opcode, const uint32_t 
 	{
 		if (length < 5)
 			return false;
-		uint32_t extension_set = args[2];
-		if (compiler.get<SPIRExtension>(extension_set).ext == SPIRExtension::SPV_AMD_shader_explicit_vertex_parameter)
+		auto &extension_set = compiler.get<SPIRExtension>(args[2]);
+		switch (extension_set.ext)
+		{
+		case SPIRExtension::GLSL:
+		{
+			auto op = static_cast<GLSLstd450>(args[3]);
+
+			switch (op)
+			{
+			case GLSLstd450InterpolateAtCentroid:
+			case GLSLstd450InterpolateAtSample:
+			case GLSLstd450InterpolateAtOffset:
+			{
+				auto *var = compiler.maybe_get<SPIRVariable>(args[4]);
+				if (var && storage_class_is_interface(var->storage))
+					variables.insert(args[4]);
+				break;
+			}
+
+			default:
+				break;
+			}
+			break;
+		}
+		case SPIRExtension::SPV_AMD_shader_explicit_vertex_parameter:
 		{
 			enum AMDShaderExplicitVertexParameter
 			{
@@ -722,6 +745,10 @@ bool Compiler::InterfaceVariableAccessHandler::handle(Op opcode, const uint32_t 
 			default:
 				break;
 			}
+			break;
+		}
+		default:
+			break;
 		}
 		break;
 	}
--- a/spirv_msl.cpp
+++ b/spirv_msl.cpp
@@ -169,7 +169,7 @@ void CompilerMSL::build_implicit_builtins()
 	     active_input_builtins.get(BuiltInInstanceIndex) || active_input_builtins.get(BuiltInBaseInstance));
 	bool need_sample_mask = msl_options.additional_fixed_sample_mask != 0xffffffff;
 	if (need_subpass_input || need_sample_pos || need_subgroup_mask || need_vertex_params || need_tesc_params ||
-	    need_multiview || need_dispatch_base || need_vertex_base_params || need_grid_params ||
+	    need_multiview || need_dispatch_base || need_vertex_base_params || need_grid_params || needs_sample_id ||
 	    needs_subgroup_invocation_id || needs_subgroup_size || need_sample_mask)
 	{
 		bool has_frag_coord = false;
@@ -235,7 +235,7 @@ void CompilerMSL::build_implicit_builtins()
 				}
 			}
 
-			if (need_sample_pos && builtin == BuiltInSampleId)
+			if ((need_sample_pos || needs_sample_id) && builtin == BuiltInSampleId)
 			{
 				builtin_sample_id_id = var.self;
 				mark_implicit_builtin(StorageClassInput, BuiltInSampleId, var.self);
@@ -414,7 +414,7 @@ void CompilerMSL::build_implicit_builtins()
 			}
 		}
 
-		if (!has_sample_id && need_sample_pos)
+		if (!has_sample_id && (need_sample_pos || needs_sample_id))
 		{
 			uint32_t offset = ir.increase_bound_by(2);
 			uint32_t type_ptr_id = offset;
@@ -1293,6 +1293,8 @@ void CompilerMSL::preprocess_op_codes()
 		needs_subgroup_invocation_id = true;
 	if (preproc.needs_subgroup_size)
 		needs_subgroup_size = true;
+	if (preproc.needs_sample_id)
+		needs_sample_id = true;
 }
 
 // Move the Private and Workgroup global variables to the entry function.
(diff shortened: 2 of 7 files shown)


spirv_msl.cpp
```diff
@@ -1464,6 +1466,27 @@ void CompilerMSL::extract_global_variables_from_function(uint32_t func_id, std::
...
+				{
+					auto op_450 = static_cast<GLSLstd450>(ops[3]);
+					switch (op_450)
+					{
+					case GLSLstd450InterpolateAtCentroid:
+					case GLSLstd450InterpolateAtSample:
+					case GLSLstd450InterpolateAtOffset:
+					{
+						// For these, we really need the stage-in block.
+						added_arg_ids.insert(stage_in_var_id);
```
HansKristian-Work, line 1482: Fallthrough. Benign, but looks weird.

HansKristian-Work: Also, why stage_in_var_id and not the variable it touches itself?

spirv_cross.cpp
```diff
@@ -699,8 +699,30 @@ bool Compiler::InterfaceVariableAccessHandler::handle(Op opcode, const uint32_t
...
+			{
+				auto *var = compiler.maybe_get<SPIRVariable>(args[4]);
+				if (var && storage_class_is_interface(var->storage))
+					variables.insert(args[4]);
+				break;
+			}
+
+			default:
+				break;
+			}
```
HansKristian-Work, line 723: Fallthrough.

spirv_msl.cpp
```diff
@@ -7903,10 +8080,67 @@ void CompilerMSL::emit_glsl_op(uint32_t result_type, uint32_t id, uint32_t eop,
...
+		// interpolate method, or refers to a local variable. We saved the interface index we need; use it to construct
+		// the base for the method call.
+		uint32_t interface_index = get_extended_decoration(args[0], SPIRVCrossDecorationInterfaceMemberIndex);
+		string component;
+		if (has_extended_decoration(args[0], SPIRVCrossDecorationInterpolantComponentExpr))
+		{
+			uint32_t index_expr = get_extended_decoration(args[0], SPIRVCrossDecorationInterpolantComponentExpr);
+			auto *c = maybe_get<SPIRConstant>(index_expr);
+			if (!c || c->specialization)
+				component = join("[", to_expression(index_expr), "]");
```
HansKristian-Work, line 8095: This is questionable. A literal could be interpreted as a SPIRConstant by mistake if that index happened to have a SPIRConstant in it. Perhaps tag the ID with 0x80000000u MSB set to mark literal vs SPIRConstant.

spirv_msl.cpp
```diff
@@ -13310,6 +13550,55 @@ bool CompilerMSL::OpCodePreprocessor::handle(Op opcode, const uint32_t *args, ui
...
+			{
+			case GLSLstd450InterpolateAtCentroid:
+			case GLSLstd450InterpolateAtSample:
+			case GLSLstd450InterpolateAtOffset:
+			{
+				if (!compiler.msl_options.supports_msl_version(2, 3))
+					SPIRV_CROSS_THROW("Pull-model interpolation requires MSL 2.3.");
+				// Fragment varyings used with pull-model interpolation need special handling,
+				// due to the way pull-model interpolation works in Metal.
+				SPIRVariable* var = compiler.maybe_get_backing_variable(args[4]);
```
HansKristian-Work, line 13569: `*` placement, also, use auto here.

spirv_msl.cpp
```diff
@@ -7903,10 +8080,67 @@ void CompilerMSL::emit_glsl_op(uint32_t result_type, uint32_t id, uint32_t eop,
...
+			uint32_t index_expr = get_extended_decoration(args[0], SPIRVCrossDecorationInterpolantComponentExpr);
+			auto *c = maybe_get<SPIRConstant>(index_expr);
+			if (!c || c->specialization)
+				component = join("[", to_expression(index_expr), "]");
+			else
+				component = join(".", index_to_swizzle(c->scalar()));
+		}
+		emit_op(result_type, id,
+		        join(to_name(stage_in_var_id), ".", to_member_name(get_stage_in_struct_type(), interface_index),
+		             ".interpolate_at_offset(", to_expression(args[1]), " + 0.4375)", component),
```
HansKristian-Work, line 8140: 0.4375 is questionable. Why not 0.5?

spirv_msl.cpp
```diff
@@ -6549,6 +6673,58 @@ void CompilerMSL::prepare_access_chain_for_scalar_access(std::string &expr, cons
...
+{
+	auto *var = maybe_get_backing_variable(ops[2]);
+	if (!var || !pull_model_inputs.count(var->self))
+		return;
+	// Get the base index.
+	uint32_t interface_index;
+	auto &var_type = get_variable_data_type(*var);
+	auto &result_type = get<SPIRType>(ops[0]);
+	auto *type = &var_type;
+	if (var_type.basetype == SPIRType::Struct)
```
HansKristian-Work, line 6687: I think this will break if apps do partial access chains. E.g. OpAccessChain a little bit, and then continue the access chain based on the first one.

spirv_msl.cpp
```diff
@@ -7144,6 +7320,7 @@ void CompilerMSL::emit_instruction(const Instruction &instruction)
 		}
 		else
 			CompilerGLSL::emit_instruction(instruction);
+		emit_interpolant_access_chain(ops, instruction.length);
```
HansKristian-Work, line 7323: Sounds like this should be called fixup interpolant access chain or something like that.

spirv_msl.cpp
HansKristian-Work: I see the argument about with 0.5 - 1/16, but that doesn't make much sense to me.

HansKristian-Work (changes requested):
Did a first pass, see comments.

spirv_msl.cpp
cdavis5e: Because if I don't, several tests in the CTS fail on Intel. I thought that was weird, too, but that's what my testing shows.

spirv_msl.cpp
cdavis5e: > Also, why stage_in_var_id and not the variable it pulls from itself?

I considered that, but supporting composites with that seemed really complicated.

spirv_msl.cpp
cdavis5e: This should never be a literal. No `OpAccessChain` instructions in SPIR-V use literals; only the artificial access chains we generate ourselves do, and I don't think we'll wind up calling this on them.

spirv_msl.cpp
cdavis5e: Not anymore, it shouldn't.

spirv_msl.cpp
cdavis5e: > > Also, why stage_in_var_id and not the variable it pulls from itself?
> 
> I considered that, but supporting composites with that seemed really complicated.

I've added a comment explaining why I did it this way.

spirv_msl.cpp
cdavis5e: I have added a comment here explaining all this.

spirv_msl.cpp
HansKristian-Work: Ok, in that case it is misleading to check for `!c` and `maybe_get`.

HansKristian-Work:
I'll get back to reviewing this on Monday.

spirv_msl.cpp
cdavis5e: But wait, what if there's an expression, but it's not a `SPIRConstant`? We have other checks like this, e.g. https://github.com/KhronosGroup/SPIRV-Cross/blob/aca9b6879af8af1704e0f54ae31b4118b11d790b/spirv_msl.cpp#L6482

HansKristian-Work: I guess ... If it's a dynamic index, then yes.

HansKristian-Work (approved):
Seems fine, don't have any obvious issues with the PR. If there are subtle issues, they can be fixed as needed later.

PR #1526, opened 2020-11-05
MSL: Emit gather component unless it's x.

js6i:
With #1504 texture gathers do not have their component emitted even when they're not depth textures. I added a test for that and changed the condition to emit the component iff it's not x, which is the default.

### DIFF
--- a/spirv_msl.cpp
+++ b/spirv_msl.cpp
@@ -8804,9 +8804,8 @@ string CompilerMSL::to_function_args(const TextureFunctionArguments &args, bool 
 		{
 			forward = forward && should_forward(args.component);
 
-			if (const auto *var = maybe_get_backing_variable(img))
-				if (!image_is_comparison(get<SPIRType>(var->basetype), var->self))
-					farg_str += ", " + to_component_argument(args.component);
+			if (evaluate_constant_u32(args.component))
+				farg_str += ", " + to_component_argument(args.component);
 		}
 	}
 
--- a/reference/shaders-msl-no-opt/frag/image-gather.asm.frag
+++ b/reference/shaders-msl-no-opt/frag/image-gather.asm.frag
@@ -1,0 +1,22 @@
+#include <metal_stdlib>
+#include <simd/simd.h>
+
+using namespace metal;
+
+struct main0_out
+{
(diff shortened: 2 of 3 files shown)


spirv_msl.cpp
```diff
@@ -8804,9 +8804,8 @@ string CompilerMSL::to_function_args(const TextureFunctionArguments &args, bool
 		{
 			forward = forward && should_forward(args.component);
 
-			if (const auto *var = maybe_get_backing_variable(img))
-				if (!image_is_comparison(get<SPIRType>(var->basetype), var->self))
-					farg_str += ", " + to_component_argument(args.component);
+			if (evaluate_constant_u32(args.component))
```
HansKristian-Work, line 8807: This seems flawed if the application somehow uses gather y/z/w on a depth2d. Right now we would just end up using X (since there's nothing else we can do), but this would break compilation.

js6i: Ah, I thought that's not a thing, do you mean a case where a gather-comparison is used on a multichannel texture? Would `image_is_comparison` check work?

HansKristian-Work: A texture2D might be used in a depth-compare context for certain scenarios, which forces depth2d type in Metal, but if applications then do textureGather(..., 1), this would break compilation since gather doesn't take component with depth2d in the first place.

HansKristian-Work: > With #1504 texture gathers do not have their component emitted even when they're not depth textures. I added a test for that and changed the condition to emit the component iff it's not x, which is the default.

I don't understand this. Are you saying there is a regression introduced by that PR?

js6i: Yes, it fails the test I added here.

HansKristian-Work: Ah, ok. I see now. I introduced a regression ... I have an alternate fix. Would you be okay if I reused your test and committed it under a different PR?

js6i: Sure, go ahead.

HansKristian-Work:
Other fix is merged, closing this, thanks for finding the issue.

PR #1527, opened 2020-11-05
GLSL: implement transpose() in GLSL 1.10 / ES 1.00

rdb:
GLSL 1.10 and ESSL 1.00 do not have `transpose()`.  However, it can be easily emulated—in most cases by flipping the `need_transpose` flag (which may just result in flipping multiplication order somewhere), or otherwise using a helper function.  It is made simpler by the fact that these versions only support square matrices, which leaves only three types to support.

I've followed the example of the HLSL inverse implementation for the helper function.

I do wonder if we should just handle OpTranspose this way for the other versions as well, as an optimisation, but that is outside the scope of this PR.  There are also other operations that are unnecessarily resolving the `need_transpose` (such as matrix-scalar multiplication, which could be performed without transposing the matrix).

### DIFF
--- a/spirv_glsl.cpp
+++ b/spirv_glsl.cpp
@@ -3814,6 +3814,33 @@ void CompilerGLSL::emit_extension_workarounds(spv::ExecutionModel model)
 		}
 		statement("");
 	}
+
+	if (requires_transpose_2x2)
+	{
+		statement("mat2 SPIRV_Cross_Transpose(mat2 m)");
+		begin_scope();
+		statement("return mat2(m[0][0], m[1][0], m[0][1], m[1][1]);");
+		end_scope();
+		statement("");
+	}
+
+	if (requires_transpose_3x3)
+	{
+		statement("mat3 SPIRV_Cross_Transpose(mat3 m)");
+		begin_scope();
+		statement("return mat3(m[0][0], m[1][0], m[2][0], m[0][1], m[1][1], m[2][1], m[0][2], m[1][2], m[2][2]);");
+		end_scope();
+		statement("");
+	}
+
+	if (requires_transpose_4x4)
+	{
+		statement("mat4 SPIRV_Cross_Transpose(mat4 m)");
+		begin_scope();
+		statement("return mat4(m[0][0], m[1][0], m[2][0], m[3][0], m[0][1], m[1][1], m[2][1], m[3][1], m[0][2], m[1][2], m[2][2], m[3][2], m[0][3], m[1][3], m[2][3], m[3][3]);");
+		end_scope();
+		statement("");
+	}
 }
 
 // Returns a string representation of the ID, usable as a function arg.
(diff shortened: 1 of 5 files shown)


spirv_glsl.cpp
```diff
@@ -3395,6 +3395,33 @@ void CompilerGLSL::emit_resources()
 		statement("");
 
 	declare_undefined_values();
+
+	if (requires_transpose_2x2)
```
HansKristian-Work, line 3399: This is a very strange to place this. Do this in emit_extension_workarounds perhaps. That's where most "wrapper" code is.

rdb: I put it here because the HLSL implementations writes the helper functions in the exact same location.  I can move it though.

spirv_glsl.cpp
```diff
@@ -12140,6 +12188,36 @@ string CompilerGLSL::convert_row_major_matrix(string exp_str, const SPIRType &ex
...
+		}
+		else if (exp_type.vecsize == 4 && exp_type.columns == 4)
+		{
+			if (!requires_transpose_4x4)
+			{
+				requires_transpose_4x4 = true;
+				force_recompile();
+			}
+		}
+		return join("SPIRV_Cross_Transpose(", exp_str, ")");
```
HansKristian-Work, line 12219: If this assumption does not hold, need SPIRV_CROSS_THROW.

spirv_glsl.cpp
```diff
@@ -10265,7 +10292,28 @@ void CompilerGLSL::emit_instruction(const Instruction &instruction)
 		break;
 
 	case OpTranspose:
-		GLSL_UFOP(transpose);
+		if (options.version < 120) // Matches GLSL 1.10 / ESSL 1.00
```
HansKristian-Work, line 10295: Isn't this the same as `is_legacy()`?

rdb: Nope, because `is_legacy()` includes GLSL 1.20, but that does have `transpose()`

spirv_glsl.cpp
```diff
@@ -10265,7 +10292,28 @@ void CompilerGLSL::emit_instruction(const Instruction &instruction)
...
+		{
+			// transpose() is not available, so instead, flip need_transpose,
+			// which can later be turned into an emulated transpose op by
+			// convert_row_major_matrix(), if necessary.
+			uint32_t result_type = ops[0];
+			uint32_t result_id = ops[1];
+			uint32_t input = ops[2];
+
+			auto &input_e = get<SPIRExpression>(input);
```
HansKristian-Work, line 10304: `get<SPIRExpression>` is always iffy, use `maybe_get` and assume that transpose state is false, if it doesn't exist.

spirv_glsl.cpp
```diff
@@ -10265,7 +10292,28 @@ void CompilerGLSL::emit_instruction(const Instruction &instruction)
...
+			bool need_transpose = input_e.need_transpose;
+			input_e.need_transpose = false;
+
+			auto &e = emit_op(result_type, result_id,
+			                  to_unpacked_row_major_matrix_expression(input),
```
HansKristian-Work, line 10309: After you have forced no transpose, just use to_expression here, to_unpacked_row_major_matrix_expression is more of a MSL-workaround-ism.

HansKristian-Work (changes requested):
See comments.

HansKristian-Work (approved):
LGTM

PR #1528, opened 2020-11-05
GLSL: Fix support for textureLod in legacy vertex shaders

rdb:
The change in #148 was made with the faulty assumption that textureLod was not supported in legacy GLSL vertex shaders.  
However, LOD sampling in the vertex shader has actually been available in both GLSL and ESSL from the beginning.  See section 8.7 of the [GLSL 1.10 spec](https://www.khronos.org/registry/OpenGL/specs/gl/GLSLangSpec.1.10.pdf) and [ESSL 1.00 spec](https://www.khronos.org/files/opengles_shading_language.pdf).

The [GL_ARB_shader_texture_lod](https://www.khronos.org/registry/OpenGL/extensions/ARB/ARB_shader_texture_lod.txt) and [GL_EXT_shader_texture_lod](https://www.khronos.org/registry/OpenGL/extensions/EXT/EXT_shader_texture_lod.txt) extensions extend the use of these existing functions to fragment shaders, and introduce the Grad variants to both vertex and fragment shaders.

This means that the fallback that has been implemented for vertex shaders (which fell back to the regular texture functions, requiring that lod=0) is entirely unnecessary.  I have removed it and instated the proper use of the texture2DLod, etc. functions that are available in legacy GLSL.

### DIFF
--- a/spirv_glsl.cpp
+++ b/spirv_glsl.cpp
@@ -5641,25 +5641,6 @@ void CompilerGLSL::emit_bitfield_insert_op(uint32_t result_type, uint32_t result
 	inherit_expression_dependencies(result_id, op3);
 }
 
-// EXT_shader_texture_lod only concerns fragment shaders so lod tex functions
-// are not allowed in ES 2 vertex shaders. But SPIR-V only supports lod tex
-// functions in vertex shaders so we revert those back to plain calls when
-// the lod is a constant value of zero.
-bool CompilerGLSL::check_explicit_lod_allowed(uint32_t lod)
-{
-	auto &execution = get_entry_point();
-	bool allowed = !is_legacy_es() || execution.model == ExecutionModelFragment;
-	if (!allowed && lod != 0)
-	{
-		auto *lod_constant = maybe_get<SPIRConstant>(lod);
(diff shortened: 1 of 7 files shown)


spirv_glsl.cpp
```diff
@@ -5691,17 +5672,22 @@ string CompilerGLSL::legacy_tex_op(const std::string &op, const SPIRType &imgtyp
...
 
-	bool use_explicit_lod = check_explicit_lod_allowed(lod);
-
-	if (op == "textureLod" || op == "textureProjLod" || op == "textureGrad" || op == "textureProjGrad")
+	// In legacy GLSL, an extension is required for textureLod in the fragment
+	// shader or textureGrad anywhere.
+	bool legacy_lod_ext = false;
+	auto &execution = get_entry_point();
+	if (op == "textureGrad" || op == "textureProjGrad" ||
+		((op == "textureLod" || op == "textureProjLod") && execution.model != ExecutionModelVertex))
```
HansKristian-Work, line 5680: Formatting, `((op == "textureLod` should align with `op == "textureGrad` using spaces.

HansKristian-Work (changes requested):
Seems okay, just a minor formatting nit.

HansKristian-Work (approved):

PR #1530, opened 2020-11-06
GLSL: Provide round/roundEven for legacy GLSL

rdb:
Despite some drivers supporting it, `round()` isn't mentioned in any of the legacy GLSL versions' specs.  The SPIR-V spec says that the rounding direction is implementation-defined, so it should be safe to fall back to `floor(x + 0.5)`.

Also, the `round()` function given by [GL_EXT_gpu_shader4](https://www.khronos.org/registry/OpenGL/extensions/EXT/EXT_gpu_shader4.txt) guarantees round-to-even semantics: 
>     Returns a value equal to the closest integer to x. If the fractional
>     portion of the operand is 0.5, the nearest even integer is returned. For
>     example, round (1.0) returns 1.0.  round(-1.5) returns -2.0. round(3.5)
>     and round (4.5) both return 4.0.
So, in legacy desktop GLSL, I've added a fallback to that extension for roundEven.  (I would've added a test, but the test harness doesn't currently support testing legacy desktop GLSL shaders—it always forces them to be ES 100.)

### DIFF
--- a/reference/opt/shaders/frag/round-even.frag
+++ b/reference/opt/shaders/frag/round-even.frag
@@ -1,0 +1,12 @@
+#version 450
+
+layout(location = 0) out vec4 FragColor;
+layout(location = 0) in vec4 vA;
+layout(location = 1) in float vB;
+
+void main()
+{
+    FragColor = roundEven(vA);
+    FragColor *= roundEven(vB);
+}
+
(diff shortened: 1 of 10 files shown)


HansKristian-Work (approved):
LGTM.

PR #1531, opened 2020-11-06
GLSL: Add error checking and extension fallback for textureGather

rdb:
Adds error checking and fallback to [GL_ARB_texture_gather](https://www.khronos.org/registry/OpenGL/extensions/ARB/ARB_texture_gather.txt) for use of textureGather().

The version provided by the extension is quite limited, only being able to return the first component and supporting no depth comparison, though it does have the Offset variant.  (More overloads are offered by GL_ARB_gpu_shader5, but no non-GL4 hardware supports that, so there's just no point.)

Note that the extension always uses the new-style syntax, even in legacy GLSL:
>     (2) What naming convention do we use for the new texture built-in
>         functions?
>
>        RESOLVED: Even though this extension is written against OpenGL
>        Shading Language 1.20, we use the new texture function naming
>        convention from OpenGL Shading Language 1.30.

I tested it manually.  I would add a test but I don't know how to tell the test harness to target a specific GLSL version.

### DIFF
--- a/spirv_glsl.cpp
+++ b/spirv_glsl.cpp
@@ -6162,6 +6162,10 @@ std::string CompilerGLSL::to_texture_op(const Instruction &i, bool sparse, bool 
 		opt = &ops[5];
 		length -= 5;
 		gather = true;
+		if (options.es && options.version < 310)
+			SPIRV_CROSS_THROW("textureGather requires ESSL 310.");
+		else if (!options.es && options.version < 400)
+			SPIRV_CROSS_THROW("textureGather with depth compare requires GLSL 400.");
 		break;
 
 	case OpImageGather:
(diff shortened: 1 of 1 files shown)


HansKristian-Work:
Needs to be rebased now.

spirv_glsl.cpp
```diff
@@ -6162,6 +6162,10 @@ std::string CompilerGLSL::to_texture_op(const Instruction &i, bool sparse, bool
 		opt = &ops[5];
 		length -= 5;
 		gather = true;
+		if (options.es && options.version < 310)
+			SPIRV_CROSS_THROW("textureGather requires ESSL 310.");
+		else if (!options.es && options.version < 400)
```
HansKristian-Work, line 6167: Don't you have to check for Dref opcode here?

HansKristian-Work (changes requested):
Nit.

spirv_glsl.cpp
rdb: Not sure what you mean; this bit is right under
```cpp
	case OpImageDrefGather:
	case OpImageSparseDrefGather:
```

HansKristian-Work: Ah, right. Derp.

HansKristian-Work (approved):

HansKristian-Work:
The test suite fails locally. Somehow, Travis was not kicked off for this PR at all for some reason.

PR #1540, opened 2020-11-12
GLSL: Require GL_ARB_vertex_attrib_64bit for double input in pre-4.10

rdb:
This extension require is missing when using double vertex inputs with pre-4.10 versions.

### DIFF
--- a/spirv_glsl.cpp
+++ b/spirv_glsl.cpp
@@ -2305,6 +2305,12 @@ void CompilerGLSL::emit_interface_block(const SPIRVariable &var)
 {
 	auto &type = get<SPIRType>(var.basetype);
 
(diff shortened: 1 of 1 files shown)


spirv_glsl.cpp
```diff
@@ -2305,6 +2305,11 @@ void CompilerGLSL::emit_interface_block(const SPIRVariable &var)
 {
 	auto &type = get<SPIRType>(var.basetype);
 
+	if (var.storage == StorageClassInput && type.basetype == SPIRType::Double &&
+	    !options.es && options.version < 410) {
```
HansKristian-Work, line 2309: Wrong formatting.

HansKristian-Work (changes requested):
Nit.

PR #1543, opened 2020-11-19
MSL: Expand subgroup support.

cdavis5e:
Add support for declaring a fixed subgroup size. Metal, like Vulkan with
`VK_EXT_subgroup_size_control`, allows the thread execution width to
vary depending on factors such as register usage. Unfortunately, this
breaks several tests that depend on the subgroup size being what the
device says it is. So we'll fix the subgroup size at the size the device
declares. The extra invocations in the subgroup will appear to be
inactive. Because of this, the ballot mask builtins are now ANDed with
the active subgroup mask.

Add support for emulating a subgroup of size 1. This is intended to be
used by Vulkan Portability implementations (e.g. MoltenVK) when the
hardware/software combo provides insufficient support for subgroups.
Luckily for us, Vulkan 1.1 only requires that the subgroup size be at
least 1.

Add support for quadgroup and SIMD-group functions which were added to
iOS in Metal 2.2 and 2.3. This will allow clients to take advantage of
expanded quadgroup and SIMD-group support in recent Metal versions and
on recent Apple GPUs (families 6 and 7).

Gut emulation of subgroup builtins in fragment shaders. It turns out
codegen for the SIMD-group functions in fragment wasn't implemented for
AMD on Mojave; it's a safe bet that it wasn't implemented for the other
drivers either. Subgroup support in fragment shaders now requires Metal
2.2.


spirv_cross_c.h
```diff
@@ -647,6 +647,10 @@ typedef enum spvc_compiler_option
 
 	SPVC_COMPILER_OPTION_HLSL_FLATTEN_MATRIX_VERTEX_INPUT_SEMANTICS = 71 | SPVC_COMPILER_OPTION_HLSL_BIT,
 
+	SPVC_COMPILER_OPTION_MSL_IOS_USE_SIMDGROUP_FUNCTIONS = 72 | SPVC_COMPILER_OPTION_MSL_BIT,
```
HansKristian-Work, line 650: Need to update header versions.

spirv_msl.cpp
```diff
@@ -11034,60 +11187,154 @@ void CompilerMSL::fix_up_shader_inputs_outputs()
...
 				});
 				break;
 			case BuiltInSubgroupLtMask:
-				if (msl_options.is_ios())
-					SPIRV_CROSS_THROW("Subgroup ballot functionality is unavailable on iOS.");
+				if (msl_options.is_ios() && !msl_options.supports_msl_version(2, 2))
+					SPIRV_CROSS_THROW("Subgroup ballot functionality requires Metal 2.2 on iOS.");
 				if (!msl_options.supports_msl_version(2, 1))
 					SPIRV_CROSS_THROW("Subgroup ballot functionality requires Metal 2.1.");
+				add_spv_func_and_recompile(SPVFuncImplSubgroupBallot);
```
HansKristian-Work, line 11320: What is the ballot used for?

spirv_msl.cpp
```diff
@@ -11034,60 +11187,154 @@ void CompilerMSL::fix_up_shader_inputs_outputs()
...
 				});
 				break;
 			case BuiltInSubgroupLeMask:
-				if (msl_options.is_ios())
-					SPIRV_CROSS_THROW("Subgroup ballot functionality is unavailable on iOS.");
+				if (msl_options.is_ios() && !msl_options.supports_msl_version(2, 2))
+					SPIRV_CROSS_THROW("Subgroup ballot functionality requires Metal 2.2 on iOS.");
 				if (!msl_options.supports_msl_version(2, 1))
 					SPIRV_CROSS_THROW("Subgroup ballot functionality requires Metal 2.1.");
+				add_spv_func_and_recompile(SPVFuncImplSubgroupBallot);
```
HansKristian-Work, line 11295: What is ballot used for?

spirv_msl.cpp
```diff
@@ -11034,60 +11187,154 @@ void CompilerMSL::fix_up_shader_inputs_outputs()
...
 				entry_func.fixup_hooks_in.push_back([=]() {
-					statement(builtin_type_decl(bi_type), " ", to_expression(var_id),
-					          " = uint4(extract_bits(0xFFFFFFFF, 0, min(",
-					          to_expression(builtin_subgroup_invocation_id_id),
-					          " + 1, 32u)), extract_bits(0xFFFFFFFF, 0, (uint)max((int)",
-					          to_expression(builtin_subgroup_invocation_id_id), " + 1 - 32, 0)), uint2(0));");
+					if (msl_options.is_ios())
+					{
+						statement(builtin_type_decl(bi_type), " ", to_expression(var_id),
```
HansKristian-Work, line 11299: I'm a little confused why iOS gets a special path here. Is this a hardcoded assumption that iOS will never get more than 32 threads in a subgroup?

HansKristian-Work (commented):
Some initial comments.

HansKristian-Work:
An FYI, Travis is down for this project atm, so please make sure the test suite runs clean on macOS on your end at least.

spirv_msl.cpp
cdavis5e: If the subgroup size is fixed higher than the actual thread execution width, the bits for the extra inactive invocations will be set when they shouldn't be. This makes sure they won't be set.

The other thing I could do is add a separate thread execution width input and use that. I'd have to make sure that this is never emitted in the varying subgroup size case, though.

spirv_msl.cpp
cdavis5e: See above.

spirv_msl.cpp
cdavis5e: Yes. The `simd_vote::vote_t` type is a 32-bit integer on iOS.

cdavis5e:
> An FYI, Travis is down for this project atm, so please make sure the test suite runs clean on macOS on your end at least.

Ugh, not here too. Oh well, I always test my changes before pushing them, especially if I make changes to the test suite.

spirv_cross_c.h
cdavis5e: Oops. Done.

PR #1553, opened 2020-11-27
msl: Don't try to use [[thread_index_in_simdgroup]] in vertex shaders.

comex:
According to the Metal Shading Language Specification, it's not supported for vertex functions in any Metal version, only fragment and kernel functions.

### DIFF
--- a/spirv_msl.cpp
+++ b/spirv_msl.cpp
@@ -13319,12 +13319,17 @@ string CompilerMSL::builtin_qualifier(BuiltIn builtin)
 				SPIRV_CROSS_THROW("thread_index_in_simdgroup requires Metal 2.2 in fragment shaders.");
 			return "thread_index_in_simdgroup";
 		}
-		else
+		else if (execution.model == ExecutionModelKernel || execution.model == ExecutionModelGLCompute ||
+		         execution.model == ExecutionModelTessellationControl ||
+		         (execution.model == ExecutionModelVertex && msl_options.vertex_for_tessellation))
 		{
+			// We are generating a Metal kernel function.
 			if (!msl_options.supports_msl_version(2))
-				SPIRV_CROSS_THROW("Subgroup builtins require Metal 2.0.");
+				SPIRV_CROSS_THROW("Subgroup builtins in kernel functions require Metal 2.0.");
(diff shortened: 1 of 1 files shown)


comex:
Updated to properly treat `ExecutionModelTessellationControl` as generating a kernel function.

spirv_msl.cpp
```diff
@@ -13319,12 +13319,17 @@ string CompilerMSL::builtin_qualifier(BuiltIn builtin)
 				SPIRV_CROSS_THROW("thread_index_in_simdgroup requires Metal 2.2 in fragment shaders.");
 			return "thread_index_in_simdgroup";
 		}
-		else
+		else if (execution.model == ExecutionModelKernel || execution.model == ExecutionModelGLCompute ||
+		         execution.model == ExecutionModelTessellationControl ||
+		         (execution.model == ExecutionModelVertex && msl_options.vertex_for_tessellation))
```
HansKristian-Work, line 13324: What about fragment shaders?

spirv_msl.cpp
```diff
@@ -13319,12 +13319,17 @@ string CompilerMSL::builtin_qualifier(BuiltIn builtin)
 				SPIRV_CROSS_THROW("thread_index_in_simdgroup requires Metal 2.2 in fragment shaders.");
 			return "thread_index_in_simdgroup";
 		}
-		else
+		else if (execution.model == ExecutionModelKernel || execution.model == ExecutionModelGLCompute ||
```
HansKristian-Work, line 13322: ExecutionModeKernel is not supported (OpenCL). GLCompute is enough.

spirv_msl.cpp
HansKristian-Work: @cdavis5e Any comments here?

spirv_msl.cpp
HansKristian-Work: Does it make more sense to check for execution.model != ExecutionModelVertex?

spirv_msl.cpp
cdavis5e: Fragment shaders were handled by the previous `if`. There are no other places I can think of where the SIMD-group builtins are supported.

HansKristian-Work (approved):
Just gonna go ahead and merge, nothing fundamentally wrong with it I think.

PR #1563, opened 2020-12-07
GLSL/HLSL: Disallow VariablePointers capability outright.

### LINKED ISSUE #1559: Generated GLSL uses pointers even though GLSL has no pointers
I'm compiling a Rust shader to SPIR-V using [`SpirvBuilder`](https://embarkstudios.github.io/rust-gpu/api/spirv_builder/struct.SpirvBuilder.html) and then compiling the SPIR-V to GLSL using [SPIRV-Cross](https://github.com/grovesNL/spirv_cross).
When the Rust code contains code like `*[&c.x, &c.y, &c.z][x]`, the resulting GLSL has this code:
```glsl
_892[0u] = &_891.x;
_892[1u] = &_891.y;
_892[2u] = &_891.z;
```
which doesn't compile.

My workaround was to change the Rust code to `[c.x, c.y, c.z][x]`, but SPIRV-Cross should either transform the pointers to code that works or it should reject the input code with an error.

HansKristian-Work:
Cannot be supported, error out early.

Fix #1559.


Boscop:
Thanks!

PR #1596, opened 2021-01-18
MSL: Metal argument buffer use enhancements

billhollings:
Add query for type of MSL texture on desc set bindings:
- When encoding textures to argument buffers, Metal require knowledge of the type of each texture. Querying from shader conversion allows us to use that knowledge when encoding an argument buffer for a specific pipeline.

Support padding between argument buffer struct members to align arguments:

- Metal requires knowledge of the entire argument buffer structure to understand where to populate resources. SPIR-V does not include descriptor set content that is not used by the shader being converted. This makes it impossible for MSL to use a single argument buffer, representing a single descriptor set, across multiple shaders.
- By adding padding between known members of the descriptor set struct, we can define the full positional content of the descriptor-set argument buffer. This allows us to use the same Metal argument buffer across multiple shaders, each of which might only access a small part of the argument buffer.


HansKristian-Work:
First comment right off the bat. Lots of merge commits in the PR. Please rebase the PR.

spirv_msl.hpp
```diff
@@ -556,6 +580,9 @@ class CompilerMSL : public CompilerGLSL
 	// by remap_constexpr_sampler(_by_binding).
 	bool is_msl_resource_binding_used(spv::ExecutionModel model, uint32_t set, uint32_t binding) const;
 
+	// Returns the MSL texture type of the variable at the desc_set and binding
+	MSLTextureType get_msl_texture_type(uint32_t desc_set, uint32_t binding);
```
HansKristian-Work, line 584: Should probably be const?

spirv_msl.cpp
```diff
@@ -12593,6 +12593,55 @@ std::string CompilerMSL::sampler_type(const SPIRType &type, uint32_t id)
...
 
+MSLTextureType CompilerMSL::get_msl_texture_type(uint32_t desc_set, uint32_t binding)
+{
+	MSLTextureType msl_tex_type = MSL_TEXTURE_TYPE_2D;
+	ir.for_each_typed_id<SPIRVariable>([&](uint32_t self, SPIRVariable &var) {
+		if (has_decoration(self, DecorationDescriptorSet) && (get_decoration(self, DecorationDescriptorSet) == desc_set) &&
+			has_decoration(self, DecorationBinding) && (get_decoration(self, DecorationBinding) == binding))
+			msl_tex_type = get_msl_texture_type(get_variable_data_type(var));
+	});
+   return msl_tex_type;
```
HansKristian-Work, line 12604: Formatting.

spirv_msl.cpp
```diff
@@ -12593,6 +12593,55 @@ std::string CompilerMSL::sampler_type(const SPIRType &type, uint32_t id)
 		return "sampler";
 }
 
+MSLTextureType CompilerMSL::get_msl_texture_type(uint32_t desc_set, uint32_t binding)
+{
+	MSLTextureType msl_tex_type = MSL_TEXTURE_TYPE_2D;
+	ir.for_each_typed_id<SPIRVariable>([&](uint32_t self, SPIRVariable &var) {
+		if (has_decoration(self, DecorationDescriptorSet) && (get_decoration(self, DecorationDescriptorSet) == desc_set) &&
+			has_decoration(self, DecorationBinding) && (get_decoration(self, DecorationBinding) == binding))
+			msl_tex_type = get_msl_texture_type(get_variable_data_type(var));
```
HansKristian-Work, line 12602: Braces for clarity.

spirv_msl.hpp
```diff
@@ -95,6 +95,21 @@ enum MSLSamplerCoord
 	MSL_SAMPLER_INT_MAX = 0x7fffffff
 };
 
+enum MSLTextureType
```
HansKristian-Work, line 98: Why is this call useful? It seems like it's a convenience query for apps?

spirv_msl.cpp
```diff
@@ -14817,10 +14866,52 @@ void CompilerMSL::analyze_argument_buffers()
...
+				uint32_t pad_cnt = resource.index - consumed_arg_slots;
+				if (pad_cnt)
+				{
+					string mbr_name = join("_m", member_index, "_pad");
+					set_member_name(buffer_type.self, member_index, mbr_name);
+
+					bool use_array = pad_cnt > 1;
+					uint32_t pad_type_id = ir.increase_bound_by(use_array ? 2 : 1);
+					auto &pad_type = set<SPIRType>(pad_type_id);
+					pad_type.basetype = SPIRType::UInt64;
```
HansKristian-Work, line 14887: This is questionable. Doesn't this basically assume that descriptors take up 8 bytes? I would normally expect this kind of interface to specify the descriptor type used for specific IDs.

HansKristian-Work (changes requested):
Comments.

spirv_msl.cpp
HansKristian-Work: AMD texture descriptors are 16 or 32 bytes for example.

billhollings:
I  pulled the texture type and handled it outside using reflection. Turns out there is also still work to be done on the question of padding. So for now, I'm going to close this PR and resubmit a variation if it turns out to be needed.

PR #1606, opened 2021-01-28
Expose position invariance.

billhollings:
Used with MSL to determine whether to compile with invariance preserved.

### DIFF
--- a/spirv_cross.hpp
+++ b/spirv_cross.hpp
@@ -498,6 +498,12 @@ public:
 	// The most common use here is to check if a buffer is readonly or writeonly.
 	Bitset get_buffer_block_flags(VariableID id) const;
 
+	// Returns whether the position output is invariant
+	bool is_position_invariant() const
+	{
+		return position_invariant;
+	}
+
 protected:
 	const uint32_t *stream(const Instruction &instr) const
 	{


cdavis5e:
I don't understand why we can't just set `MTLCompileOptions.preserveInvariance` always when available. AFAICT, it only has an effect when the `[[position]]` output of a vertex shader (incl. post-tessellation vertex shader) has `[[invariant]]`, i.e. it doesn't affect the output when the shader doesn't have the attribute, so it should be safe to turn it on always.

HansKristian-Work (approved):
I don't mind this getting merged since it's trivial. Up to you whether to merge to close this ref the comment from @cdavis5e .

billhollings:
> I don't understand why we can't just set `MTLCompileOptions.preserveInvariance` always when available. AFAICT, it only has an effect when the `[[position]]` output of a vertex shader (incl. post-tessellation vertex shader) has `[[invariant]]`, i.e. it doesn't affect the output when the shader doesn't have the attribute, so it should be safe to turn it on always.

Interesting point. But it begs the question about why `MTLCompileOptions.preserveInvariance` exists at all, and why it is disabled by default. Why doesn't the MSL compiler just handle the `[[invariant]]` qualifier, and avoid the need for the additional compiler setting?

I suppose `MTLCompileOptions.preserveInvariance` could exist just as a way of reusing the same shader in both invariant and non-invariant use-cases (something like a "don't worry about invariance this time" option).

On the other hand, this kinda scares me:

>This is replaced with a conservative invariant model where the compiler marks operations that doesn’t go into an invariant calculation. This will guarantee anything that is invariant calculation remains invariant. This option may reduce performance as it may prevent certain optimizations to preserve invariance.

That, and the fact that `MTLCompileOptions.preserveInvariance` is disabled by default, makes me wonder if there are more fundamental differences in how the compiler approaches overall optimization if this switch is enabled.

Is there a downside to being conservative and enabling `MTLCompileOptions.preserveInvariance` only when the shader reports `is_position_invariant()` (and/or potentially other future invariance support)? 

I have the MoltenVK code to do just that ready, so unless there is a downside, let's take that approach, since it conservatively handles both cases (ie. the compiler switch does or does not make a difference to non-invariant shader code). And certainly as far as SPRIV-Cross goes, this change has no downside.

cdavis5e:
> Is there a downside to being conservative and enabling `MTLCompileOptions.preserveInvariance` only when the shader reports `is_position_invariant()` (and/or potentially other future invariance support)?

I suppose not.

PR #1615, opened 2021-02-15
c: Add missing IOS_SUPPORT_BASE_VERTEX_INSTANCE option.

### LINKED ISSUE #1612: MSL: No way to set msl_options.ios_support_base_vertex_instance from the C API
Hello, not sure if  I missed it, but It doesn't look like `msl_options.ios_support_base_vertex_instance` can be set form the C api.

Looking at `spirv_cross_c.cpp` I don't see it in the switch statement inside of `spvc_compiler_options_set_uint`.

I'd like to use it as otherwise I get the error: `BaseInstance requires Metal 1.1 and Mac or Apple A9+ hardware` when generating MSL with an IOS target.

HansKristian-Work:
Fix #1612.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -331,7 +331,7 @@ if (SPIRV_CROSS_STATIC)
 endif()
 
 set(spirv-cross-abi-major 0)
-set(spirv-cross-abi-minor 44)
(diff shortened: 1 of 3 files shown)


HansKristian-Work:
Trivial, just merging.

PR #1630, opened 2021-02-26
MSL: Deal with pointer-to-pointer qualifier ordering.

### LINKED ISSUE #1628: [MSL] C pointer type syntax is not (trivially) concatenable, like SPIRV-Cross assumes.
There are a bunch of places in `spirv_msl.cpp` in which types/declarations are built up by concatenation:
<details>
<summary>(click to expand a **non-exhaustive** series of examples - there's more I've skipped)</summary>

https://github.com/KhronosGroup/SPIRV-Cross/blob/621884d70917038caf7509f7b1b3c143807ff43f/spirv_msl.cpp#L6111
https://github.com/KhronosGroup/SPIRV-Cross/blob/621884d70917038caf7509f7b1b3c143807ff43f/spirv_msl.cpp#L6139
https://github.com/KhronosGroup/SPIRV-Cross/blob/621884d70917038caf7509f7b1b3c143807ff43f/spirv_msl.cpp#L10986
https://github.com/KhronosGroup/SPIRV-Cross/blob/621884d70917038caf7509f7b1b3c143807ff43f/spirv_msl.cpp#L10986
https://github.com/KhronosGroup/SPIRV-Cross/blob/621884d70917038caf7509f7b1b3c143807ff43f/spirv_msl.cpp#L11207-L11208
https://github.com/KhronosGroup/SPIRV-Cross/blob/621884d70917038caf7509f7b1b3c143807ff43f/spirv_msl.cpp#L12552
</details>

These all assume that you can substitute `T` in `some_addrspace T*` (or `some_addrspace T global;`) with any type and have the result mean "a pointer/global in `some_addrspace` pointing to `T`", but that only works in limited circumstances.

One of the simpler examples from what we've seen is (easiest way to get these is `OpUndef`s, FWIW) this type:
```
thread thread float2**
```

If C type syntax allowed for parenthesis, it'd be easy to see how we got there (and avoid the problem in the first place):
```
thread (thread float2*)*
```

But because the pointer syntax steals all of the relevant qualifiers before the `*`, what happens is:
* the first `*` takes *both* `thread`s
* the second `*` ends up with none

The simplest solution (for this example) is to *never* try and add anything *to the left* of a type, but instead *only add to the right* - that is, `T some_addrspace*` (or `T some_addrspace global;`), e.g.:
```
float2 thread* thread*
```

The order is irrelevant (as long as you don't move anything before/after an `*`) so this means the same as if you tried to go for preserving `thread float2*` in the "simple" case, but is easier to do than trying to account for everything that could be found in the nested type you're adding to.

It also has the nice property of making `some_addrspace*` look like a special form of `*`, but that's subjective.

<hr/>

We've also seen `constant texture2d<float>` show up as a type, how does that work? If the address space does indeed apply to `texture2d<...>` successfully, it probably has to be kept *before* `texture2d<...>`?

HansKristian-Work:
There might be more cases to consider, but this should take care of the common cases.
Pointer-to-pointer is very rare and cannot happen for resources, only variable expressions, so the scope to consider should be somewhat limited.

Also makes sure to set pointer_depth properly when we synthesize pointer types, it matters now since type_is_pointer needs to know.

Fix #1628.

### DIFF
--- a/spirv_msl.cpp
+++ b/spirv_msl.cpp
@@ -399,6 +399,7 @@ void CompilerMSL::build_implicit_builtins()
 				SPIRType vec4_type_ptr;
 				vec4_type_ptr = vec4_type;
 				vec4_type_ptr.pointer = true;
+				vec4_type_ptr.pointer_depth++;
 				vec4_type_ptr.parent_type = type_id;
 				vec4_type_ptr.storage = StorageClassInput;
 				auto &ptr_type = set<SPIRType>(type_ptr_id, vec4_type_ptr);
@@ -420,6 +421,7 @@ void CompilerMSL::build_implicit_builtins()
 				SPIRType uint_type_ptr;
 				uint_type_ptr = get_uint_type();
 				uint_type_ptr.pointer = true;
+				uint_type_ptr.pointer_depth++;
 				uint_type_ptr.parent_type = get_uint_type_id();
 				uint_type_ptr.storage = StorageClassInput;
 				auto &ptr_type = set<SPIRType>(type_ptr_id, uint_type_ptr);
@@ -441,6 +443,7 @@ void CompilerMSL::build_implicit_builtins()
 				SPIRType uint_type_ptr;
 				uint_type_ptr = get_uint_type();
 				uint_type_ptr.pointer = true;
+				uint_type_ptr.pointer_depth++;
 				uint_type_ptr.parent_type = get_uint_type_id();
 				uint_type_ptr.storage = StorageClassInput;
 				auto &ptr_type = set<SPIRType>(type_ptr_id, uint_type_ptr);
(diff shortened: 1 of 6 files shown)


spirv_msl.cpp
```diff
@@ -12548,8 +12629,22 @@ string CompilerMSL::type_to_glsl(const SPIRType &type, uint32_t id)
...
-		type_name = join(get_type_address_space(type, id), " ", type_to_glsl(get<SPIRType>(type.parent_type), id));
+
+		auto type_address_space = get_type_address_space(type, id);
+		auto type_decl = type_to_glsl(get<SPIRType>(type.parent_type), id);
+
+		// Work around C pointer qualifier rules. If glsl_type is a pointer type as well
+		// we'll need to emit the address space to the right.
+		// We could always go this route, but it makes the code unnatural.
+		// Prefer emitting thread T *foo over T thread* foo since it's more readable,
+		// but we'll have to emit thread T * thread * T constant bar; for example.
```
eddyb, line 12643: It would make the implementation 10x smaller to never do the previous ordering, but since you've gone through the trouble, might as well *shrug*.

HansKristian-Work: Main reason is maintenance, I don't want to review 1000+ reference file changes, and I find `device T *foo` 100x more readable than `T device * foo` personally.

eddyb: Thanks, that clears it up for me, I missed the test files.

PR #1642, opened 2021-03-25
MSL: Support long ulong types in buffers in 2.3+.

mehmetoguzderin:
[Metal Shading Language Specification](https://developer.apple.com/metal/Metal-Shading-Language-Specification.pdf) states that:

> Note: As of Metal 2.3, Metal supports buffers that contain long or ulong data types.

This PR aims to support long and ulong data types in buffers for Metal 2.3+.

### DIFF
--- a/spirv_msl.cpp
+++ b/spirv_msl.cpp
@@ -14092,12 +14092,6 @@ uint32_t CompilerMSL::get_declared_type_alignment_msl(const SPIRType &type, bool
 	case SPIRType::Sampler:
 		SPIRV_CROSS_THROW("Querying alignment of opaque object.");
 
-	case SPIRType::Int64:
-		if (!msl_options.supports_msl_version(2, 3))
-			SPIRV_CROSS_THROW("long types in buffers are only supported in MSL 2.3 and above.");
-	case SPIRType::UInt64:
-		if (!msl_options.supports_msl_version(2, 3))
-			SPIRV_CROSS_THROW("ulong types in buffers are only supported in MSL 2.3 and above.");
 	case SPIRType::Double:
 		SPIRV_CROSS_THROW("double types are not supported in buffers in MSL.");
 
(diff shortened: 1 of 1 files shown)


spirv_msl.cpp
```diff
@@ -14093,9 +14093,11 @@ uint32_t CompilerMSL::get_declared_type_alignment_msl(const SPIRType &type, bool
 		SPIRV_CROSS_THROW("Querying alignment of opaque object.");
 
 	case SPIRType::Int64:
-		SPIRV_CROSS_THROW("long types are not supported in buffers in MSL.");
+		if (!msl_options.supports_msl_version(2, 3))
+			SPIRV_CROSS_THROW("long types in buffers are only supported in MSL 2.3 and above.");
```
HansKristian-Work, line 14097: Need break;

spirv_msl.cpp
```diff
@@ -14093,9 +14093,11 @@ uint32_t CompilerMSL::get_declared_type_alignment_msl(const SPIRType &type, bool
...
 	case SPIRType::UInt64:
-		SPIRV_CROSS_THROW("ulong types are not supported in buffers in MSL.");
+		if (!msl_options.supports_msl_version(2, 3))
+			SPIRV_CROSS_THROW("ulong types in buffers are only supported in MSL 2.3 and above.");
```
HansKristian-Work, line 14100: Need break;

HansKristian-Work (changes requested):
Bugs

mehmetoguzderin:
@HansKristian-Work wouldn't `break` block execution from visiting `default` case?

HansKristian-Work:
> @HansKristian-Work wouldn't `break` block execution from visiting `default` case?

It never would, since it falls through to Double. You need to move the Int64/Uint64 checks to the default block.

mehmetoguzderin:
@HansKristian-Work then we don't add any breaks but throw if type is found to be 64-bit integer _inside_ default case?

mehmetoguzderin:
I assume break has no impact after throw because that aborts.

spirv_msl.cpp
mehmetoguzderin: PTAL at the new commit

spirv_msl.cpp
mehmetoguzderin: PTAL at the new commit

HansKristian-Work (approved):

PR #1643, opened 2021-03-26
Add some interfaces to aid with LTO-style optimization

### LINKED ISSUE #1639: MSL: gl_Position is undetectable in tessellation stage IO using the SPIRV-Cross reflection API
At Unity we've working on adding Metal tessellation support using SPIRV-Cross but have hit a limitation, hopefully there's already a way of doing this that we're unaware of.

The issue is that there does not seem to be a way to tell when and at what offset the generated MSL will contain a `gl_Position` member in IO interface for the Vertex or Hull kernels or the Domain vertex function as it's not reported along with the other `SPVC_RESOURCE_TYPE_STAGE_INPUT` or `SPVC_RESOURCE_TYPE_STAGE_OUTPUT` resource list members  even though it's in the generated MSL.

This results in the following problems when trying to validate and link up the different Metal stages:
1) To be able to set up control-point attribute data (`MTLVertexAttributeDescriptor` data in `MTLVertexDescriptor` ) for the post-tess vertex (domain) stage it is necessary to know the layout of the Hull kernel's control point output buffer. As the data transfer is done via buffer it does not seem possible to work around this and get the data layout from the Metal reflection for the Domain `MTLFunction`.
2) If we don't know that `gl_Position` is in the vertex kernel output or the hull kernel output, we cannot know the control point size/stride to be able to allocate minimally sized buffers for communication between the tessellation stages (this can be worked around by always assuming `gl_Position` is there and padding each point with  a `float4` though is suboptimal)
3) As we use `--msl-multi-patch-workgroup` the communication between the vertex and hull kernels happens directly through a buffer. We'd like to validate that the format is identical but since we cannot know if only one of them contains `gl_Position` this can't be done for all cases. Note that if `--msl-multi-patch-workgroup` is not used, stage IO validation is still not possible since Vertex stage output may or may not contain `gl_Position` even though it could be detected in the Hull stage input using the Metal runtime reflection (but not using SPIRV-Cross)
4) Similar as point 3, we'd like to be able to verify that the `gl_Position` post-tess vertex (domain) function input attribute was written to by the previous stage after compiling the two stages with SPIRV-Cross

Currently the only solution seems to be to manually parse out the `OpEntryPoint` instruction and to go over the list of `Interface <id>`s and only use SPIRV-Cross to check if any of them have a `BuiltIn` decoration of `Position`. This also requires us to assume that the `gl_Position` member will always be places at the end of the IO structs, as there's no way to get its offset in the IO buffers as well. As the placement looks like an arbitrary SPIRV-Cross implementation detail it does not seem like something we should rely on.

* The following is the source for the vert/hull/domain stages:
[tess_source.txt](https://github.com/KhronosGroup/SPIRV-Cross/files/6189619/tess_source.txt)

* This is the associated SPIR-V after compiling with DXC (binaries downloaded from [shader playground](http://shader-playground.timjones.io/26226e408b07100cfc2d8351cf3d0184), `.o` files renamed to `.spv.txt` to upload to github)
[Vert.spv.txt](https://github.com/KhronosGroup/SPIRV-Cross/files/6189613/Vert.spv.txt)
[Hull.spv.txt](https://github.com/KhronosGroup/SPIRV-Cross/files/6189612/Hull.spv.txt)
[Domain.spv.txt](https://github.com/KhronosGroup/SPIRV-Cross/files/6189611/Domain.spv.txt)

* The following is the MSL generated with SPIRV-Cross
[Vert.msl.txt](https://github.com/KhronosGroup/SPIRV-Cross/files/6189618/Vert.msl.txt) `./spirv-cross --msl --msl-vertex-for-tessellation Vert.spv.txt --output Vert.msl.txt`
[Hull.msl.txt](https://github.com/KhronosGroup/SPIRV-Cross/files/6189617/Hull.msl.txt) `./spirv-cross --msl --msl-multi-patch-workgroup Hull.spv.txt --output Hull.msl.txt`
[Domain.msl.txt](https://github.com/KhronosGroup/SPIRV-Cross/files/6189616/Domain.msl.txt) `./spirv-cross --msl Domain.spv.txt --output Domain.msl.txt`

HansKristian-Work:
This PR supports better cross-stage matching between vertex <-> tesc <-> tese stages by masking stage outputs.

While adding this, I found a lot of horrible edge cases in tessellation which had to be fixed to properly support this feature.

- Fix stage sorting by handling sorting of builtins correctly. Previously, location sorting would ignore builtins, which made robust cross-stage linking impossible.
- Fix several bugs with more complicated types in tessellation, especially passing function arguments.
- Remove unnecessary flattening cruft for multi-patch TESC. There are still many cases which are impossible to express in the legacy path. One consideration is that we cannot rewrite types this way, but the Vulkan spec only allows mismatch between stages for plain scalars or vectors, not composites.
- We still partially flatten Block decorated variables out of necessity, however, only the top-most member is flattened in multi-patch/capture output modes.
- Add API to mask location/component or builtin. Only supported in MSL, but could be implemented in HLSL as well later.

Fix #1639.


HansKristian-Work:
@cdavis5e @billhollings This will likely impact MoltenVK. Would be great if you could have a look as well.

PR #1648, opened 2021-04-13
MSL: Support padding Metal argument buffer entries based on argument index.

billhollings:
If `CompilerMSL::Options::pad_argument_buffer_resources` enabled, Metal argument buffer
struct members are positionally aligned to their argument indexes by adding synthetic
padding members when needed. The types and sizes of these synthetic members are
identified in the `resource_bindings` vector provided through the API.

Add `CompilerMSL::Options::pad_argument_buffer_resources` to enable padding
Metal argument buffer structs to positionally match members to argument indexes.
Add `MSLResourceBinding::base_type` to identify resource type through API.

### DIFF
--- a/spirv_msl.cpp
+++ b/spirv_msl.cpp
@@ -78,9 +78,22 @@ void CompilerMSL::add_msl_resource_binding(const MSLResourceBinding &binding)
 	arg_idx_tuple.binding = binding.msl_##rez; \
 	resource_arg_buff_idx_to_binding_number[arg_idx_tuple] = binding.binding
 
-		switch (binding.base_type)
+		switch (binding.basetype)
 		{
 		case SPIRType::Void:
+		case SPIRType::Boolean:
+		case SPIRType::SByte:
+		case SPIRType::UByte:
+		case SPIRType::Short:
+		case SPIRType::UShort:
+		case SPIRType::Int:
+		case SPIRType::UInt:
+		case SPIRType::Int64:
+		case SPIRType::UInt64:
+		case SPIRType::AtomicCounter:
+		case SPIRType::Half:
+		case SPIRType::Float:
+		case SPIRType::Double:
 			ADD_ARG_IDX_TO_BINDING_NUM_LOOKUP(buffer);
 			break;
 		case SPIRType::Image:
@@ -15140,22 +15153,35 @@ void CompilerMSL::analyze_argument_buffers()
 				while (resource.index > next_arg_buff_index)
 				{
 					auto &rez_bind = get_argument_buffer_resource(desc_set, next_arg_buff_index);
-					switch (rez_bind.base_type)
+					switch (rez_bind.basetype)
 					{
 					case SPIRType::Void:
-						add_argument_buffer_padding_buffer_type(buffer_type, member_index, next_arg_buff_index, rez_bind.count);
+					case SPIRType::Boolean:
+					case SPIRType::SByte:
+					case SPIRType::UByte:
+					case SPIRType::Short:
+					case SPIRType::UShort:
+					case SPIRType::Int:
+					case SPIRType::UInt:
+					case SPIRType::Int64:
+					case SPIRType::UInt64:
+					case SPIRType::AtomicCounter:
+					case SPIRType::Half:
+					case SPIRType::Float:
+					case SPIRType::Double:
+						add_argument_buffer_padding_buffer_type(buffer_type, member_index, next_arg_buff_index, rez_bind);
 						break;
 					case SPIRType::Image:
-						add_argument_buffer_padding_image_type(buffer_type, member_index, next_arg_buff_index, rez_bind.count);
+						add_argument_buffer_padding_image_type(buffer_type, member_index, next_arg_buff_index, rez_bind);
 						break;
 					case SPIRType::Sampler:
-						add_argument_buffer_padding_sampler_type(buffer_type, member_index, next_arg_buff_index, rez_bind.count);
+						add_argument_buffer_padding_sampler_type(buffer_type, member_index, next_arg_buff_index, rez_bind);
 						break;
 					case SPIRType::SampledImage:
 						if (next_arg_buff_index == rez_bind.msl_sampler)
-							add_argument_buffer_padding_sampler_type(buffer_type, member_index, next_arg_buff_index, rez_bind.count);
+							add_argument_buffer_padding_sampler_type(buffer_type, member_index, next_arg_buff_index, rez_bind);
 						else
-							add_argument_buffer_padding_image_type(buffer_type, member_index, next_arg_buff_index, rez_bind.count);
+							add_argument_buffer_padding_image_type(buffer_type, member_index, next_arg_buff_index, rez_bind);
 						break;
 					default:
 						break;
@@ -15296,13 +15322,13 @@ MSLResourceBinding &CompilerMSL::get_argument_buffer_resource(uint32_t desc_set,
 // Adds an argument buffer padding argument buffer type as one or more members of the struct type at the member index.
 // Metal does not support arrays of buffers, so these are emitted as multiple struct members.
 void CompilerMSL::add_argument_buffer_padding_buffer_type(SPIRType &struct_type, uint32_t &mbr_idx,
-                                                          uint32_t &arg_buff_index, uint32_t count)
+                                                          uint32_t &arg_buff_index, MSLResourceBinding &rez_bind)
 {
 	if (!argument_buffer_padding_buffer_type_id)
 	{
 		uint32_t buff_type_id = ir.increase_bound_by(2);
 		auto &buff_type = set<SPIRType>(buff_type_id);
-		buff_type.basetype = SPIRType::Void;
+		buff_type.basetype = rez_bind.basetype;
 		buff_type.storage = StorageClassUniformConstant;
 
 		uint32_t ptr_type_id = buff_type_id + 1;
@@ -15315,13 +15341,13 @@ void CompilerMSL::add_argument_buffer_padding_buffer_type(SPIRType &struct_type,
 		argument_buffer_padding_buffer_type_id = ptr_type_id;
 	}
 
-	for (uint32_t rez_idx = 0; rez_idx < count; rez_idx++)
+	for (uint32_t rez_idx = 0; rez_idx < rez_bind.count; rez_idx++)
 		add_argument_buffer_padding_type(argument_buffer_padding_buffer_type_id, struct_type, mbr_idx, arg_buff_index, 1);
 }
 
 // Adds an argument buffer padding argument image type as a member of the struct type at the member index.
 void CompilerMSL::add_argument_buffer_padding_image_type(SPIRType &struct_type, uint32_t &mbr_idx,
-                                                         uint32_t &arg_buff_index, uint32_t count)
+                                                         uint32_t &arg_buff_index, MSLResourceBinding &rez_bind)
 {
 	if (!argument_buffer_padding_image_type_id)
 	{
@@ -15347,12 +15373,12 @@ void CompilerMSL::add_argument_buffer_padding_image_type(SPIRType &struct_type, 
 		argument_buffer_padding_image_type_id = img_type_id;
 	}
 
-	add_argument_buffer_padding_type(argument_buffer_padding_image_type_id, struct_type, mbr_idx, arg_buff_index, count);
+	add_argument_buffer_padding_type(argument_buffer_padding_image_type_id, struct_type, mbr_idx, arg_buff_index, rez_bind.count);
 }
 
 // Adds an argument buffer padding argument sampler type as a member of the struct type at the member index.
 void CompilerMSL::add_argument_buffer_padding_sampler_type(SPIRType &struct_type, uint32_t &mbr_idx,
-                                                           uint32_t &arg_buff_index, uint32_t count)
+                                                           uint32_t &arg_buff_index, MSLResourceBinding &rez_bind)
 {
 	if (!argument_buffer_padding_sampler_type_id)
 	{
@@ -15364,7 +15390,7 @@ void CompilerMSL::add_argument_buffer_padding_sampler_type(SPIRType &struct_type
 		argument_buffer_padding_sampler_type_id = samp_type_id;
 	}
 
-	add_argument_buffer_padding_type(argument_buffer_padding_sampler_type_id, struct_type, mbr_idx, arg_buff_index, count);
+	add_argument_buffer_padding_type(argument_buffer_padding_sampler_type_id, struct_type, mbr_idx, arg_buff_index, rez_bind.count);
 }
 
 // Adds the argument buffer padding argument type as a member of the struct type at the member index.
--- a/spirv_msl.cpp
+++ b/spirv_msl.cpp
@@ -72,26 +72,33 @@ void CompilerMSL::add_msl_resource_binding(const MSLResourceBinding &binding)
 	// arg buffer indexes, also maintain a lookup by argument buffer index.
 	if (msl_options.pad_argument_buffer_resources)
 	{
-		uint32_t arg_buff_idx = k_unknown_component;
+		StageSetBinding arg_idx_tuple = { binding.stage, binding.desc_set, k_unknown_component };
+
+#define ADD_ARG_IDX_TO_BINDING_NUM_LOOKUP(rez)  \
+	arg_idx_tuple.binding = binding.msl_ ##rez; \
+	resource_arg_buff_idx_to_binding_number[arg_idx_tuple] = binding.binding
+
 		switch (binding.base_type)
 		{
 		case SPIRType::Void:
-			arg_buff_idx = binding.msl_buffer;
+			ADD_ARG_IDX_TO_BINDING_NUM_LOOKUP(buffer);
 			break;
 		case SPIRType::Image:
-		case SPIRType::SampledImage:
-			arg_buff_idx = binding.msl_texture;
+			ADD_ARG_IDX_TO_BINDING_NUM_LOOKUP(texture);
 			break;
 		case SPIRType::Sampler:
-			arg_buff_idx = binding.msl_sampler;
+			ADD_ARG_IDX_TO_BINDING_NUM_LOOKUP(sampler);
+			break;
+		case SPIRType::SampledImage:
+			ADD_ARG_IDX_TO_BINDING_NUM_LOOKUP(texture);
+			ADD_ARG_IDX_TO_BINDING_NUM_LOOKUP(sampler);
 			break;
 		default:
 			SPIRV_CROSS_THROW("Unexpected argument buffer resource base type. When padding argument buffer elements, "
 			                  "all descriptor set resources must be supplied with a base type by the app.");
 			break;
 		}
-		StageSetBinding arg_idx_tuple = { binding.stage, binding.desc_set, arg_buff_idx };
-		resource_arg_buff_idx_to_binding_number[arg_idx_tuple] = binding.binding;
+#undef ADD_ARG_IDX_TO_BINDING_NUM_LOOKUP
 	}
 }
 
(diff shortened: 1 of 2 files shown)


spirv_msl.hpp
```diff
@@ -1027,6 +1051,11 @@ class CompilerMSL : public CompilerGLSL
 
 	void analyze_argument_buffers();
 	bool descriptor_set_is_argument_buffer(uint32_t desc_set) const;
+	MSLResourceBinding& get_argument_buffer_resource(uint32_t desc_set, uint32_t arg_idx);
```
HansKristian-Work, line 1054: `&` binds right.

spirv_msl.hpp
```diff
@@ -1027,6 +1051,11 @@ class CompilerMSL : public CompilerGLSL
...
+	void add_argument_buffer_padding_buffer_type(SPIRType& struct_type, uint32_t& mbr_idx, uint32_t count);
+	void add_argument_buffer_padding_image_type(SPIRType& struct_type, uint32_t& mbr_idx, uint32_t count);
+	void add_argument_buffer_padding_sampler_type(SPIRType& struct_type, uint32_t& mbr_idx, uint32_t count);
+	void add_argument_buffer_padding_type(uint32_t mbr_type_id, SPIRType& struct_type, uint32_t& mbr_idx, uint32_t count);
```
HansKristian-Work, line 1058: `&` binds right.

spirv_msl.cpp
```diff
@@ -15196,6 +15235,133 @@ void CompilerMSL::analyze_argument_buffers()
...
+
+		argument_buffer_padding_sampler_type_id = samp_type_id;
+	}
+
+	add_argument_buffer_padding_type(argument_buffer_padding_sampler_type_id, struct_type, mbr_idx, count);
+}
+
+// Adds the argument buffer padding argument type as a member of the struct type at the member index.
+// Returns the number of argument buffer slots consumed.
+void CompilerMSL::add_argument_buffer_padding_type(uint32_t mbr_type_id, SPIRType& struct_type, uint32_t& mbr_idx, uint32_t count)
```
HansKristian-Work, line 15345: `&` binds right. Probably other instances of this nit.

spirv_msl.cpp
```diff
@@ -15196,6 +15235,133 @@ void CompilerMSL::analyze_argument_buffers()
...
+		base_type.basetype = SPIRType::Float;
+		base_type.width = 32;
+
+		uint32_t img_type_id = base_type_id + 1;
+		auto &img_type = set<SPIRType>(img_type_id);
+		img_type.basetype = SPIRType::Image;
+		img_type.storage = StorageClassUniformConstant;
+
+		img_type.image.type = base_type_id;
+		img_type.image.dim = Dim2D;
```
HansKristian-Work, line 15312: Does this actually work? In Metal, the size of an image descriptor can change based on state like dimension, MSAA, etc. Afaik, the type must match exactly.

spirv_msl.cpp
```diff
@@ -15196,6 +15235,133 @@ void CompilerMSL::analyze_argument_buffers()
...
+		uint32_t ptr_type_id = buff_type_id + 1;
+		auto &ptr_type = set<SPIRType>(ptr_type_id);
+		ptr_type = buff_type;
+		ptr_type.pointer = true;
+		ptr_type.pointer_depth++;
+		ptr_type.parent_type = buff_type_id;
+
+		argument_buffer_padding_buffer_type_id = ptr_type_id;
+	}
+	for (uint32_t rez_idx = 0; rez_idx < count; rez_idx++) {
```
HansKristian-Work, line 15291: Formatting

spirv_msl.cpp
```diff
@@ -15196,6 +15235,133 @@ void CompilerMSL::analyze_argument_buffers()
...
 }
 
+// Return the resource type of the app-provided resources for the descriptor set,
+// that matches the resource index of the argument buffer index.
+MSLResourceBinding& CompilerMSL::get_argument_buffer_resource(uint32_t desc_set, uint32_t arg_idx)
+{
+	for (auto itr = resource_bindings.begin(); itr != resource_bindings.end(); itr++)
+	{
+		auto& rez_bind = itr->second.first;
+		uint32_t rez_idx = ~0;
```
HansKristian-Work, line 15245: Integer cast warnings.

spirv_msl.cpp
```diff
@@ -15196,6 +15235,133 @@ void CompilerMSL::analyze_argument_buffers()
 	}
 }
 
+// Return the resource type of the app-provided resources for the descriptor set,
+// that matches the resource index of the argument buffer index.
+MSLResourceBinding& CompilerMSL::get_argument_buffer_resource(uint32_t desc_set, uint32_t arg_idx)
+{
+	for (auto itr = resource_bindings.begin(); itr != resource_bindings.end(); itr++)
```
HansKristian-Work, line 15242: Use range for loop.

spirv_msl.cpp
```diff
@@ -15094,10 +15094,49 @@ void CompilerMSL::analyze_argument_buffers()
...
 			auto &var = *resource.var;
 			auto &type = get_variable_data_type(var);
+
+			// If needed, synthesize and add padding members.
+			// member_index and next_arg_buff_index are incremented when padding members are added.
+			if (msl_options.pad_argument_buffer_resources)
+			{
+				while (resource.index > next_arg_buff_index)
+				{
+					auto& rez_bind = get_argument_buffer_resource(desc_set, next_arg_buff_index);
```
HansKristian-Work, line 15109: O(N) looping over all binds feels awkward. Probably more straight forward to loop over the bindings and build a map from argument buffer ID -> base type, and then plug holes with trivial argument index -> basetype lookups.

HansKristian-Work (changes requested):

spirv_msl.hpp
billhollings: Fixed.

spirv_msl.cpp
billhollings: Fixed.

spirv_msl.hpp
billhollings: Fixed.

spirv_msl.cpp
billhollings: Fixed.

spirv_msl.cpp
billhollings: Replaced with map lookup.

spirv_msl.cpp
billhollings: Fixed.

spirv_msl.cpp
billhollings: I've added `resource_arg_buff_idx_to_binding_number` as a parallel secondary map that maps resource arg index to resource binding. 

I designed this as a two-stage lookup (`arg-idx -> binding# -> resource binding`), rather than linking to a `MSLResourceBinding*` from within `resource_arg_buff_idx_to_binding_number`, because using a direct pointer would point to content with `resource_bindings` map, which I felt was too much entanglement, and would leave dangling pointers if future changes removed content from `resource_bindings`.

spirv_msl.cpp
billhollings: >Does this actually work?

This definitely works on all the macOS devices I've tested (AMD, Intel, M1). On macOS, all argument descriptors are the same size (8 bytes). And all app testing with this design I've done has been successful so far on macOS.

However, this is flaky on iOS, and I'm still working through it. 

>In Metal, the size of an image descriptor can change based on state like dimension, MSAA, etc. Afaik, the type must match exactly.

What is the source of this statement? Is this documented somewhere that I haven't found yet, or is it an assumption based on the requirement for `MTLTextureType` as part of the argument encoder configuration?

On iOS, buffer, image, and sampler arg descriptors have different sizes from one another, but measurements I've done by asking `[MTLArgumentEncoder encodedLength]` produces a consistent 128 bytes for all `MTLTextureTypes`. 

Based on that, I've been working under the assumption that the need for `MTLTextureType` was not related to size specifically, but encoding. In that case, after taking into consideration resource type, a standard padding space for image args independent of `MTLTextureType` _**should**_ work. But I must admit I haven't got this working entirely properly yet on iOS, and perhaps size does matter here after all (even though it is not obviously reported through the size measurement API).

The advantages of getting this approach working wherever possible are compelling, because it associates the Metal argument buffer with a descriptor set, rather than a pipeline shader stage, and separates descriptor set updates from pipeline encoding, as Vulkan intends. Once written to, a descriptor set Metal argument buffer can be bound to multiple pipelines across multiple pipeline stages, without having to repeatedly dynamically write to a separate Metal argument buffer for each pipeline stage.

So...I'd like to include this approach as an option to enable on platforms like macOS that will support it. On iOS, we'll see, and on iOS we might have to turn this off and swap in some form of dynamic writing to per-pipeline-stage Metal argument buffers. Very non-Vulkan and very inefficient. Ugggh!

@cdavis5e You did a chunk of work on this earlier, and any insight you have would be appreciated as well.

billhollings:
What's the nature of the CI/appveyor failure? I'm not familiar with that environment, and there is no feedback on code issues:

>`C:\projects\spirv-cross\spirv_msl.cpp : fatal error C1128: number of sections exceeded object file format limit: compile with /bigobj [C:\projects\spirv-cross\_builds\spirv-cross-msl.vcxproj]`

spirv_msl.cpp
HansKristian-Work: > even though it is not obviously reported through the size measurement API

128 seems excessive. I'm going to assume this is some kind of alignment. Probably try taking the size of an array of 1024 images of different types?

HansKristian-Work: To be clear, I'm not against this PR. If you can ship it in MoltenVK, then it's good enough for me.

HansKristian-Work: > `MTLTextureType` as part of the argument encoder configuration?

I recall in portability discussions that you had to add in texture type in Vulkan somehow to be able to use IAB, but my memory is hazy.

HansKristian-Work:
> What's the nature of the CI/appveyor failure? I'm not familiar with that environment, and there is no feedback on code issues:
> 
> > `C:\projects\spirv-cross\spirv_msl.cpp : fatal error C1128: number of sections exceeded object file format limit: compile with /bigobj [C:\projects\spirv-cross\_builds\spirv-cross-msl.vcxproj]`

Rebase against master. MSVC is broken and I had to add /bigobj to workaround spirv_msl.cpp being too big.

spirv_msl.cpp
```diff
@@ -15094,10 +15120,53 @@ void CompilerMSL::analyze_argument_buffers()
...
+						next_arg_buff_index += add_argument_buffer_padding_buffer_type(buffer_type, member_index, rez_bind.count);
+						break;
+					case SPIRType::Image:
+						next_arg_buff_index += add_argument_buffer_padding_image_type(buffer_type, member_index, rez_bind.count);
+						break;
+					case SPIRType::Sampler:
+						next_arg_buff_index += add_argument_buffer_padding_sampler_type(buffer_type, member_index, rez_bind.count);
+						break;
+					case SPIRType::SampledImage:
+						next_arg_buff_index += add_argument_buffer_padding_image_type(buffer_type, member_index, rez_bind.count);
```
HansKristian-Work, line 15148: Are SampledImages always packed together in IABs? iirc, it's possible to setup a layout in SPIRV-Cross that looks like:
```
struct Foo { texture a [[id(10)]]; sampler a_smplr [[id(400)]]; }
```
so I wonder if we have to emit dummy binds for either the texture side or sampler side depending on how this works.

spirv_msl.cpp
```diff
@@ -67,6 +67,32 @@ void CompilerMSL::add_msl_resource_binding(const MSLResourceBinding &binding)
...
+	{
+		uint32_t arg_buff_idx = k_unknown_component;
+		switch (binding.base_type)
+		{
+		case SPIRType::Void:
+			arg_buff_idx = binding.msl_buffer;
+			break;
+		case SPIRType::Image:
+		case SPIRType::SampledImage:
+			arg_buff_idx = binding.msl_texture;
```
HansKristian-Work, line 83: I think we need to split SampledImage into binding.msl_texture and binding.msl_sampler if I recall the implementation correctly.

spirv_msl.cpp
```diff
@@ -15196,6 +15265,130 @@ void CompilerMSL::analyze_argument_buffers()
...
+	if (arg_itr != end(resource_arg_buff_idx_to_binding_number))
+	{
+		StageSetBinding bind_tuple = { stage, desc_set, arg_itr->second };
+		auto bind_itr = resource_bindings.find(bind_tuple);
+		if (bind_itr != end(resource_bindings))
+			return bind_itr->second.first;
+	}
+	SPIRV_CROSS_THROW("Argument buffer resource base type could not be determined. When padding argument buffer "
+	                  "elements, all descriptor set resources must be supplied with a base type by the app.");
+	static MSLResourceBinding unkwn_rez;
```
HansKristian-Work, line 15286: unreachable code.

spirv_msl.cpp
billhollings: >128 seems excessive. I'm going to assume this is some kind of alignment. Probably try taking the size of an array of 1024 images of different types?

It does seem excessive, doesn't it? But it seems to be real on iOS:

```
[mvk-info] Metal arg buff for 1024 Buffer storage is 8200.
[mvk-info] Metal arg buff for 1024 Sampler storage is 16392.
[mvk-info] Metal arg buff for 1024 MTLTextureType1D storage is 131080.
[mvk-info] Metal arg buff for 1024 MTLTextureType1DArray storage is 131080.
[mvk-info] Metal arg buff for 1024 MTLTextureType2D storage is 131080.
[mvk-info] Metal arg buff for 1024 MTLTextureType2DArray storage is 131080.
[mvk-info] Metal arg buff for 1024 MTLTextureType2DMultisample storage is 131080.
[mvk-info] Metal arg buff for 1024 MTLTextureType2DMultisampleArray storage is 131080.
[mvk-info] Metal arg buff for 1024 MTLTextureTypeCube storage is 131080.
[mvk-info] Metal arg buff for 1024 MTLTextureTypeCubeArray storage is 131080.
[mvk-info] Metal arg buff for 1024 MTLTextureType3D storage is 131080.
[mvk-info] Metal arg buff for 1024 MTLTextureTypeTextureBuffer storage is 131080.
```

Each arg buffer has an 8-byte "header" or equivalent. Other than that, the remaining content is 128 bytes per for textures.

Contrast with macOS. In this case, Apple Silicon M1, which make the above even more odd, since the architectures are similar:

```
[mvk-info] Metal arg buff for 1024 Buffer storage is 8192.
[mvk-info] Metal arg buff for 1024 Sampler storage is 8192.
[mvk-info] Metal arg buff for 1024 MTLTextureType1D storage is 8192.
[mvk-info] Metal arg buff for 1024 MTLTextureType1DArray storage is 8192.
[mvk-info] Metal arg buff for 1024 MTLTextureType2D storage is 8192.
[mvk-info] Metal arg buff for 1024 MTLTextureType2DArray storage is 8192.
[mvk-info] Metal arg buff for 1024 MTLTextureType2DMultisample storage is 8192.
[mvk-info] Metal arg buff for 1024 MTLTextureType2DMultisampleArray storage is 8192.
[mvk-info] Metal arg buff for 1024 MTLTextureTypeCube storage is 8192.
[mvk-info] Metal arg buff for 1024 MTLTextureTypeCubeArray storage is 8192.
[mvk-info] Metal arg buff for 1024 MTLTextureType3D storage is 8192.
[mvk-info] Metal arg buff for 1024 MTLTextureTypeTextureBuffer storage is 8192.
```
No "header" and 8-bytes per resource descriptor, regardless of type.

spirv_msl.cpp
billhollings: I wasn't sure what happens with `SPIRV_CROSS_EXCEPTIONS_TO_ASSERTIONS`. 

Does the assertion ever get compiled out (say in a prod build)?

spirv_msl.cpp
billhollings: Hmmm...the `MSLResourceBinding` for that would have both texture and sampler. MoltenVK ensures they are packed together in consecutive arguments...but I suppose another client library might have them laid out differently for some reason. We'd have to somehow track during the lookup that "that time I was after the texture, this time I'm after the sampler", etc.

Unless we have need to make it more complicated at the moment, perhaps we could go with this simpler assumption for now, and grow it out if other clients need to support a different approach to adding both textures and samplers from a single `SampledImage` resource.

spirv_msl.cpp
billhollings: I don't think that's needed. However...in thinking more about your question above about distribution of ID's between the texture and sampler component...we might be able to. Let me think about this and see if a simple option exists.

spirv_msl.cpp
cdavis5e: > This definitely works on all the macOS devices I've tested (AMD, Intel, M1). On macOS, all argument descriptors are the same size (8 bytes). And all app testing with this design I've done has been successful so far on macOS.
> 
> > In Metal, the size of an image descriptor can change based on state like dimension, MSAA, etc. Afaik, the type must match exactly.
> 
> What is the source of this statement?

Me. When *I* tested it on my AMD Radeon Pro 460, buffer descriptors were 16 bytes; and texture descriptors were 32 bytes, except for 2D MS, which was 64 bytes--but strangely not 2D MS array, which was again 32 bytes.

Maybe it depends on your OS version? I tested on Mojave. That you tested on M1 suggests you are using Big Sur. Maybe they're actually working on my request to make this agnostic of texture type?

cdavis5e:
> > What's the nature of the CI/appveyor failure? I'm not familiar with that environment, and there is no feedback on code issues:
> > > `C:\projects\spirv-cross\spirv_msl.cpp : fatal error C1128: number of sections exceeded object file format limit: compile with /bigobj [C:\projects\spirv-cross\_builds\spirv-cross-msl.vcxproj]`
> 
> MSVC is broken and I had to add /bigobj to workaround spirv_msl.cpp being too big.
> 
> EDIT: I think that commit is stuck on my branch. I'll ignore appveyor failures for now. You can cherry-pick
> [8f95a3d](https://github.com/KhronosGroup/SPIRV-Cross/commit/8f95a3de1e04595b29a39f34412f4e59ab13acec)

More specifically, COFF by default only supports 65536 sections in an object file. Normally, you would have far fewer than that (`.text`, `.data`, `.rdata`, `.bss`, `.pdata`, `.xdata`, `.drectve` are the most common ones in object files), but with C++ `COMDAT`s (e.g. templates, inline functions), each of which gets its own section, it's easy to exceed that limit with a large source file. I don't know why MSVC isn't smart enough to figure this out on its own and enable `/bigobj` automatically.

spirv_msl.cpp
cdavis5e: > > This definitely works on all the macOS devices I've tested (AMD, Intel, M1). On macOS, all argument descriptors are the same size (8 bytes). And all app testing with this design I've done has been successful so far on macOS.
> > > In Metal, the size of an image descriptor can change based on state like dimension, MSAA, etc. Afaik, the type must match exactly.
> > 
> > 
> > What is the source of this statement?
> 
> Me. When _I_ tested it on my AMD Radeon Pro 460, buffer descriptors were 16 bytes; and texture descriptors were 32 bytes, except for 2D MS, which was 64 bytes--but strangely not 2D MS array, which was again 32 bytes.
> 
> Maybe it depends on your OS version? I tested on Mojave. That you tested on M1 suggests you are using Big Sur. Maybe they're actually working on my request to make this agnostic of texture type?

It occurs to me that you tested multiple descriptors per buffer, but I only tested one. You found that the buffer had an 8-byte header, because the descriptor count doesn't go evenly into the buffer size. That might account for some of the different sizes we saw.

spirv_msl.cpp
billhollings: Latest adds separate lookups for the image and sampler components of `SampledImage`.

spirv_msl.cpp
billhollings: Fixed.

spirv_msl.cpp
billhollings: Latest adds separate lookups for the image and sampler components of `SampledImage`.

spirv_msl.cpp
cdavis5e: So I tried it with multiple elements this time, both with `arrayLength` greater than one, and with multiple descriptors... and I found the same results as before: buffers are 16 bytes, textures are 32 bytes, 2D MS textures (but not 2D MS array) are 64 bytes. I tried samplers, too (I didn't last time); they are also 16 bytes.

Yeah, I think it depends on your OS version.

billhollings: >Maybe it depends on your OS version? I tested on Mojave. That you tested on M1 suggests you are using Big Sur. Maybe they're actually working on my request to make this agnostic of texture type?

Yes. Big Sur in all cases. You might be right about the OS advances changing things. On macOS, all descriptors, independent of type, are 8 bytes now. And for textures and samplers those seem to be just numeric indexes into some kind of lookup. Perhaps they are caching the actual descriptors themselves somewhere else, and not in the actual `MTLBuffer` used as the argument buffer?

And odd that iOS has such a huge size for texture descriptors. Assuming macOS M1 hardware would be similar to iOS, that's another indication that some kind of reference abstraction might now be happening within the driver on macOS when it comes to arg buffer content.

cdavis5e: I also tried it on my Intel GPU on Mojave. There I see the same results as @billhollings: buffers, textures, and samplers are all 8 bytes. Are you sure you tested using the AMD GPU? I wonder if Apple has started doing this for all GPUs because programs have begun to depend on this behavior.

cdavis5e: I also tried it on my M1 on Big Sur, and I get the same results as Bill there, too.

billhollings: Yes...I tested all three macOS GPU's I have (AMD, Intel, and M1). AMD is actually the default on the machine I run CTS on.

Maybe they had to create an abstraction for Intel, and are now standardizing it to all drivers in order to make it consistent and not so dependent on types (per your filed request).

And...don't get me started on the requirement to specify usage for every resource on every command encoder. That almost completely negates the call reduction gained by submitting the resources in an argument buffer.

billhollings:
@HansKristian-Work 

I tweaked this a bit over the weekend, to add some flexibility, in my attempts to get Metal arg buffs working on iOS under MoltenVK.

I'm still working on iOS (and earlier non-Intel macOS per @cdavis5e findings) for MoltenVK, but I think we're good to go here as far as our needs from SPIRV-Cross now. 

If everything looks good, I'd like to get this pulled in ASAP so that I can move to push a "beta" version of Metal arg buffer support to MoltenVK, to support (for now) at least modern-macOS, this week ahead of the SDK next week.

@cdavis Per your retesting above...thanks for Intel on Mojave...and are you able to try on Catalina by any chance? Just trying to sort out what to restrict this to for now.

cdavis5e:
> and are you able to try on Catalina by any chance?

No, sorry. I'm running low on disk space, and I upgraded my Catalina install to Big Sur.

PR #1662, opened 2021-04-26
MSL: Fix '--msl-multi-patch-workgroup' out of bounds reads when dispatching more threads than control points

Dredhog:
We had been working aound this by using:
`[computeEncoder dispatchThreads: threadGridDims threadsPerThreadgroup: threadsPerThreadgroup];`
instead of:
`[computeEncoder dispatchThreadgroups: threadgroups threadsPerThreadgroup: threadsPerThreadgroup];`
as we had experienced flickering and undefined behaviour with `dispatchThreadgroups` when the last threadgroup did not have enough data to fill all `threadsPerThreadgroup` and would result in writes passed the end of the output buffer and some off by one issues.

But it appears that the `dispatchThreads` version is less widely supported (in terms of tvOS and others) so I'm adding bounds checks on the compiler side to be able to use the `dispatchThreadgroups` version safely.

**Edit:** currently removed the changed which fixed the out of bounds _writes_ based on PR feedback (only leaving fix for out of bounds _reads_)

### DIFF
--- a/spirv_msl.cpp
+++ b/spirv_msl.cpp
@@ -3462,17 +3462,17 @@ uint32_t CompilerMSL::add_interface_block(StorageClass storage, bool patch)
 					{
 						entry_func.fixup_hooks_in.push_back([=]() {
 							statement("device ", to_name(ir.default_entry_point), "_", ib_var_ref, "& ", ib_var_ref,
-							          " = ", patch_output_buffer_var_name, "[", to_expression(builtin_invocation_id_id),
-							          ".x / ", get_entry_point().output_vertices, "];");
+							          " = ", patch_output_buffer_var_name, "[min(", to_expression(builtin_invocation_id_id),
+							          ".x / ", get_entry_point().output_vertices, ", spvIndirectParams[1] - 1)];");
 						});
 					}
 					else
 					{
 						entry_func.fixup_hooks_in.push_back([=]() {
+							uint32_t output_vertices = get_entry_point().output_vertices;
 							statement("device ", to_name(ir.default_entry_point), "_", ib_var_ref, "* gl_out = &",
-							          output_buffer_var_name, "[", to_expression(builtin_invocation_id_id), ".x - ",
-							          to_expression(builtin_invocation_id_id), ".x % ",
-							          get_entry_point().output_vertices, "];");
+							          output_buffer_var_name, "[min(", to_expression(builtin_invocation_id_id), ".x / ",
+							          output_vertices, ", ", "spvIndirectParams[1] - 1) * ", output_vertices, "];");
 						});
 					}
 				}
@@ -11796,7 +11796,7 @@ void CompilerMSL::fix_up_shader_inputs_outputs()
 				entry_func.fixup_hooks_in.push_back([=]() {
 					statement(builtin_type_decl(bi_type), " ", to_expression(var_id), " = min(",
 					          to_expression(builtin_invocation_id_id), ".x / ", this->get_entry_point().output_vertices,
-					          ", spvIndirectParams[1]);");
+					          ", spvIndirectParams[1] - 1);");
 				});
 				break;
 			case BuiltInPatchVertices:
--- a/spirv_msl.cpp
+++ b/spirv_msl.cpp
@@ -3462,17 +3462,17 @@ uint32_t CompilerMSL::add_interface_block(StorageClass storage, bool patch)
 					{
 						entry_func.fixup_hooks_in.push_back([=]() {
 							statement("device ", to_name(ir.default_entry_point), "_", ib_var_ref, "& ", ib_var_ref,
-							          " = ", patch_output_buffer_var_name, "[min(", to_expression(builtin_invocation_id_id),
-							          ".x / ", get_entry_point().output_vertices, ", spvIndirectParams[1] - 1)];");
+							          " = ", patch_output_buffer_var_name, "[", to_expression(builtin_invocation_id_id),
+							          ".x / ", get_entry_point().output_vertices, "];");
 						});
 					}
 					else
 					{
 						entry_func.fixup_hooks_in.push_back([=]() {
-							uint32_t output_vertices = get_entry_point().output_vertices;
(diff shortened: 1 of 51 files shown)


spirv_msl.cpp
```diff
@@ -3462,17 +3462,17 @@ uint32_t CompilerMSL::add_interface_block(StorageClass storage, bool patch)
 					{
 						entry_func.fixup_hooks_in.push_back([=]() {
 							statement("device ", to_name(ir.default_entry_point), "_", ib_var_ref, "& ", ib_var_ref,
-							          " = ", patch_output_buffer_var_name, "[", to_expression(builtin_invocation_id_id),
-							          ".x / ", get_entry_point().output_vertices, "];");
+							          " = ", patch_output_buffer_var_name, "[min(", to_expression(builtin_invocation_id_id),
```
HansKristian-Work, line 3465: The intention is that you have to pad out the gl_out buffer to handle workgroup alignment (and only clamp gl_PrimitiveID). With min() here, you will risk having multiple workgroup each write to the same output block, and cause potential hazards, especially with control point synchronization.

spirv_msl.cpp
```diff
@@ -11796,7 +11796,7 @@ void CompilerMSL::fix_up_shader_inputs_outputs()
 				entry_func.fixup_hooks_in.push_back([=]() {
 					statement(builtin_type_decl(bi_type), " ", to_expression(var_id), " = min(",
 					          to_expression(builtin_invocation_id_id), ".x / ", this->get_entry_point().output_vertices,
-					          ", spvIndirectParams[1]);");
+					          ", spvIndirectParams[1] - 1);");
```
HansKristian-Work, line 11799: This seems like a correct fix.

HansKristian-Work (changes requested):

spirv_msl.cpp
Dredhog: Hmm, so adding the buffers is a possible option, but I could probably come up with example shaders which won't work as expected if the redundant threads are run in the first place e.g. If I'm writing out some data for each control point or each patch to a buffer I'll get excess data written there (not saying that my current solution addresses this, it doesn't)

My intention here was to follow the pattern currently used by SPIRV-Cross as by using `min()` when computing `gl_PrimitiveID` it was already resulting in the out of bounds threads all overwriting a single tessellation factor multiple times.

However, to me it seems like the true solution would be to just add something like `if(gl_GlobalInvocationID.x >= spvIndirectParams[1]*output_point_count) return;` at the top of the function. This would avoid any hazzards as you say, would avoid writing to the last tessellation factor multiple times, would avoid any sideffects of the excess threads and would not require to pad the patch and control point output buffers. Would you have any objections to this?

HansKristian-Work: Yes, returning is also broken since any subsequent barrier() is now undefined behavior.

HansKristian-Work: > buffer I'll get excess data written there

Are you thinking if the redundant threads use global side effects ala atomics on a buffer? I guess that is a problem, but there's generally no rule in most graphics APIs that vertex shaders run only once, so a shader relying on this is a little bit shaky. Tess/Geom side effects is optional I think.

HansKristian-Work: simdgroup barriers would avoid most of the problems here, but there is no guarantee that a simdgroup covers an entire patch (32 control points in worst case).

Dredhog: >Yes, returning is also broken since any subsequent barrier() is now undefined behavior, and will hang the GPU

Oh... Well I was afraid it might be more complicated.

>Are you thinking if the redundant threads use global side effects ala atomics on a buffer?

Yeah, pretty much, I was thinking about things like HLSL append/consume buffers.

>Tess/Geom side effects is optional I think.

I'm not completely sure what you mean right here here

I'm just spitballing here, but couldn't the whole control point section (until the hull point -> hull const barrier) be placed in the opposite if block instead of returning? i.e. something like
```
gl_PrimitiveID = ...
... Prepare patch pointers and temporaries
... Read data into temporary control point structs 
bool out_of_bounds = gl_GlobalInvocationID.x >= spvIndirectParams[1]*output_point_count;
if(!out_of_bounds)
{
... Hull control point processing
}

Barrier

if(!out_of_bounds && gl_Position == 0)
{
.. Hull patch processing
}
```
Would this just be difficult or would it be impossible for some reason?

Dredhog: I would imagine if there's no way for additional barriers to end up in the `!out_of_bounds` branches that this should work, or am I missing something?

HansKristian-Work: > I'm not completely sure what you mean right here here
Basically, at least in Vulkan, being able to store to arbitrary buffers in vertex/tess/geom is not a guaranteed feature, so I would be somewhat surprised if you require that functionality to be available.

> I'm just spitballing here, but couldn't the whole control point section (until the hull point -> hull const barrier) be placed in the opposite if block instead of returning? i.e. something like

The strict HLSL flow of control point -> barrier -> patch constant is a HLSL-ism that does not exist in SPIR-V. I think the idea of keeping all code in conditional control flow and then leave it -> barrier -> go back to conditional control flow could actually work in all sensible scenarios, at least any code HLSL would generate. The edge case is when barrier() itself happens in uniform control flow, but I don't think even that will work right now ... It's not allowed in GLSL either, but I think SPIR-V technically allows it. I wouldn't be against adding a mode like this.

Dredhog: Sorry, haven't gotten around to looking deeper into this, but glad it's not a hard no.

I'll probably undo the changes to the output range index for now (hopefully tomorrow), so that at least the `gl_Position` fix below could go in.

HansKristian-Work: > but glad it's not a hard no

Yes, it should be fine. I'll accept the fix if it's required.

Dredhog: Ok, I've undone these now

Dredhog:
@HansKristian-Work tests are green after removing the change to output indices, feel free to merge when you are ready

HansKristian-Work (approved):

PR #1680, opened 2021-05-26
Add comment after inf/nan float number for clarifying.

xndcn:
Outputs like:
glsl: `float v = uintBitsToFloat(0xff800000u /* -inf */);`
msl: `float v = as_type<float>(0x7f800000u /* inf */);`

### DIFF
--- a/spirv_glsl.cpp
+++ b/spirv_glsl.cpp
@@ -4948,7 +4948,13 @@ string CompilerGLSL::convert_float_to_string(const SPIRConstant &c, uint32_t col
 
 			char print_buffer[32];
 			sprintf(print_buffer, "0x%xu", c.scalar(col, row));
-			res = join(bitcast_glsl_op(out_type, in_type), "(", print_buffer, ")");
+
+			const char *comment = "inf";
+			if (float_value == -numeric_limits<float>::infinity())
+				comment = "-inf";
+			else if (std::isnan(float_value))
+				comment = "nan";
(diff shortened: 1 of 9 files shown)


spirv_glsl.cpp
```diff
@@ -4948,7 +4948,16 @@ string CompilerGLSL::convert_float_to_string(const SPIRConstant &c, uint32_t col
 
 			char print_buffer[32];
 			sprintf(print_buffer, "0x%xu", c.scalar(col, row));
+#ifndef NDEBUG
```
HansKristian-Work, line 4951: Just do this unconditionally.

HansKristian-Work (changes requested):

spirv_glsl.cpp
```diff
@@ -4948,7 +4948,13 @@ string CompilerGLSL::convert_float_to_string(const SPIRConstant &c, uint32_t col
 
 			char print_buffer[32];
 			sprintf(print_buffer, "0x%xu", c.scalar(col, row));
-			res = join(bitcast_glsl_op(out_type, in_type), "(", print_buffer, ")");
+
+			string comment = "inf";
```
HansKristian-Work, line 4952: `const char *`, no need to dynamically allocate the comment here.

HansKristian-Work (changes requested):
Nit

spirv_glsl.cpp
xndcn: Brilliant! Thanks!

HansKristian-Work:
Need to update reference test output as well.

HansKristian-Work (approved):

PR #1689, opened 2021-06-08
MSL:  convert depth buffer fetch to regular texture read.

atyuwen:
Metal doesn't support depth buffer fetch, so here gives a chance to convert those subpassLoads from depth buffer to regular texture reads.

### DIFF
--- a/spirv_cross.cpp
+++ b/spirv_cross.cpp
@@ -4163,6 +4163,8 @@ void Compiler::analyze_image_and_sampler_usage()
 
 	comparison_ids = move(handler.comparison_ids);
 	need_subpass_input = handler.need_subpass_input;
+	need_subpass_color_input = handler.need_subpass_color_input;
+	need_subpass_depth_input = handler.need_subpass_depth_input;
 
 	// Forward information from separate images and samplers into combined image samplers.
 	for (auto &combined : combined_image_samplers)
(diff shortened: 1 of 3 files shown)


HansKristian-Work:
I don't really understand what this is trying to solve. Please add some tests.

cdavis5e:
> I don't really understand what this is trying to solve.

I suspect it's for when a depth `SubpassInput` is used with `use_framebuffer_fetch_subpasses`. Metal can't yet do framebuffer fetch with a D/S attachment.

atyuwen:
> > I don't really understand what this is trying to solve.
> 
> I suspect it's for when a depth `SubpassInput` is used with `use_framebuffer_fetch_subpasses`. Metal can't yet do framebuffer fetch with a D/S attachment.

Exactly. Unfortunately in standard Vulkan GLSL there is no way to distinguish whether a `SubpassInput` is depth or not. This pr currently uses `SPIRType.image.depth` to specify a depth `SubpassInput`, and assumes it's the SPIR-V codegen's responsibility to set the `SPIRType.image.depth` field correctly.

However, the `SPIRType.image.depth`  field is always `false` if the SPIR-V is generated by Glslang, which makes it hard to write a unit test. I'm considering whether it is better to add an interface like `CompilerMSL::mark_depth_subpass_input(VariableID id)` to specify a depth `SubpassInput`, though the implementation will be slightly more complicated.

HansKristian-Work:
Subpass attachments in Vulkan do not know if the attachment is depth or not, so yes, we will need some way to mark if an input_attachment_location is depth or not.

atyuwen:
Ok, finally I realized it might be an improper feature because subpassInput can be passed as function arguments. for example, the following function can be called with a color subpassInput and also depth subpassInput.
```
vec4 load_subpasses(mediump subpassInput uInput)
{
	return subpassLoad(uInput);
}
```
To handle it correctly, the compiler must generate 2 overloads: one for color subpassInput and one for depth subpassInput respectively, which makes things too complicated.

PR #1692, opened 2021-06-21
Fix IVariant -Wdeprecated-copy-with-dtor

Kangz:
Since C++11 the generation of implicit copy constructor is deprecated if
the type has a user-defined constructor or destructor (even if defaulted).

Fix this by adding a defaulted copy-constructor for IVariant that remove
the implicit default constructor, so add it too.

### DIFF
--- a/spirv_common.hpp
+++ b/spirv_common.hpp
@@ -417,6 +417,11 @@ struct IVariant
 	virtual ~IVariant() = default;
 	virtual IVariant *clone(ObjectPoolBase *pool) = 0;
 	ID self = 0;
+
+protected:
+	IVariant() = default;
+	IVariant(const IVariant&) = default;
+	IVariant &operator=(const IVariant&) = default;
 };
 
 #define SPIRV_CROSS_DECLARE_CLONE(T)                                \


spirv_common.hpp
```diff
@@ -414,7 +414,11 @@ using ID = TypedID<TypeNone>;
 // Helper for Variant interface.
 struct IVariant
 {
+	IVariant() = default;
 	virtual ~IVariant() = default;
+	IVariant(const IVariant&) = default;
```
HansKristian-Work, line 419: It feels very wrong to copy a virtual interface. It must use virtual clone(). Where is the copy constructor actually triggered?

Kangz: This is the error I got:

```
In file included from ../../third_party/vulkan-deps/spirv-cross/src/spirv_cross.hpp:28:
In file included from ../../third_party/vulkan-deps/spirv-cross/src/spirv_cfg.hpp:27:
../../third_party/vulkan-deps/spirv-cross/src/spirv_common.hpp:418:10: error: definition of implicit copy constructor for 'IVariant' is deprecated because it has a user-declared destructor [-Werror,-Wdeprecated-copy-with-dtor]
        virtual ~IVariant() = default;
                ^
../../third_party/vulkan-deps/spirv-cross/src/spirv_common.hpp:430:8: note: in implicit copy constructor for 'spirv_cross::IVariant' first required here
struct SPIRUndef : IVariant
       ^
../../third_party/vulkan-deps/spirv-cross/src/spirv_cross_containers.hpp:579:13: note: in implicit copy constructor for 'spirv_cross::SPIRUndef' first required here
                new (ptr) T(std::forward<P>(p)...);
                          ^
../../third_party/vulkan-deps/spirv-cross/src/spirv_common.hpp:443:2: note: in instantiation of function template specialization 'spirv_cross::ObjectPool<spirv_cross::SPIRUndef>::allocate<spirv_cross::SPIRUndef &>' requested here
        SPIRV_CROSS_DECLARE_CLONE(SPIRUndef)
        ^
../../third_party/vulkan-deps/spirv-cross/src/spirv_common.hpp:427:46: note: expanded from macro 'SPIRV_CROSS_DECLARE_CLONE'
                return static_cast<ObjectPool<T> *>(pool)->allocate(*this); \
                                                           ^
1 error generated.
```

HansKristian-Work: I see, so it's required implicitly by a base class doing the construction. I guess it would be better then to make the default copy (dtors) `protected:` then? I don't want copy constructors being called here except from clone().

Kangz: Sounds good, I'll do that.

HansKristian-Work (approved):

Kangz:
Thank you!

PR #1703, opened 2021-07-01
[MSL] Fix as_type casts being used to bitcast to different sizes

GenuineAster:
SPIRV-Cross incorrectly generates MSL `as_type` casts for source/target types that have different sizes, which is illegal in MSL.

We were encountering this issue with expressions such as `foo = uint64_t(my_64bit_int | my_32bit_int);` which would generate an `as_type` cast for `my_32bit_int` to a 64-bit integer.

Please let me know if there is a better solution to this issue, and how I can add a regression test for this issue.

### DIFF
--- a/spirv_msl.cpp
+++ b/spirv_msl.cpp
@@ -13898,15 +13898,15 @@ string CompilerMSL::bitcast_glsl_op(const SPIRType &out_type, const SPIRType &in
 	bool integral_cast = type_is_integral(out_type) && type_is_integral(in_type);
 	bool same_size_cast = out_type.width == in_type.width;
 
-	if (integral_cast && same_size_cast)
+	if (same_size_cast)
 	{
 		// Trivial bitcast case, casts between integers.
-		return type_to_glsl(out_type);
+		return "as_type<" + type_to_glsl(out_type) + ">";
 	}
 	else
 	{
 		// Fall back to the catch-all bitcast in MSL.
-		return "as_type<" + type_to_glsl(out_type) + ">";
+		return type_to_glsl(out_type);
 	}
 }
 


HansKristian-Work:
How would you handle the case of bitcasting say `u16vec2` <-> `uint`? Isn't that the point of `as_type<T>`

GenuineAster:
> How would you handle the case of bitcasting say `u16vec2` <-> `uint` then? Isn't that the point of `as_type<T>`?

Shouldn't that work with the current code? They have the same width so it should use `as_type`.

cdavis5e:
> > How would you handle the case of bitcasting say `u16vec2` <-> `uint` then? Isn't that the point of `as_type<T>`?
> 
> Shouldn't that work with the current code? They have the same width so it should use `as_type`.

I'm afraid that `width` member may be a bit confusing. It's the width of the *element* type, not the type as a whole. So for an `ivec4`, you'd expect it to have a `width` of 128, but `width` is really 32, because the element type is `int`.

GenuineAster:
Ah ok! I'll fix that.

HansKristian-Work:
When does SPIRV-Cross actually hit this case at all? OpBitcast of different sizes is not legal to begin with.

GenuineAster:
It's the case I wrote in the PR description: `foo = uint64_t(my_64bit_int | my_32bit_int);`

HansKristian-Work:
> It's the case I wrote in the PR description: `foo = uint64_t(my_64bit_int | my_32bit_int);`

Do you have a reproducer here? This seems to work fine for me at least:

```
#version 450
#extension GL_ARB_gpu_shader_int64 : require
layout(local_size_x = 1) in;

layout(set = 0, binding = 0) buffer SSBO
{
	uint64_t v;
	int a;
	int64_t b;
};

void main()
{
	v = uint64_t(a | b);
}
```

```
#include <metal_stdlib>
#include <simd/simd.h>

using namespace metal;

struct SSBO
{
    ulong v;
    int a;
    long b;
};

constant uint3 gl_WorkGroupSize [[maybe_unused]] = uint3(1u);

kernel void main0(device SSBO& _11 [[buffer(0)]])
{
    _11.v = ulong(long(_11.a) | _11.b);
}
```

GenuineAster:
Hi! Sorry for the delay, I haven't had much time to turn on my Mac lately but I will try to move this forward this week or next and provide a complete use case!

billhollings:
Hmmm...I was unaware of this outstanding PR and it looks like I dealt with this issue in #1724 as well. 

The check in #1724 might be a little bit more robust than the check here.

billhollings:
I've added a SPIRV-Cross test shader to #1724 to illustrate and test the fix.

HansKristian-Work:
Closed as duplicate now.

PR #1709, opened 2021-07-14
MSL: Adjust gl_SampleMaskIn for sample-shading and/or fixed sample mask.

billhollings:
Vulkan specifies that the Sample Mask Test occurs before fragment shading.
This means `gl_SampleMaskIn` should be influenced by both sample-shading and
`VkPipelineMultisampleStateCreateInfo::pSampleMask`.

CTS tests `dEQP-VK.pipeline.multisample_shader_builtin.*` bear this out.

For sample-shading, `gl_SampleMaskIn` should only have a single bit set,
Since Metal does not filter for this, apply a bitmask based on gl_SampleID.

For a fixed sample mask, since Metal is unaware of
`VkPipelineMultisampleStateCreateInfo::pSampleMask`, we need to ensure that
we apply it to both `gl_SampleMaskIn` and `gl_SampleMask`. This has the side
effect of a redundant application of `pSampleMask` if the shader already
includes `gl_SampleMaskIn` when setting `gl_SampleMask`, but I don't see an
easy way around this.

Also, simplify the logic for including the fixed sample mask in `gl_ShaderMask`,
and print the fixed sample mask as a hex value for readability of bits.

### DIFF
--- a/reference/opt/shaders-msl/frag/sample-mask-in-and-out.fixed-sample-mask.force-sample.frag
+++ b/reference/opt/shaders-msl/frag/sample-mask-in-and-out.fixed-sample-mask.force-sample.frag
@@ -9,11 +9,11 @@ struct main0_out
     uint gl_SampleMask [[sample_mask]];
 };
 
-fragment main0_out main0()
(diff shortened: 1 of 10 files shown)


HansKristian-Work (approved):
LGTM

PR #1724, opened 2021-08-12
MSL: Fix casting in constant expressions with different sizes.

billhollings:
Previous casting in constant expressions allowed `as_type<>` between
types of different overall sizes.
Add check for overall size (width * vecsize) to ensure `as_type<>` will work,
otherwise use regular cast. Also beef up test of integer values to also check
vecsize, and use regular casts for those.

TBH...I'm not sure why we always go to bitcasting when emitting constant expressions. It seems most constant expressions would not have the need for bitcasting. Here I have just done the minimal surgery required to fix the issue at hand, which was to avoid using `as_type<>` when say casting a 64-bit expression value to a 32-bit result, etc.


HansKristian-Work (approved):
LGTM

PR #1725, opened 2021-08-16
MSL: Fix duplicate gl_Position outputs when gl_Position defined but unused.

billhollings:
When `gl_Position` is defined by SPIR-V, but neither used nor initialized, it appeared twice in the MSL output, as `gl_Position` and `glPosition_1`.

The existing tests for whether an output is active check only that it is used by an op, or initialized. Adding the implicit `gl_Position` also marked the existing `gl_Position` as active, duplicating the output variable.

Fix is that when checking for the need to add an implicit `gl_Position` output, also check if the var is already defined in the shader, and just needs to be marked as active.
Add test shader.

### DIFF
--- a/reference/opt/shaders-msl/vert/unused-position.vert
+++ b/reference/opt/shaders-msl/vert/unused-position.vert
@@ -1,0 +1,18 @@
+#include <metal_stdlib>
+#include <simd/simd.h>
+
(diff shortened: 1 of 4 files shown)


HansKristian-Work (approved):
LGTM

PR #1735, opened 2021-09-03
Add support for additional ops in OpSpecConstantOp.

billhollings:
MSL: Support op `OpQuantizeToF16` in `OpSpecConstantOp`.
All: Support op `OpSRem` in `OpSpecConstantOp`.

I was not able add new tests, as these would require SPIR-V Assembly shaders, which I'm not familiar with creating.


cdavis5e:
I have some experience with that. Hang on a moment...

cdavis5e:
There is an existing test, `shaders-msl/asm/vert/spec-constant-op-composite.asm.vert`. I have attached a patch which adds some tests to that file.
[spec-const-tests.diff.txt](https://github.com/KhronosGroup/SPIRV-Cross/files/7109150/spec-const-tests.diff.txt)

billhollings:
>I have attached a patch which adds some tests to that file.

Thanks very much! Tests added to PR.

HansKristian-Work (approved):
LGTM

PR #1738, opened 2021-09-12
MSL: Return fragment function value even when last SPIR-V Op is discard (OpKill).

billhollings:
The terminator Op in a fragment function can be an `OpKill`. In GLSL, where the function returns `void`, this can simply finish the function, but in MSL, this can happen even when the fragment function expects to return a value. So for MSL, if needed, return the value on an `OpKill` that terminates the final entry function block.

Add test shader for new functionality.
Add legacy test reference shader for unrelated buffer-bitcast test, that doesn't seem to have been added previously.

### DIFF
--- a/spirv_msl.cpp
+++ b/spirv_msl.cpp
@@ -3467,6 +3467,7 @@ uint32_t CompilerMSL::add_interface_block(StorageClass storage, bool patch)
 		// Add the output interface struct as a local variable to the entry function.
 		// If the entry point should return the output struct, set the entry function
 		// to return the output interface struct, otherwise to return nothing.
+		// Watch out for the rare case where the terminator of the last entry point block is a Kill, instead of a Return.
 		// Indicate the output var requires early initialization.
 		bool ep_should_return_output = !get_is_rasterization_disabled();
 		uint32_t rtn_id = ep_should_return_output ? ib_var_id : 0;
@@ -3476,7 +3477,7 @@ uint32_t CompilerMSL::add_interface_block(StorageClass storage, bool patch)
 			for (auto &blk_id : entry_func.blocks)
 			{
 				auto &blk = get<SPIRBlock>(blk_id);
-				if (blk.terminator == SPIRBlock::Return)
+				if (blk.terminator == SPIRBlock::Return || (blk.terminator == SPIRBlock::Kill && blk_id == entry_func.blocks.back()))
 					blk.return_value = rtn_id;
 			}
 			vars_needing_early_declaration.push_back(ib_var_id);
--- a/spirv_msl.cpp
+++ b/spirv_msl.cpp
@@ -3467,7 +3467,9 @@ uint32_t CompilerMSL::add_interface_block(StorageClass storage, bool patch)
 		// Add the output interface struct as a local variable to the entry function.
 		// If the entry point should return the output struct, set the entry function
 		// to return the output interface struct, otherwise to return nothing.
-		// Watch out for the rare case where the terminator of the last entry point block is a Kill, instead of a Return.
+		// Watch out for the rare case where the terminator of the last entry point block is a
+		// Kill, instead of a Return. Based on SPIR-V's block-domination rules, we assume that
+		// any block that has a Kill will also have a terminating Return, except the last block.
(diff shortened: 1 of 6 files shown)


spirv_msl.cpp
```diff
@@ -3476,7 +3477,7 @@ uint32_t CompilerMSL::add_interface_block(StorageClass storage, bool patch)
 			for (auto &blk_id : entry_func.blocks)
 			{
 				auto &blk = get<SPIRBlock>(blk_id);
-				if (blk.terminator == SPIRBlock::Return)
+				if (blk.terminator == SPIRBlock::Return || (blk.terminator == SPIRBlock::Kill && blk_id == entry_func.blocks.back()))
```
HansKristian-Work, line 3480: Looking for last block seems a bit iffy, but I suppose this exploits the SPIR-V rule where a block must be declared before a block it dominates?

How are semantics of discard_fragment() in MSL? Is it demote semantics and not actually kill?

billhollings: >Looking for last block seems a bit iffy, but I suppose this exploits the SPIR-V rule where a block must be declared before a block it dominates?

Indeed. I was not sure about this approach, as I don't have a complete understanding on how SPIRV-Cross organizes blocks relative to control flow within a function.

In `CompilerGLSL::emit_block_chain(SPIRBlock &block)`, where the `return_value` is consumed, to determine the final terminating block, the `case SPIRBlock::Return:` uses the approach:

    if (!cfg.node_terminates_control_flow_in_sub_graph(current_function->entry_block, block.self) ||
        block.loop_dominator != BlockID(SPIRBlock::NoDominator))

which might be more accurate. If that's the case, can we use that approach in the code above? If not, perhaps we could add it to `CompilerGLSL::emit_block_chain(SPIRBlock &block)` in the `case SPIRBlock::Kill:`.

billhollings: >How are semantics of discard_fragment() in MSL? Is it demote semantics and not actually kill?

MSL `discard_fragment()` used to behave as a kill, but now it behaves as a demote-to-helper. From the current MSL spec:

>Multiple fragment threads or helper threads associated with a fragment thread execute together to compute derivatives. Prior to Metal 2.3, if any (but not all) of these threads executes the discard_fragment function, the thread is terminated and the behavior of any derivative computations (explicit or implicit) is undefined. Since Metal 2.3, discard_fragment marks the fragment is terminated while continuing to execute in parallel to make derivative computations well defined. Even though execution continues, the write behavior remains the same as before. The fragment will discard the fragment output and discard all writes to buffer or texture after discard_fragment.

HansKristian-Work: node_terminates_control_flow_in_sub_graph is used to essentially check if we are the outermost scope. I think the last block thing should work, but it needs comments explaining the rationale behind this code since it is a little weird.

> MSL discard_fragment() used to behave as a kill, but now it behaves as a demote-to-helper. From the current MSL spec:

Fun. I suppose we only return after discard_fragment() to satisfy the compiler then? What if the last block of a leaf function calls discard_fragment() (or demote), would that also require workarounds like these?

billhollings: >I suppose we only return after discard_fragment() to satisfy the compiler then?

Yes. In the test shader I included, we got caught because the shader sets a member in the output variable, so we expect to return a value, but the Kill without return caused us not to emit `return out;` when encountered. And that of course caused a standard _missing return value in non-void function_ error in the MSL compiler.

I did think about forcing the function to be a `void` function once the Kill was encountered, but figured there could be code paths within a shader that might make a return value required.

>the last block thing should work, since OpKill is a terminating instruction, 

Okay. I assumed any blocks other than the last would need to contain a Return command in addition to a Kill, but as I mentioned, I wasn't entirely sure about the organization of block flow.

> but it needs comments explaining the rationale behind this code since it is a little weird.

Okay. I'll add some additional comments there.

billhollings: > I'll add some additional comments there.

Done.

billhollings:
Not sure what's going on with the CI builds.

HansKristian-Work (approved):
LGTM

PR #1739, opened 2021-09-13
HLSL: added support for domain shaders (aka tesselation evaluation)

Try:
+ Added implementation to cover most of use cases
+ Added 4 testes (copied from MSL suite)
TODO: add support for I/O blocks
#906


Try:
Some info about overall approach:
Solution is largely bases on what spirv-cross is already doing for metal-msl: OutputPatch<> is forwarded to every function in HLSL code. Meanwhile per-patch variables are done in same way as all varying were handled before.

Testing:
Unfortunately spirv-cross test system doesn't work on windows robustly (even HEAD of master fails to pass), so I've used units tests in my game engine.
Test cases far should 'just work', but issues with white spaces may arise.

HansKristian-Work:
I'll look at this eventually ...

HansKristian-Work:
What is the motivation to implement only evaluation stage, without domain shaders (which is the really difficult part of HLSL translation)?

HansKristian-Work:
> Unfortunately spirv-cross test system doesn't work on windows robustly (even HEAD of master fails to pass), so I've used units tests in my game engine.

How are you running the tests, and which toolchain? The tests pass Windows MSVC CI here.

Try:
> How are you running the tests, and which toolchain?

MinGW 8.1.0 64-bit (installed as part of Qt toolkit, if that important)
environment: MSYS 1.0
host: Windows10
 
There are a minor issues with script to begin with:
`./test_shaders.sh` doesn't work as-is, so I had to specify python expicitly:
`python ./test_shaders.py shaders ${OPTS} --spirv-cross "$SPIRV_CROSS_PATH"` 

And then, with the actual testing:
```bash
...
Testing shader: shaders\asm\frag\hlsl-sample-cmp-level-zero-cube.asm.frag
C:\Users\ALEXAN~1\AppData\Local\Temp\tmpf_5v02uuvkhlsl-sample-cmp-level-zero-cube.asm.frag
Reference shader path: reference/shaders\asm\frag\hlsl-sample-cmp-level-zero-cube.asm.frag
Generated source code in C:\Users\ALEXAN~1\AppData\Local\Temp\tmppz_586q0hlsl-sample-cmp-level-zero-cube.asm.frag does not match reference reference/shaders\asm\frag\hlsl-sample-cmp-level-zero-cube.asm.frag!

Generated:
======================
#version 450

uniform samplerCubeShadow SPIRV_Cross_CombinedpointLightShadowMapshadowSamplerPCF;

layout(location = 0) out float _entryPointOutput;

float _main()
{
    vec4 _33 = vec4(0.10000000149011612, 0.10000000149011612, 0.10000000149011612, 0.5);
    return textureGrad(SPIRV_Cross_CombinedpointLightShadowMapshadowSamplerPCF, vec4(_33.xyz, _33.w), vec3(0.0), vec3(0.0));
}

void main()
{
    _entryPointOutput = _main();
}


======================

Error: Does not match reference
```

On line 
`vec4 _33 = vec4(0.10000000149011612, 0.10000000149011612, 0.10000000149011612, 0.5);` the generated code has more digits:
`vec4 _33 = vec4(0.100000001490116119384765625, 0.100000001490116119384765625, 0.100000001490116119384765625, 0.5);`

Try:
> What is the motivation to implement only evaluation stage, without hull shaders (which is the really difficult part of HLSL translation)?

Implementing all together is expected to be too large of commit to review, so I decide to split thing up.
Also I don't have great ideas on how to approach HULL at this moment, most likely 100% is not going to be possible, but we can work on partial solution to cover common cases.

Idea I had so far is to detect pattern, when glsl code:
1. Writes out varying's like `var_out[gl_InvocationID] = ***`
2. Writes per-patch data like `if(gl_InvocationID==constant) patch_out = ...`
3. No `barrier` and shared-memory access in glsl code (not supported in `hs_5_0`)
In that case we can split main function into hlsl-main+path function more-or-less cleanly

HansKristian-Work:
I do not want to commit to supporting anything tessellation until we have a solid idea how we're going to do it well. Especially TESC is fiendish due to the various problems of:

- Where does control point phase begin?
- Where does patch phase begin?
- How are non-IO data passed from control point to patch phase?
- How to correctly handle barrier()?
- Pretty sure it is 100% impossible to support full SPIR-V, since we get one implicit barrier() in HLSL (control point -> patch phase), but SPIR-V can have multiple. How to robustly detect these failure cases? Can we actually support a useful subset of shaders that engines emit?

These problems contribute to why tess support has not happened yet, it gets mind meltingly complex. It's complicated enough, that it cannot start with hammering out code, this needs to be well thought out and designed first.

Try:
> I do not want to commit to supporting anything tessellation until we have a solid idea how we're going to do it well

I see your point.
Let me take a time to build some raw prototype if TESC, so we can continue talk about what is possible to hanle and what is not.

> Can we actually support a useful subset of shaders that engines emit?

If we would be able to handle SPIR-V code generated from HLSL, then it should be good enough, I think.

Try:
Started a new PR: https://github.com/KhronosGroup/SPIRV-Cross/pull/1904
(this branch is too outdated - so it was a hassle to rebase it)

Issue #1599, opened 2021-01-19
Text representation of float literals is not-minimal


baryluk:
Example:

```glsl
#version 460

layout (location = 0) out vec4 fragColor;

void main()
{
	fragColor = vec4(0.4, 0.4, 0.8, 1.0);
}
```

-> glslang from trunk into Vulkan 1.0 SPIR-V.
-> SPIRV-Cross from trunk into HLSL SM 50

```hlsl
static float4 fragColor;

struct SPIRV_Cross_Output
{
    float4 fragColor : SV_Target0;
};

void frag_main()
{
    fragColor = float4(0.4000000059604644775390625f, 0.4000000059604644775390625f, 0.800000011920928955078125f, 1.0f);
}

SPIRV_Cross_Output main()
{
    frag_main();
    SPIRV_Cross_Output stage_output;
    stage_output.fragColor = fragColor;
    return stage_output;
}
```

This is ugly, annoying, and makes debugging a bit harder.

0.4000000059604644775390625 - 0.4  (in double precision) = 5.960464455334602e-09

relative error: 5.960464455334602e-09 / 0.4 (in double precision) = 1.4901161138336505e-08

That is below what is representable in 32-bit single precision float.

The single precision number has 23 bit of mantissa, so the last bit represents 2^{-23} ≈ 1.1920928955078125e-07, which is more than the relative error of these two literals.

I believe `0.4000000059604644775390625f` and `0.4f`, parse to the same exact single precission number.

It looks like there is a fixed accuracy somewhere in SPIRV-Cross, maybe something like `"%.25g"`, where for for floats, "%.8g" is probably enough for not denormalized numbers. A more smart approach is to use better algorithms, which are known around (but usually not exposed in C or C++ standard libraries).

HansKristian-Work:
I would prefer a better algorithm since it is indeed ugly, but I don't know how to write one and it's not available in C++11 standard library.

baryluk:
The Ryū algorithm by Ulf Adams is usually used these days to do this. Sometimes an older grisu2 algorithm is used.

Some Ryu references:

 * https://github.com/ulfjack/ryu
 * Ulf Adams. 2019. Ryū revisited: printf floating point conversion. Proc. ACM Program. Lang. 3, OOPSLA, Article 169 (October 2019), 23 pages. DOI:https://doi.org/10.1145/3360595
 * https://doi.org/10.5281/zenodo.3366212
[...]

PR #1743, opened 2021-09-17
Introduce Ryu for float-point literal generating.

chirsz-ever:
Try to solve #1599 . The Ryu sources is from https://github.com/ulfjack/ryu .

I don't think this is a breaking change, because the float-point literals generated before and after this PR have the same binary form.

Something I can't determine:
- The current (73a0ec4480daaf91f5837b7c997f941b78aff08c) code always generate scientific format, would this be pretty?
- Take a choose between `ryu/d2s_full_table.h` and `ryu/d2s_small_table.h`. The `ryu/d2s_small_table.h` is slower, but with a smaller size.
- How to deal with the `SPIRV_CROSS_FLT_FMT` micro?


chirsz-ever:
Oh, Visual Studio 12 does not support C99 ...

HansKristian-Work:
One immediate question is licensing. Is this taking Ryu's official source code? The reason I'm asking this is because the CLA has some requirements w.r.t. ownership.

SPIRV-Cross is also dual licensed outbound with MIT + Apache 2, and I'm not sure if we can re-license code that has not been contributed without the original author's consent.

chirsz-ever:
I did not notice `LICENSE/MIT.txt`, maybe it's better to mention it on `README.md`?

Yes, the sources in `ryu` is directly copied from Ryu's official source code.

Technically the license is just be applied to your own source code of which you have the entire copyright. The licenses of third parties' are simply passed down.

Or you can choose to use Ryu under [the boost license](https://github.com/ulfjack/ryu/blob/master/LICENSE-Boost), which allow the License files to be ignored if just distributing binary products:

> The copyright notices in the Software and this entire statement, including
the above license grant, this restriction and the following disclaimer,
must be included in all copies of the Software, in whole or in part, and
all derivative works of the Software, unless such copies or derivative
works are solely in the form of machine-executable object code generated by
a source language processor.

This PR did not pass the tests, and may not obey the section 8 of CLA, so I will close it.

PR #1744, opened 2021-09-20
MSL: Fix location and component variable matching between shader stages.

billhollings:
Consolidate derivation of Metal `user(locnL_C)` output/input location
and component attribute qualifier, to establish SVOT across stages.

The `comp != 0` test was not performed uniformly between inputs and outputs, resulting in linkage breakage between attribute qualifiers `user(locn0_0)` for outputs and `user(locn0)` for inputs when component is zero.

I've consolidated to `user(locn0)` for the zero component to make it visibly simpler for the common case where no component is provided, or is assumed to be zero.

Since this was a consolidation, I did not include any new test cases as this was a simple change, and both inputs and outputs now perform identically and any subsequent changes that alters that will then break current tests.

### DIFF
--- a/spirv_msl.cpp
+++ b/spirv_msl.cpp
@@ -10666,15 +10666,9 @@ string CompilerMSL::member_attribute_qualifier(const SPIRType &type, uint32_t in
 				return "";
 			}
 		}
-		uint32_t comp;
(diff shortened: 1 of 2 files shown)


HansKristian-Work (approved):
Nice catch, LGTM.

PR #1745, opened 2021-09-20
MSL: Track location component to match vecsize between shader stages.

billhollings:
Matching output/input struct member types between shader stages could fail if
a location is shared between members, each using different components of that
location, because the member vecsize was only stored once for the location.

Add `MSLShaderInput::component` member.
Use `LocationComponentPair` to key `inputs_by_location`, instead of just location.
`ensure_correct_input_type()` pass component value as well as location.

Any tests would require ability to pass location and component info. Not sure how to do that.

@cdavis5e Inviting your review here regarding impact on tessellation stages.

### DIFF
--- a/spirv_msl.cpp
+++ b/spirv_msl.cpp
@@ -58,7 +58,7 @@ CompilerMSL::CompilerMSL(ParsedIR &&ir_)
 
 void CompilerMSL::add_msl_shader_input(const MSLShaderInput &si)
 {
-	inputs_by_location[si.location] = si;
+	inputs_by_location[{si.location, si.component}] = si;
 	if (si.builtin != BuiltInMax && !inputs_by_builtin.count(si.builtin))
 		inputs_by_builtin[si.builtin] = si;
 }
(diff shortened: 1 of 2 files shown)


spirv_msl.cpp
```diff
@@ -3602,7 +3606,7 @@ uint32_t CompilerMSL::add_interface_block(StorageClass storage, bool patch)
 		// the struct containing them is the correct size and layout.
 		for (auto &input : inputs_by_location)
 		{
-			if (location_inputs_in_use.count(input.first) != 0)
+			if (location_inputs_in_use.count(input.first.location) != 0)
```
HansKristian-Work, line 3609: Here we lose component information. is_msl_shader_input_used() is missing a component member as well.

HansKristian-Work (changes requested):
Comment

spirv_msl.cpp
HansKristian-Work: To stay compatible I think we'd need two entry points. One where we return true if the Location is used at all (any component is fine) and one where we specify the matching Component as well, i.e. a narrower match.

HansKristian-Work:
CI runs should be fixed if you rebase against master. Github Actions nuked Ubuntu 16.04 support.

spirv_msl.cpp
billhollings: Hmmm...this was actually deliberate. 

I took the perspective that location is the lowest granularity of independence, analogous to a pixel, and if any component of a location is used, the entire location is considered used. Consequently, I didn't see a need to track usage below the level of a location.

But my assumptions might be wrong. Is there a usecase to track component usage?

HansKristian-Work: For tessellation inputs where we need to synthesize fake inputs, we'd need to know at a per-variable granularity, and that would require per-component granularity usage.

HansKristian-Work: I'm not 100% sure, but if you don't see a problem with this on your end in MoltenVK, I guess I can leave it for now. Can always be fixed later.

HansKristian-Work (approved):

PR #1746, opened 2021-09-22
MSL: Selectively enable fast-math in MSL code to match Vulkan CTS results.

billhollings:
Based on CTS testing, math optimizations between MSL and Vulkan are inconsistent.
In some cases, enabling MSL's fast-math compilation option matches Vulkan's math
results. In other cases, disabling it does. Broadly enabling or disabling fast-math
across all shaders results in some CTS test failures either way.

To fix this, selectively enable/disable fast-math optimizations in the MSL code,
using `metal::fast` and `metal::precise` function namespaces, where supported, and
the `[[clang::optnone]]` function attribute otherwise.

Adjust SPIRV-Cross unit test reference shaders to accommodate these changes.


HansKristian-Work (approved):
This feels a little bit magical, but fast math is always black magic anyways.

PR #1749, opened 2021-09-23
MSL: Honor infinities in OpQuantizeToF16 when compiling using fast-math.

billhollings:
Add `spvQuantizeToF16()` family of synthetic functions to convert from float to half and back again, and add function attribute `[[clang::optnone]]` to honor infinities during conversions.

Adjust SPIRV-Cross unit test reference shaders to accommodate these changes.

### DIFF
--- a/spirv_msl.cpp
+++ b/spirv_msl.cpp
@@ -5006,9 +5006,7 @@ void CompilerMSL::emit_custom_functions()
 			// SpvHalfTypeSelector is used to match the half* template type to the float* template type.
 			statement("template <typename F> struct SpvHalfTypeSelector;");
 			statement("template <> struct SpvHalfTypeSelector<float> { public: using H = half; };");
-			statement("template <> struct SpvHalfTypeSelector<float2> { public: using H = half2; };");
-			statement("template <> struct SpvHalfTypeSelector<float3> { public: using H = half3; };");
-			statement("template <> struct SpvHalfTypeSelector<float4> { public: using H = half4; };");
+			statement("template<uint N> struct SpvHalfTypeSelector<vec<float, N>> { using H = vec<half, N>; };");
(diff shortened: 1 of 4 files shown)


spirv_msl.cpp
```diff
@@ -5002,6 +5002,33 @@ void CompilerMSL::emit_custom_functions()
 			statement("");
 			break;
 
+		case SPVFuncImplQuantizeToF16:
+			// Ensure fast-math is disabled to match Vulkan results.
+			statement("[[clang::optnone]] float spvQuantizeToF16(float val)");
```
HansKristian-Work, line 5007: Can't this be made into a template?

HansKristian-Work (changes requested):
Should be templated, otherwise fine.

spirv_msl.cpp
billhollings: I've made this into a template now. It's still a bit messy because of the need for a separate struct template to align the `half*` with the `float*` types, but it was the most suitable option I could come up with for accomplishing that. I'm happy to hear alternate approaches to handling that mapping.

I had originally tried to come up with a template, but gave up because of that. Thanks for the prod to force me to have a second look at the problem.

cdavis5e: Partial specialization with an NTTP:
```c++
template<typename F> struct spvHalfTypeSelector;
template<uint N> struct spvHalfTypeSelector<vec<float, N>> { using H = vec<half, N>; };
```
This should make it a little less messy.

billhollings: Thanks for thinking of `vec<T, n>`. It's a clever solution for n > 1, and saves a couple of template lines.

HansKristian-Work (approved):
Insane template hackery, but LGTM.

PR #1752, opened 2021-09-28
MSL: Support synthetic functions in function constants.

billhollings:
- Emit synthetic functions before function constants.
- Support use of `spvQuantizeToF16()` in function constants for numerical behavior consistency with the op code.
- Ensure subnormal results from `OpQuantizeToF16` are flushed to zero per SPIR-V spec.
- Adjust SPIRV-Cross unit test reference shaders to accommodate these changes. Any MSL reference shader that includes a synthetic function is affected, since the location it is emitted has changed.

Moving the synthetic functions before function constants also consolidates them with synthetic templates, and consolidates function constants with resource struct definitions, all of which visually cleans up the MSL code layout structure.

This PR resolves all outstanding CTS OpQuantizeToF16 failures.

### DIFF
--- a/reference/opt/shaders-msl/asm/comp/atomic-decrement.asm.comp
+++ b/reference/opt/shaders-msl/asm/comp/atomic-decrement.asm.comp
@@ -7,17 +7,17 @@
 
 using namespace metal;
 
-struct u0_counters
-{
-    uint c;
-};
-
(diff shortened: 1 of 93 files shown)


HansKristian-Work (approved):
LGTM

PR #1757, opened 2021-09-30
Improve handling of INT_MIN/INT64_MIN literals.

### LINKED ISSUE #1754: MSL: min int32 value incorrectly represented, causing error
SPIRV-Cross emits a Metal shader where -2147483648 is used for bit operations:

http://shader-playground.timjones.io/8387d8f36e0b71a4b26fd1bb4c29e96b

-2147483648 can be correctly represented as a SPIR-V OpConstant of type `int`. However, in source code, the numeric literal is first interpreted as a 32-bit int and then negated (see, f.i. GLSL 4.60.7 spec p. 30). Since in the case of -2147483648 the maximum positive value of 2147483647 is exceeded, the value is promoted to `long`.

To fix this, SPIRV-Cross could encode the value with the hex constant 0x80000000.

----------------------

Offline example:

msl_error.hlsl
```hlsl
[[vk::binding(0)]]
RWStructuredBuffer<int> in_data;
[[vk::binding(1)]]
RWStructuredBuffer<float> out_data;

[numthreads(32, 1, 1)]
void CSMain()
{
  out_data[0] = asfloat(in_data[0] & 0x80000000);
}
```

comp.sh
```sh
dxc -E CSMain -T cs_6_6 -spirv -fspv-target-env=vulkan1.1 msl_error.hlsl -Fo msl_error.spv
#glslangValidator -D -e CSMain -S comp --target-env vulkan1.1 -o msl_error.spv msl_error.hlsl
spirv-cross msl_error.spv --msl --output msl_error.metal
xcrun -sdk macosx metal -std=osx-metal1.1 -c -o msl_error.air msl_error.metal
```

----------------------

Vulkan SDK 1.2.189.0
Apple LLVM version 31001.160 (metalfe-31001.160.1)
MacOS Catalina 10.15.7

HansKristian-Work:
We cannot naively convert these to decimal literals. C/C++ (and thus
MSL) has extremely awkward literal promotion rules.

Fix #1754.

### DIFF
--- a/spirv_common.hpp
+++ b/spirv_common.hpp
@@ -211,6 +211,28 @@ inline std::string convert_to_string(const T &t)
 	return std::to_string(t);
 }
 
+static inline std::string convert_to_string(int32_t value)
+{
+	// INT_MIN is ... special on some backends. If we use a decimal literal, and negate it, we
+	// could accidentally promote the literal to long first, then negate.
+	// To workaround it, emit int(0x80000000) instead.
+	if (value == std::numeric_limits<int32_t>::min())
+		return "int(0x80000000)";
+	else
+		return std::to_string(value);
+}
+
+static inline std::string convert_to_string(int64_t value, const std::string &int64_type, bool long_long_literal_suffix)
+{
+	// INT64_MIN is ... special on some backends.
+	// If we use a decimal literal, and negate it, we might overflow the representable numbers.
+	// To workaround it, emit int(0x80000000) instead.
+	if (value == std::numeric_limits<int64_t>::min())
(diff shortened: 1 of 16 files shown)


spirv_common.hpp
```diff
@@ -211,6 +211,28 @@ inline std::string convert_to_string(const T &t)
...
+		return "int(0x80000000)";
+	else
+		return std::to_string(value);
+}
+
+static inline std::string convert_to_string(int64_t value, const std::string &int64_type, bool long_long_literal_suffix)
+{
+	// INT64_MIN is ... special on some backends.
+	// If we use a decimal literal, and negate it, we might overflow the representable numbers.
+	// To workaround it, emit int(0x80000000) instead.
```
pablode, line 229: Adjust comment to 64-bit

PR #1769, opened 2021-10-14
Add support for SPM without modifying directory structure

wtholliday:
This change only adds Package.swift.

Makes it really easy to add SPIRV-Cross to an Xcode project.


HansKristian-Work (approved):
LGTM

PR #1773, opened 2021-10-19
Support Metal 2.4 Intersection Query, Implement GL_EXT_ray_query.

congyue1977:


spirv_msl.cpp
```diff
@@ -8373,6 +8378,95 @@ void CompilerMSL::emit_instruction(const Instruction &instruction)
 			SPIRV_CROSS_THROW("Raster order groups require MSL 2.0.");
 		break; // Nothing to do in the body
 
+	case OpRayQueryInitializeKHR:
```
HansKristian-Work, line 8381: OpConvertUToAccelerationStructureKHR is missing, but I suppose that is expected, since Metal does not support BDA ... Either way, this opcode should have a `SPIRV_CROSS_THROW` case.

spirv_msl.cpp
```diff
@@ -8373,6 +8378,95 @@ void CompilerMSL::emit_instruction(const Instruction &instruction)
...
+	{
+		flush_variable_declaration(ops[0]);
+
+		statement("intersection_params _intersection_params_;");
```
HansKristian-Work, line 8385: What does this do? Also, there will be conflicts with naming if multiple such variables are declared in same scope.

spirv_msl.cpp
```diff
@@ -8373,6 +8378,95 @@ void CompilerMSL::emit_instruction(const Instruction &instruction)
...
+		          ", _intersection_params_);");
+		break;
+	}
+	case OpRayQueryProceedKHR:
+	{
+		flush_variable_declaration(ops[0]);
+		emit_op(ops[0], ops[1], join(to_expression(ops[2]), ".next()"), false);
+		break;
+	}
+#define MSL_RAY_QUERY_IS_CANDIDATE to_expression(ops[3]) == "0u"
```
HansKristian-Work, line 8397: This is goofy. Use `get<SPIRConstant>(ops[3]).scalar_u32() == 0` instead.

spirv_msl.cpp
```diff
@@ -8373,6 +8378,95 @@ void CompilerMSL::emit_instruction(const Instruction &instruction)
...
+		MSL_RAY_QUERY_GET_OP2(IntersectionObjectRayOrigin, ray_origin);
+		MSL_RAY_QUERY_GET_OP2(IntersectionObjectRayDirection, ray_direction);
+		MSL_RAY_QUERY_GET_OP2(IntersectionObjectToWorld, object_to_world_transform);
+		MSL_RAY_QUERY_GET_OP2(IntersectionWorldToObject, world_to_object_transform);
+		MSL_RAY_QUERY_IS_OP2(IntersectionFrontFace, triangle_front_facing);
+
+	case OpRayQueryGetIntersectionTypeKHR:
+		flush_variable_declaration(ops[2]);
+		if (MSL_RAY_QUERY_IS_CANDIDATE)
+			emit_op(ops[0], ops[1], join("((uint)", to_expression(ops[2]), ".get_candidate_intersection_type()) - 1"),
```
HansKristian-Work, line 8434: Why `(uint)v` cast? `uint(v)` make a bit more sense to me and avoids having to enclose the scope of ops[2].

HansKristian-Work: Why - 1?

spirv_msl.cpp
```diff
@@ -1512,6 +1512,11 @@ void CompilerMSL::preprocess_op_codes()
 	    (is_sample_rate() && (active_input_builtins.get(BuiltInFragCoord) ||
 	                          (need_subpass_input && !msl_options.use_framebuffer_fetch_subpasses))))
 		needs_sample_id = true;
+
+	if (preproc.ray_tracing)
```
HansKristian-Work, line 1516: Should validate MSL version somewhere. This seems like a natural place to do so.

spirv_msl.cpp
```diff
@@ -8373,6 +8378,95 @@ void CompilerMSL::emit_instruction(const Instruction &instruction)
...
+		MSL_RAY_QUERY_IS_OP2(IntersectionFrontFace, triangle_front_facing);
+
+	case OpRayQueryGetIntersectionTypeKHR:
+		flush_variable_declaration(ops[2]);
+		if (MSL_RAY_QUERY_IS_CANDIDATE)
+			emit_op(ops[0], ops[1], join("((uint)", to_expression(ops[2]), ".get_candidate_intersection_type()) - 1"),
+			        false);
+		else
+			emit_op(ops[0], ops[1], join("(uint)", to_expression(ops[2]), ".get_committed_intersection_type()"), false);
+
```
HansKristian-Work, line 8438: Weird newline.

spirv_msl.cpp
HansKristian-Work: Also not seeing an implementation for `OpRayQueryGetIntersectionInstanceShaderBindingTableRecordOffsetKHR` either.

spirv_msl.cpp
```diff
@@ -11773,6 +11867,10 @@ void CompilerMSL::entry_point_args_discrete_descriptors(string &ep_args)
 			}
 			break;
 		}
+		case SPIRType::AccelerationStructure:
+			ep_args += ", " + type_to_glsl(type, var_id) + " " + r.name;
```
HansKristian-Work, line 11871: Do acceleration structures work inside argument buffers?

spirv_msl.cpp
```diff
@@ -13283,6 +13381,12 @@ string CompilerMSL::type_to_glsl(const SPIRType &type, uint32_t id)
 	case SPIRType::Double:
 		type_name = "double"; // Currently unsupported
 		break;
+	case SPIRType::AccelerationStructure:
+		type_name = "acceleration_structure<instancing>";
+		break;
+	case SPIRType::RayQuery:
+		type_name = "intersection_query<instancing, triangle_data>";
```
HansKristian-Work, line 13388: Does this only work with triangle_data? What about procedural primitives?

spirv_msl.cpp
```diff
@@ -14987,6 +15091,10 @@ bool CompilerMSL::OpCodePreprocessor::handle(Op opcode, const uint32_t *args, ui
 		break;
 	}
 
+	case OpTypeRayQueryKHR:
```
HansKristian-Work, line 15094: Don't think we'll see types here. The ray tracing flag should be set based on SPIR-V capabilities I think, not based on opcodes.

spirv_msl.cpp
congyue1977: No MSL function corresponding to OpRayQueryGetIntersectionInstanceShaderBindingTableRecordOffsetKHR in the Intersection Query.

spirv_msl.cpp
congyue1977: Fixed.

spirv_msl.cpp
congyue1977: Fixed.

spirv_msl.cpp
congyue1977: The Intersection Query supports procedural primitives by default, and there are no such procedural tags.

spirv_msl.cpp
congyue1977: Unlike rayQueryGetIntersectionTypeEXT, the get_candidate_intersection_type and get_committed_intersection_type functions in MSL return the same enumeration value, while GLSL and HLSL return different enumeration values. This processing is for compatibility. If there is a better way to deal with it, please tell me to modify it.

spirv_msl.cpp
```diff
@@ -8373,6 +8381,98 @@ void CompilerMSL::emit_instruction(const Instruction &instruction)
 			SPIRV_CROSS_THROW("Raster order groups require MSL 2.0.");
 		break; // Nothing to do in the body
 
+	case OpConvertUToAccelerationStructureKHR:
+		if (!msl_options.supports_msl_version(2, 4))
+			SPIRV_CROSS_THROW("Intersection Query require MSL 2.4.");
+		break; // Nothing to do in the body
```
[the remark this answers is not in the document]
congyue1977: Fixed.

PR #1774, opened 2021-10-19
INTEL_fragment_shader_ordering support

Erfan-Ahmadi:


spirv_glsl.cpp
```diff
@@ -778,16 +772,28 @@ void CompilerGLSL::emit_header()
 		{
 			if (options.version < 310)
 				SPIRV_CROSS_THROW("At least ESSL 3.10 required for fragment shader interlock.");
-			require_extension_internal("GL_NV_fragment_shader_interlock");
+			statement("#extension GL_NV_fragment_shader_interlock : enable");
```
HansKristian-Work, line 775: Is there any reason this is not using the require_extension_internal mechanism? I'd expect this code to request GL_ARB_fragment_shader_interlock, and then the code that emits extensions will handle the ifdefs. This is similar to how other conditional extensions do it.

spirv_glsl.cpp
```diff
@@ -778,16 +772,28 @@ void CompilerGLSL::emit_header()
...
+			statement("#define SPIRV_Cross_beginInvocationInterlock() beginInvocationInterlockNV()");
+			statement("#define SPIRV_Cross_endInvocationInterlock() endInvocationInterlockNV()");
 		}
 		else
 		{
 			if (options.version < 420)
-				require_extension_internal("GL_ARB_shader_image_load_store");
-			require_extension_internal("GL_ARB_fragment_shader_interlock");
+				statement("#extension GL_ARB_shader_image_load_store : enable");
```
HansKristian-Work, line 782: This is problematic if some other code used require_extension_internal (earlier in emit_header, this is exactly the case), we'd end up with double declaration, so this should also be require_extension_internal.

spirv_glsl.cpp
```diff
@@ -778,16 +772,28 @@ void CompilerGLSL::emit_header()
...
+
+		    statement("#ifdef GL_ARB_fragment_shader_interlock");
```
HansKristian-Work, line 784: These ifdef ladders can be handled in the code that emits extensions below. I'd prefer consistency in this case even if it's not strictly necessary for correctness.

spirv_glsl.cpp
```diff
@@ -1041,14 +1047,32 @@ void CompilerGLSL::emit_header()
 		if (execution.flags.get(ExecutionModePostDepthCoverage))
 			inputs.push_back("post_depth_coverage");
 
+		if (execution.flags.get(ExecutionModePixelInterlockOrderedEXT) ||
```
HansKristian-Work, line 1050: Should probably combine this check into a bool so it's not redundantly checked below.

spirv_glsl.cpp
```diff
@@ -1041,14 +1047,32 @@ void CompilerGLSL::emit_header()
...
+			execution.flags.get(ExecutionModePixelInterlockUnorderedEXT) ||
+			execution.flags.get(ExecutionModeSampleInterlockOrderedEXT) ||
+			execution.flags.get(ExecutionModeSampleInterlockUnorderedEXT))
+		{
+			statement("#if defined(GL_ARB_fragment_shader_interlock)");
+		}
+
 		if (execution.flags.get(ExecutionModePixelInterlockOrderedEXT))
-			inputs.push_back("pixel_interlock_ordered");
+			statement("layout(pixel_interlock_ordered) in;");
```
HansKristian-Work, line 1059: I assume it's allowed to declare:
```
layout(pixel_interlock_ordered) in;
layout(early_fragment_tests) in;
```
where both execution modes are set?

HansKristian-Work (changes requested):
The approach seems fine, but some nits.

spirv_glsl.cpp
Erfan-Ahmadi: Yes, I've checked generated code, both modes are set in this case.

spirv_glsl.cpp
Erfan-Ahmadi: You're right, require extension may be called for GL_ARB_shader_image_load_store earlier.

spirv_glsl.cpp
devshgraphicsprogramming: can you give an example?

spirv_glsl.cpp
devshgraphicsprogramming: why are we even concerned with `GL_ARB_shader_image_load_store` ?

PR #1783, opened 2021-10-21
MSL: Support more usecases for unpacked vectors.

### LINKED ISSUE #1772: MSL: Struct member mis-declared as int4 or float4 vector array when should be int or float scalar array
SPIR-V conversion to MSL fails compilation because an `int` or `float` scalar struct member array is mis-declared as array of vector `int4` or float4.

The following SPIR-V:

```
%6 = OpTypeInt 32 1
%7 = OpTypeInt 32 0
%8 = OpConstant %7 3
%12 = OpTypeArray %6 %8
%13 = OpTypeStruct %12
...
%36 = OpTypeFloat 32
%43 = OpConstant %36 0
%44 = OpTypeArray %36 %8
%45 = OpTypeStruct %44
```

results in the (incorrect) MSL:
```
struct buf1
{
    int4 _GLF_uniform_int_values[3];
};

struct buf0
{
    float4 _GLF_uniform_float_values[3];
};
```

where the vector `int4` should be a scalar `int`.

The same SPIR-V results in the (correct) GLSL:
```
layout(set = 0, binding = 1, std140) uniform buf1
{
    int _GLF_uniform_int_values[3];
} _15;

layout(set = 0, binding = 0, std140) uniform buf0
{
    highp float _GLF_uniform_float_values[3];
} _47;
```

These CTS test fail as a result:
```
dEQP-VK.graphicsfuzz.cov-access-array-dot
dEQP-VK.graphicsfuzz.cov-apfloat-determinant-for-if
dEQP-VK.graphicsfuzz.cov-apfloat-unpackunorm-loop
dEQP-VK.graphicsfuzz.cov-array-accesses-clamp
dEQP-VK.graphicsfuzz.cov-array-copies-loops-with-limiters
...total 110 tests
```

SPIR-V file: 
[incorrect-vec-array-declaration-from-scalar-array.spv.zip](https://github.com/KhronosGroup/SPIRV-Cross/files/7368504/incorrect-vec-array-declaration-from-scalar-array.spv.zip)

billhollings:
Additional usecases include array and vector indexing, variable declarations, loop initializers, function return values, switch statement evaluations, and various specialized MSL operations.

Ultimately, we might consider refactoring to move the current `CompilerGLSL::to_unpacked_expression()` to a new `CompilerMSL::to_expression()` to always take into consideration possible unpacking behavior, and remove `CompilerGLSL::to_unpacked_expression()` altogether. 

I did not do that here, because I wanted to get further input from others in the team (@HansKristian-Work @cdavis5e) before taking that step. For example, I noticed that HLSL uses unpacking in a couple of places too, but didn't investigate further at this time.

Refactor `CompilerGLSL::to_enclosed_unpacked_expression()` for conciseness and consistency with similar functionality.

Fixes #1772.

### DIFF
--- a/spirv_glsl.cpp
+++ b/spirv_glsl.cpp
@@ -4375,19 +4375,7 @@ string CompilerGLSL::to_unpacked_expression(uint32_t id, bool register_expressio
 
 string CompilerGLSL::to_enclosed_unpacked_expression(uint32_t id, bool register_expression_read)
 {
-	// If we need to transpose, it will also take care of unpacking rules.
-	auto *e = maybe_get<SPIRExpression>(id);
-	bool need_transpose = e && e->need_transpose;
-	bool is_remapped = has_extended_decoration(id, SPIRVCrossDecorationPhysicalTypeID);
-	bool is_packed = has_extended_decoration(id, SPIRVCrossDecorationPhysicalTypePacked);
-	if (!need_transpose && (is_remapped || is_packed))
-	{
-		return unpack_expression_type(to_expression(id, register_expression_read), expression_type(id),
-		                              get_extended_decoration(id, SPIRVCrossDecorationPhysicalTypeID),
-		                              has_extended_decoration(id, SPIRVCrossDecorationPhysicalTypePacked), false);
-	}
-	else
-		return to_enclosed_expression(id, register_expression_read);
+	return enclose_expression(to_unpacked_expression(id, register_expression_read));
 }
 
 string CompilerGLSL::to_dereferenced_expression(uint32_t id, bool register_expression_read)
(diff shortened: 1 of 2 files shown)


HansKristian-Work:
It might be worth it to refactor to_expression to to_raw_expression or something like this, and then do a pass over all uses of it. Then promote to_unpacked_expression to to_expression.

billhollings:
>Just thinking of what we'd call the non-unpacked expression thing.

Ah. Okay.

I think my point was that AFAIK unpacking is only used by MSL and not GLSL. So I was thinking that GLSL would _**only**_ have a default `to_expression()` that did not take into consideration unpacking, and MSL would override `to_expression()` to basically _**always**_ check for unpacking is required, and call the parent GLSL version if not.

Effectively...I was thinking there would be no separate non-unpacking function at either class level.

But happy to hear how GLSL and HLSL might need that.

billhollings:
I've created issue #1785, which we can use for further discussion about these ideas.

PR #1792, opened 2021-10-28
MSL: Remove over-zealous check for struct packing compatibility.

billhollings:
Previous test for `SPIRVCrossDecorationPhysicalTypePacked` on parent struct when unpacking member struct was too restrictive, and not needed as long as padding compensates.

Added unit test.


HansKristian-Work (approved):
Seems OK I guess.

PR #1794, opened 2021-10-29
Add 64 bit support for OpSwitch

### LINKED ISSUE #1768: MSL & GLSL: OpBranchConditional throws nullptr error from the True Label
These CTS test fail after SPIR-V conversion to MSL throws `nullptr` error from the _True Label_:

```
dEQP-VK.spirv_assembly.type.scalar.i64.switch_vert
dEQP-VK.spirv_assembly.type.scalar.i64.switch_tessc
dEQP-VK.spirv_assembly.type.scalar.i64.switch_tesse
dEQP-VK.spirv_assembly.type.scalar.i64.switch_frag
dEQP-VK.spirv_assembly.type.scalar.i64.switch_comp
dEQP-VK.spirv_assembly.type.scalar.u64.switch_vert
dEQP-VK.spirv_assembly.type.scalar.u64.switch_tessc
dEQP-VK.spirv_assembly.type.scalar.u64.switch_tesse
dEQP-VK.spirv_assembly.type.scalar.u64.switch_frag
dEQP-VK.spirv_assembly.type.scalar.u64.switch_comp
```

The related conversion to GLSL also fails.

SPIR-V files: 
[nullptr_on_branch_conditional_true_label.spv.zip](https://github.com/KhronosGroup/SPIRV-Cross/files/7341037/nullptr_on_branch_conditional_true_label.spv.zip)

etra0:
Hello! this is my first PR.

According to the spec, if the `condition` has a type wider than 32 bits,
the literals to be compared with will be of that size as well.

This caused some misalignments if the `condition` is bigger than 32,
causing a nullptr return without further explanation.

Currently neither GLSL nor MSL supports uint64 as the condition but the
SPIRV allows it anyway. With this fix the error message explicitly
states that a switch with 64 bit sized variables can't be used in both GLSL
and MSL.

I couldn't find any "global" array containing type information at this stage,
so I had to add a map to some IDs in order to check their type.

This also fixes #1768.

### DIFF
--- a/spirv_cfg.cpp
+++ b/spirv_cfg.cpp
@@ -135,8 +135,9 @@ bool CFG::post_order_visit(uint32_t block_id)
 		break;
 
 	case SPIRBlock::MultiSelect:
-		compiler.fix_switch_branches(block);
-		for (auto &target : block.cases)
+	{
+		const auto &cases = compiler.get_case_list(block);
+		for (const auto &target : cases)
 		{
 			if (post_order_visit(target.block))
 				add_branch(block_id, target.block);
@@ -144,6 +145,7 @@ bool CFG::post_order_visit(uint32_t block_id)
 		if (block.default_block && post_order_visit(block.default_block))
 			add_branch(block_id, block.default_block);
 		break;
+	}
 	default:
 		break;
 	}
@@ -385,7 +387,9 @@ void DominatorBuilder::lift_continue_block_dominator()
 		break;
 
 	case SPIRBlock::MultiSelect:
-		for (auto &target : block.cases)
+	{
+		auto &cases = cfg.get_compiler().get_case_list(block);
+		for (auto &target : cases)
 		{
 			if (cfg.get_visit_order(target.block) > post_order)
 				back_edge_dominator = true;
@@ -393,6 +397,7 @@ void DominatorBuilder::lift_continue_block_dominator()
 		if (block.default_block && cfg.get_visit_order(block.default_block) > post_order)
 			back_edge_dominator = true;
 		break;
+	}
 
 	default:
 		break;
--- a/spirv_cfg.cpp
+++ b/spirv_cfg.cpp
@@ -135,6 +135,7 @@ bool CFG::post_order_visit(uint32_t block_id)
 		break;
 
 	case SPIRBlock::MultiSelect:
+		compiler.fix_switch_branches(block);
 		for (auto &target : block.cases)
 		{
 			if (post_order_visit(target.block))
@@ -143,7 +144,6 @@ bool CFG::post_order_visit(uint32_t block_id)
 		if (block.default_block && post_order_visit(block.default_block))
 			add_branch(block_id, block.default_block);
 		break;
-
 	default:
 		break;
 	}
--- a/spirv_common.hpp
+++ b/spirv_common.hpp
@@ -854,7 +854,7 @@ struct SPIRBlock : IVariant
 
 	struct Case
 	{
-		uint32_t value;
+		uint64_t value;
 		BlockID block;
 	};
 	SmallVector<Case> cases;
--- a/spirv_common.hpp
+++ b/spirv_common.hpp
@@ -857,7 +857,8 @@ struct SPIRBlock : IVariant
 		uint64_t value;
 		BlockID block;
 	};
-	SmallVector<Case> cases;
+	mutable SmallVector<Case> cases;
+	SmallVector<Case> cases_64bit;
 
 	// If we have tried to optimize code for this block but failed,
 	// keep track of this.
--- a/spirv_common.hpp
+++ b/spirv_common.hpp
@@ -857,7 +857,7 @@ struct SPIRBlock : IVariant
 		uint64_t value;
 		BlockID block;
 	};
-	mutable SmallVector<Case> cases;
+	SmallVector<Case> cases_32bit;
 	SmallVector<Case> cases_64bit;
 
 	// If we have tried to optimize code for this block but failed,
--- a/spirv_cross.cpp
+++ b/spirv_cross.cpp
@@ -1626,6 +1626,17 @@ SPIRBlock::ContinueBlockType Compiler::continue_block_type(const SPIRBlock &bloc
 	}
 }
 
+void Compiler::fix_switch_branches(const SPIRBlock &block) const {
+	auto search = ir.load_types.find(block.condition);
+	if (search == ir.load_types.end()) {
+		SPIRV_CROSS_THROW("Use of undeclared variable on a switch statement.");
+	}
+
+	const auto& type = search->second;
+	if (type.width > 32)
+		block.cases = std::move(block.cases_64bit);
+}
+
 bool Compiler::traverse_all_reachable_opcodes(const SPIRBlock &block, OpcodeHandler &handler) const
 {
 	handler.set_current_block(block);
--- a/spirv_cross.cpp
+++ b/spirv_cross.cpp
@@ -1626,9 +1626,11 @@ SPIRBlock::ContinueBlockType Compiler::continue_block_type(const SPIRBlock &bloc
 	}
 }
 
-void Compiler::fix_switch_branches(const SPIRBlock &block) const {
+void Compiler::fix_switch_branches(const SPIRBlock &block) const
+{
 	auto search = ir.load_type_width.find(block.condition);
-	if (search == ir.load_type_width.end()) {
+	if (search == ir.load_type_width.end())
+	{
 		SPIRV_CROSS_THROW("Use of undeclared variable on a switch statement.");
 	}
 
--- a/spirv_cross.cpp
+++ b/spirv_cross.cpp
@@ -1627,13 +1627,13 @@ SPIRBlock::ContinueBlockType Compiler::continue_block_type(const SPIRBlock &bloc
 }
 
 void Compiler::fix_switch_branches(const SPIRBlock &block) const {
-	auto search = ir.load_types.find(block.condition);
-	if (search == ir.load_types.end()) {
+	auto search = ir.load_type_width.find(block.condition);
+	if (search == ir.load_type_width.end()) {
 		SPIRV_CROSS_THROW("Use of undeclared variable on a switch statement.");
 	}
 
-	const auto& type = search->second;
-	if (type.width > 32)
+	const uint32_t width = search->second;
+	if (width > 32)
 		block.cases = std::move(block.cases_64bit);
 }
 
--- a/spirv_cross.cpp
+++ b/spirv_cross.cpp
@@ -1628,13 +1628,31 @@ SPIRBlock::ContinueBlockType Compiler::continue_block_type(const SPIRBlock &bloc
 
 const SmallVector<SPIRBlock::Case> &Compiler::get_case_list(const SPIRBlock &block) const
 {
-	auto search = ir.load_type_width.find(block.condition);
-	if (search == ir.load_type_width.end())
+	uint32_t width = 0;
+
+	// First we check if we can get the type directly from the block.condition
+	// since it can be a SPIRConstant or a SPIRVariable.
+	if (const auto *constant = maybe_get<SPIRConstant>(block.condition))
 	{
-		SPIRV_CROSS_THROW("Use of undeclared variable on a switch statement.");
+		const auto &type = get<SPIRType>(constant->constant_type);
+		width = type.width;
+	}
+	else if (const auto *var = maybe_get<SPIRVariable>(block.condition))
+	{
+		const auto &type = get<SPIRType>(var->basetype);
+		width = type.width;
+	}
+	else
+	{
+		auto search = ir.load_type_width.find(block.condition);
+		if (search == ir.load_type_width.end())
+		{
+			SPIRV_CROSS_THROW("Use of undeclared variable on a switch statement.");
+		}
+
+		width = search->second;
 	}
 
-	const uint32_t width = search->second;
 	if (width > 32)
 		return block.cases_64bit;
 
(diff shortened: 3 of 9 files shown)


HansKristian-Work (changes requested):
I really don't understand the need for all this complexity. When emitting the switch statement, just check the type of the condition and emit 64-bit literals or not based on that. I don't want to hack up the parser like this.

etra0:
> I really don't understand the need for all this complexity. When emitting the switch statement, just check the type of the condition and emit 64-bit literals or not based on that. I don't want to hack up the parser like this.

The main issue is that since wider types are allowed in the `OpSwitch` statement, the literals gets compiled as wider types as well, which causes a misaligned read causing to store wrong labels at this specific stage (which causes the bug #1768, which stores 0 as a label). 

At this stage I couldn't find any track of the types of previously parsed operations so that's why I resorted to manually adding the types to a map.

Another solution I can think is as you said, add the check at the check in the future when there's enough information about types but that would carry on with the misaligned labels. 

Let me know what you think!

HansKristian-Work:
Ugh, right ... The parsing now becomes context sensitive apparently. Well, in that case, I'd prefer if the parsing parsed the Switch twice, and emit a case list for 32-bit (if literal list length aligns to 2) and 64-bit (if literal list length aligns to 3), and then the appropriate case list can be selected at runtime. I really want to avoid any case sensitive parsing in the general parser.

etra0:
I see, I'll rework the PR then.

etra0:
Hmm, I'm not sure how to deal with the tracking of types. The fix should be applied before building the CFG (or while building it) but at that stage I don't see type information. Do you have any suggestion on where to look at?

HansKristian-Work:
> Hmm, I'm not sure how to deal with the tracking of types. The fix should be applied before building the CFG (or while building it) but at that stage I don't see type information, the condition is usually of `TypeNone`. Do you have any suggestion on where to look at?

When emitting the switch you can use `expression_type(condition.id)`.

etra0:
So I tried moving the logic out of the parser (I still store type information in that process but I'd guess it's fine?) and now we do the fix in the CFG stage since we can't do it after because that's where the nullptr is deref. I think I could do the fixup before the CFG stage by manually iterating the blocks as well.

Let me know what you think.

spirv_cross_parsed_ir.hpp
```diff
@@ -78,6 +78,13 @@ class ParsedIR
 	SmallVector<ID> ids_for_constant_or_type;
 	SmallVector<ID> ids_for_constant_or_variable;
 
+	// We need to keep track of the width the Ops that contains a type for the
+	// OpSwitch instruction, since this one doesn't contains the type in the
+	// instruction itself. And in some case we need to cast the condition to
+	// wider types. We only need the width to do the branch fixup since the
+	// type check itself can be done at runtime
+	std::unordered_map<ID, uint32_t> load_type_width;
```
HansKristian-Work, line 86: This is unnecessary.

spirv_parser.cpp
```diff
@@ -771,6 +771,8 @@ void Parser::parse(const Instruction &instruction)
 
 		uint32_t result_type = ops[0];
 		uint32_t id = ops[1];
+		auto &type = get<SPIRType>(result_type);
+		ir.load_type_width.insert({ id, type.width });
```
HansKristian-Work, line 775: Tracking load_type_width is unnecessary.

spirv_parser.cpp
```diff
@@ -975,9 +978,16 @@ void Parser::parse(const Instruction &instruction)
 		current_block->condition = ops[0];
 		current_block->default_block = ops[1];
 
+		uint32_t remaining_ops = length - 2;
 		for (uint32_t i = 2; i + 2 <= length; i += 2)
 			current_block->cases.push_back({ ops[i], ops[i + 1] });
 
+		if (remaining_ops % 3)
```
HansKristian-Work, line 985: Formatting, see clang-format.

spirv_common.hpp
```diff
@@ -854,10 +854,11 @@ struct SPIRBlock : IVariant
 
 	struct Case
 	{
-		uint32_t value;
+		uint64_t value;
 		BlockID block;
 	};
-	SmallVector<Case> cases;
+	mutable SmallVector<Case> cases;
```
HansKristian-Work, line 860: This is unnecessary, we should not mutate the cases after parse.

spirv_cross.cpp
```diff
@@ -1626,6 +1626,17 @@ SPIRBlock::ContinueBlockType Compiler::continue_block_type(const SPIRBlock &bloc
 	}
 }
 
+void Compiler::fix_switch_branches(const SPIRBlock &block) const {
+	auto search = ir.load_type_width.find(block.condition);
```
HansKristian-Work, line 1630: No need.

spirv_parser.cpp
```diff
@@ -975,9 +978,16 @@ void Parser::parse(const Instruction &instruction)
 		current_block->condition = ops[0];
 		current_block->default_block = ops[1];
 
+		uint32_t remaining_ops = length - 2;
```
HansKristian-Work, line 981: Don't parse 32-bit case labels if it does not align to remaining_ops % 2 == 0

spirv_parser.cpp
HansKristian-Work, line 985: Shouldn't this be remaining_ops % 3 == 0?

HansKristian-Work (changes requested):
There's only a few parts required here:

- In parser, parse the 32-bit and/or 64-bit depending on number of arguments.
- https://github.com/KhronosGroup/SPIRV-Cross/blob/master/spirv_glsl.cpp#L14756 is when the switch block is emitted and we see the type of the condition. There is already an error case for 64-bit switch labels here. Based on the type we can ...
- replace `block.cases` in https://github.com/KhronosGroup/SPIRV-Cross/blob/master/spirv_glsl.cpp#L14792 with either picking the 32-bit case list or 64-bit case list.

etra0:
I think I haven't done a very well job explaining myself, I'm sorry for that. 
Take the following SPIR-V as an example:
```asm
; SPIR-V
; Version: 1.0
; Generator: Google Shaderc over Glslang; 10
; Bound: 18
; Schema: 0
               OpCapability Shader
               OpCapability Int64
          %1 = OpExtInstImport "GLSL.std.450"
               OpMemoryModel Logical GLSL450
               OpEntryPoint Fragment %main "main"
               OpExecutionMode %main OriginUpperLeft
               OpSource GLSL 330
               OpSourceExtension "GL_GOOGLE_cpp_style_line_directive"
               OpSourceExtension "GL_GOOGLE_include_directive"
               OpName %main "main"
               OpName %sw "sw"
               OpName %result "result"
       %void = OpTypeVoid
          %3 = OpTypeFunction %void
        %int = OpTypeInt 32 1
        %long = OpTypeInt 64 1
%_ptr_Function_int = OpTypePointer Function %int
%_ptr_Function_long = OpTypePointer Function %long
     %int_42 = OpConstant %int 42
      %int_0 = OpConstant %int 0
    %int_420 = OpConstant %int 420
     %long_42 = OpConstant %long 42
       %main = OpFunction %void None %3
          %5 = OpLabel
         %sw = OpVariable %_ptr_Function_long Function
     %result = OpVariable %_ptr_Function_int Function
               OpStore %sw %long_42
               OpStore %result %int_0
         %12 = OpLoad %long %sw
               OpSelectionMerge %15 None
               OpSwitch %12 %15 -42 %13 420 %14
         %13 = OpLabel
               OpStore %result %int_42
               OpBranch %14
         %14 = OpLabel
               OpStore %result %int_420
               OpBranch %15
         %15 = OpLabel
               OpReturn
               OpFunctionEnd
```
The OpSwitch loads the id `%12` which contains itself a long (in this case, 64 bit wide). I compiled this spirv with `glslc` and another nearly similar by just changing the type of the `%sw` to int to see how they differ at the binary level and I got the following results:

image

where the left is the corresponding 32 bit and the right is the 64 bit. A the offset `0x1FC`, the sign extended version of the value `-42` is larger in the 64 version because of the casting of the switch (0xFFFFFFD6 vs 0xFFFFFFFFFFFFFFD6) which causes the misaligned read so we can't postpone the fix after the control flow graph building because a nullptr deref occurs otherwise (in this example it would try to get the `0xFFFFF FFFFF` block which doesn't exists, and it causes a segfault). The casting is 'embedded' at the compilation time if that makes sense.

spirv_parser.cpp
etra0: Yes, it should. I'm sorry for this.

etra0: Formatted in the latest commit. Thanks.

spirv_parser.cpp
etra0: I was thinking on having the `remaining_ops % 2` be the fallback case always, and when doing the branch fix see if it's necessary to override that case. Technically we "don't" have the option to choose since when one case is valid the other one is immediately invalid, so that's why I thought this way. Let me know what you think.

spirv_cross.cpp
```diff
@@ -1626,6 +1626,19 @@ SPIRBlock::ContinueBlockType Compiler::continue_block_type(const SPIRBlock &bloc
 	}
 }
 
+void Compiler::fix_switch_branches(const SPIRBlock &block) const
+{
+	auto search = ir.load_type_width.find(block.condition);
+	if (search == ir.load_type_width.end())
+	{
+		SPIRV_CROSS_THROW("Use of undeclared variable on a switch statement.");
```
HansKristian-Work, line 1634: This breaks if the condition is a constant or spec constant.

spirv_parser.cpp
```diff
@@ -771,6 +771,8 @@ void Parser::parse(const Instruction &instruction)
 
 		uint32_t result_type = ops[0];
 		uint32_t id = ops[1];
+		auto &type = get<SPIRType>(result_type);
```
HansKristian-Work, line 774: Would make more sense to me if the code that queries load_type_width would check for `maybe_get<SPIRVariable>` and `maybe_get<SPIRConstant>` before falling back to normal SSA variable query.

spirv_cfg.cpp
```diff
@@ -135,6 +135,7 @@ bool CFG::post_order_visit(uint32_t block_id)
 		break;
 
 	case SPIRBlock::MultiSelect:
+		compiler.fix_switch_branches(block);
```
HansKristian-Work, line 138: Should be a constant `get_case_list()` instead of mutating like this.

HansKristian-Work (changes requested):
Ok, ran some shaders myself and I understand the problem better now.

spirv_parser.cpp
```diff
@@ -1134,6 +1145,14 @@ void Parser::parse(const Instruction &instruction)
 	// Actual opcodes.
 	default:
 	{
+		if (ops)
```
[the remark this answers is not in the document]
etra0: Noted.

spirv_cross.cpp
etra0: On that note -- is there a way to test shaders that would throw an exception?

HansKristian-Work: No. To be conformant, I think we actually have to support 64-bit switch statements don't we? That's a bit outside the scope of this PR though, I can test a bit locally before accepting the PR.

etra0: Hmm, at least from what I can see in the latest spec ([GLSL lang spec 4.60](https://www.khronos.org/registry/OpenGL/specs/gl/GLSLangSpec.4.60.pdf)), `int` and `uint` are explicitly defined as 32 bits types [1] and in the definition of the `switch` statement it says the condition must be a scalar integer and if it's not, it's implicitly casted [2].

[1] Section 4.1.3 in the spec.
[2] Section 6.2

HansKristian-Work: We'd have to rewrite the switch to if/else ladders in this scenario.

spirv_parser.cpp
etra0: If I understand you correctly, what you're asking for is to only track for IDs and get the types at the function `get_case_list()`. I tried following this approach but usually when the `block.condition` points to an `OpLoad`, the `type_id` ends up being a `TypeNone` which makes visually impossible to get the actual type (with helpers like `expression_type` etc.). That's why I've decided to track the type that early on. If you see any other way please let me know!

spirv_parser.cpp
```diff
@@ -975,9 +978,17 @@ void Parser::parse(const Instruction &instruction)
 		current_block->condition = ops[0];
 		current_block->default_block = ops[1];
 
+		uint32_t remaining_ops = length - 2;
 		for (uint32_t i = 2; i + 2 <= length; i += 2)
 			current_block->cases.push_back({ ops[i], ops[i + 1] });
 
+		if ((remaining_ops % 3) == 0)
```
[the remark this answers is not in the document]
etra0: done!

etra0:
I did change the mutating approach to returning the corresponding case from the `block.condition` width. I tried using `expression_type` to get the type of the instructions but sometimes it returns `TypeNone` which makes it fails in cases where it shouldn't. Maybe I'm missing a spot where the type is populated after the CFG.

spirv_parser.cpp
HansKristian-Work: Yes, but certain things are defined up front, like PHI nodes (SPIRVariable), constants and spec constants (SPIRConstant). SPIRExpression is found late as you say, and we'll need the map for that scenario.

HansKristian-Work (changes requested):
Starting to look good. Just need to rebase, get existing tests to pass and check SPIRVariable/SPIRConstant directly when possible.

etra0:
_huh_, I was sure I was running `test_shaders.sh` before doing commits. I'll manually check the one it's failing then.

HansKristian-Work (approved):
LGTM, thanks.

PR #1798, opened 2021-11-03
MSL: Correctly emit user(clip/cullN) for clip/cull builtins in tess output struct.

billhollings:
Only emit `user(locnN)` for tess builtin input variables, and allow output builtin to emit `user(clip/cullN)`. Previously, output builtin would emit location if input builtin also existed.

Fixes the following CTS tests:

```
dEQP-VK.clipping.user_defined.clip_distance.vert_tess.1_fragmentshader_read
dEQP-VK.clipping.user_defined.clip_distance.vert_tess.2_fragmentshader_read
dEQP-VK.clipping.user_defined.clip_distance.vert_tess.3_fragmentshader_read
dEQP-VK.clipping.user_defined.clip_distance.vert_tess.4_fragmentshader_read
dEQP-VK.clipping.user_defined.clip_distance.vert_tess.5_fragmentshader_read
dEQP-VK.clipping.user_defined.clip_distance.vert_tess.6_fragmentshader_read
dEQP-VK.clipping.user_defined.clip_distance.vert_tess.7_fragmentshader_read
dEQP-VK.clipping.user_defined.clip_distance.vert_tess.8_fragmentshader_read
dEQP-VK.clipping.user_defined.clip_cull_distance.vert_tess.8_fragmentshader_read
```

@cdavis5e Please review for tess compatibility

### DIFF
--- a/spirv_msl.cpp
+++ b/spirv_msl.cpp
@@ -2251,7 +2251,7 @@ void CompilerMSL::add_plain_variable_to_interface_block(StorageClass storage, co
 			set_member_decoration(ib_type.self, ib_mbr_idx, DecorationComponent, comp);
 		mark_location_as_used_by_shader(locn, get<SPIRType>(type_id), storage);
 	}
-	else if (is_builtin && is_tessellation_shader() && inputs_by_builtin.count(builtin))
+	else if (is_builtin && is_tessellation_shader() && storage == StorageClassInput && inputs_by_builtin.count(builtin))
 	{
 		uint32_t locn = inputs_by_builtin[builtin].location;
 		set_member_decoration(ib_type.self, ib_mbr_idx, DecorationLocation, locn);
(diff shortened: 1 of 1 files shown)


cdavis5e:
Don't we only want to do this for tessellation evaluation shaders? What happens if it's written from a TCS and read in a TES?

billhollings:
> Don't we only want to do this for tessellation evaluation shaders? What happens if it's written from a TCS and read in a TES?

What action do you mean by "do this"?

For the most part, `[[user(locN)]]` is covered by a previous test for explicit location decorations. Would that cover the TSC -> TSE matching you raise?

Hmmm...looks like `[[user(clip/cullN)]]` should probably be used for outputs from VS and TES, and inputs to FS, and `[[user(locN)]]` should be used otherwise. So perhaps we could flip the last two logic tests around to test explicitly for VS & TES outputs and FS inputs to apply `[[user(clip/cullN)]]`, then fall into `[[user(locN)]]` if that test fails?

HansKristian-Work:
VS and TES potentially need to emit 2 attributes as output:
- [clip_distance]
- [user(clipN)] (only used if FS reads clip distance)

Any write from TCS goes through a buffer, so it never uses locations or attributes I think.

HansKristian-Work (approved):
LGTM, seems like a strict fix either way.

PR #1799, opened 2021-11-03
ESSL: Generate correct precision qualifiers & type literal suffix.

sindney:
When enabling 16 bit types, current implementation uses for example 
'GL_EXT_shader_explicit_arithmetic_types_float16' to support different types. 

It requires an extension('EXT_shader_explicit_arithmetic_types') to be supported.

But for mobile platforms, it's [sufficient](https://developer.arm.com/documentation/101897/0200/shader-code/minimize-precision) to use precision qualifiers to mark the width of float types.
We only need 16bit float & vectors.

So for practical reasons, i made some ESSL related changes to support this behavior: 
```glsl
precision mediump float;
precision highp int;

layout(location = 0) in highp vec4 in_var_TEXCOORD10_centroid;
layout(location = 1) in highp vec4 in_var_TEXCOORD11_centroid;
layout(location = 2) in vec4 in_var_COLOR0;
layout(location = 3) in highp vec4 in_var_TEXCOORD0[2];
layout(location = 5) in vec4 in_var_TEXCOORD7;
layout(location = 6) in highp vec4 in_var_TEXCOORD8;
layout(location = 0) out vec4 out_var_SV_Target0;
layout(location = 1) out vec4 out_var_SV_Target1;
layout(location = 2) out vec4 out_var_SV_Target2;

void main()
{
    highp vec4 _309 = gl_FragCoord;
    _309.w = 0.0 / gl_FragCoord.w;
    highp vec4 param_var_SvPosition = _309;
    vec3 _329 = vec3(in_var_TEXCOORD10_centroid.xyz);
    vec4 _330 = vec4(in_var_TEXCOORD11_centroid);
    vec4 _296 = _330;
    vec3 _333 = _330.xyz;
    mat3 _338 = mat3(_329, cross(_333, _329) * _296.w, _333);
    highp vec2 _228 = vec2(in_var_TEXCOORD0[1u].x, in_var_TEXCOORD0[1u].y);
```

### DIFF
--- a/spirv_glsl.cpp
+++ b/spirv_glsl.cpp
@@ -400,21 +400,30 @@ void CompilerGLSL::find_static_extensions()
 		}
 		else if (type.basetype == SPIRType::Half)
 		{
-			require_extension_internal("GL_EXT_shader_explicit_arithmetic_types_float16");
-			if (options.vulkan_semantics)
-				require_extension_internal("GL_EXT_shader_16bit_storage");
+			if (!options.es)
+			{
+				require_extension_internal("GL_EXT_shader_explicit_arithmetic_types_float16");
+				if (options.vulkan_semantics)
+					require_extension_internal("GL_EXT_shader_16bit_storage");
+			}
 		}
 		else if (type.basetype == SPIRType::SByte || type.basetype == SPIRType::UByte)
 		{
-			require_extension_internal("GL_EXT_shader_explicit_arithmetic_types_int8");
-			if (options.vulkan_semantics)
-				require_extension_internal("GL_EXT_shader_8bit_storage");
+			if (!options.es)
+			{
+				require_extension_internal("GL_EXT_shader_explicit_arithmetic_types_int8");
+				if (options.vulkan_semantics)
+					require_extension_internal("GL_EXT_shader_8bit_storage");
+			}
 		}
 		else if (type.basetype == SPIRType::Short || type.basetype == SPIRType::UShort)
 		{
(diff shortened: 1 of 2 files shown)


HansKristian-Work:
I don't think it's sufficient to use mediump here, because there is no guarantee that mediump is implemented as 16-bit, it *might* be 16-bit, but that's no guarantee and we cannot implement Float16 cap with it.

sindney:
Hi, Hans, game engines used to do cross compiling using mesa3d's hlslcc, but now spirv seems to be a better solution(optimization pass works great). And full mobile support on the new cross compiler is the last missing part.

Yes, the qualifier is just a hint. But on mobile platforms, we need to support a wide range of devices, and since their drivers are not update-able, if we require this extension in our ESSL shaders, we need to provide fallback methods for the 16 bit extension. I know it's better and reasonable to use extension for this. But in reality, at least on mobile platforms, we still need this legacy qualifier way to use 16 bit types for performance reasons, and it's critical for reducing register pressure.

So, how about we add a new option like 'enablelegacy16bitsupport' or some thing like that, to support this behavior?
Then it won't affect the standard behavior.

HansKristian-Work:
Yes, FP16 is important on mobile, but I'm not sure why the SPIR-V needs to emit FP16 directly. Is there anything stopping you from using RelaxedPrecision SPIR-V? That will compile down to mediump as well. If you're cross compiling from HLSLcc I don't see how you'd end up with raw FP16 code to begin with.

sindney:
Thanks, I'll figure out how to use relaxed precision to achieve the same thing.

PR #1815, opened 2021-11-23
MSL: Add 64 bit switch support

etra0:
This _actually_ adds 64 bit support for MSL (previous PR ended up fixing a segfault). I made sure it also passes the tests posted by @billhollings with VK-GL-CTS:

```
Test case 'dEQP-VK.spirv_assembly.type.scalar.i64.switch_vert'..

Test case 'dEQP-VK.spirv_assembly.type.scalar.i64.switch_tessc'..

Test case 'dEQP-VK.spirv_assembly.type.scalar.i64.switch_tesse'..

Test case 'dEQP-VK.spirv_assembly.type.scalar.i64.switch_frag'..

Test case 'dEQP-VK.spirv_assembly.type.scalar.i64.switch_comp'..
  Pass (Output match with expected)

Test case 'dEQP-VK.spirv_assembly.type.scalar.u64.switch_vert'..

Test case 'dEQP-VK.spirv_assembly.type.scalar.u64.switch_tessc'..

Test case 'dEQP-VK.spirv_assembly.type.scalar.u64.switch_tesse'..

Test case 'dEQP-VK.spirv_assembly.type.scalar.u64.switch_frag'..

Test case 'dEQP-VK.spirv_assembly.type.scalar.u64.switch_comp'..
  Pass (Output match with expected)
```

### DIFF
--- a/spirv_glsl.cpp
+++ b/spirv_glsl.cpp
@@ -14768,13 +14768,13 @@ void CompilerGLSL::emit_block_chain(SPIRBlock &block)
 	case SPIRBlock::MultiSelect:
 	{
 		auto &type = expression_type(block.condition);
-		bool unsigned_case =
-		    type.basetype == SPIRType::UInt || type.basetype == SPIRType::UShort || type.basetype == SPIRType::UByte;
+		bool unsigned_case = type.basetype == SPIRType::UInt || type.basetype == SPIRType::UShort ||
+		                     type.basetype == SPIRType::UByte || type.basetype == SPIRType::UInt64;
 
 		if (block.merge == SPIRBlock::MergeNone)
 			SPIRV_CROSS_THROW("Switch statement is not structured");
 
-		if (type.basetype == SPIRType::UInt64 || type.basetype == SPIRType::Int64)
+		if (!backend.support_64bit_switch && (type.basetype == SPIRType::UInt64 || type.basetype == SPIRType::Int64))
 		{
 			// SPIR-V spec suggests this is allowed, but we cannot support it in higher level languages.
 			SPIRV_CROSS_THROW("Cannot use 64-bit switch selectors.");
@@ -14783,6 +14783,10 @@ void CompilerGLSL::emit_block_chain(SPIRBlock &block)
 		const char *label_suffix = "";
 		if (type.basetype == SPIRType::UInt && backend.uint32_t_literal_suffix)
 			label_suffix = "u";
+		else if (type.basetype == SPIRType::Int64 && backend.support_64bit_switch)
+			label_suffix = "l";
+		else if (type.basetype == SPIRType::UInt64 && backend.support_64bit_switch)
+			label_suffix = "ul";
 		else if (type.basetype == SPIRType::UShort)
 			label_suffix = backend.uint16_t_literal_suffix;
 		else if (type.basetype == SPIRType::Short)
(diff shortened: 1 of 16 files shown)


spirv_glsl.cpp
```diff
@@ -14865,11 +14865,11 @@ void CompilerGLSL::emit_block_chain(SPIRBlock &block)
 
 		size_t num_blocks = block_declaration_order.size();
 
-		const auto to_case_label = [](uint32_t literal, bool is_unsigned_case) -> string {
-			return is_unsigned_case ? convert_to_string(literal) : convert_to_string(int32_t(literal));
+		const auto to_case_label = [](uint64_t literal, bool is_unsigned_case) -> string {
+			return is_unsigned_case ? convert_to_string(literal) : convert_to_string(int64_t(literal));
```
HansKristian-Work, line 14869: This seems broken for 32-bit int case literals, we won't get sign extension I think?

etra0: I'll write some tests to check that!

HansKristian-Work: There's also the cases below for line 14913 where we have to emit some branches when "complex OpPhi" is used. There are some generated conditionals there that might be affected with 64-bit switch.

HansKristian-Work: It uses to_case_label, so it *should* be fine, but good to sanity check it ...

etra0: Just wrote a quick test and you're right, we're loosing sign extension for 32bit values. I'll include the test in the next commit too.

etra0:
I added a test for the sign extension, I'm not sure how to test the OpPhi case. If you want a rebase let me know!

spirv_glsl.cpp
```diff
@@ -14865,11 +14865,19 @@ void CompilerGLSL::emit_block_chain(SPIRBlock &block)
 
 		size_t num_blocks = block_declaration_order.size();
 
-		const auto to_case_label = [](uint32_t literal, bool is_unsigned_case) -> string {
-			return is_unsigned_case ? convert_to_string(literal) : convert_to_string(int32_t(literal));
+		// Precompute the is_32bit_int since it's block based, not per-case
+		// based so we then move the bool to the lambda.
+		const bool is_32bit_int = (type.basetype == SPIRType::Int);
+		const auto to_case_label = [=](uint64_t literal, bool is_unsigned_case) -> string {
```
HansKristian-Work, line 14871: Could just be `uint64_t literal, uint32_t width, bool is_unsigned_case` and then just roll out the 4 cases.

spirv_glsl.cpp
```diff
@@ -14865,11 +14865,19 @@ void CompilerGLSL::emit_block_chain(SPIRBlock &block)
 
 		size_t num_blocks = block_declaration_order.size();
 
-		const auto to_case_label = [](uint32_t literal, bool is_unsigned_case) -> string {
-			return is_unsigned_case ? convert_to_string(literal) : convert_to_string(int32_t(literal));
+		// Precompute the is_32bit_int since it's block based, not per-case
+		// based so we then move the bool to the lambda.
+		const bool is_32bit_int = (type.basetype == SPIRType::Int);
```
HansKristian-Work, line 14870: This misses the case where the switch condition is 16-bit.

HansKristian-Work (changes requested):

spirv_glsl.cpp
etra0: Maybe I should open other issue for this but, [the following spirv is valid](https://gist.github.com/etra0/0900c18651dfa61092a77e8f92d0ad3d#file-switch-spvasm), which spirv-cross will generate the frag just fine, but then it won't compile back since `glslc` requires explicitly that switch conditions are int, and in the spec I think it's specified that it should be `int` or `uint`

HansKristian-Work (approved):
LGTM.

Issue #1778, opened 2021-10-19
MSL & GLSL: Support OpAccessChain indexes that are out of bounds?


billhollings:
<sigh>

CTS includes tests that appear to play with the fact that `OpAccessChain` supports indexes that are out of bounds and possibly even negative, allowing sideways access to neighboring elements. 

WTF? Why? Why would this be designed in as an option? 🤦🏻‍♂️

Can we do this for MSL & GLSL?

The specific example here is access into a matrix element, with a final (vector swizzle) index of `4`, resulting in the following conversion error from SPIRV-Cross, for both MSL and GLSL:

```
Swizzle index out of range
```

Would replacing the swizzle to `xyza` with `[4]` automatically work for GLSL and MSL? It's quite probable the MSL compiler would rightly complain with a type index validation error if the `4` is emitted into MSL as a literal.

These CTS test fail as a result:
```
dEQP-VK.graphicsfuzz.two-loops-mat-add
```

SPIR-V file: 
[opaccesschain-out-of-bounds.spv.zip](https://github.com/KhronosGroup/SPIRV-Cross/files/7376540/opaccesschain-out-of-bounds.spv.zip)

Issue #1784, opened 2021-10-21
MSL & GLSL: Clamp vector element access to vector size


billhollings:
SPIR-V conversion to MSL fails Metal compilation when shader accesses vector element outside vector size.

The following SPIR-V code accesses a vector element outside the size of the vector:
```
      OpName %9 "v"
 %6 = OpTypeFloat 32
 %7 = OpTypeVector %6 3
 %8 = OpTypePointer Function %7
 %9 = OpVariable %8 Function
%10 = OpTypeInt 32 1
%26 = OpConstant %10 3
...
%48 = OpAccessChain %47 %9 %26
      OpStore %48 %46
```

The resulting MSL code is:
```
v.w = _43._GLF_uniform_float_values[0];
```

which is attempting to set the a vector value at index `3` (`.w`), even the vector only has 3 components (`float3`).

This is similar to #1778, but generalized to treat swizzle access outside size of vector as undefined behavior. Clamp swizzle to allow compilation, but invoke undefined behavior at runtime.

One possible solution is to pass the vector size to `CompilerGLSL::index_to_swizzle()` and clamp the returned value to the vector size (which would also fix what #1778 fixed. However, this wouldn't handle indexed element access using a constant value (`v[3]`). So perhaps an alternative that would solve both would be to clamp the size of any constant access to a vector component to the vector size, before using either `[n]` or `index_to_swizzle()`.

These CTS test fail as a result:
```
dEQP-VK.graphicsfuzz.cov-vector-illegal-index-never-executed
```

SPIR-V files:
[out-of-vector-bounds.spv.zip](https://github.com/KhronosGroup/SPIRV-Cross/files/7392841/out-of-vector-bounds.spv.zip)

etra0:
The CTS was a bit more complex than just checking the out of bound access so here's a minimal repro: [out-of-bounds.spv.gz](https://github.com/KhronosGroup/SPIRV-Cross/files/7513816/out-of-bounds.spv.gz)

HansKristian-Work:
This can probably be closed now?

etra0:
I can confirm master pass the CTS test in MoltenVK.

PR #1817, opened 2021-11-23
Clamp vector element access to vector size.

etra0:
I relied on the things discussed in https://github.com/KhronosGroup/SPIRV-Cross/issues/1778.

Not sure if I should add a special case for negative numbers. I'm also very open
to discuss the implementation to be extra sure this is the way to go.

### DIFF
--- a/spirv_glsl.cpp
+++ b/spirv_glsl.cpp
@@ -8894,29 +8894,36 @@ string CompilerGLSL::access_chain_internal(uint32_t base, const uint32_t *indice
 				                                       is_packed);
 			}
 
-			if (is_literal && !is_packed && !row_major_matrix_needs_conversion)
+			if (is_literal)
 			{
-				expr += ".";
-				expr += index_to_swizzle(index);
+				bool out_of_bounds = (index >= type->vecsize);
+
+				if (!is_packed && !row_major_matrix_needs_conversion)
+				{
+					expr += ".";
+					expr += index_to_swizzle(out_of_bounds ? 0 : index);
+				}
+				else
+				{
+					// For packed vectors, we can only access them as an array, not by swizzle.
+					expr += join("[", out_of_bounds ? 0 : index, "]");
+				}
 			}
 			else if (ir.ids[index].get_type() == TypeConstant && !is_packed && !row_major_matrix_needs_conversion)
 			{
 				auto &c = get<SPIRConstant>(index);
+				bool out_of_bounds = (c.scalar() >= type->vecsize);
+
 				if (c.specialization)
 				{
 					// If the index is a spec constant, we cannot turn extract into a swizzle.
-					expr += join("[", to_expression(index), "]");
+					expr += join("[", out_of_bounds ? "0" : to_expression(index), "]");
 				}
 				else
 				{
 					expr += ".";
-					expr += index_to_swizzle(c.scalar());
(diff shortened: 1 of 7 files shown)


spirv_glsl.cpp
```diff
@@ -8889,30 +8889,35 @@ string CompilerGLSL::access_chain_internal(uint32_t base, const uint32_t *indice
...
-				expr += index_to_swizzle(index);
+				if (!is_packed && !row_major_matrix_needs_conversion)
+				{
+					expr += ".";
+					expr += index_to_swizzle(index);
+				}
+				else
+				{
+					// For packed vectors, we can only access them as an array, not by swizzle.
+					expr += join("[", index, "]");
```
HansKristian-Work, line 8902: Can a literal be out of bounds?

spirv_glsl.cpp
```diff
@@ -8889,30 +8889,35 @@ string CompilerGLSL::access_chain_internal(uint32_t base, const uint32_t *indice
...
+				}
 			}
 			else if (ir.ids[index].get_type() == TypeConstant && !is_packed && !row_major_matrix_needs_conversion)
 			{
 				auto &c = get<SPIRConstant>(index);
+				bool out_of_bounds = (c.scalar() >= type->vecsize);
```
HansKristian-Work, line 8908: This can get spicy with OpSpecConstantOp. In that we cannot directly look at the scalar value, but the MSL compiler might still complain ... You might want to add a test for that and consider looking at `Compiler::evaluate_spec_constant_u32()`.

HansKristian-Work: Or hm ... we won't know at compile time what the result is going to be. Perhaps do a dynamic "out_of_bounds" check on the spec constant op, e.g.

```
foo[spec_constant_op_result >= size ? 0 : spec_constant_op_result]
```

HansKristian-Work:
> special case for negative numbers

Do we need that? the u32 scalar result will be > 0x80000000, so it should trigger any check.

spirv_glsl.cpp
etra0: It seems like the `OpSpecConstantOp` is of type `TypeConstantOp` so it would be excluded of that branch specifically (it would fall in the `else` condition).

I was targeting the `OpAccessChain` mainly, but now that I think about it, when unrolling the indexes there's no check for `OpAccessChain` or `OpInBoundsAccessChain`.  I would say that adding the dynamic check would be necessary only in the latter but I'm willing to change it too.

Let me know what you think!

HansKristian-Work: Wouldn't it be necessary in both? InBoundsAccessChain guarantees that it's in bound, but I guess only if it's executed.

HansKristian-Work:
FWIW, I'm going on vacation after today, so if it's time critical to land this before holidays @billhollings or @cdavis5e should be able to review and merge. The PR seems fine as is.

etra0:
Sounds good. I added the literal check just in case (I didn't managed to create a test for it -- couldn't find more detailed info about the literal bit), tested it in linux and macos for regressions and nothing so far. I've also tested the fuzz test in https://github.com/KhronosGroup/SPIRV-Cross/issues/1784 and also passes.

No rush in merging this. Have some happy holidays!

HansKristian-Work (approved):

PR #1822, opened 2021-12-09
test_shaders: Add the option to generate diff instead.

etra0:
I personally feel like displaying a `diff` instead of the output is useful in some cases to identify the differences faster.


HansKristian-Work (approved):
Seems useful, thanks.

PR #1823, opened 2021-12-16
Check how many times a variable is forwarded for binary ops.

etra0:
Avoid the abuse of forwarded variables by limiting it at 255.
MSL throws a compile error and I think there's no harm by limiting it at 255. No regression was found and it also passes the following CTS tests:
```
Test case 'dEQP-VK.spirv_assembly.instruction.compute.spirv_ids_abuse.lots_ids'..
  Pass (Output match with expected)

Test case 'dEQP-VK.spirv_assembly.instruction.graphics.spirv_ids_abuse.lots_ids_vert'..

Test case 'dEQP-VK.spirv_assembly.instruction.graphics.spirv_ids_abuse.lots_ids_tessc'..

Test case 'dEQP-VK.spirv_assembly.instruction.graphics.spirv_ids_abuse.lots_ids_tesse'..

Test case 'dEQP-VK.spirv_assembly.instruction.graphics.spirv_ids_abuse.lots_ids_frag'..
```


etra0:
In the case this matters, the test can be heavily reduced to 400 lines, those should be enough to check regressions in the future since we would have two expressions of sums (i.e. it breaks just once the forwarding chain).

In the testing suite we don't get a significant slowdown, it's more related to the size.

billhollings:
Should fix issue #1823. Need to confirm.

spirv_glsl.cpp
```diff
@@ -5847,6 +5847,34 @@ bool CompilerGLSL::emit_complex_bitcast(uint32_t result_type, uint32_t id, uint3
 	return true;
 }
 
+static bool check_forward_limit_exceeded(uint32_t op0, uint32_t op1)
```
HansKristian-Work, line 5850: This is not thread safe and way too hacky. I'll try to see how to best solve this.

etra0: Hmm, I think it would be possible to avoid using static and use context dependent variables (like, add them to the CompilerGLSL class), but I agree this is a hacky way.

HansKristian-Work: I think we can use the expression use counter system somehow.

HansKristian-Work:
I'll try to make time for this next week.

Issue #1762, opened 2021-10-07
MSL: gl_TessCoord mis-declared as a float2 instead of correct float3


billhollings:
SPIR-V conversion to MSL fails compilation due to `gl_TessCoord` mis-declared as a `float2` instead of correct `float3`:

These CTS test fail as a result:
```
dEQP-VK.spirv_assembly.instruction.graphics.cross_stage.basic_type.flat
dEQP-VK.spirv_assembly.instruction.graphics.cross_stage.basic_type.no_perspective
dEQP-VK.spirv_assembly.instruction.graphics.cross_stage.basic_type.relaxedprecision
dEQP-VK.spirv_assembly.instruction.graphics.cross_stage.interface_blocks.flat
dEQP-VK.spirv_assembly.instruction.graphics.cross_stage.interface_blocks.no_perspective
dEQP-VK.spirv_assembly.instruction.graphics.cross_stage.interface_blocks.relaxedprecision
```
SPIR-V file: 
[gl_TessCoord_as_float2_bad.spv.zip](https://github.com/KhronosGroup/SPIRV-Cross/files/7304932/gl_TessCoord_as_float2_bad.spv.zip)

etra0:
Hmm, from what I can see, the only criteria to choose between `float2` and `float3` is the execution mode:
https://github.com/KhronosGroup/SPIRV-Cross/blob/e9cc6403341baf0edd430a4027b074d0a06b782f/spirv_msl.cpp#L14583-L14584

Looking at the uploaded SPIR-V and a test that expects a `float2` instead of a `float3` (for example: https://github.com/KhronosGroup/SPIRV-Cross/blob/master/shaders-msl/asm/tese/unnamed-builtin-array.asm.tese) I can't see a specific factor besides the index access and the execution mode being `SpacingFractionalEven` vs the uploaded file which is `SpacingEqual`. 

Adding the `|| execution.flags.get(ExecutionModeSpacingEqual)` to the condition makes all test pass with the uploaded file included but I **seriously** doubt that's the correct way of doing it, and I don't have much experience with MSL either.

Do you have any idea what other criteria could be used to choose the correct type?

cfnptr:
Looks like the problem is that the SPIRV-MSL translator sets all three `gl_TessCoord` values (x, y, z) no matter what type of patch (Triangles or Quads) is used in the shader.

[...]

PR #1824, opened 2021-12-17
Fix incorrect MSL gl_TessCoord variable type

cfnptr:
Fix for this issue #1762.

Adds "In" postfix to the "gl_TessCoord" on entry point arguments decompilation stage if built in type is "BuiltInTessCoord" and execution mode is "ExecutionModeQuads". And declares shadow  variable "gl_TessCoord" on entry point declarations.

I have compiled and tested it with VK-GL-CTS, looks like it now works as expected.

### DIFF
--- a/reference/opt/shaders-msl/asm/tese/unnamed-builtin-array.asm.tese
+++ b/reference/opt/shaders-msl/asm/tese/unnamed-builtin-array.asm.tese
@@ -55,8 +55,9 @@ struct main0_patchIn
     float2 gl_TessLevelInner [[attribute(1)]];
 };
 
-[[ patch(quad, 0) ]] vertex main0_out main0(main0_patchIn patchIn [[stage_in]], float2 gl_TessCoord [[position_in_patch]])
+[[ patch(quad, 0) ]] vertex main0_out main0(main0_patchIn patchIn [[stage_in]], float2 gl_TessCoordIn [[position_in_patch]])
 {
+    float3 gl_TessCoord = float3(gl_TessCoordIn.x, gl_TessCoordIn.y, 0.0);
     main0_out out = {};
     spvUnsafeArray<float, 2> gl_TessLevelInner = {};
     spvUnsafeArray<float, 4> gl_TessLevelOuter = {};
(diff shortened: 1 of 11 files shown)


cfnptr:
Hmm, looks like test fails because fix is not checking if `gl_TessCoord` is present in function arguments.

billhollings:
> Hmm, looks like test fails because fix is not checking if `gl_TessCoord` is present in the function arguments.

SPIRV-Cross has a large number of unit tests, and the changes are breaking some of them.

When making changes to SPIRV-Cross, you always need to run `./test_shaders.sh`, and if the revised reference shaders look good, replace them with the new reference shaders. The easiest way to do this is to simply delete the old reference shader and let the test case automatically regenerate the new version.

You can find more complete info in the [_Contributing_](https://github.com/KhronosGroup/SPIRV-Cross#contributing) section of the `README.md`.

If more than one reference shader needs to be replaced, you'll need to run `./test_shaders.sh` iteratively, which can take a while. Since you're only concerned with the MSL shaders, I typically temporarily edit `test_shaders.sh` to comment out all but the lines that run the MSL tests.

cfnptr:
I have fixed problem with `gl_TessCoord` presence in the function arguments.
Now all tests from the `./test_shaders.sh` passes on my machine.

billhollings (approved):
LGTM now.

cdavis5e:
> > Hmm, looks like test fails because fix is not checking if `gl_TessCoord` is present in the function arguments.
> 
> SPIRV-Cross has a large number of unit tests, and the changes are breaking some of them.
> 
> When making changes to SPIRV-Cross, you always need to run `./test_shaders.sh`, and if the revised reference shaders look good, replace them with the new reference shaders. The easiest way to do this is to simply delete the old reference shader and let the test case automatically regenerate the new version.
> 
> You can find more complete info in the [_Contributing_](https://github.com/KhronosGroup/SPIRV-Cross#contributing) section of the `README.md`.
> 
> If more than one reference shader needs to be replaced, you'll need to run `./test_shaders.sh` iteratively, which can take a while. Since you're only concerned with the MSL shaders, I typically temporarily edit `test_shaders.sh` to comment out all but the lines that run the MSL tests.

Or, you can run `update_test_shaders.sh`, which replaces the shaders automatically. After it's finished, you can inspect the diffs with `git diff`.

billhollings:
Fixes #1762.

Issue #1775, opened 2021-10-19
MSL & GLSL: Conversion error - Over 3 compilation loops detected.  


billhollings:
SPIR-V conversion to MSL fails compilation because of an internal "Over 3 compilation loops detected" error.

These CTS test fail as a result:
```
dEQP-VK.graphicsfuzz.mat-array-deep-control-flow
dEQP-VK.graphicsfuzz.spv-dead-break-and-unroll
```

SPIR-V files:
[over_3_compilation_loops-deep-control-flow.spv.zip](https://github.com/KhronosGroup/SPIRV-Cross/files/7376006/over_3_compilation_loops-deep-control-flow.spv.zip)
[over_3_compilation_loops-dead-break-and-unroll.spv.zip](https://github.com/KhronosGroup/SPIRV-Cross/files/7376008/over_3_compilation_loops-dead-break-and-unroll.spv.zip)

etra0:
Did a quick test by relaxing the condition to 4 instead of 3 and both test passes. This would make the CTS pass and I don't see any regression.

It would be a bit of a silly PR, but would help with the CTS conformance. Any opinions?

billhollings:
> Did a quick test by relaxing the condition to 4 instead of 3 and both test passes. This would make the CTS pass and I don't see any regression.
> 
> It would be a bit of a silly PR, but would help with the CTS conformance. Any opinions?

@etra0 Thanks for checking that option out.

@HansKristian-Work I think this idea is best reviewed by you. Can you comment on why 4 loops might work instead of 3 in this case, and whether there is anything magical about setting the limit to 3, or dangerous about raising it to 4?

HansKristian-Work:
Yes, this is kinda hairy ... I think we'll need to forgo the idea of a fixed loop count and consider that certain force_recompiles can be considered to guarantee forward progress.

PR #1825, opened 2021-12-22
Relax compilation loop detection.

etra0:
As discussed in the issue https://github.com/KhronosGroup/SPIRV-Cross/issues/1775, I didn't noticed any regression by augmenting the limit by one, and it feels a bit unnecessary but it helps with CTS compliance by passing two new tests:

```
Test case 'dEQP-VK.graphicsfuzz.mat-array-deep-control-flow'..
  Pass (Pass)

Test case 'dEQP-VK.graphicsfuzz.spv-dead-break-and-unroll'..
  Pass (Pass)
```


HansKristian-Work:
This needs to be investigated why 4 loops works but not 3. That is a bug.

PR #1862, opened 2022-02-12
HLSL: Make --flatten-ubo work correctly

### LINKED ISSUE #1827: HLSL: packoffset is produced for SM3.0 but fxc does not honor it
Hi, I'm still developing Direct3D 9 application in order to make my program run on old PCs. I'm trying to port my shader written for Vulkan to SM3.0, and facing a problem where constants are not packed as the code produced by spirv-cross is saying like `packoffset(c0.z)`. This is because fxc ignores `packoffset` if the shader profile is SM3.0, so it's not a bug. But I think it would be nicer if spirv-cross had an option to combine multiple constants into one and make the constant layout identical to as if `packoffset` is applied. Here's example:

The source code

```hlsl
struct MyShaderParams {
    float2 alpha;
    float beta;
};
[[vk::push_constant]]
MyShaderParams params;
float4 main() : SV_Position {
    return float4(params.alpha, params.beta, 1.0);
}
```

passing to dxc and spirv-cross, we get

```hlsl
cbuffer type_PushConstant_MyShaderParams
{
    float2 params_alpha : packoffset(c0);
    float params_beta : packoffset(c0.z);
};

uniform float4 gl_HalfPixel;

static float4 gl_Position;
struct SPIRV_Cross_Output
{
    float4 gl_Position : POSITION;
};

void vert_main()
{
    gl_Position = float4(params_alpha, params_beta, 1.0f);
    gl_Position.x = gl_Position.x - gl_HalfPixel.x * gl_Position.w;
    gl_Position.y = gl_Position.y + gl_HalfPixel.y * gl_Position.w;
}

SPIRV_Cross_Output main()
{
    vert_main();
    SPIRV_Cross_Output stage_output;
    stage_output.gl_Position = gl_Position;
    return stage_output;
}
```
It says `float params_beta : packoffset(c0.z);` but fxc ignores `packoffset`, so it's assigned to c1 actually.

In my idea, `params_alpha` and `params_beta` will be combined into one member variable. The output will be like below:

```hlsl
cbuffer type_PushConstant_MyShaderParams
{
    float3 params_alpha_and_params_beta;
};

uniform float4 gl_HalfPixel;

static float4 gl_Position;
struct SPIRV_Cross_Output
{
    float4 gl_Position : POSITION;
};

void vert_main()
{
    gl_Position = float4(params_alpha_and_params_beta.xy, params_alpha_and_params_beta.z, 1.0f);
    gl_Position.x = gl_Position.x - gl_HalfPixel.x * gl_Position.w;
    gl_Position.y = gl_Position.y + gl_HalfPixel.y * gl_Position.w;
}

SPIRV_Cross_Output main()
{
    vert_main();
    SPIRV_Cross_Output stage_output;
    stage_output.gl_Position = gl_Position;
    return stage_output;
}
```

Is there anyone else interested in?

skhrshin:
Resolves #1827.

Tests are taken from existing ones for GLSL.

One thing I'm worrying about is a redundant local variable which exists in reference/shaders-hlsl/flatten/struct.rowmajor.flatten.vert. This is not easy to eliminate because the first local variable is emitted from AccessChain instruction and the second one is from Load instruction.


spirv_glsl.cpp
```diff
@@ -10239,7 +10244,14 @@ void CompilerGLSL::emit_instruction(const Instruction &instruction)
 		// If an expression is mutable and forwardable, we speculate that it is immutable.
 		AccessChainMeta meta;
 		bool ptr_chain = opcode == OpPtrAccessChain;
-		auto e = access_chain(ops[2], &ops[3], length - 3, get<SPIRType>(ops[0]), &meta, ptr_chain);
+		auto target_type = get<SPIRType>(ops[0]);
+		auto e = access_chain(ops[2], &ops[3], length - 3, target_type, &meta, ptr_chain);
+
+		if (should_emit_temporary(ops[2], target_type))
+		{
+			statement(declare_temporary(ops[0], ops[1]), move(e), ";");
```
HansKristian-Work, line 10252: This is incorrect. The more correct approach here is to block any forwarding in the SPIRExpression instead (see last argument). You can use OpCompositeConstruct as a guideline here.

spirv_hlsl.cpp
```diff
@@ -4412,6 +4420,51 @@ void CompilerHLSL::emit_access_chain(const Instruction &instruction)
...
+}
+
+std::string CompilerHLSL::flattened_access_chain_struct(uint32_t base, const uint32_t *indices, uint32_t count,
+                                                        const SPIRType &target_type, uint32_t offset)
+{
+	std::string expr;
+
+	expr += "{";
+
+	for (uint32_t i = 0; i < uint32_t(target_type.member_types.size()); ++i)
```
HansKristian-Work, line 4435: It might be possible to just modify CompilerGLSL implementation with a check to backend.can_declare_struct_inline. This seems like duplicated code.

HansKristian-Work (changes requested):
See comments.

spirv_glsl.cpp
skhrshin: Do you mean that the statement shouldn't be emitted from OpAccessChain? I have ever tried making `SPIRExpression::immutable` `true` and that results emitting not stuffs like `Light _51 = {UBO[_96 * 2 + 4].xyz, UBO[_96 * 2 + 4].w, UBO[_96 * 2 + 5]};` but like `float3 _101 = ({UBO[_96 * 2 + 4].xyz, UBO[_96 * 2 + 4].w, UBO[_96 * 2 + 5]}).Position;`. The OpCompositeConstruct case in `CompilerGLSL::emit_instruction` is calling `emit_op`, and `emit_op` calls `statement` if the given `forwarding` is `false`. It looks not using the last argument of `SPIRExpression` to me. I didn't understand what I should refer to.

HansKristian-Work: Right, emit_op() is probably the thing to call here. This is so that end up binding the expression to a temporary, which is the only way HLSL can generate struct composites.

skhrshin:
I changed the code so that it sees `backend.can_declare_struct_inline`, and calls `emit_op` instead of `statement`. Sorry if I have not understood what you are pointing out.

spirv_glsl.cpp
```diff
@@ -10239,9 +10244,13 @@ void CompilerGLSL::emit_instruction(const Instruction &instruction)
 		// If an expression is mutable and forwardable, we speculate that it is immutable.
 		AccessChainMeta meta;
 		bool ptr_chain = opcode == OpPtrAccessChain;
-		auto e = access_chain(ops[2], &ops[3], length - 3, get<SPIRType>(ops[0]), &meta, ptr_chain);
+		auto target_type = get<SPIRType>(ops[0]);
+		auto e = access_chain(ops[2], &ops[3], length - 3, target_type, &meta, ptr_chain);
 
-		auto &expr = set<SPIRExpression>(ops[1], move(e), ops[0], should_forward(ops[2]));
+		auto &expr = (!backend.can_declare_struct_inline && flattened_buffer_blocks.count(ops[2]) &&
```
HansKristian-Work, line 10250: Make the boolean condition here into a temporary. Also add a comment about the rationale, that we might not be able to build composites inline on some backends, and they have to be bound to a temporary.

spirv_glsl.cpp
```diff
@@ -10239,9 +10244,13 @@ void CompilerGLSL::emit_instruction(const Instruction &instruction)
 		// If an expression is mutable and forwardable, we speculate that it is immutable.
 		AccessChainMeta meta;
 		bool ptr_chain = opcode == OpPtrAccessChain;
-		auto e = access_chain(ops[2], &ops[3], length - 3, get<SPIRType>(ops[0]), &meta, ptr_chain);
+		auto target_type = get<SPIRType>(ops[0]);
```
HansKristian-Work, line 10247: `auto &target_type`

spirv_glsl.cpp
```diff
@@ -10239,9 +10244,13 @@ void CompilerGLSL::emit_instruction(const Instruction &instruction)
...
+		auto e = access_chain(ops[2], &ops[3], length - 3, target_type, &meta, ptr_chain);
 
-		auto &expr = set<SPIRExpression>(ops[1], move(e), ops[0], should_forward(ops[2]));
+		auto &expr = (!backend.can_declare_struct_inline && flattened_buffer_blocks.count(ops[2]) &&
+		              target_type.basetype == SPIRType::Struct) ?
+                         emit_op(ops[0], ops[1], e, false) :
```
HansKristian-Work, line 10252: `move(e)`.

HansKristian-Work (approved):
LGTM. To merge when CI passes.

PR #1877, opened 2022-03-02
Suppress -Wunqualified-std-cast-call.

mysterymath:
Wunqualified-std-cast-call was added to Clang in
https://reviews.llvm.org/D119670, which using this project when building
with -Werror. This change ignores this error as a stopgap; eventually
the unqualified std::move() instances should be qualified.


mysterymath:
Upon further reflection, it's easier to just fix these.

PR #1883, opened 2022-03-03
Support OpTerminateInvocation as synonym of the deprecated OpKill.

RandomShaper:
Reference: https://www.khronos.org/registry/SPIR-V/specs/unified1/SPIRV.html#OpKill


HansKristian-Work:
I already merged this exact fix earlier today: https://github.com/KhronosGroup/SPIRV-Cross/pull/1880.

PR #1886, opened 2022-03-04
HLSL: Fix Helper Invocation wrongly allowed in compute.

RandomShaper:
I'm sorry. I just realized this.

### DIFF
--- a/spirv_hlsl.cpp
+++ b/spirv_hlsl.cpp
@@ -729,7 +729,7 @@ void CompilerHLSL::emit_builtin_inputs_in_struct()
 			break;
 
 		case BuiltInHelperInvocation:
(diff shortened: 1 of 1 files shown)


spirv_hlsl.cpp
```diff
@@ -5603,7 +5603,7 @@ void CompilerHLSL::emit_instruction(const Instruction &instruction)
 	}
 
 	case OpIsHelperInvocationEXT:
-		if (hlsl_options.shader_model < 50 || (get_entry_point().model != ExecutionModelFragment && get_entry_point().model != ExecutionModelGLCompute))
+		if (hlsl_options.shader_model < 50 || get_entry_point().model != ExecutionModelFragment)
```
HansKristian-Work, line 5606: Also remove it from the other place where this is checked, i.e. the builtin. This is the EXT opcode.

RandomShaper: Done.

HansKristian-Work (approved):

Issue #1763, opened 2021-10-07
MSL: Error trying to cast spvUnsafeArray<float, 3> to float[3] array


billhollings:
SPIR-V conversion to MSL fails compilation in an attempt to cast `spvUnsafeArray<float, 3>` to `float[3]` array:

These CTS test fail as a result:
```
dEQP-VK.spirv_assembly.instruction.graphics.decoration_group.same_decoration_group_on_multiple_types_vert
dEQP-VK.spirv_assembly.instruction.graphics.decoration_group.same_decoration_group_on_multiple_types_tessc
dEQP-VK.spirv_assembly.instruction.graphics.decoration_group.same_decoration_group_on_multiple_types_tesse
dEQP-VK.spirv_assembly.instruction.graphics.decoration_group.same_decoration_group_on_multiple_types_frag
```
SPIR-V file: 
[bad_cast_from_spvUnsafeArray.spv.zip](https://github.com/KhronosGroup/SPIRV-Cross/files/7305033/bad_cast_from_spvUnsafeArray.spv.zip)

cfnptr:
Decompiled `bad_cast_from_spvUnsafeArray.spv.zip` MSL shader:

```
#include <metal_stdlib>
#include <simd/simd.h>

using namespace metal;

template<typename T, size_t Num>
struct spvUnsafeArray
{
    T elements[Num ? Num : 1];
    
    thread T& operator [] (size_t pos) thread
    {
        return elements[pos];
    }
    constexpr const thread T& operator [] (size_t pos) const thread
    {
        return elements[pos];
    }
    
    device T& operator [] (size_t pos) device
    {
        return elements[pos];
    }
    constexpr const device T& operator [] (size_t pos) const device
    {
        return elements[pos];
    }
    
    constexpr const constant T& operator [] (size_t pos) const constant
    {
        return elements[pos];
    }
    
    threadgroup T& operator [] (size_t pos) threadgroup
    {
        return elements[pos];
    }
    constexpr const threadgroup T& operator [] (size_t pos) const threadgroup
    {
        return elements[pos];
    }
};

struct _14
{
    float _m0[3];
};

struct _15
{
    float _m0[3];
};

constant spvUnsafeArray<float, 3> _93 = spvUnsafeArray<float, 3>({ 1.0, 2.0, 1.0 });
constant spvUnsafeArray<float, 3> _94 = spvUnsafeArray<float, 3>({ -1.0, -2.0, -1.0 });

[...]

PR #1894, opened 2022-03-16
Add corner case for structs containing templated arrays

etra0:
This adds a special corner case when there's a struct with only one member that contains
the offset decorator set to 0, we can ignore said decorator so we can emit
the templated array.

I'm not completely sure about this approach but any feedback is welcomed.

**NOTE:** This needs to run `./test_shaders.sh --update` but I've not pushed them to make it easier to review.

This fixes https://github.com/KhronosGroup/SPIRV-Cross/issues/1763

```
Test case 'dEQP-VK.spirv_assembly.instruction.graphics.decoration_group.same_decoration_group_on_multiple_types_vert'..

Test case 'dEQP-VK.spirv_assembly.instruction.graphics.decoration_group.same_decoration_group_on_multiple_types_tessc'..

Test case 'dEQP-VK.spirv_assembly.instruction.graphics.decoration_group.same_decoration_group_on_multiple_types_tesse'..

Test case 'dEQP-VK.spirv_assembly.instruction.graphics.decoration_group.same_decoration_group_on_multiple_types_frag'..
```

### DIFF
--- a/reference/opt/shaders-msl/asm/frag/offset-decorator-can-be-ignored.asm.frag
+++ b/reference/opt/shaders-msl/asm/frag/offset-decorator-can-be-ignored.asm.frag
@@ -1,0 +1,9 @@
+#include <metal_stdlib>
+#include <simd/simd.h>
+
+using namespace metal;
+
+fragment void main0()
+{
+}
+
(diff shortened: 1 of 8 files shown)


HansKristian-Work:
I think this is solved much more directly by: https://github.com/KhronosGroup/SPIRV-Cross/pull/1897.

etra0:
Got it, thanks!

PR #1896, opened 2022-03-22
Use types have same widths in loop condition.

Roy-AMD:
In case comparisons between types of different widths in a loop condition
caused the loop to behave unexpectedly.


spirv_msl.cpp
```diff
@@ -4953,7 +4953,7 @@ void CompilerMSL::emit_custom_functions()
 			{
 				uint32_t dimensions = spv_func - SPVFuncImplArrayCopyMultidimBase;
 				string tmp = "template<typename T";
-				for (uint8_t i = 0; i < dimensions; i++)
+				for (uint32_t i = 0; i < dimensions; i++)
 				{
 					tmp += ", uint ";
 					tmp += 'A' + i;
```
HansKristian-Work, line 4959: Should probably explicitly cast this to `char('A' + i)`. Implicit type promotion here seems dicey.

spirv_msl.cpp
```diff
@@ -4953,7 +4953,7 @@ void CompilerMSL::emit_custom_functions()
 			{
 				uint32_t dimensions = spv_func - SPVFuncImplArrayCopyMultidimBase;
 				string tmp = "template<typename T";
-				for (uint8_t i = 0; i < dimensions; i++)
+				for (uint32_t i = 0; i < dimensions; i++)
```
HansKristian-Work, line 4956: Which problem does this solve? It makes sense to have uint32_t here since dimensions is uint32_t, but I'd like to understand why.

Roy-AMD: Use  different widths in a loop condition may cause some issue, and if keep uint32_t here, the code can't pass codeQL scan.
In addition to, the `SPVFuncImplArrayCopyMultidimBase` is `SPVFuncImpl` enum, and the enum have a 32bit length, if we keep the enum growth, and no where to check it, the value of dimensions may overside of uint8::max. it will cause more issue. 

May be it's not the best way to modify uint8_t to uint32_t, we also can consider to use `SPVFuncImpl : uint8_t`, and keep `dimensions` as uint8_t here.

spirv_msl.cpp
Roy-AMD: Yes, and it may be need a `assert(i <= 0xff);`

HansKristian-Work (changes requested):
Needs `char('A' + i)` fix and potentially assert check as well, other than that, it's fine.

Roy-AMD:
> Needs `char('A' + i)` fix and potentially assert check as well, other than that, it's fine.

how about the new change? we keep all the type to uint8_t.

HansKristian-Work (approved):

PR #1899, opened 2022-03-23
Allow volatile vars that are not builtins to be forwarded.

billhollings:
This fixes a regression introduced by 93b0dc7, where all volatile variables were not allowed to be forwarded. This doesn't work well for volatile memory object variables like images or buffer blocks, because it forces local variables
to be defined, which is unnecessary, and sometimes results in wrong types.

This patch restricts volatile builtin variables (eg. `HelperInvocation`) from being forwarded, but allows other volatile variables to be forwarded, as before.

### DIFF
--- a/spirv_glsl.cpp
+++ b/spirv_glsl.cpp
@@ -9600,8 +9600,8 @@ bool CompilerGLSL::should_forward(uint32_t id) const
 	auto *var = maybe_get<SPIRVariable>(id);
 	if (var)
 	{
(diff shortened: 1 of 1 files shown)


HansKristian-Work (approved):
I guess this is okay since it fixes regressions, but it does break volatile loads from Volatile buffer objects. Those loads cannot be forwarded. I guess the proper fix would be more careful about which types we're blocking forward for.

PR #1924, opened 2022-04-26
cli: display missing memory qualifiers for reflect and dump-resources

stefalie:
The CLI options `--dump-resources` and `--reflect` seem to be missing some of the memory qualifiers available in GLSL:

- `--dump-resources` doesn't print `restrict`, `coherent`, and `volatile`.
- `--reflect` only prints qualifiers for SSBOs. I don't think there is any harm in displaying that info for other resources as well. For images, for example, knowing `readonly`/`writeonly` is useful when trying to do automatic synchronization/barriers.

This PR adds these missing qualifiers to the output of the two mentioned command lines tools.


HansKristian-Work:
The change makes sense, but the reference test output needs to be updated.

stefalie:
Thx. I fixed the test reference outputs and locally everything seems to pass.

stefalie:
Argh, adding that one missing line directly with "edit in Github" also added and newline at the end of file. Sorry about that, will take care it tonight.

PR #1954, opened 2022-06-03
Fix missing function in README sample

vkaytsanov:


HansKristian-Work:
LGTM, but I don't see the CLA being signed. Just a technical formality. I don't see the popup for that though.

HansKristian-Work:
LGTM, but I don't see the CLA being signed. Just a technical formality. I don't see the popup for that though.

vkaytsanov:
> LGTM, but I don't see the CLA being signed. Just a technical formality. I don't see the popup for that though.

It should be signed now.

PR #1961, opened 2022-06-08
HLSL: Implement bit casts between Half and UShort.

RandomShaper:
I'm trying to get AMD FidelityFX shader through SPIRV-Cross. For some reasons, I have to compile to SPIR-V first. (I know it'd be better to use the HLSL version directly, but for now this flow is more convenient for me.)

The matter is that this function in the generated HLSL gives compile errors:
```
half APrxLoRsqH1(half a)
{
    uint16_t param = 22947u;
    uint16_t param_1 = 1u;
    return (AW1_x(param) - ((a) >> AW1_x(param_1)));
}
```
> `error: int or unsigned int type required`
> `   return (AW1_x(param) - ((a) >> AW1_x(param_1)));`

For reference, this is the corresponding SPIR-V dissasembly of the function:
```
%APrxLoRsqH1_f161_ = OpFunction %half None %32 ; 0x000041a0
        %a_7 = OpFunctionParameter %_ptr_Function_half ; 0x000041b4
         %90 = OpLabel ; 0x000041c0
    %param_9 = OpVariable %_ptr_Function_ushort Function ; 0x000041c8
   %param_10 = OpVariable %_ptr_Function_ushort Function ; 0x000041d8
               OpStore %param_9 %ushort_22947 ; 0x000041e8
        %328 = OpFunctionCall %ushort %AW1_x_u161_ %param_9 ; 0x000041f4
        %329 = OpLoad %half %a_7 ; 0x00004208
        %330 = OpBitcast %ushort %329 ; 0x00004218
               OpStore %param_10 %ushort_1 ; 0x00004228
        %333 = OpFunctionCall %ushort %AW1_x_u161_ %param_10 ; 0x00004234
        %334 = OpShiftRightLogical %ushort %330 %333 ; 0x00004248
        %335 = OpISub %ushort %328 %334 ; 0x0000425c
        %336 = OpBitcast %half %335 ; 0x00004270
               OpReturnValue %336 ; 0x00004280
               OpFunctionEnd ; 0x00004288
```

This PR extends the support for bit casts so the function at hand is generated correctly for HLSL, like this:
```
half APrxLoRsqH1(half a)
{
    uint16_t param = 22947u;
    uint16_t param_1 = 1u;
    return f16tof32(AW1_x(param) - (f32tof16(a) >> AW1_x(param_1)));
}
```

The compiler will need to be passed the `-enable-16bit-types` switch, which will only work on SM 6.2+, but the intrinsics used were introduced in SM 5.0, so that's the one checked for.

### DIFF
--- a/spirv_hlsl.cpp
+++ b/spirv_hlsl.cpp
@@ -3598,6 +3598,18 @@ string CompilerHLSL::bitcast_glsl_op(const SPIRType &out_type, const SPIRType &i
 		}
 		return "spvPackFloat2x16";
 	}
+	else if (out_type.basetype == SPIRType::UShort && in_type.basetype == SPIRType::Half)
+	{
+		if (hlsl_options.shader_model < 40)
+			SPIRV_CROSS_THROW("Half to UShort requires Shader Model 4.");
+		return "(" + type_to_glsl(out_type) + ")f32tof16";
+	}
+	else if (out_type.basetype == SPIRType::Half && in_type.basetype == SPIRType::UShort)
+	{
+		if (hlsl_options.shader_model < 40)
+			SPIRV_CROSS_THROW("UShort to Half requires Shader Model 4.");
+		return "(" + type_to_glsl(out_type) + ")f16tof32";
+	}
 	else
 		return "";
 }


spirv_hlsl.cpp
```diff
@@ -3598,6 +3598,18 @@ string CompilerHLSL::bitcast_glsl_op(const SPIRType &out_type, const SPIRType &i
 		}
 		return "spvPackFloat2x16";
 	}
+	else if (out_type.basetype == SPIRType::UShort && in_type.basetype == SPIRType::Half)
+	{
+		if (hlsl_options.shader_model < 50)
+			SPIRV_CROSS_THROW("Half to UShort requires Shader Model 5.");
```
HansKristian-Work, line 3604: This requires SM 6.2 I think.

spirv_hlsl.cpp
```diff
@@ -3598,6 +3598,18 @@ string CompilerHLSL::bitcast_glsl_op(const SPIRType &out_type, const SPIRType &i
...
+		return "f32tof16";
+	}
+	else if (out_type.basetype == SPIRType::Half && in_type.basetype == SPIRType::UShort)
+	{
+		if (hlsl_options.shader_model < 50)
+			SPIRV_CROSS_THROW("UShort to Half requires Shader Model 5.");
+		return "f16tof32";
```
HansKristian-Work, line 3611: Also needs narrowing cast to half.

HansKristian-Work (changes requested):
Comments

spirv_hlsl.cpp
RandomShaper: Both `f16tof32()` and `f32tof16()` are actually supported from as low as SM 4.0 (via emuation; natively from 5.0). Therefore, this code (generated by SPIRV-Cross using SM 4.0) compiles on DXC with just `-T cs_4_0` (the rest of the file is omitted for convenience):
```
min16uint AW1_x(min16uint a)
{
    return a;
}

min16float APrxLoRsqH1(min16float a)
{
    min16uint param = 22947u;
    min16uint param_1 = 1u;
    return f16tof32(AW1_x(param) - (f32tof16(a) >> AW1_x(param_1)));
}
```

So, I've changed the check to be happy even with SM 4.0 for now.

spirv_hlsl.cpp
```diff
@@ -3598,6 +3598,18 @@ string CompilerHLSL::bitcast_glsl_op(const SPIRType &out_type, const SPIRType &i
 		}
 		return "spvPackFloat2x16";
 	}
+	else if (out_type.basetype == SPIRType::UShort && in_type.basetype == SPIRType::Half)
+	{
+		if (hlsl_options.shader_model < 50)
+			SPIRV_CROSS_THROW("Half to UShort requires Shader Model 5.");
+		return "f32tof16";
```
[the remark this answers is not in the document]
RandomShaper: I've checked every intrinsic suspicious of providing this in the HLSL reference, these are the only ones I've found that are able to deal with 16-bit float representation.

I've added the cast (supporting both 16-bit types enabled or not).

The function above is now generated like this (and warnings are gone):
```
half APrxLoRsqH1(half a)
{
    uint16_t param = 22947u;
    uint16_t param_1 = 1u;
    return (half)f16tof32(AW1_x(param) - ((uint16_t)f32tof16(a) >> AW1_x(param_1)));
}
```

spirv_hlsl.cpp
RandomShaper: Likewise.

RandomShaper:
I understand this is not as efficient as one could wish, but at least does the trick. Please take a look to the updated code.

spirv_hlsl.cpp
HansKristian-Work: It's not f16tof32 I'm concerned about, it's UShort and Half types. These are explicit 16-bit types in SPIR-V and cannot be supported properly before SM 6.2. min16float/min16int is basically RelaxedPrecision.

RandomShaper: To be honest, I'm not sure I undersand the full implications of that. I can just change the check to the following and so be safe:
```
	if (!hlsl_options.enable_16bit_types) // Implies SM 6.2
		SPIRV_CROSS_THROW("[...] to [...] requires native 16-bit type support.");
```

What do you think?

HansKristian-Work: Yes, that looks better. I forgot there was a enable_16bit_types option in SPIRV-Cross, but yes. Bitcasts in HLSL on min16float types would fail in DXC, so this is more correct.

RandomShaper: Just in case, before making the change, in the current state of this PR SPIRV-Cross can generate this (with SM 4.0 and 16-bit types disabled):
```
min16uint AW1_x(min16uint a)
{
    return a;
}

min16float APrxLoRsqH1(min16float a)
{
    min16uint param = 22947u;
    min16uint param_1 = 1u;
    return (min16float)f16tof32(AW1_x(param) - ((min16uint)f32tof16(a) >> AW1_x(param_1)));
}

min16float APrxLoRcpH1(min16float a)
{
    min16uint param = 30596u;
    return (min16float)f16tof32(AW1_x(param) - (min16uint)f32tof16(a));
}
```

And it compiles fine on DXC. No warnings, no errors. And the shader works fine at runtime.

PR #1966, opened 2022-06-21
Fix compilation errors with deprecated sprintf function

fred-wang:
This commit modifies sprintf calls to fix the following compilation
errors using XCode 14.0 beta on MacOS:

> 'sprintf' is deprecated: This function is provided for compatibility
> reasons only.  Due to security concerns inherent in the design of
> sprintf(3), it is highly recommended that you use snprintf(3) instead.

### DIFF
--- a/spirv_glsl.cpp
+++ b/spirv_glsl.cpp
@@ -5275,7 +5275,7 @@ string CompilerGLSL::convert_float_to_string(const SPIRConstant &c, uint32_t col
 			in_type.width = 32;
 
 			char print_buffer[32];
-			sprintf(print_buffer, "0x%xu", c.scalar(col, row));
+			snprintf(print_buffer, sizeof(print_buffer), "0x%xu", c.scalar(col, row));
 
 			const char *comment = "inf";
 			if (float_value == -numeric_limits<float>::infinity())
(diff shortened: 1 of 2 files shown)


HansKristian-Work:
This is deliberate.

- It is statically known that these cannot overflow.
- snprintf does not exist or is buggy on older MSVC versions.

fred-wang:
@HansKristian-Work Thanks, so what would be the recommended way to fix the build failure on newer XCode version? disable the warning? use #ifdef?

Incidentally, not sure if there is some kind of consistency rules between KhronosGroup repos, but I noticed a similar commit was done for SPIRV-Tools: https://github.com/KhronosGroup/SPIRV-Tools/pull/4811

HansKristian-Work:
> build failure on newer XCode version

Is it a straight up build failure? I guess you can ifndef _MSC_VER and comment about compatibility with old MSVC.

spirv_glsl.cpp
```diff
@@ -5213,9 +5213,9 @@ string CompilerGLSL::constant_expression(const SPIRConstant &c, bool inside_bloc
 	}
 }
 
-#ifdef _MSC_VER
-// sprintf warning.
-// We cannot rely on snprintf existing because, ..., MSVC.
+#ifdef _WIN32
```
HansKristian-Work, line 5216: This pragma is MSVC specific.

spirv_glsl.cpp
```diff
@@ -5399,7 +5408,7 @@ std::string CompilerGLSL::convert_double_to_string(const SPIRConstant &c, uint32
 	return res;
 }
 
-#ifdef _MSC_VER
+#ifdef _WIN32
```
HansKristian-Work, line 5411: MSVC specific.

fred-wang:
OK did the change, but i suspect we'd like to switch to snprintf at some point for MinGW too, if we can't disable the warning.

On macOS, the build failure is due to the deprecation warning.

fred-wang:
Thanks for the review!

fred-wang:
Thanks for the review!

PR #1989, opened 2022-08-02
GLSL fixes for OpImageRead and OpImageQuerySize with uncommon image types.

ashleyharris-maptek-com-au:
Few fixes for corner cases with unloddable and unsampled images passed into OpImageRead or OpImageQuerySize, which is allowed by the SPIR-V spec, but can result in the need to call a different variation of a glsl function.

OpImageQuerySize may map to either:
- textureSize(sampler)
- textureSize(sampler, lodLevel)

Depending on the "dim" of the sampled image. Rect, Buffer, and MS images must use the first variation, all others must use the second.

Same with OpImageRead, it may map to either:
- texelFetch(sampledImage, coords)
- imageLoad(unsampledImage, coords)
- texelFetch(sampler, coords, lod)
- (Plus the sparse and MS cases)

Depending on whether "sampled" in the image type is either 0 or 2, and whether it's loddable or not and needs the additional lod argument.

### DIFF
--- a/spirv_glsl.cpp
+++ b/spirv_glsl.cpp
@@ -12662,7 +12662,33 @@ void CompilerGLSL::emit_instruction(const Instruction &instruction)
 					    join("imageLoad(", to_non_uniform_aware_expression(ops[2]), ", ", coord_expr, ", ", to_expression(samples), ")");
 				}
 				else
-					imgexpr = join("imageLoad(", to_non_uniform_aware_expression(ops[2]), ", ", coord_expr, ")");
+				{
+					// 'imageLoad' can only be used if the image type is "image" (as per
+					// https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/imageLoad.xhtml) which is only true
+					// if image_type_glsl(...) returns a string containing "image", which is only true if
+					// type.image.sampled == 2. However type.image.sampled can be 0 in OpImageRead calls as per:
+					// https://www.khronos.org/registry/SPIR-V/specs/unified1/SPIRV.html#OpTypeImage , so that needs to
+					// be handled, as that may have an extra LOD argument:
+
+					if (type.image.sampled == 2)
+					{
+						// We're using an image, not a sampler - use imageLoad.
+						imgexpr = join("imageLoad(", to_non_uniform_aware_expression(ops[2]), ", ", coord_expr, ")");
+					}
+					else if (type.image.dim == Dim::DimBuffer || type.image.dim == Dim::DimRect)
+					{
+						// We're using a sampler so can't use imageLoad, despite our overloads of texelFetch having the
+						// same arguments.
+						imgexpr = join("texelFetch(", to_non_uniform_aware_expression(ops[2]), ", ", coord_expr, ")");
+					}
+					else
+					{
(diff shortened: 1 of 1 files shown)


spirv_glsl.cpp
```diff
@@ -12662,7 +12662,33 @@ void CompilerGLSL::emit_instruction(const Instruction &instruction)
...
-					imgexpr = join("imageLoad(", to_non_uniform_aware_expression(ops[2]), ", ", coord_expr, ")");
+				{
+					// 'imageLoad' can only be used if the image type is "image" (as per
+					// https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/imageLoad.xhtml) which is only true
+					// if image_type_glsl(...) returns a string containing "image", which is only true if
+					// type.image.sampled == 2. However type.image.sampled can be 0 in OpImageRead calls as per:
+					// https://www.khronos.org/registry/SPIR-V/specs/unified1/SPIRV.html#OpTypeImage , so that needs to
+					// be handled, as that may have an extra LOD argument:
+
+					if (type.image.sampled == 2)
```
HansKristian-Work, line 12673: sampled == 0 is a CL thing. Vulkan environment does not support sampled 0. What is the reason to add cases to handle this? If sampled == 0, we don't even know how to emit the type properly, since it could be image2D or texture2D.

HansKristian-Work (changes requested):
Missing:
- Rationale for why this has to be supported.
- Tests.

spirv_glsl.cpp
ashleyharris-maptek-com-au: I must've missed in the spec where it says "Sampled == 0" doesn't work. There's a few references to it in spec that (to me at least?) imply it's supported, eg SampledRect Capability in section 3.31.37. What am I missing that says it can't be done?

I've got shaders with Sampled == 0 in them that seem to be valid, and pass validation. (Doesn't mean they're valid of course - open to being surprised and needing to patch the validator). The shaders seem to work, and also export to HLSL fine and work with Direct3D11. With GLSL it just emits the wrong function or choses the wrong number of arguments, resulting in uncompilable GLSL.

HansKristian-Work: Table 21. Shader Resource and Storage Class Correspondence declares which resource types match with which descriptor type, and there it says sampled = 1, sampled = 2 explicitly.

The SPIR-V spec here is more general here than the environment it is used in. SPIRV-Cross is designed for the shader environment that is supported by Vulkan and friends, i.e. any kernel specific environment capabilities are not supported.

It's plausible there is missing validation here. Did you generate the SPIR-V yourself?

> The shaders seem to work, and also export to HLSL fine and work with Direct3D11.

How would it? It would be ambiguous if we need to emit RWTexture2D or Texture2D based on sampled == 1 vs sampled == 2 when declaring the resource. Perhaps it just picked the correct one by luck?

ashleyharris-maptek-com-au: The SPIRV seems to be working in my Vulkan fragment shader, and the converted HLSL is definitely working under D3D11, and a recent build of the validator passes it. Sounds like it might be luck.

I wasn't aware of that table. I had to google it as it wasn't in the SPIR-V spec but thank you for pointing me towards it.

PR #1994, opened 2022-08-09
CTS shared composite variables

gpx1000:
CTS Portability:  @cdavis5e is primary author on these changes.  I have reviewed them and internally LGTM;

### DIFF
--- a/spirv_msl.cpp
+++ b/spirv_msl.cpp
@@ -1966,6 +1966,13 @@ void CompilerMSL::mark_packable_structs()
 			    (has_decoration(type.self, DecorationBlock) || has_decoration(type.self, DecorationBufferBlock)))
 				mark_as_packable(type);
 		}
+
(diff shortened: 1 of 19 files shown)


spirv_msl.cpp
```diff
@@ -4828,6 +4867,10 @@ void CompilerMSL::add_typedef_line(const string &line)
 // Template struct like spvUnsafeArray<> need to be declared *before* any resources are declared
 void CompilerMSL::emit_custom_templates()
 {
+	static const string address_spaces[] = {
```
HansKristian-Work, line 4870: Should use `static const char *` here to avoid magic static initialization.

HansKristian-Work:
Just fixed the trivial style nit on my end. Looked overall good to me, but hard to review correctness in detail. Passing CTS is a good enough indicator that it works for me.

PR #2005, opened 2022-08-19
Don't rename remapped variables like 'gl_LastFragDepthARM'

atyuwen:
The name fixing logic makes shaders with depth fetch broken (`gl_LastFragDepthARM` is renamed to `_RESERVED_IDENTIFIER_FIXUP_gl_LastFragDepthARM`). We should prevent remapped vars from being renamed.

### DIFF
--- a/spirv_cross_parsed_ir.cpp
+++ b/spirv_cross_parsed_ir.cpp
@@ -330,6 +330,10 @@ void ParsedIR::fixup_reserved_names()
 {
 	for (uint32_t id : meta_needing_name_fixup)
 	{
+		// Don't rename remapped variables like 'gl_LastFragDepthARM'.
+		if (ids[id].get_type() == TypeVariable && get<SPIRVariable>(id).remapped_variable)
+			continue;
+
 		auto &m = meta[id];
 		sanitize_identifier(m.decoration.alias, false, false);
 		for (auto &memb : m.members)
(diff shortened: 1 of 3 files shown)


atyuwen:
We also found that the row-major-load workaround causes significant performance degeneration on some Android devices. Disable this workaround and the performance returns.

spirv_glsl.cpp
```diff
@@ -649,7 +649,7 @@ string CompilerGLSL::compile()
 	{
 		// only NV_gpu_shader5 supports divergent indexing on OpenGL, and it does so without extra qualifiers
 		backend.nonuniform_qualifier = "";
-		backend.needs_row_major_load_workaround = true;
+		backend.needs_row_major_load_workaround = !options.es;
```
HansKristian-Work, line 652: IIRC, this workaround is needed on PowerVR as well?

spirv_cross_parsed_ir.cpp
```diff
@@ -342,6 +342,8 @@ void ParsedIR::set_name(ID id, const string &name)
 {
 	auto &m = meta[id];
 	m.decoration.alias = name;
+	if (ids[id].get_type() == TypeVariable && get<SPIRVariable>(id).remapped_variable)
```
HansKristian-Work, line 345: This is a little brittle since it relies on calling order of set_remapped_variable_state and set_name. It would be better to defer this check until we iterate over meta_needing_name_fixup.

HansKristian-Work (approved):
LGTM

PR #2014, opened 2022-09-06
MSL: Add a mechanism to fix up shader outputs.

cdavis5e:
This is analogous to the existing support for fixing up shader inputs. It is intended to be used with tessellation to add implicit builtins that are read from a later stage, despite not being written in an earlier stage. (Believe it or not, this is in fact legal in Vulkan.)

Helps fix 8 CTS tests under `dEQP-VK.pipeline.*.no_position`. (Eight other tests work solely by accident without this change.)

cc @gpx1000

### DIFF
--- a/spirv_msl.cpp
+++ b/spirv_msl.cpp
@@ -56,11 +56,18 @@ CompilerMSL::CompilerMSL(ParsedIR &&ir_)
 {
 }
 
-void CompilerMSL::add_msl_shader_input(const MSLShaderInput &si)
+void CompilerMSL::add_msl_shader_input(const MSLShaderInterfaceVariable &si)
 {
 	inputs_by_location[{si.location, si.component}] = si;
 	if (si.builtin != BuiltInMax && !inputs_by_builtin.count(si.builtin))
 		inputs_by_builtin[si.builtin] = si;
+}
+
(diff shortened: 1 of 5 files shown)


spirv_msl.cpp
```diff
@@ -3862,6 +3923,67 @@ uint32_t CompilerMSL::add_interface_block(StorageClass storage, bool patch)
...
+			uint32_t offset = ir.increase_bound_by(4);
+			uint32_t type_id = offset;
+			uint32_t array_type_id = offset + 1;
+			uint32_t ptr_type_id = offset + 2;
+			uint32_t var_id = offset + 3;
+
+			SPIRType type;
+			switch (output.second.format)
+			{
+			case MSL_SHADER_INPUT_FORMAT_UINT16:
```
billhollings, line 3947: Should this new code (and others near here) use `MSL_SHADER_VARIABLE_FORMAT_*`?

billhollings (changes requested):
Suggest minor correctness fix in enum use.

spirv_msl.cpp
cdavis5e: Ah, you are of course correct.

billhollings (approved):

HansKristian-Work (approved):
Seems okay.

PR #2018, opened 2022-09-13
MSL: only fix up gl_FragCoord if really necessary.

atyuwen:
Currently  `gl_FragCoord` will be fixed up to get per-sample semantics when `SubpassInput` or `SubpassInputMS`  is detected:
```
gl_FragCoord.xy += get_sample_position(gl_SampleID) - 0.5;
```
 But Old iOS (e.g. iOS 9) does not allow to use `[[sample_id]]` in shader if the pipeline is not multisampled, so the above code may not compile. To solve this issue, we have to check if the fix-up is really needed.

### DIFF
--- a/reference/opt/shaders-ue4/asm/frag/subpass-input.ios.framebuffer-fetch.asm.frag
+++ b/reference/opt/shaders-ue4/asm/frag/subpass-input.ios.framebuffer-fetch.asm.frag
@@ -197,10 +197,9 @@ struct main0_out
     float4 out_var_SV_Target0 [[color(0)]];
 };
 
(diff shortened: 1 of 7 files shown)


HansKristian-Work:
I think this is correct. The Vulkan spec talks about this scenario:

> The MS operand of the OpTypeImage must be 0 if the samples field of the corresponding [VkAttachmentDescription](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#VkAttachmentDescription) is VK_SAMPLE_COUNT_1_BIT and 1 otherwise.

So if only subpassInput is used, this must be a single sampled subpass, and MS must be used for a multisampled subpass, meaning that it's safe to promote to sample shading in that case (using sample_id).

HansKristian-Work (approved):

PR #2027, opened 2022-09-23
MSL: Emit correct address space when casting during OpStore.

billhollings:
When storing to local variable (eg. `OpCopyLogical`), the default device address space used during casts is illegal.

Determine correct address space based on variable type.


billhollings:
@HansKristian-Work Ping. This has been sitting here for over a week, and I expect it should be a fairly simple review. Would you mind reviewing and pulling in when you get a chance soon, please? Thanks.

HansKristian-Work:
Sorry about the delay.

Issue #2032, opened 2022-10-01
[HLSL] `SPIRV_Cross_VertexInfo` is not relieble


Try:
Face issue with my pet-game-engine:

1. spirv-cross emits code, like:
```c++
cbuffer SPIRV_Cross_VertexInfo
{
   int SPIRV_Cross_BaseVertex;
   int SPIRV_Cross_BaseInstance; 
};     
```
2. Since no particular binding register is specified, it gonna be assigned automatically by DXIL.
Normally it's lowest positive value that is unused by other `cbuffer`s
3. Register that assigned automatically may overlap with some else `cbuffer` in fragment shader, causing link/runtime issues.

Problem is somewhat outside of spirv-cross scope, and apparently there is no perfect solution to it. My current idea is to add new parameter `CompilerHLSL::Options::vertexInfoRegister`, that can be provided by game engine to specify exact register.
(Note: engine also can't provide sensible value, but can specify "magic" value like 255 or something)
What do you think about such solution? Thanks!

PR #2036, opened 2022-10-04
HLSL: add support for user-defined register/space id for push_constant(root constants)

Try:
Followup to #2032

Explicit binding for `push_constant`, helps engine to avoid link/runtime errors in generated hlsl shader-pipeline.
Note: not using AUX-buffer concept intentionally, since push_constant is not auxiliary item.

syntax:
`--hlsl-push-constant-binding <register> <space>`

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -715,6 +715,8 @@ struct CLIArguments
 	bool hlsl_base_vertex_index_explicit_binding = false;
 	uint32_t hlsl_base_vertex_index_register_index = 0;
 	uint32_t hlsl_base_vertex_index_register_space = 0;
+	uint32_t hlsl_push_constant_register_index = ResourceBindingPushConstantBinding;
+	uint32_t hlsl_push_constant_register_space = ResourceBindingPushConstantDescriptorSet;
 
 	bool hlsl_force_storage_buffer_as_uav = false;
 	bool hlsl_nonwritable_uav_texture_as_srv = false;
@@ -1377,6 +1379,7 @@ static string compile_iteration(const CLIArguments &args, std::vector<uint32_t> 
 		hlsl_opts.flatten_matrix_vertex_input_semantics = args.hlsl_flatten_matrix_vertex_input_semantics;
 		hlsl->set_hlsl_options(hlsl_opts);
 		hlsl->set_resource_binding_flags(args.hlsl_binding_flags);
+		hlsl->set_hlsl_push_constant_binding(args.hlsl_push_constant_register_index, args.hlsl_push_constant_register_space);
 		if (args.hlsl_base_vertex_index_explicit_binding)
 		{
 			hlsl->set_hlsl_aux_buffer_binding(HLSL_AUX_BINDING_BASE_VERTEX_INSTANCE,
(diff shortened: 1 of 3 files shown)


spirv_hlsl.cpp
```diff
@@ -3470,8 +3476,8 @@ string CompilerHLSL::to_resource_binding(const SPIRVariable &var)
 		return "";
 
 	uint32_t desc_set =
-	    resource_flags == HLSL_BINDING_AUTO_PUSH_CONSTANT_BIT ? ResourceBindingPushConstantDescriptorSet : 0u;
```
HansKristian-Work, line 3473: You're supposed to set up a binding using the normal API here. `add_hlsl_resource_binding` where you use the ResourceBindingPushConstantDescriptorSet/Binding as a pseudo-binding point. Does that not work? It's fine adding a CLI option for this, but I don't think we need to change the API.

spirv_hlsl.hpp
```diff
@@ -54,8 +54,6 @@ enum HLSLBindingFlagBits
 	HLSL_BINDING_AUTO_NONE_BIT = 0,
 
 	// Push constant (root constant) resources will be declared as CBVs (b-space) without a register() declaration.
-	// A register will be automatically assigned by the D3D compiler, but must therefore be reflected in D3D-land.
```
HansKristian-Work, line 57: This comment is still correct. If AUTO_PUSH_CONSTANT_BIT is used that is the intent.

main.cpp
```diff
@@ -1550,6 +1553,10 @@ static int main_inner(int argc, char *argv[])
 		args.hlsl_base_vertex_index_register_index = parser.next_uint();
 		args.hlsl_base_vertex_index_register_space = parser.next_uint();
 	});
+	cbs.add("--hlsl-push-constant-binding", [&args](CLIParser &parser) {
```
HansKristian-Work, line 1556: Missing --help log explaining the option.

HansKristian-Work (changes requested):
Comments.

spirv_hlsl.cpp
Try: Thanks for pointing to `add_hlsl_resource_binding` - I was not aware of this api.
Overall this almost works for me, except stage flags:
```
HLSLResourceBinding push_binding;
push_binding.stage = ???; // can't specify anything until shader is parsed
push_binding.desc_set = ResourceBindingPushConstantDescriptorSet;
push_binding.binding = 0;
push_binding.cbv.register_binding = args.hlsl_push_constant_register_index;
push_binding.cbv.register_space = args.hlsl_push_constant_register_space;
hlsl->add_hlsl_resource_binding(push_binding);
```
Can't specify anything at `push_binding.stage ` until shader is parsed. Fine with my engine, but why `.stage` is needed in the first place? AFAIK spirv-cross processes one shader at a time, so no need to use different settings per stage.

HansKristian-Work: You can query the execution mode. The stage there was probably not the best idea ever, but it would be useful if multi-entry point codegen was supported at some point.

Try: Thanks!
Already added `HLSLResourceBinding` to my game engine - all works well. 
Since issue is resolved, we can discard PR

PR #2039, opened 2022-10-14
Raw buffer tess input

cdavis5e:
This set of changes adds support for tessellation evaluation shaders accessing their inputs using raw buffer accesses, instead of using vertex-style input. This is in support of a Khronos Group project to reduce the number of CTS failures in MoltenVK.


main.cpp
```diff
@@ -886,6 +887,9 @@ static void print_help_msl()
 	                "\t\t<format> can be 'any32', 'any16', 'u16', 'u8', or 'other', to indicate a 32-bit opaque value, 16-bit opaque value, 16-bit unsigned integer, 8-bit unsigned integer, "
 	                "or other-typed variable. <size> is the vector length of the variable, which must be greater than or equal to that declared in the shader.\n"
 	                "\t\tUseful if shader stage interfaces don't match up, as pipeline creation might otherwise fail.\n"
+					"\t[--msl-raw-buffer-tess-input]:\n\t\tUse raw buffers for tessellation evaluation input.\n"
```
HansKristian-Work, line 890: Would it be better to call this `msl-raw-buffer-tese-input`? This option only applies to TESE, and afaik, the multi-patch-workgroup is the equivalent for TESC.

spirv_cross.hpp
```diff
@@ -371,6 +371,9 @@ class Compiler
 	spv::ExecutionModel get_execution_model() const;
 
 	bool is_tessellation_shader() const;
+	bool is_tesc_shader() const;
```
HansKristian-Work, line 374: These seem like they should be private helpers.

spirv_cross_c.h
```diff
@@ -347,13 +347,29 @@ typedef struct spvc_msl_shader_interface_var
...
 SPVC_PUBLIC_API void spvc_msl_shader_input_init(spvc_msl_shader_input *input);
 
+/* Maps to C++ API. */
+typedef struct spvc_msl_shader_interface_var_2
+{
+	unsigned location;
+	spvc_msl_vertex_format format;
+	SpvBuiltIn builtin;
+	unsigned vecsize;
+	spvc_bool patch;
```
HansKristian-Work, line 365: Instead of a boolean patch, I would prefer using an enum for { vertex, patch, per-primitive } to be a bit more future proof for Mesh shaders when we get around to that at some point and not having to make an interface_var_3.

spirv_msl.cpp
```diff
@@ -11972,9 +12039,17 @@ string CompilerMSL::get_type_address_space(const SPIRType &type, uint32_t id, bo
...
 
 	case StorageClassInput:
-		if (get_execution_model() == ExecutionModelTessellationControl && var &&
-		    var->basevariable == stage_in_ptr_var_id)
-			addr_space = msl_options.multi_patch_workgroup ? "constant" : "threadgroup";
+		if (is_tesc_shader() && var && var->basevariable == stage_in_ptr_var_id)
+			addr_space = msl_options.multi_patch_workgroup ? "const device" : "threadgroup";
+		// Don't pass tessellation levels in the device AS; we load and convert them
+		// to float manually.
+		if (is_tese_shader() && msl_options.raw_buffer_tess_input && var &&
```
HansKristian-Work, line 12046: This if is incomprehensible, please refactor out booleans.

spirv_msl.hpp
```diff
@@ -69,6 +69,7 @@ struct MSLShaderInterfaceVariable
 	MSLShaderVariableFormat format = MSL_SHADER_VARIABLE_FORMAT_OTHER;
 	spv::BuiltIn builtin = spv::BuiltInMax;
 	uint32_t vecsize = 0;
+	bool patch = false;
```
HansKristian-Work, line 72: This should be an enum to allow for mesh shaders at some point.

spirv_msl.hpp
```diff
@@ -387,6 +389,11 @@ class CompilerMSL : public CompilerGLSL
 		// builtins are processed, but should result in more efficient usage of the GPU.
 		bool multi_patch_workgroup = false;
 
+		// Use storage buffers instead of vertex-style attributes for tessellation evaluation
+		// input. This may require conversion of inputs in the generated post-tessellation
+		// vertex shader, but allows the use of nested arrays.
+		bool raw_buffer_tess_input = false;
```
HansKristian-Work, line 395: raw_buffer_tese_input perhaps?

spirv_cross_c.h
```diff
@@ -336,7 +336,7 @@ typedef struct spvc_msl_vertex_attribute
  */
 SPVC_PUBLIC_API void spvc_msl_vertex_attribute_init(spvc_msl_vertex_attribute *attr);
 
-/* Maps to C++ API. */
+/* Maps to C++ API. Deprecated; use spvc_msl_shader_interface_var_2. */
```
HansKristian-Work, line 339: When updating the C API headers, we also need to update

`SPVC_C_API_VERSION_MINOR` and spirv-cross-abi-minor in CMakeLists.txt.

HansKristian-Work (changes requested):
Mostly nits. It's arguably too much to review in detail, so if this passes the CTS tests, I think that's what I care more about.

main.cpp
cdavis5e: Done.

spirv_cross.hpp
cdavis5e: I've hoisted them into `CompilerMSL`.

spirv_cross_c.h
cdavis5e: Good catch.

spirv_msl.cpp
cdavis5e: This should hopefully be clearer now.

spirv_msl.hpp
cdavis5e: Done.

spirv_cross_c.h
cdavis5e: Oops. Done.

main.cpp
```diff
@@ -878,14 +879,19 @@ static void print_help_msl()
...
-	                "\t[--msl-shader-output <index> <format> <size>]:\n\t\tSpecify the format of the shader output at <index>.\n"
+	                "\t[--msl-shader-output <index> <format> <size> <rate>]:\n\t\tSpecify the format of the shader output at <index>.\n"
 	                "\t\t<format> can be 'any32', 'any16', 'u16', 'u8', or 'other', to indicate a 32-bit opaque value, 16-bit opaque value, 16-bit unsigned integer, 8-bit unsigned integer, "
-	                "or other-typed variable. <size> is the vector length of the variable, which must be greater than or equal to that declared in the shader.\n"
+	                "or other-typed variable. <size> is the vector length of the variable, which must be greater than or equal to that declared in the shader. <rate> can be 'vertex', "
+					"'primitive', or 'patch' to indicate a per-vertex, per-primitive, or per-patch variable.\n"
 	                "\t\tUseful if shader stage interfaces don't match up, as pipeline creation might otherwise fail.\n"
+					"\t[--msl-raw-buffer-tese-input]:\n\t\tUse raw buffers for tessellation evaluation input.\n"
```
HansKristian-Work, line 892: Indent.

main.cpp
```diff
@@ -878,14 +879,19 @@ static void print_help_msl()
 	                "\t[--msl-disable-frag-stencil-ref-builtin]:\n\t\tDisable FragStencilRef output. Useful if pipeline does not enable stencil output, as pipeline creation might otherwise fail.\n"
 	                "\t[--msl-enable-frag-output-mask <mask>]:\n\t\tOnly selectively enable fragment outputs. Useful if pipeline does not enable fragment output for certain locations, as pipeline creation might otherwise fail.\n"
 	                "\t[--msl-no-clip-distance-user-varying]:\n\t\tDo not emit user varyings to emulate gl_ClipDistance in fragment shaders.\n"
-	                "\t[--msl-shader-input <index> <format> <size>]:\n\t\tSpecify the format of the shader input at <index>.\n"
+	                "\t[--msl-shader-input <index> <format> <size> <rate>]:\n\t\tSpecify the format of the shader input at <index>.\n"
```
HansKristian-Work, line 882: Breaking the CLI like this can be very frustrating from a user perspective. Is it possible to keep the existing command working and add a new one? E.g `--msl-add-shader-input` and `--msl-add-shader-output`.

spirv_msl.cpp
```diff
@@ -14355,23 +14476,26 @@ bool CompilerMSL::variable_decl_is_remapped_storage(const SPIRVariable &variable
...
+		return variable.storage == StorageClassOutput && is_tesc_shader() && is_stage_output_variable_masked(variable);
 	}
 	else if (storage == StorageClassStorageBuffer)
 	{
+		// These builtins are passed directly; we don't want to use remapping
+		// for them.
+		auto builtin = (BuiltIn)get_decoration(variable.self, DecorationBuiltIn);
+		if (is_tese_shader() && is_builtin_variable(variable) &&
+		    (builtin == BuiltInTessCoord || builtin == BuiltInPrimitiveId))
+			return false;
```
HansKristian-Work, line 14497: weird formatting.

HansKristian-Work (changes requested):
Nits

main.cpp
cdavis5e: Good point.

main.cpp
cdavis5e: Fixed.

HansKristian-Work (approved):
LGTM, thanks.

PR #2050, opened 2022-11-01
MSL: Implement `CompositeInsert` `OpSpecConstantOp`.

cdavis5e:
This op creates a new composite constant with one element replaced. So, we reconstruct the `SPIRConstant` for the composite constant, but with one of the IDs replaced. Constant initializer lists are memoized for when the result of a `CompositeInsert` is used in another `CompositeInsert`.

(I wanted to add a test case for GLSL as well, but for two things:
1. `glslang` in Vulkan mode chokes on the first constant array, insisting that its initializer needs to be a constant. [Bug in glslang?]
2. The declarations for the buffers used by the shader aren't emitted, regardless of whether Vulkan mode is enabled.)

Fixes five tests under
`dEQP-VK.spirv_assembly.instruction.*.opspecconstantop.vector_related`.

### DIFF
--- a/spirv_cross_containers.hpp
+++ b/spirv_cross_containers.hpp
@@ -210,7 +210,8 @@ public:
 		buffer_capacity = N;
 	}
 
-	SmallVector(const T *arg_list_begin, const T *arg_list_end) SPIRV_CROSS_NOEXCEPT : SmallVector()
+	template <typename U>
+	SmallVector(const U *arg_list_begin, const U *arg_list_end) SPIRV_CROSS_NOEXCEPT : SmallVector()
 	{
 		auto count = size_t(arg_list_end - arg_list_begin);
 		reserve(count);
@@ -219,7 +220,13 @@ public:
 		this->buffer_size = count;
 	}
 
-	SmallVector(std::initializer_list<T> init) SPIRV_CROSS_NOEXCEPT : SmallVector(init.begin(), init.end())
+	template <typename U>
+	SmallVector(std::initializer_list<U> init) SPIRV_CROSS_NOEXCEPT : SmallVector(init.begin(), init.end())
+	{
+	}
+
+	template <typename U, size_t M>
+	SmallVector(const U (&init)[M]) SPIRV_CROSS_NOEXCEPT : SmallVector(init, init + M)
 	{
 	}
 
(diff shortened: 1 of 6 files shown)


spirv_cross_containers.hpp
```diff
@@ -219,7 +220,13 @@ class SmallVector : public VectorView<T>
...
 	}
 
-	SmallVector(std::initializer_list<T> init) SPIRV_CROSS_NOEXCEPT : SmallVector(init.begin(), init.end())
+	template <typename U>
+	SmallVector(std::initializer_list<U> init) SPIRV_CROSS_NOEXCEPT : SmallVector(init.begin(), init.end())
+	{
+	}
+
+	template <typename U, size_t M>
+	SmallVector(U (&init)[M]) SPIRV_CROSS_NOEXCEPT : SmallVector(init, &init[M])
```
HansKristian-Work, line 229: `const U`?

HansKristian-Work: `&init[M]` is not a valid dereference. I've had problems where Clang inferred things from just this before. `init + M` seems better here to avoid that potential problem.

spirv_glsl.cpp
```diff
@@ -4930,6 +4930,79 @@ string CompilerGLSL::to_expression(uint32_t id, bool register_expression_read)
...
+void CompilerGLSL::set_composite_constant(ConstantID const_id, TypeID type_id,
+                                          const SmallVector<ConstantID> &initializers)
+{
+	if (maybe_get<SPIRConstantOp>(const_id))
+	{
+		const_composite_insert_ids[const_id] = initializers;
+		return;
+	}
+
+	auto &constant = set<SPIRConstant>(type_id);
```
HansKristian-Work, line 4988: type_id?

HansKristian-Work (changes requested):

spirv_cross_containers.hpp
cdavis5e: > `const U`?

I was afraid that wouldn't compile. Apparently, it does.

> `&init[M]` is not a valid dereference. I've had problems where Clang inferred things from just this before. `init + M` seems better here to avoid that potential problem.

I was actually torn between these two forms. Ironically, I avoided the latter because I was afraid *that* was undefined behavior! But I think you're right and I got that backwards.

spirv_glsl.cpp
cdavis5e: Oops. Forgot the ID to set. Fixed.

Apparently, that test I got from the CTS doesn't go deeper than one level. I've added a test for that and fixed all the problems that uncovered.

HansKristian-Work (approved):
LGTM

PR #2056, opened 2022-11-08
MSL: Prevent stores to storage resources in discarded fragments.

### LINKED ISSUE #2054: Fix `shaderFragmentDepth32FloatMultisample` failures in the Vulkan CTS
This is in support of a Khronos contract to reduce CTS failures in MoltenVK, leading up to certification as a Vulkan Portability implementation.

These failures are caused by a bug in Metal, but not one the name of the Vulkan Portability feature implies. It's actually caused by some Metal devices--notably, Apple Silicon GPUs, and possibly others--continuing to allow writes to storage resources, even if the fragment were discarded. Luckily, we can work around this, and the workaround isn't even that intrusive. A boolean flag will be added, which is set when the fragment is discarded, and checked when a resource is written to.

I have filed feedback FB11736212 to cajole Apple into fixing this on their end. In anticipation of this, the workaround is controlled by an option.

Fixes at least three CTS tests, under `dEQP-VK.glsl.builtin_var.fragdepth.*`, and possibly others.

cdavis5e:
Some Metal devices have a bug where storage resources can still be written to even if the fragment is discarded. This is obviously a bug in Metal, but bothering Apple to fix it will only fix it for newer versions; therefore, a workaround is needed for older versions. I have made this an option so that, in case the bug is ever fixed, the workaround can be disabled.

This workaround is simple: if a fragment shader may discard its fragment and does either of the following:

 * writes to a storage resource; or
 * checks if it's a helper invocation,

an additional boolean flag is created and passed to all functions. The flag is set when the fragment is discarded, checked on all resource writes, and ORed to the result of `simd_is_helper_thread()`.

Fixes at least 3 failures in the CTS.

Closes #2054.


HansKristian-Work:
> ORed to the result of simd_is_helper_thread()

Does simd_is_helper_thread() not update based on helper lane state? This is a requirement of OpIsHelperInvocationEXT at least.

spirv_msl.cpp
```diff
@@ -1588,6 +1589,25 @@ void CompilerMSL::preprocess_op_codes()
 	                          (need_subpass_input_ms && !msl_options.use_framebuffer_fetch_subpasses))))
 		needs_sample_id = true;
 
+	// OpKill is removed by the parser, so we need to identify those by inspecting
+	// blocks.
+	ir.for_each_typed_id<SPIRBlock>([&preproc](uint32_t, SPIRBlock &block) {
+		if (block.terminator == SPIRBlock::Kill)
+			// FIXME: This doesn't distinguish between OpKill and OpTerminateInvocation!
```
HansKristian-Work, line 1596: OpKill == OpTerminateInvocation, so this is fine I think. Kill is just looser where it gives a little leeway to support demote because broken apps/drivers, but that was never the intention.

spirv_msl.cpp
```diff
@@ -1588,6 +1589,25 @@ void CompilerMSL::preprocess_op_codes()
...
+			preproc.uses_discard = true;
+	});
+
+	// Fragment shaders that both write to storage resources and discard fragments
+	// need checks on the writes, to work around Metal allowing these writes despite
+	// the fragment being dead. Similarly, fragment shaders that use OpIsHelperInvocationEXT,
+	// or read a volatile BuiltInHelperInvocation, need this extra check, because
+	// despite that keeping derivatives well-defined is the sole purpose of allowing a
+	// fragment thread to run after discard, simd_is_helper_thread() doesn't return
+	// true for discarded fragment threads.
```
HansKristian-Work, line 1606: ... what?

spirv_msl.cpp
```diff
@@ -8616,9 +8643,16 @@ void CompilerMSL::emit_instruction(const Instruction &instruction)
 		args.base.is_fetch = true;
 		args.coord = coord_id;
 		args.lod = lod;
+		if (needs_frag_discard_checks())
```
HansKristian-Work, line 8646: This will break if emitted inside a continue block, control flow constructs is banned there.

HansKristian-Work: Alternatively, it might be possible to detect any complicated stores inside a continue block, and force the "complex loop" paths.

spirv_msl.cpp
```diff
@@ -9020,14 +9065,19 @@ void CompilerMSL::emit_instruction(const Instruction &instruction)
...
+			statement(frag_discarded_var_name, " = true;");
 		break;
 
 	case OpIsHelperInvocationEXT:
 		if (msl_options.is_ios() && !msl_options.supports_msl_version(2, 3))
 			SPIRV_CROSS_THROW("simd_is_helper_thread() requires MSL 2.3 on iOS.");
 		else if (msl_options.is_macos() && !msl_options.supports_msl_version(2, 1))
 			SPIRV_CROSS_THROW("simd_is_helper_thread() requires MSL 2.1 on macOS.");
-		emit_op(ops[0], ops[1], "simd_is_helper_thread()", false);
+		if (needs_frag_discard_checks())
```
HansKristian-Work, line 9077: Rather than having two checks here, is it possible to solve this and volatile IsHelperInvocation builtin in one fell swoop, by using reusing the "is helper" builtin here?

BuiltInHelperInvocation is translated to `simd_is_helper_thread()` inline, but if that is lowered to a temporary instead it could unify the IsHelperInvocationEXT and volatile HelperInvocation paths I think.

main.cpp
```diff
@@ -672,6 +672,7 @@ struct CLIArguments
 	bool msl_emulate_subgroups = false;
 	uint32_t msl_fixed_subgroup_size = 0;
 	bool msl_force_sample_rate_shading = false;
+	bool msl_check_discarded_frag_stores = true;
```
HansKristian-Work, line 675: Should this be true? Realistically I think only MoltenVK will care about this. This is a bug that only appears for very specific CTS use cases, right?

reference/shaders-msl/vulkan/frag/demote-to-helper.vk.nocompat.msl23.frag
```diff
@@ -5,8 +5,10 @@ using namespace metal;
 
 fragment void main0()
 {
+    bool spvIsDiscarded = false;
```
HansKristian-Work, line 8: I think a more direct way is that `bool spvIsDiscarded = simd_is_helper_thread();` and then that variable can be updated and passed down.

HansKristian-Work (changes requested):
Comments.

cdavis5e:
> Does simd_is_helper_thread() not update based on helper lane state? This is a requirement of OpIsHelperInvocationEXT at least.

It doesn't. I've filed FB11768488 about this.

spirv_msl.cpp
cdavis5e: Yes, you read that right.

main.cpp
HansKristian-Work: Maybe we need to split the workarounds into two, there's two problems that need to be solved:

- simd_is_helper_thread() being broken. The workaround for that is not invasive. I think we can apply it by default until a fix is known.
- Masking stores is exceptionally invasive and ugly.

spirv_msl.cpp
cdavis5e: One day, though, I'll need to do this differently, in order to properly support `SPV_KHR_terminate_invocation`. Basically, we'll need to set a boolean, like now, only now we immediately return from all functions when we determine an `OpTerminateInvocation` was executed.

spirv_msl.cpp
cdavis5e: Well, that part of the comment's gone now, since I split this up into two options, but my point still stands.

spirv_msl.cpp
cdavis5e: I mostly did the former, but for `OpStore` I was forced to do the latter, due to the way `emit_instruction()` works.

cdavis5e: And it just occurred to me I forgot to add a test case for that. I meant to do that before updating...

spirv_msl.cpp

main.cpp
cdavis5e: Done.

spirv_msl.cpp
cdavis5e: Added.

spirv_msl.cpp
```diff
@@ -1461,6 +1496,20 @@ string CompilerMSL::compile()
 	preprocess_op_codes();
 	build_implicit_builtins();
 
+	if (needs_manual_helper_invocation_updates() &&
+	    (active_input_builtins.get(BuiltInHelperInvocation) || needs_helper_invocation))
+	{
+		string discard_expr =
```
HansKristian-Work, line 1502: Cute hack :p

HansKristian-Work (approved):
Annoying that we have to do this, but oh well.

Issue #2079, opened 2023-01-09
New loop behavior in glslang not detected by `block_is_loop_candidate`


loic-sharma:
Disclaimer: I'm new to this area, please excuse any mistakes!

## Background

After https://github.com/KhronosGroup/glslang/commit/7a914ce9261dd591ec793df42b5aba11d6c02848, glslang emits a new `OpLine` for for loops.

I believe this new behavior breaks `block_is_loop_candidate` as `block.ops` now contains a single `OpLine` instruction:

https://github.com/KhronosGroup/SPIRV-Cross/blob/cded61dde3f78a8d2a563a99501ac855b2e4ffab/spirv_cross.cpp#L1528

This results in suboptimal GLSL when doing GLSL -> SPIRV -> GLSL roundtrip on a for loop using gslang.

## Reproduction

Compiling the following SPIR-V assembly produces unexpected GLSL ES when using these steps:

1. `spirv-as.exe input.spvasm`
2. `spirv-cross.exe --version 310 --es out.spv --output test.comp`

<details>
<summary>Input SPIR-V assembly...</summary>

```
; SPIR-V
; Version: 1.0
; Generator: Google Shaderc over Glslang; 10
; Bound: 33
; Schema: 0
               OpCapability Shader
          %2 = OpExtInstImport "GLSL.std.450"
               OpMemoryModel Logical GLSL450
               OpEntryPoint Fragment %main "main" %fragColor
               OpExecutionMode %main OriginLowerLeft
          %1 = OpString "C:/Code/f/engine/src/flutter/lib/ui/fixtures/shaders/supported_op_shaders/300_Test.frag"
               OpSource GLSL 460 %1 "// OpModuleProcessed entry-point i_300_Test_fragment_main
// OpModuleProcessed auto-map-bindings
// OpModuleProcessed auto-map-locations
// OpModuleProcessed client opengl100
// OpModuleProcessed target-env opengl
// OpModuleProcessed entry-point main
#line 1
#version 320 es

// Copyright 2013 The Flutter Authors. All rights reserved.
// Use of this source code is governed by a BSD-style license that can be
// found in the LICENSE file.

precision highp float;

layout(location = 0) out vec4 fragColor;

layout(location = 0) uniform float a;

void main() {
  float sum = 0.0;
  for (float i = 0.0; i < 6.0; i++) {
    sum += 1;
  }
  fragColor = vec4(0.0, sum, 0.0, 1.0);
}
"
[...]

PR #2080, opened 2023-01-10
Accept glslang's new loop behavior

loic-sharma:
After https://github.com/KhronosGroup/glslang/commit/7a914ce9261dd591ec793df42b5aba11d6c02848, glslang emits a new `OpLine` in for loops. This can be safely ignored.

Relevant specification docs:
1. https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#_miscellaneous_instructions
2. https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#_debug_instructions

Addresses https://github.com/KhronosGroup/SPIRV-Cross/issues/2079

### DIFF
--- a/spirv_cross.cpp
+++ b/spirv_cross.cpp
@@ -1468,6 +1468,39 @@ bool Compiler::get_binary_offset_for_decoration(VariableID id, spv::Decoration d
 	return true;
 }
 
+bool Compiler::block_is_noop(const SPIRBlock &block) const
+{
+	if (block.terminator != SPIRBlock::Direct)
+		return false;
+
+	// Verify all instructions have no semantic impact.
+	for (auto &i : block.ops)
+	{
+		auto op = static_cast<Op>(i.op);
+
+		switch (op)
+		{
+		// Non-Semantic instructions.
+		case OpNop:
+		case OpSourceContinued:
+		case OpSource:
+		case OpSourceExtension:
+		case OpName:
+		case OpMemberName:
+		case OpString:
+		case OpLine:
+		case OpNoLine:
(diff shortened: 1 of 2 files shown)


spirv_cross.cpp
```diff
@@ -1468,6 +1468,39 @@ bool Compiler::get_binary_offset_for_decoration(VariableID id, spv::Decoration d
...
+		case OpNoLine:
+		case OpModuleProcessed:
+			break;
+
+		default:
+			return false;
+		}
+	}
+
+	return true;
```
loic-sharma, line 1501: `Compiler::execution_is_noop` checks the next block's phi variables and verifies their parents aren't the current block. Should I add that here as well?

spirv_cross.cpp
```diff
@@ -1468,6 +1468,39 @@ bool Compiler::get_binary_offset_for_decoration(VariableID id, spv::Decoration d
 	return true;
 }
 
+bool Compiler::block_is_noop(const SPIRBlock &block) const
+{
+	if (block.terminator != SPIRBlock::Direct)
```
loic-sharma, line 1473: Should I include `SPIRBlock::Unreachable`? It's not necessary for `Compiler::block_is_loop_candidate` but one could argue an unreachable block should be no-op.

loic-sharma:
I will double-check that my employer doesn't have any restrictions with me signing the CLA. Apologies for the delay, I will sign it ASAP.

HansKristian-Work:
Yes, I have set aside some time tomorrow to merge this and look at the other PR.

PR #2086, opened 2023-01-14
GLSL/HLSL: Add legacy handling for int vertex attributes

rdb:
Hi, I'm back with more legacy fixes :)

spirv-cross rightly rejects uint vertex attributes, because uints are not supported by legacy GLSL at all.  However it does pass through *signed* integer inputs, even though they are not supported by legacy GLSL (the spec is very explicit that attributes must be float, even if this is popularly accepted, see also KhronosGroup/glslang#3111).

This PR changes these attributes to be declared as float and then automatically generates a typecast to convert them to int/ivec at OpLoad time.  The assumption is that the engine will just use plain old glVertexAttribPointer (since glVertexAttribIPointer is not available in legacy GL anyway) with the normalized flag set to GL_FALSE to supply these attributes.

This change is necessary for our vertex skinning shaders to be portable to legacy targets.

Same change for HLSL, except this case is a little bit simpler as it already generates code to copy the input variables to static globals, so that's where the (implicit, which seems to be OK in HLSL) type cast happens.

### DIFF
--- a/reference/opt/shaders-hlsl/vert/legacy-int-attribute.sm30.vert
+++ b/reference/opt/shaders-hlsl/vert/legacy-int-attribute.sm30.vert
@@ -1,0 +1,36 @@
+uniform float4 gl_HalfPixel;
+
+static float4 gl_Position;
+static int4 attr_int4;
+static int attr_int1;
+
+struct SPIRV_Cross_Input
+{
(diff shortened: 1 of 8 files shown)


HansKristian-Work (approved):
LGTM

PR #2087, opened 2023-01-16
GLSL: Support GL_NV_shader_noperspective_interpolation in GLES

rdb:
The noperspective qualifier isn't in ESSL, so add a check to throw if it is used.

There is a GL_NV_shader_noperspective_interpolation (available for ES 3+) that adds support for it.


HansKristian-Work (approved):
LGTM.

PR #2088, opened 2023-01-17
HLSL: Support PointSize output in vertex shader in SM 3.0

rdb:
Thank you for accepting my other PRs.

This one adds support for the `BuiltInPointSize` output in HLSL SM 3.0, which had a `PSIZE` semantic that matches the behaviour of `gl_PointSize`.

Tested to work with a shader that now shows identical results between DirectX 9 and OpenGL.

The `point_size_compat` option is now ignored if `shader_model <= 30`.


HansKristian-Work (approved):
LGTM

PR #2089, opened 2023-01-18
Win10 doesn't like std::max

warmenhoven:


cdavis5e:
10:1 defining `NOMINMAX` before including `<windows.h>` will also solve this.

HansKristian-Work:
> 10:1 defining NOMINMAX before including <windows.h> will also solve this.

iirc this particular projects builds spirv-cross by including every cpp into a big translation unit, so that would pollute things.

PR #2090, opened 2023-01-19
GLSL: Add support for trunc, modf, isnan and isinf in legacy GLSL

rdb:
These functions are missing in legacy GLSL but added in 130 / 300es.  More coming soon (sinh and friends still missing) but wanted to get this out for review first.

I wanted to implement isinf as `v == v && v * 0.0 != v * 0.0` but that doesn't seem to work reliably on the implementations I tested.  The formula `v != 0.0 && 2.0 * v == v` , which is used in Cg, does work.  I didn't want to simply compare to `1.0 / 0.0` since the result of dividing by zero is unspecified in legacy GLSL, so it might produce false positives somewhere.

Unfortunately in GLSL the && operator doesn't work component-wise and it doesn't have an `and()` function, so I have to write out the isinf expression per component.  I'm not really worried about performance in any case, I'm just adding this function for completeness' sake.

As for GLSLstd450Modf, I couldn't figure out how to prevent it from writing out a temporary for the result.  I am personally not really bothered by it (the driver will probably just optimize them out anyway) but if you want me to fix that, I could use a hint.

trunc, modf, isinf, isnan tested to produce correct results with Intel drivers on Linux.  I haven't tested GLSLstd450ModfStruct because I don't know how to produce that in the SPIR-V.

### DIFF
--- a/spirv_glsl.cpp
+++ b/spirv_glsl.cpp
@@ -7863,8 +7863,22 @@ void CompilerGLSL::emit_glsl_op(uint32_t result_type, uint32_t id, uint32_t eop,
 		break;
 
 	case GLSLstd450Trunc:
-		emit_unary_func_op(result_type, id, args[0], "trunc");
+		if (!is_legacy())
+			emit_unary_func_op(result_type, id, args[0], "trunc");
+		else
+		{
+			// Implement by value-casting to int and back.
+			bool forward = should_forward(args[0]);
+			auto op0 = to_unpacked_expression(args[0]);
+			auto &op0_type = expression_type(args[0]);
+			auto via_type = op0_type;
+			via_type.basetype = SPIRType::Int;
+			auto expr = join(type_to_glsl(op0_type), "(", type_to_glsl(via_type), "(", op0, "))");
+			emit_op(result_type, id, expr, forward);
+			inherit_expression_dependencies(id, args[0]);
+		}
 		break;
+
 	case GLSLstd450SAbs:
 		emit_unary_func_op_cast(result_type, id, args[0], "abs", int_type, int_type);
 		break;
@@ -7906,18 +7920,47 @@ void CompilerGLSL::emit_glsl_op(uint32_t result_type, uint32_t id, uint32_t eop,
 		else
 			emit_trinary_func_op(result_type, id, args[0], args[1], args[2], "fma");
 		break;
+
 	case GLSLstd450Modf:
 		register_call_out_argument(args[1]);
-		forced_temporaries.insert(id);
-		emit_binary_func_op(result_type, id, args[0], args[1], "modf");
+		if (!is_legacy())
+		{
+			forced_temporaries.insert(id);
+			emit_binary_func_op(result_type, id, args[0], args[1], "modf");
+		}
+		else
+		{
+			//NB. legacy GLSL doesn't have trunc() either, so we do a value cast
+			auto &op1_type = expression_type(args[1]);
+			auto via_type = op1_type;
+			via_type.basetype = SPIRType::Int;
+			statement(to_expression(args[1]), " = ",
+			          type_to_glsl(op1_type), "(", type_to_glsl(via_type),
+			          "(", to_expression(args[0]), "));");
+			emit_binary_op(result_type, id, args[0], args[1], "-");
+		}
 		break;
 
 	case GLSLstd450ModfStruct:
 	{
 		auto &type = get<SPIRType>(result_type);
 		emit_uninitialized_temporary_expression(result_type, id);
-		statement(to_expression(id), ".", to_member_name(type, 0), " = ", "modf(", to_expression(args[0]), ", ",
-		          to_expression(id), ".", to_member_name(type, 1), ");");
+		if (!is_legacy())
+		{
+			statement(to_expression(id), ".", to_member_name(type, 0), " = ", "modf(", to_expression(args[0]), ", ",
+			          to_expression(id), ".", to_member_name(type, 1), ");");
+		}
+		else
+		{
+			//NB. legacy GLSL doesn't have trunc() either, so we do a value cast
+			auto &op0_type = expression_type(args[0]);
+			auto via_type = op0_type;
+			via_type.basetype = SPIRType::Int;
+			statement(to_expression(id), ".", to_member_name(type, 1), " = ", type_to_glsl(op0_type),
+			          "(", type_to_glsl(via_type), "(", to_expression(args[0]), "));");
+			statement(to_expression(id), ".", to_member_name(type, 0), " = ", to_enclosed_expression(args[0]), " - ",
+			          to_expression(id), ".", to_member_name(type, 1), ";");
+		}
 		break;
 	}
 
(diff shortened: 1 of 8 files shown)


spirv_glsl.cpp
```diff
@@ -12091,19 +12187,27 @@ void CompilerGLSL::emit_instruction(const Instruction &instruction)
...
 		uint32_t op0 = ops[2];
 		uint32_t op1 = ops[3];
 
 		// Needs special handling.
 		bool forward = should_forward(op0) && should_forward(op1);
-		auto expr = join(to_enclosed_expression(op0), " - ", to_enclosed_expression(op1), " * ", "trunc(",
-		                 to_enclosed_expression(op0), " / ", to_enclosed_expression(op1), ")");
+		std::string expr;
+		if (!is_legacy())
+			expr = join(to_enclosed_expression(op0), " - ", to_enclosed_expression(op1), " * ", "trunc(",
```
HansKristian-Work, line 12199: Multi-line, need braces.

spirv_glsl.cpp
```diff
@@ -11813,11 +11868,52 @@ void CompilerGLSL::emit_instruction(const Instruction &instruction)
...
+
+					if (i + 1 < type.vecsize)
+						expr += ", ";
+				}
+				expr += ')';
+			}
+			else
+			{
+				// Need explicit flush since we only call to_expression once
+				flush_variable_declaration(operand);
```
HansKristian-Work, line 11909: flush_variable_declaration does not do what you expect here. track_expression_read() is probably what you want instead. You can call that after to_enclosed_expression.

spirv_glsl.cpp
```diff
@@ -7892,18 +7906,45 @@ void CompilerGLSL::emit_glsl_op(uint32_t result_type, uint32_t id, uint32_t eop,
...
 		break;
 
 	case GLSLstd450ModfStruct:
 	{
 		auto &type = get<SPIRType>(result_type);
 		emit_uninitialized_temporary_expression(result_type, id);
-		statement(to_expression(id), ".", to_member_name(type, 0), " = ", "modf(", to_expression(args[0]), ", ",
-		          to_expression(id), ".", to_member_name(type, 1), ");");
+		if (!is_legacy())
+			statement(to_expression(id), ".", to_member_name(type, 0), " = ", "modf(", to_expression(args[0]), ", ",
```
HansKristian-Work, line 7935: Multi-line, need braces.

spirv_glsl.cpp
```diff
@@ -7892,18 +7906,45 @@ void CompilerGLSL::emit_glsl_op(uint32_t result_type, uint32_t id, uint32_t eop,
...
+			forced_temporaries.insert(id);
+			emit_binary_func_op(result_type, id, args[0], args[1], "modf");
+		}
+		else
+		{
+			//NB. legacy GLSL doesn't have trunc() either, so we do a value cast
+			auto &op1_type = expression_type(args[1]);
+			auto via_type = op1_type;
+			via_type.basetype = SPIRType::Int;
+			statement(to_expression(args[1], false), " = ",
```
HansKristian-Work, line 7923: What you want here is emit_uninitialized_temporary_expression and then you can add statements to fill in the temporary as needed.

spirv_glsl.cpp
```diff
@@ -11813,11 +11868,52 @@ void CompilerGLSL::emit_instruction(const Instruction &instruction)
 
 	// ALU
 	case OpIsNan:
-		GLSL_UFOP(isnan);
+		if (!is_legacy())
+			GLSL_UFOP(isnan);
+		else if (expression_type(ops[2]).vecsize > 1)
+			// Check if the number doesn't equal itself
```
HansKristian-Work, line 11874: Misleading indentation, wrap in braces.

spirv_glsl.cpp
```diff
@@ -11813,11 +11868,52 @@ void CompilerGLSL::emit_instruction(const Instruction &instruction)
 
 	// ALU
 	case OpIsNan:
-		GLSL_UFOP(isnan);
+		if (!is_legacy())
+			GLSL_UFOP(isnan);
+		else if (expression_type(ops[2]).vecsize > 1)
```
HansKristian-Work, line 11873: Can check ops[0] type instead instead of expression_type.

spirv_glsl.cpp
```diff
@@ -11813,11 +11868,52 @@ void CompilerGLSL::emit_instruction(const Instruction &instruction)
...
+			emit_binary_op(ops[0], ops[1], ops[2], ops[2], "!=");
 		break;
 
 	case OpIsInf:
-		GLSL_UFOP(isinf);
+		if (!is_legacy())
+			GLSL_UFOP(isinf);
+		else
+		{
+			// inf * 2 == inf by IEEE 754 rules, note this also applies to 0.0
```
HansKristian-Work, line 11885: A bit questionable without precise, but since this is only for legacy targets, sure.

spirv_glsl.cpp
```diff
@@ -12091,19 +12187,27 @@ void CompilerGLSL::emit_instruction(const Instruction &instruction)
...
 		bool forward = should_forward(op0) && should_forward(op1);
-		auto expr = join(to_enclosed_expression(op0), " - ", to_enclosed_expression(op1), " * ", "trunc(",
-		                 to_enclosed_expression(op0), " / ", to_enclosed_expression(op1), ")");
+		std::string expr;
+		if (!is_legacy())
+			expr = join(to_enclosed_expression(op0), " - ", to_enclosed_expression(op1), " * ", "trunc(",
+			            to_enclosed_expression(op0), " / ", to_enclosed_expression(op1), ")");
+		else
+		{
+			// Legacy GLSL has no trunc, emulate by casting to int and back
```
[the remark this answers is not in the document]
rdb: I think it's OK, because the GLSL 1.20 / ESSL 1.00 specs define the behaviour of the int cast as truncating:
> When constructors are used to convert a **float** to an **int**, the fractional part of the floating-point value is dropped.

Alternative is implementing with `floor(abs(x)) * sign(x)` but I think the behaviour of cast is defined well enough as-is (we can revisit if we find actual examples of drivers that disagree).

Thanks for the review, I'll get to work on addressing the other points.

spirv_glsl.cpp
rdb: This doesn't seem to work.  I get an exception "Overwriting a variant with new type."

I tried:
```cpp
emit_uninitialized_temporary_expression(op1_type.self, args[1]);
```

Note that this is an out argument (via OpVariable), not a result.  I assumed `register_call_out_argument` was handling this case, and it seemed to be working.

HansKristian-Work: Right, this is targeting args[1]. I was a bit confused since you did to_expression(foo, false). What was the purpose of the "false" here?

rdb: Just saw your reply.

I added the "false" because while I was trying to figure out why it was always emitting the result temporary, I looked at the argument, it said "register read", thought "well it's not a read", then set it in the hope it'd fix that.  But it didn't.  I'll remove it.

rdb:
I believe I've addressed all the issues except if you want further changes regarding modf.  Checks pass.

HansKristian-Work (approved):
LGTM

PR #2093, opened 2023-01-19
GLSL/HLSL: Add emulated fallbacks for hyperbolic math functions

rdb:
Hyperbolics were added in GLSL 1.30.  Furthermore, the inverse functions are not supported in any version of HLSL.  This implements them.  The math has been tested.

I would appreciate a review of the GLSLstd450Tanh logic in particular since it creates intermediate temporaries.  I'm not sure I understand the `extra_sub_expressions` logic and whether I need to use it here.

### DIFF
--- a/spirv_glsl.cpp
+++ b/spirv_glsl.cpp
@@ -7978,22 +7978,77 @@ void CompilerGLSL::emit_glsl_op(uint32_t result_type, uint32_t id, uint32_t eop,
 		emit_unary_func_op(result_type, id, args[0], "atan");
 		break;
 	case GLSLstd450Sinh:
-		emit_unary_func_op(result_type, id, args[0], "sinh");
+		if (!is_legacy())
+			emit_unary_func_op(result_type, id, args[0], "sinh");
+		else
+		{
+			bool forward = should_forward(args[0]);
+			auto expr = join("(exp(", to_expression(args[0]), ") - exp(-", to_enclosed_expression(args[0]), ")) * 0.5");
+			emit_op(result_type, id, expr, forward);
+			inherit_expression_dependencies(id, args[0]);
+		}
 		break;
 	case GLSLstd450Cosh:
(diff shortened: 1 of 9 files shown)


spirv_glsl.cpp
```diff
@@ -7964,22 +7964,69 @@ void CompilerGLSL::emit_glsl_op(uint32_t result_type, uint32_t id, uint32_t eop,
...
+		}
 		break;
 	case GLSLstd450Tanh:
-		emit_unary_func_op(result_type, id, args[0], "tanh");
+		if (!is_legacy())
+			emit_unary_func_op(result_type, id, args[0], "tanh");
+		else
+		{
+			// Create temporaries to store the result of exp(arg) and exp(-arg).
+			uint32_t epos_id = ir.increase_bound_by(2);
```
HansKristian-Work, line 7994: The more proper way to do this is to make use of the `extra_sub_expressions` map.

E.g. `uint32_t &ids = extra_sub_expressions[result_id]` if ids is 0, allocate 2 IDs, then inherit the RelaxedPrecision.

spirv_glsl.cpp
```diff
@@ -7964,22 +7964,69 @@ void CompilerGLSL::emit_glsl_op(uint32_t result_type, uint32_t id, uint32_t eop,
...
+			uint32_t eneg_id = epos_id + 1;
+
+			// Inherit precision qualifiers.
+			ir.meta[epos_id] = ir.meta[id];
```
HansKristian-Work, line 7998: Copying the entire meta struct is wasteful. Just check `if (has_decoration()) set_decoration()` instead.

spirv_glsl.cpp
HansKristian-Work: This is to avoid endless reallocation of IDs on recompilation steps.

HansKristian-Work (changes requested):

HansKristian-Work (approved):
LGTM.

PR #2095, opened 2023-01-21
GLSL: Multiplying matrix with scalar shouldn't force transpose

rdb:
Sorry for bombarding you with PRs lately!

This one is simple, it just prevents forcing the transpose when multiplying a matrix times a scalar.  We can just set need_transpose on the result.  It should be rare to actually need a forced transpose.

This comes up in eg. skinning shaders where a matrices might be linearly interpolated based on a set of blend weights, then multiplied with a vector.  No need to force transpose on all those matrices when we can just flip the multiplication order in the end.


HansKristian-Work (approved):
LGTM

PR #2100, opened 2023-01-27
GLSL: Add matrix inverse/determinant polyfills, outerProduct()

rdb:
Phew, I think this completes the set of functions for older GLSL versions:

* `outerProduct` was added in GLSL 1.20 / ESSL 3.00
* `inverse` was added in GLSL 1.40 / ESSL 3.00
* `determinant` was added in GLSL 1.50 / ESSL 3.00

I thought it was maybe cleaner to generalise the polyfill mechanism a bit, I can change it if you don't like it.

outerProduct was copied from the HLSL version and tested.

Sorry about the big scary cloud of math in spvInverse, I assure you that I double-checked and tested it.

This also changes inverse/determinant to no longer unnecessarily force a transpose.


spirv_glsl.cpp
```diff
@@ -8024,11 +8130,69 @@ void CompilerGLSL::emit_glsl_op(uint32_t result_type, uint32_t id, uint32_t eop,
...
+		{
+			auto &type = get<SPIRType>(result_type);
+			assert(type.vecsize >= 2 && type.vecsize <= 4);
+			assert(type.vecsize == type.columns);
+
+			// ARB_gpu_shader_fp64 needs GLSL 150, other types are invalid
+			if (type.basetype != SPIRType::Float)
+				SPIRV_CROSS_THROW("Unsupported type for matrix inverse");
+
+			require_polyfill((Polyfill)(PolyfillMatrixInverse2x2 << (type.vecsize - 2)));
```
HansKristian-Work, line 8181: `static_cast<Polyfill>` instead of C cast.

spirv_glsl.cpp
```diff
@@ -4429,6 +4444,97 @@ void CompilerGLSL::emit_extension_workarounds(spv::ExecutionModel model)
...
+		                      "m[0][2] * m[1][0] - m[0][0] * m[1][2], "
+		                      "t[2], "
+		                      "m[0][1] * m[2][0] - m[0][0] * m[2][1], "
+		                      "m[0][0] * m[1][1] - m[0][1] * m[1][0]) "
+		                      "* (1.0 / dot(m[0], t));");
+		end_scope();
+		statement("");
+	}
+
+	if (required_polyfills & PolyfillMatrixInverse4x4) {
```
HansKristian-Work, line 4509: Brace.

spirv_glsl.cpp
```diff
@@ -4429,6 +4444,97 @@ void CompilerGLSL::emit_extension_workarounds(spv::ExecutionModel model)
...
+		statement("float spvDeterminant(mat3 m)");
+		begin_scope();
+		statement("return dot(m[0], vec3(m[1][1] * m[2][2] - m[1][2] * m[2][1], "
+		                                "m[1][2] * m[2][0] - m[1][0] * m[2][2], "
+		                                "m[1][0] * m[2][1] - m[1][1] * m[2][0]));");
+		end_scope();
+		statement("");
+	}
+
+	if (required_polyfills & PolyfillDeterminant4x4) {
```
HansKristian-Work, line 4468: Brace.

spirv_glsl.cpp
```diff
@@ -8024,11 +8130,69 @@ void CompilerGLSL::emit_glsl_op(uint32_t result_type, uint32_t id, uint32_t eop,
...
+		{
+			auto type = expression_type(args[0]);
+			assert(type.vecsize >= 2 && type.vecsize <= 4);
+			assert(type.vecsize == type.columns);
+
+			// ARB_gpu_shader_fp64 needs GLSL 150, other types are not valid
+			if (type.basetype != SPIRType::Float)
+				SPIRV_CROSS_THROW("Unsupported type for matrix determinant");
+
+			require_polyfill((Polyfill)(PolyfillDeterminant2x2 << (type.vecsize - 2)));
```
HansKristian-Work, line 8150: `static_cast<Polyfill>`

spirv_glsl.cpp
```diff
@@ -4401,8 +4413,11 @@ void CompilerGLSL::emit_extension_workarounds(spv::ExecutionModel model)
 		}
 		statement("");
 	}
+}
 
-	if (requires_transpose_2x2)
+void CompilerGLSL::emit_polyfills()
+{
+	if (required_polyfills & PolyfillTranspose2x2)
```
HansKristian-Work, line 4420: Note that there are problems here with mediump vs high. If a highp matrix is used in a fragment shader, it will be demoted to mediump here. See row major load workaround where it needs to stamp out two variants that are highp and mediump.

The existing transpose code is also bugged, so I can fix it up later if it's a hassle.

HansKristian-Work (changes requested):
Small nits. I'll assume the math is correct.

spirv_glsl.cpp
rdb: Will fix across the board while I'm at it.  I assume we don't care about `lowp`?

HansKristian-Work: No. SPIR-V does not support lowp.

spirv_glsl.cpp
```diff
@@ -701,6 +712,11 @@ string CompilerGLSL::compile()
 		emit_resources();
 		emit_extension_workarounds(get_execution_model());
 
+		if (required_polyfills != 0)
+			emit_polyfills(required_polyfills, false);
+		if (options.es && required_polyfills_relaxed != 0)
```
HansKristian-Work, line 717: Vulkan GLSL also needs highp vs mediump.

HansKristian-Work (approved):
LGTM, thanks.

PR #2103, opened 2023-02-01
MSL: Add a workaround for broken `level()` arguments.

cdavis5e:
Some Metal devices have a bug with depth array textures using comparison
with explicit LoD, where the LoD given will be biased by some amount.
For these devices, we can use a gradient instead, which does not exhibit
this problem. As with the fragment demote workaround, this is only
expected to be needed until the bug is fixed in Metal.

Fixes two CTS tests `dEQP-VK.glsl.texture_functions.texturelod{,offset}.sampler1darrayshadow_fragment`.

### DIFF
--- a/spirv_msl.cpp
+++ b/spirv_msl.cpp
@@ -10823,7 +10823,8 @@ string CompilerMSL::to_function_args(const TextureFunctionArguments &args, bool 
 			// We will detect a compile-time constant 0 value for gradient and promote that to level(0) on MSL.
 			bool constant_zero_x = !grad_x || expression_is_constant_null(grad_x);
 			bool constant_zero_y = !grad_y || expression_is_constant_null(grad_y);
-			if (constant_zero_x && constant_zero_y)
+			if (constant_zero_x && constant_zero_y &&
+			    (!imgtype.image.arrayed || !msl_options.sample_dref_lod_array_as_grad))
 			{
 				lod = 0;
 				grad_x = 0;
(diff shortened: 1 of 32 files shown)


spirv_msl.cpp
```diff
@@ -10869,6 +10870,57 @@ string CompilerMSL::to_function_args(const TextureFunctionArguments &args, bool
...
+			// Some Metal devices have a bug where the LoD is erroneously biased upward
+			// when using a level() argument. Since this doesn't happen with gradient2d(),
+			// if we perform the LoD calculation in reverse, we can pass a gradient
+			// instead.
+			// lod = log2(rhoMax/eta) -> exp2(lod) = rhoMax/eta
+			// If we make all of the scale factors the same, eta will be 1 and
+			// exp2(lod) = rho.
+			// rhoX = dP/dx * extent; rhoY = dP/dy * extent
+			// Therefore, dP/dx = dP/dy = exp2(lod)/extent.
+			// (Subtracting 0.5 before exponentiation gives better results.)
```
HansKristian-Work, line 10887: Subtracting 0.5 here seems a bit arbitrary and hacky. That would imply a -0.5 lod bias. Is this required to avoid rounding shenanigans when landing on the mid-point between two levels, e.g. 1.5 or 2.5?

spirv_msl.cpp
```diff
@@ -10869,6 +10870,57 @@ string CompilerMSL::to_function_args(const TextureFunctionArguments &args, bool
...
+					grad_opt = "cube";
+					extent = join("float3(", to_expression(img), ".get_width())");
+				}
+				break;
+			default:
+				grad_opt = "unsupported_gradient_dimension";
+				extent = "float3(1.0)";
+				break;
+			}
+			farg_str += join(", gradient", grad_opt, "(exp2(", to_expression(lod), " - 0.5) / ", extent, ", exp2(",
```
HansKristian-Work, line 10921: This gradient is 2D only it seems. Does it work with Cube as expected?

spirv_msl.cpp
```diff
@@ -10869,6 +10870,57 @@ string CompilerMSL::to_function_args(const TextureFunctionArguments &args, bool
...
+			{
+			case Dim1D:
+				grad_opt = "2d";
+				extent = join("float2(", to_expression(img), ".get_width(), 1.0)");
+				break;
+			case Dim2D:
+				grad_opt = "2d";
+				extent = join("float2(", to_expression(img), ".get_width(), ", to_expression(img), ".get_height())");
+				break;
+			case Dim3D:
```
HansKristian-Work, line 10899: Do we need to handle 3D here if depth images cannot be 3D?

HansKristian-Work (changes requested):
Comments.

spirv_msl.cpp
cdavis5e: No. It's actually required to make the LoD come out right. The CTS test actually fails without this. And I agree--that 0.5 fudge factor is kludgy.

spirv_msl.cpp
cdavis5e: `gradientcube` only takes two arguments, one for dP/dx and one for dP/dy. The difference is that P has three elements for `cube`, but only two for `2d`.

spirv_msl.cpp
cdavis5e: I actually debated whether or not to leave that in, in case 3D depth images ever become possible. I'll remove it.

spirv_msl.cpp
HansKristian-Work: Right, it's just the dimension of the gradient that changes.

spirv_msl.cpp
HansKristian-Work: Ok, since this is just a hacky workaround, I guess we can go with it ...

HansKristian-Work (approved):

PR #2105, opened 2023-02-05
add support for OpAtomicFAddEXT atomic add, and also Min, Max

rcoreilly:
These are now available in Metal 3.0, which is only enabled on MacOS 13.x in MoltenVK, where OpAtomicFAddEXT has been tested all the way through, per this discussion: https://github.com/KhronosGroup/MoltenVK/discussions/1616#discussioncomment-4868519

Can also run the metal compiler on code generated using spirv-cross under 12.x and it works fine.

In my testing, it appeared to already have all the necessary logic to require the https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VK_EXT_shader_atomic_float.html and associated SPV extensions to actually enable the use of atomic add on float types, so all that was needed was to put the parsing in the right place.

### DIFF
--- a/test_shaders.py
+++ b/test_shaders.py
@@ -124,7 +124,9 @@ def msl_compiler_supports_version(version):
 
 def path_to_msl_standard(shader):
     if '.ios.' in shader:
-        if '.msl2.' in shader:
+        if '.msl3.' in shader:
+            return '-std=metal3.0'
+        elif '.msl2.' in shader:
             return '-std=ios-metal2.0'
         elif '.msl21.' in shader:
             return '-std=ios-metal2.1'
@@ -141,7 +143,9 @@ def path_to_msl_standard(shader):
         else:
             return '-std=ios-metal1.2'
     else:
-        if '.msl2.' in shader:
+        if '.msl3.' in shader:
+            return '-std=metal3.0'
+        elif '.msl2.' in shader:
             return '-std=macos-metal2.0'
         elif '.msl21.' in shader:
             return '-std=macos-metal2.1'
@@ -157,7 +161,9 @@ def path_to_msl_standard(shader):
             return '-std=macos-metal1.2'
 
 def path_to_msl_standard_cli(shader):
-    if '.msl2.' in shader:
+    if '.msl3.' in shader:
+        return '30000'
+    elif '.msl2.' in shader:
         return '20000'
     elif '.msl21.' in shader:
         return '20100'
--- a/test_shaders.py
+++ b/test_shaders.py
@@ -110,17 +110,30 @@ def print_msl_compiler_version():
         pass
 
 def msl_compiler_supports_version(version):
-    try:
-        subprocess.check_call(['xcrun', '--sdk', 'macosx', 'metal', '-x', 'metal', '-std=macos-metal' + version, '-'],
+    if float(version) >= 3.0:
+        try:
+            subprocess.check_call(['xcrun', '--sdk', 'macosx', 'metal', '-x', 'metal', '-std=metal' + version, '-'],
                 stdin = subprocess.DEVNULL, stdout = subprocess.DEVNULL, stderr = subprocess.DEVNULL)
-        print('Current SDK supports MSL {0}. Enabling validation for MSL {0} shaders.'.format(version))
-        return True
-    except OSError as e:
-        print('Failed to check if MSL {} is not supported. It probably is not.'.format(version))
-        return False
-    except subprocess.CalledProcessError:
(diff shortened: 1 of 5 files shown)


TellowKrinkle:
From MSL Spec section 6.15.2.5: "These functions are applicable to any atomic object of type atomic_int, atomic_uint, and atomic_bool. Add and sub are supported for atomic_float."

I can confirm that the Metal frontend won't compile min/max for atomic floats.

rcoreilly:
Ok so should I remove the Min / Max cases then?

cdavis5e:
> Ok so should I remove the Min / Max cases then?

Yes. They aren't required by `SPV_EXT_shader_atomic_float_add`, and thus not by `VK_EXT_shader_atomic_float`. That functionality isn't provided until `SPV_EXT_shader_atomic_float_min_max`/`VK_EXT_shader_atomic_float2`.

cdavis5e:
Also, this should really have a test case showing that `SPV_EXT_shader_atomic_float_add` functionality is successfully converted to MSL.

rcoreilly:
any tips on how to add tests?

cdavis5e:
> any tips on how to add tests?

Just add a GLSL file under `shaders-msl/comp` that exercises the functionality of the `GL_EXT_shader_atomic_float` extension. Then run `test_shaders.sh` in the SPIRV-Cross root; it will generate a reference output for you, and verify that the output compiles with Apple's compiler. (This will also validate that you didn't break anything else.)

rcoreilly:
`test_shaders.py` does not go up to msl version 3.0.  should I try to figure out how to make that work or perhaps someone with greater familiarity could get that part working?  I know it does work with msl3.0 so someone could merge this PR and then fix that?  I can push my test .comp file but it fails.  Also the test system would have to be running on MacOS 13.x or otherwise have metal 3.0 installed.

cdavis5e:
> `test_shaders.py` does not go up to msl version 3.0. should I try to figure out how to make that work or perhaps someone with greater familiarity could get that part working?

The code you need to change starts [here](https://github.com/KhronosGroup/SPIRV-Cross/blob/4e2fdb25671c742a9fbe93a6034eb1542244c7e1/test_shaders.py#L125). Just add a case for `.msl3.` to each of those three functions. The string you need for `path_to_msl_standard_cli()` is `'30000'`. (In general, it's major * 10000 + minor * 100.) 

> Also the test system would have to be running on MacOS 13.x or otherwise have metal 3.0 installed.

Not macOS 13. Just Xcode 14. But yes. For this reason, the test harness also [checks what versions the installed Metal compiler supports](https://github.com/KhronosGroup/SPIRV-Cross/blob/4e2fdb25671c742a9fbe93a6034eb1542244c7e1/test_shaders.py#L112), and [disables tests that require too new Metal](https://github.com/KhronosGroup/SPIRV-Cross/blob/4e2fdb25671c742a9fbe93a6034eb1542244c7e1/test_shaders.py#L1022). You should change that part, too.

rcoreilly:
ok should be good -- passes test.

cdavis5e:
> For this reason, the test harness also [checks what versions the installed Metal compiler supports](https://github.com/KhronosGroup/SPIRV-Cross/blob/4e2fdb25671c742a9fbe93a6034eb1542244c7e1/test_shaders.py#L112), and [disables tests that require too new Metal](https://github.com/KhronosGroup/SPIRV-Cross/blob/4e2fdb25671c742a9fbe93a6034eb1542244c7e1/test_shaders.py#L1022). **You should change that part, too.** [emphasis added]

rcoreilly:
ok got that part too I think.

test_shaders.py
```diff
@@ -110,21 +110,36 @@ def print_msl_compiler_version():
         pass
 
 def msl_compiler_supports_version(version):
-    try:
-        subprocess.check_call(['xcrun', '--sdk', 'macosx', 'metal', '-x', 'metal', '-std=macos-metal' + version, '-'],
+    if float(version) >= 3.0:
```
HansKristian-Work, line 113: This is a bit too awkward. Duplicated code path is also not great. The function should just be:

```
msl_compiler_supports_version('-std=macos-metal2.0')
msl_compiler_supports_version('-std=macos-metal2.1')
msl_compiler_supports_version('-std=metal3.0')
```
That avoids the duplicated path and the bizarre float cast.

HansKristian-Work (changes requested):
Comment

test_shaders.py
rcoreilly: good idea -- cleaner.

PR #2108, opened 2023-02-17
MSL: Deduplicate function constants.

cdavis5e:
It is possible in SPIR-V to declare multiple specialization constants with the same constant ID. The most common cause of this in GLSL is defining a spec constant, then declaring the workgroup size to use that spec constant by its ID. But, MSL forbids defining multiple function constants with the same function constant ID. So, we must only emit one definition of the actual function constant (with the `[[function_constant(id)]]` attribute); but we can point the other variables at this one definition.

Fixes three tests in the Vulkan CTS under
`dEQP-VK.compute.basic.max_local_size_*`.

### DIFF
--- a/spirv_msl.cpp
+++ b/spirv_msl.cpp
@@ -7344,6 +7344,15 @@ void CompilerMSL::emit_specialization_constants_and_structs()
 	emitted = false;
 	declared_structs.clear();
 
+	// It is possible to have multiple spec constants that use the same spec constant ID.
+	// The most common cause of this is defining spec constants in GLSL while also declaring
+	// the workgroup size to use those spec constants. But, Metal forbids declaring more than
+	// one variable with the same function constant ID.
+	// In this case, we must only declare one variable with the [[function_constant(id)]]
+	// attribute, and use its initializer to initialize all the spec constants with
+	// that ID.
+	std::unordered_map<uint32_t, ConstantID> unique_func_constants;
+
 	for (auto &id_ : ir.ids_for_constant_undef_or_type)
 	{
 		auto &id = ir.ids[id_];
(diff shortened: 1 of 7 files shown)


spirv_msl.cpp
```diff
@@ -7344,6 +7344,20 @@ void CompilerMSL::emit_specialization_constants_and_structs()
...
 	declared_structs.clear();
 
+	// It is possible to have multiple spec constants that use the same spec constant ID.
+	// The most common cause of this is defining spec constants in GLSL while also declaring
+	// the workgroup size to use those spec constants. But, Metal forbids declaring more than
+	// one variable with the same function constant ID.
+	// In this case, we must only declare one variable with the [[function_constant(id)]]
+	// attribute, and use its initializer to initialize all the spec constants with
+	// that ID.
+	std::unordered_map<uint32_t, ConstantID> unique_func_constants;
```
HansKristian-Work, line 7354: This seems a bit inefficient, and problematic if get_specialization_constants iterates in a different order than ids_for_constant_undef_or_type. I don't think it does right now, but no need to risk it after refactors down the line.

It would make more sense to me to move the unique check logic to where the specialization constant is actually emitted. If we have seen the ID before, refer to that, otherwise emit a new function constant.

HansKristian-Work: Is it valid to specify the same constant ID but with different types? Might be worth asserting that scenario I think.

HansKristian-Work (changes requested):

spirv_msl.cpp
cdavis5e: > It would make more sense to me to move the unique check logic to where the specialization constant is actually emitted. If we have seen the ID before, refer to that, otherwise emit a new function constant.

Done.

> Is it valid to specify the same constant ID but with different types? Might be worth asserting that scenario I think.

GLSL won't let you declare multiple spec constants with the same ID at all; it only generates this scenario in the specific case of defining spec constants, then using them to define the workgroup size. SPIR-V itself, however, *will* let you do this, even with different types. Nothing in the SPIR-V or Vulkan specs forbids this, and a manually constructed shader that does this is accepted by `spirv-val`. This seems to me to be an underspecified corner case.

HansKristian-Work (approved):

PR #2110, opened 2023-02-23
GLSL: Don't create temporaries for sampled image types in GLSL.

LukasBanana:
This fixes a bug that was exposed when targeting GLSL with Vulkan semantics while 'force_temporary' is enabled.

### DIFF
--- a/spirv_glsl.cpp
+++ b/spirv_glsl.cpp
@@ -6467,7 +6467,10 @@ void CompilerGLSL::emit_unary_func_op(uint32_t result_type, uint32_t result_id, 
 void CompilerGLSL::emit_binary_func_op(uint32_t result_type, uint32_t result_id, uint32_t op0, uint32_t op1,
                                        const char *op)
 {
-	bool forward = should_forward(op0) && should_forward(op1);
+	// Opaque types (e.g. OpTypeSampledImage) must always be forwarded in GLSL
+	const auto &type = get_type(result_type);
+	bool must_forward = type_is_opaque_value(type);
+	bool forward = must_forward || (should_forward(op0) && should_forward(op1));
 	emit_op(result_type, result_id, join(op, "(", to_unpacked_expression(op0), ", ", to_unpacked_expression(op1), ")"),
 	        forward);
 	inherit_expression_dependencies(result_id, op0);


spirv_glsl.cpp
```diff
@@ -6467,7 +6467,10 @@ void CompilerGLSL::emit_unary_func_op(uint32_t result_type, uint32_t result_id,
 void CompilerGLSL::emit_binary_func_op(uint32_t result_type, uint32_t result_id, uint32_t op0, uint32_t op1,
                                        const char *op)
 {
-	bool forward = should_forward(op0) && should_forward(op1);
+	// Sampled image types must always be forwarded in GLSL unless they are a uniform variable
+	const auto &type = get_type(result_type);
+	bool must_forward = type.basetype == SPIRType::SampledImage && type.storage != spv::StorageClassUniform;
```
HansKristian-Work, line 6472: This check is not complete. Need to consider Sampler and Image as well. There should be a helper somewhere for "is_opaque_type" or something like that.

HansKristian-Work (changes requested):

spirv_glsl.cpp
LukasBanana: Thanks for the hint, I added `type_is_opaque_value` to the condition.

spirv_glsl.cpp
```diff
@@ -6467,7 +6467,10 @@ void CompilerGLSL::emit_unary_func_op(uint32_t result_type, uint32_t result_id,
 void CompilerGLSL::emit_binary_func_op(uint32_t result_type, uint32_t result_id, uint32_t op0, uint32_t op1,
                                        const char *op)
 {
-	bool forward = should_forward(op0) && should_forward(op1);
+	// Opaque types (e.g. OpTypeSampledImage) must always be forwarded in GLSL unless they are a uniform variable
+	const auto &type = get_type(result_type);
+	bool must_forward = type_is_opaque_value(type) && type.storage != spv::StorageClassUniform;
```
HansKristian-Work, line 6472: This comment is a bit misleading because StorageClassUniform (UBO / SSBO in SPIR-V 1.0) != StorageClassUniformConstant (descriptors). I think opaque types always have to be forwarded no matter what, just checking if type is opaque should be enough?

HansKristian-Work (changes requested):

spirv_glsl.cpp
LukasBanana: I think so, I wasn't sure if the uniform storage class is handled differently here.

HansKristian-Work (approved):
LGTM if CI passes.

PR #2121, opened 2023-03-09
adds Khronos license to LICENSES folder.

Keenuts:
This license is in spirv.h, but absent from the LICENSES file.


HansKristian-Work:
@oddhack Is this something you need to look at?

oddhack:
> @oddhack Is this something you need to look at?

I suggest using a more qualified SPDX identifier (LicenseRef-KhronosFreeUse? I'm not sure what we actually call this specific MIT variant or indeed if we have a formal name for it).

Otherwise, it looks like this (whitespace changes aside) is just putting a copy of the license statement in LICENSES/ as REUSE requires, which is good. You could make the headers a bit shorter by replacing the full license text with '// SPDX-License-Identifier: LicenseRef-Khronos\<whatever>'. But that would probably require corresponding changes in the generator scripts, which I think are in another repo, so best to leave that complication for another PR.

Keenuts:
Thanks for looking into that 😊

> I suggest using a more qualified SPDX identifier (LicenseRef-KhronosFreeUse? I'm not sure what we actually call this specific MIT variant or indeed if we have a formal name for it).

done!


> Otherwise, it looks like this (whitespace changes aside)
Is the whitespace change an issue? (Just thought I'd clean this at the same occasion)

oddhack:
> Thanks for looking into that blush
> 
> > I suggest using a more qualified SPDX identifier (LicenseRef-KhronosFreeUse? I'm not sure what we actually call this specific MIT variant or indeed if we have a formal name for it).
> 
> done!
> 
> > Otherwise, it looks like this (whitespace changes aside)
> > Is the whitespace change an issue? (Just thought I'd clean this at the same occasion)

I'm not very familiar with the SPIR repository ecosystem but I had the impression that this header is generated elsewhere and copied in here, so any local changes could be overwritten eventually? If not then it's not an issue.

oddhack (approved):

Keenuts:
>  Ahh yes, probably copied from SPIRV-Headers.
Reverting this so it's clean.

Issue #1974, opened 2022-07-05
[HLSL] Add support for `SPV_NV_mesh_shader`


Try:
DirectX12 now has similar(more rich actually) functionality out of the box, so it would be nice to translate `mesh` and `task` shaders.
Links:
https://github.com/KhronosGroup/GLSL/blob/master/extensions/nv/GLSL_NV_mesh_shader.txt
https://microsoft.github.io/DirectX-Specs/d3d/MeshShader.html#checkfeaturesupport

Core functionality is almost identical, but there are corner-cases with `points` and others.

#### General
 1. `local_size_*` and `[numthreads]`
   Vulkan:  `local_size_x<=32; local_size_y=1; local_size_=1`
   DirectX12: `numthreads X*Y*Z <= 128` 
2. Intrisics:
   `gl_WorkGroupSize`, `gl_WorkGroupID`, and others - same as compute shader. 

#### Mesh
 1. Output data size:
   Vulkan: compile time known by `layout(triangles, max_vertices = *, max_primitives = *) out;` 
   DirectX12:  determinate by array size of `out` data.
 2. Output topology:
   Vulkan: compile time known by `layout(triangles) out`. Possible values are: `points`, `lines`, `triangles`
   DirectX12: compile time known by: `[OutputTopology("triangle")]`. Possible values are: `"lines"`, `"triangles"`

#### Task
Task is mainly compute shader, except special builtin outputs:
1. `gl_TaskCountNV` -> doesn't maps directly, should be passed to `DispatchMesh`
2. `taskNV` -> also parameter to `DispatchMesh`. Mesh stage receives it as `in payload`.
NOTE: tested locally task translation, since it's quite simple: apparently there is no version of `DispatchMesh` in HLSL that doesn't takes `payload_t` input and `payload_t` cannot be empty struct. This brings extra complexity to translation.

#### Builtin variables
1. `gl_WorkGroupSize` ->  N/A, but it's not an issue
4. `gl_WorkGroupID` -> `SV_GroupID`
5. `gl_LocalInvocationID` -> `SV_GroupThreadID`
6. `gl_GlobalInvocationID` -> `SV_DispatchThreadID`
7. `gl_LocalInvocationIndex` -> `SV_GroupIndex`
8. `gl_TaskCountNV` -> `DispatchMesh(count,1,1,usrTaskData)` at tail of `amplification` stage
[...]

PR #2124, opened 2023-03-18
HLSL: Add task(amplification) shader support

Try:
#1974

### DIFF
--- a/spirv_glsl.cpp
+++ b/spirv_glsl.cpp
@@ -14316,6 +14316,10 @@ void CompilerGLSL::emit_instruction(const Instruction &instruction)
 		statement("SetMeshOutputsEXT(", to_unpacked_expression(ops[0]), ", ", to_unpacked_expression(ops[1]), ");");
 		break;
 
+	case OpEmitMeshTasksEXT:
+		statement("EmitMeshTasksEXT(", to_unpacked_expression(ops[0]), ", ", to_unpacked_expression(ops[1]), ", ", to_unpacked_expression(ops[2]), ");");
+		break;
+
 	case OpReadClockKHR:
 	{
 		auto &type = get<SPIRType>(ops[0]);
@@ -16873,10 +16877,6 @@ void CompilerGLSL::emit_block_chain(SPIRBlock &block)
 		break;
 
 	case SPIRBlock::EmitMeshTasks:
-		statement("EmitMeshTasksEXT(",
-		          to_unpacked_expression(block.mesh.groups[0]), ", ",
-		          to_unpacked_expression(block.mesh.groups[1]), ", ",
-		          to_unpacked_expression(block.mesh.groups[2]), ");");
 		break;
 
 	default:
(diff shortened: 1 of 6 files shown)


spirv_hlsl.cpp
```diff
@@ -6355,6 +6361,19 @@ void CompilerHLSL::emit_instruction(const Instruction &instruction)
...
+	case OpEmitMeshTasksEXT:
+	{
+		if (instruction.length == 4)
+		{
+			statement("DispatchMesh(", to_unpacked_expression(ops[0]), ", ", to_unpacked_expression(ops[1]), ", ",
+			    to_unpacked_expression(ops[2]), ", ", to_unpacked_expression(ops[3]), ");");
+		}
+		else
+		{
+			SPIRV_CROSS_THROW("Amplification shader in HLSL must have payload");
```
Try, line 6373: I've tested with DXC: empty struct as payload compiles, yet produces invalid byte-code
`bcf5408759f3.hlsl:17:5: error: For amplification shader with entry 'main', payload size 4 is greater than declared size of 0 bytes`
https://shader-playground.timjones.io/8d8dbf3b63aa67aa3d000327b56915dd

spirv_parser.cpp
```diff
@@ -1123,16 +1123,21 @@ void Parser::parse(const Instruction &instruction)
 		break;
 
 	case OpEmitMeshTasksEXT:
+	{
 		if (!current_block)
 			SPIRV_CROSS_THROW("Trying to end a non-existing block.");
+
+		const auto *type = maybe_get<SPIRType>(ops[0]);
```
HansKristian-Work, line 1130: This looks wrong. EmitMeshTasksEXT does not return anything.

spirv_hlsl.cpp
```diff
@@ -2931,8 +2936,8 @@ void CompilerHLSL::emit_hlsl_entry_point()
 
 	switch (execution.model)
 	{
+	case spv::ExecutionModelTaskEXT:
```
HansKristian-Work, line 2939: Drop spv::

spirv_glsl.cpp
```diff
@@ -16873,10 +16877,6 @@ void CompilerGLSL::emit_block_chain(SPIRBlock &block)
 		break;
 
 	case SPIRBlock::EmitMeshTasks:
-		statement("EmitMeshTasksEXT(",
```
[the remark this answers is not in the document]
HansKristian-Work: If the goal is to have a separate implementation for HLSL, you can add a `virtual void emit_mesh_tasks(const SPIRBlock &block)` to handle backend differences.

spirv_hlsl.cpp
HansKristian-Work: Could you just make a dummy struct with one int?

Try: Haven't tested yet, but probably 'one-int' wont work, because task-payload should match mesh-payload in order to link.

HansKristian-Work (approved):

Issue #1479, opened 2020-10-08
GLSL: Support `GL_KHR_shader_subgroup_arithmetic` 


devshgraphicsprogramming:
Similar to #1350, #1351 and #1352

TODO: Issue Description and implementation plan.

I can assign myself / my company.

FunMiles:
Is there any progress on this? What is the actual status? Trying to look through some of the other issues mentioned, it seems that some of it should be working but not all?
Such operations are extremely important tools for efficient Vulkan Compute shaders.

In MoltenVK I got an error with glslang not accepting arithmetic operations (`'subgroup op' : requires SPIR-V 1.3 `).
Is this relevant here or a totally different issue?

HansKristian-Work:
This is for desktop GL fallback paths. Vulkan GLSL and Metal paths should work fine.

FunMiles:
@HansKristian-Work thanks for the quick reply.
As it turns out, I just had to set a flag on glslangValidator and it worked. The code runs correctly, so the translation SPIR-V to MSL must have worked.

PR #2133, opened 2023-04-02
GLSL: Support GL_KHR_shader_subgroup_arithmetic (WIP)

georgeouzou:
Support OpenGL workarounds for subgroup arithmetic group operations.

NVIDIA workarounds are based on  GL_NV_shader_thread_shuffle.
AMD workarounds will be based on GL_AMD_shader_ballot.

Related issue  #1479

### DIFF
--- a/spirv_glsl.cpp
+++ b/spirv_glsl.cpp
@@ -4036,6 +4036,20 @@ void CompilerGLSL::emit_subgroup_arithmetic_workaround(std::string func, Op op, 
 		break;
 	}
 
+	case OpGroupNonUniformFAdd:
+	{
+		type_infos.emplace_back(TypeInfo{ "float", "0.0f" });
+		type_infos.emplace_back(TypeInfo{ "vec2", "vec2(0.0f)" });
+		type_infos.emplace_back(TypeInfo{ "vec3", "vec3(0.0f)" });
+		type_infos.emplace_back(TypeInfo{ "vec4", "vec4(0.0f)" });
+		// ARB_gpu_shader_fp64 is required in GL4.0 which in turn is required by NV_thread_shuffle
+		type_infos.emplace_back(TypeInfo{ "double", "0.0LF" });
+		type_infos.emplace_back(TypeInfo{ "dvec2", "dvec2(0.0LF)" });
+		type_infos.emplace_back(TypeInfo{ "dvec3", "dvec3(0.0LF)" });
+		type_infos.emplace_back(TypeInfo{ "dvec4", "dvec4(0.0LF)" });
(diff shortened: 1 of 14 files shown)


spirv_glsl.cpp
```diff
@@ -3992,6 +3993,168 @@ void CompilerGLSL::emit_output_variable_initializer(const SPIRVariable &var)
 	}
 }
 
+void CompilerGLSL::emit_subgroup_arithmetic_workaround(std::string func, Op op, GroupOperation group_op)
```
HansKristian-Work, line 3996: const reference for func seems more appropriate.

spirv_glsl.cpp
```diff
@@ -3992,6 +3993,168 @@ void CompilerGLSL::emit_output_variable_initializer(const SPIRVariable &var)
...
+	if (op_is_addition)
+	{
+		op_symbol = "+=";
+	}
+	else if (op_is_multiplication)
+	{
+		op_symbol = "*=";
+	}
+
+	for (const TypeInfo& t : type_infos) {
```
HansKristian-Work, line 4095: Brace on next line. `const TypeInfo &t`.

HansKristian-Work (changes requested):
Hard to review the logic without tests.

Could you add arithmetic tests as well to demonstrate the codegen?

`shaders-no-opt/comp/subgroups_basicvoteballot.vk.comp` and
`reference/shaders-no-opt/comp/subgroups_basicvoteballot.vk.comp` is a test for the existing emulation I think.

HansKristian-Work:
CI fails with a build error due to warning:

```
 /home/runner/work/SPIRV-Cross/SPIRV-Cross/spirv_glsl.cpp: In static member function ‘static bool spirv_cross::CompilerGLSL::is_supported_subgroup_op_in_opengl(spv::Op, const uint32_t*)’:
/home/runner/work/SPIRV-Cross/SPIRV-Cross/spirv_glsl.cpp:7354:3: error: this statement may fall through [-Werror=implicit-fallthrough=]
 7354 |   }
      |   ^
/home/runner/work/SPIRV-Cross/SPIRV-Cross/spirv_glsl.cpp:7356:2: note: here
 7356 |  default:
      |  ^~~~~~~
```

HansKristian-Work (approved):
Thanks for the tests, now it's more obvious what it's doing. Seems okay.

georgeouzou:
@HansKristian-Work  thanks, i will add the other operations too. (max, min, etc)

HansKristian-Work:
You can follow up in another PR with more opcodes.

PR #2137, opened 2023-04-24
Fixed generating Metal shaders that make use of fragment barycentric coords

jay3d:
The title says all.

### DIFF
--- a/spirv_msl.cpp
+++ b/spirv_msl.cpp
@@ -11959,7 +11959,7 @@ string CompilerMSL::member_attribute_qualifier(const SPIRType &type, uint32_t in
 		else
 			quals = member_location_attribute_qualifier(type, index);
 
-		if (builtin == BuiltInBaryCoordKHR || builtin == BuiltInBaryCoordNoPerspKHR)
+		if (builtin == BuiltInBaryCoordKHR)
 		{
 			if (has_member_decoration(type.self, index, DecorationFlat) ||
 			    has_member_decoration(type.self, index, DecorationCentroid) ||
@@ -11977,35 +11977,38 @@ string CompilerMSL::member_attribute_qualifier(const SPIRType &type, uint32_t in
 		// FragCoord builtin; it's always noperspective on Metal.
 		if (!type_is_integral(mbr_type) && (!is_builtin || builtin != BuiltInFragCoord))
 		{
-			if (has_member_decoration(type.self, index, DecorationFlat))
+			if (builtin != BuiltInBaryCoordKHR && builtin != BuiltInBaryCoordNoPerspKHR)
 			{
-				if (!quals.empty())
-					quals += ", ";
-				quals += "flat";
-			}
-			else if (has_member_decoration(type.self, index, DecorationCentroid))
-			{
-				if (!quals.empty())
-					quals += ", ";
-				if (has_member_decoration(type.self, index, DecorationNoPerspective))
-					quals += "centroid_no_perspective";
-				else
-					quals += "centroid_perspective";
-			}
-			else if (has_member_decoration(type.self, index, DecorationSample))
-			{
-				if (!quals.empty())
-					quals += ", ";
-				if (has_member_decoration(type.self, index, DecorationNoPerspective))
-					quals += "sample_no_perspective";
-				else
-					quals += "sample_perspective";
-			}
-			else if (has_member_decoration(type.self, index, DecorationNoPerspective))
-			{
-				if (!quals.empty())
-					quals += ", ";
-				quals += "center_no_perspective";
+				if (has_member_decoration(type.self, index, DecorationFlat))
+				{
+					if (!quals.empty())
+						quals += ", ";
+					quals += "flat";
+				}
+				else if (has_member_decoration(type.self, index, DecorationCentroid))
+				{
+					if (!quals.empty())
+						quals += ", ";
+					if (has_member_decoration(type.self, index, DecorationNoPerspective))
+						quals += "centroid_no_perspective";
+					else
+						quals += "centroid_perspective";
+				}
+				else if (has_member_decoration(type.self, index, DecorationSample))
+				{
+					if (!quals.empty())
+						quals += ", ";
+					if (has_member_decoration(type.self, index, DecorationNoPerspective))
+						quals += "sample_no_perspective";
+					else
+						quals += "sample_perspective";
+				}
+				else if (has_member_decoration(type.self, index, DecorationNoPerspective))
+				{
+					if (!quals.empty())
+						quals += ", ";
+					quals += "center_no_perspective";
+				}
 			}
 		}
 


spirv_msl.cpp
```diff
@@ -11959,7 +11959,7 @@ string CompilerMSL::member_attribute_qualifier(const SPIRType &type, uint32_t in
 		else
 			quals = member_location_attribute_qualifier(type, index);
 
-		if (builtin == BuiltInBaryCoordKHR || builtin == BuiltInBaryCoordNoPerspKHR)
+		if (builtin == BuiltInBaryCoordKHR)
 		{
 			if (has_member_decoration(type.self, index, DecorationFlat) ||
 			    has_member_decoration(type.self, index, DecorationCentroid) ||
```
cdavis5e, line 11965: This looks wrong.

cdavis5e (changes requested):
Changes' descriptions should, you know, describe what the change actually does and, more important, why. 😃 It isn't obvious that this change is necessary or even desirable.

Are you seeing fragment shaders with barycentric inputs that have one of the interpolation decorations? What app/game?

`Flat` and `NoPerspective` decorations on these should never be seen. `Flat` cannot be used at all with barycentrics, and `NoPerspective` is of course handled by `BaryCoordNoPerspKHR`. Your change will silently drop `Centroid` and `Sample` decorations, which will make shaders that use them "work" but cause incorrect rendering with MSAA when these decorations are used.

spirv_msl.cpp
jay3d: Thanks for the comment.

Without my changes, when compiling GLSL shaders to SPIR-V which has `gl_BaryCoordNoPersp`, then using SPIRV-Cross to generate Metal shaders, this error will be displayed:
```
Flat, Centroid, Sample, NoPerspective decorations are not supported for BaryCoord inputs.
```
Which is not entirely true because Metal supports `Centroid` and `NoPerspective` decorations for `BaryCoord` input according to this spec:
https://developer.apple.com/metal/Metal-Shading-Language-Specification.pdf
at page 103:

> A fragment function input declared with the `[[barycentric_coord]]` attribute can only be
> declared with either the `center_perspective` (default) or `center_no_perspective`
> sampling and interpolation attributes. The barycentric coordinates and per-pixel primitive ID
> can be passed as fragment function input in structures organized as shown in these examples:
> ```
> struct FragmentInput0 {
>     uint primitive_id [[primitive_id]];
>     // [[center_perspective]] is the default, so it can be omitted.
>     float3 barycentric_coord [[barycentric_coord, center_perspective]];
> };
> ```
> ```
> struct FragmentInput1 {
>     uint primitive_id [[primitive_id]];
>     float2 linear_barycentric_coord [[barycentric_coord,
>     center_no_perspective]];
> };
> ```

So there's clearly a bug there; My changes may not be 100% perfect but they work in my use case.
You may take a look further and advice accordingly.
PS: This PR will complement this: https://github.com/KhronosGroup/glslang/pull/3198

HansKristian-Work:
Is there any forward progress on this PR?

rcaridade145:
If DecorationNoPerspective is allowed this line https://github.com/KhronosGroup/SPIRV-Cross/blob/main/spirv_msl.cpp#L11981 should not be here.
Not exactly sure where center_perspective would come from.

jay3d:
> Are you seeing fragment shaders with barycentric inputs that have one of the interpolation decorations? What app/game?
My app, and that's enough for me: https://www.youtube.com/watch?v=_omNIB8q4vM

I'm all ears... How do you suggest this could be fixed?

jay3d:
@cdavis5e @rcaridade145 @HansKristian-Work 
It's been more that one year since this issue was brought up, and no fix, and no work or discussion for my PR. This is frustrating, this makes this utility _**unusable**_ albeit for generating GLSL and very outdated and limited MSL or HLSL, despite specifying latest generator profiles set for both (MSL and HLSL). All what I got was that my code was called "wrong", but it indeed works and fixes a major issue, at least with my case, and I have example renderings to show. In the end I discovered that my code was not exactly "wrong", but it was a hack to an actually extremely broken MSL generator.

Allow me to elaborate on that in more detail:

Let take this very simple HLSL code:
```hlsl
struct PSInput
{
    linear float3 baryCoord : SV_BARYCENTRICS0;
    noperspective float3 baryCoord1 : SV_BARYCENTRICS1;
};

float4 PSMain(PSInput input) : SV_TARGET
{
	float3 baryCoord = input.baryCoord;
	float3 baryCoord1 = input.baryCoord1;

    [[branch]] if (baryCoord1.x > 0.1)
    {
        return float4(baryCoord1.x, baryCoord1.y, 1.0f - baryCoord1.x - baryCoord1.y, 1.0f);
    }

    return float4(baryCoord.x, baryCoord.y, 1.0f - baryCoord.x - baryCoord.y, 1.0f);
}
```
While it's useless shader and doesn't make any sense, but it serves as test case for MSL and HLSL generator in SPV-Cross.
Now perspective and non-perspective correct (`linear` and `noperspective` respectively in the code above) barycentrics can _co-exist_ in the same fragment shader in GLSL, HLSL and MSL. Now when this code is compiled with DXC with this command:

`dxc .\test.hlsl -T ps_6_1 -E PSMain -O3 -spirv -fspv-target-env="universal1.5"`

Then this spirv disassembly is generated for reference:
```
; SPIR-V
; Version: 1.5
; Generator: Google spiregg; 0
; Bound: 36
; Schema: 0
               OpCapability Shader
               OpCapability FragmentBarycentricKHR
               OpExtension "SPV_KHR_fragment_shader_barycentric"
               OpMemoryModel Logical GLSL450
               OpEntryPoint Fragment %PSMain "PSMain" %2 %3 %out_var_SV_TARGET
               OpExecutionMode %PSMain OriginUpperLeft
               OpSource HLSL 610
               OpName %out_var_SV_TARGET "out.var.SV_TARGET"
               OpName %PSMain "PSMain"
               OpDecorate %2 BuiltIn BaryCoordKHR
               OpDecorate %3 BuiltIn BaryCoordNoPerspKHR
               OpDecorate %out_var_SV_TARGET Location 0
      %float = OpTypeFloat 32
    %float_1 = OpConstant %float 1
%float_0_100000001 = OpConstant %float 0.100000001
    %v3float = OpTypeVector %float 3
%_ptr_Input_v3float = OpTypePointer Input %v3float
    %v4float = OpTypeVector %float 4
%_ptr_Output_v4float = OpTypePointer Output %v4float
       %void = OpTypeVoid
         %13 = OpTypeFunction %void
       %bool = OpTypeBool
          %2 = OpVariable %_ptr_Input_v3float Input
          %3 = OpVariable %_ptr_Input_v3float Input
%out_var_SV_TARGET = OpVariable %_ptr_Output_v4float Output
       %uint = OpTypeInt 32 0
     %uint_0 = OpConstant %uint 0
     %PSMain = OpFunction %void None %13
         %17 = OpLabel
         %18 = OpLoad %v3float %2
         %19 = OpCompositeExtract %float %18 0
         %20 = OpCompositeExtract %float %18 1
         %21 = OpLoad %v3float %3
         %22 = OpCompositeExtract %float %21 0
         %23 = OpCompositeExtract %float %21 1
               OpSelectionMerge %24 None
               OpSwitch %uint_0 %25
         %25 = OpLabel
         %26 = OpFOrdGreaterThan %bool %22 %float_0_100000001
               OpSelectionMerge %27 DontFlatten
               OpBranchConditional %26 %28 %27
         %28 = OpLabel
         %29 = OpFSub %float %float_1 %22
         %30 = OpFSub %float %29 %23
         %31 = OpCompositeConstruct %v4float %22 %23 %30 %float_1
               OpBranch %24
         %27 = OpLabel
         %32 = OpFSub %float %float_1 %19
         %33 = OpFSub %float %32 %20
         %34 = OpCompositeConstruct %v4float %19 %20 %33 %float_1
               OpBranch %24
         %24 = OpLabel
         %35 = OpPhi %v4float %31 %28 %34 %27
               OpStore %out_var_SV_TARGET %35
               OpReturn
               OpFunctionEnd
```

Now SPIRV-Cross will generate this GLSL correctly:

```glsl
#version 450
#extension GL_EXT_fragment_shader_barycentric : require
#if defined(GL_EXT_control_flow_attributes)
#extension GL_EXT_control_flow_attributes : require
#define SPIRV_CROSS_FLATTEN [[flatten]]
#define SPIRV_CROSS_BRANCH [[dont_flatten]]
#define SPIRV_CROSS_UNROLL [[unroll]]
#define SPIRV_CROSS_LOOP [[dont_unroll]]
#else
#define SPIRV_CROSS_FLATTEN
#define SPIRV_CROSS_BRANCH
#define SPIRV_CROSS_UNROLL
#define SPIRV_CROSS_LOOP
#endif

layout(location = 0) out vec4 out_var_SV_TARGET;

void main()
{
    vec4 _35;
    do
    {
        SPIRV_CROSS_BRANCH
        if (gl_BaryCoordNoPerspEXT.x > 0.100000001490116119384765625)
        {
            _35 = vec4(gl_BaryCoordNoPerspEXT.xy, (1.0 - gl_BaryCoordNoPerspEXT.x) - gl_BaryCoordNoPerspEXT.y, 1.0);
            break;
        }
        _35 = vec4(gl_BaryCoordEXT.xy, (1.0 - gl_BaryCoordEXT.x) - gl_BaryCoordEXT.y, 1.0);
        break;
    } while(false);
    out_var_SV_TARGET = _35;
}
```

But then it won't for HLSL using this command:

`.\spirv-cross.exe .\test.spirv --hlsl --shader-model 61`

And it's giving this confusing error:

`SPIRV-Cross threw an exception: Unsupported builtin in HLSL: 5286`

???? 🤔 ????

Also when I try to generate MSL using this command:

`.\spirv-cross.exe .\test.spirv --msl --msl-version 30100`

It gives this error:

`SPIRV-Cross threw an exception: Cannot declare both BaryCoordNV and BaryCoordNoPerspNV in same shader in MSL.`

Well how do you know that? There's nothing in the Metal specs (3.1 here as specified by the profile) that prevents you from doing that. A prove of that is that Apple's own `metal-shaderconverter` can convert the DXIL generated from DXC and render it correctly:

`dxc .\test.hlsl -T ps_6_1 -E PSMain -O3 -Fo test.dxil`
And then:
`metal-shaderconverter.exe .\test.dxil --validateAll -o test.mil`

And it renders correctly.

Until these are fixed for both HLSL and MSL, SPIRV-Cross should be called SPIRV-GLSL.

HansKristian-Work:
This fell off the radar. I'll have a look next week.

HansKristian-Work:
I implemented HLSL barycentrics here https://github.com/KhronosGroup/SPIRV-Cross/pull/2391

But MSL is a broken language that cannot support Vulkan barycentrics.

Two problems.

- PerVertexKHR cannot be implemented in any reasonable way.
> Well how do you know that?

Because Metal shader compiler rejects:

```
#include <metal_stdlib>
#include <simd/simd.h>

using namespace metal;

struct main0_out
{
    float3 value [[color(0)]];
};

struct main0_in
{
    float3 gl_BaryCoordEXT [[barycentric_coord, center_perspective]];
    float3 gl_BaryCoordNoPerspEXT [[barycentric_coord, center_no_perspective]];
};

{
    main0_out out = {};
    out.value = in.gl_BaryCoordEXT + in.gl_BaryCoordNoPerspEXT;
    return out;
}
```

```
Z:/home/maister/git/SPIRV-Cross/test.msl:14:12: error: declaration with attribute 'barycentric_coord' already specified
    float3 gl_BaryCoordNoPerspEXT [[barycentric_coord, center_no_perspective]];
    ~~~~~~~^~~~~~~~~~~~~~~~~~~~~~
Z:/home/maister/git/SPIRV-Cross/test.msl:13:12: note: previous declaration with attribute 'barycentric_coord' here
    float3 gl_BaryCoordEXT [[barycentric_coord, center_perspective]];
```

The reason Apple's DXIL conversion can work is because they bypass the language they expose to developers and emit IR directly. That's why they can make it work, but we cannot.

jay3d:
I think in https://developer.apple.com/metal/Metal-Shading-Language-Specification.pdf
They describe what seems to be a solution to that, if I'm not mistaken, starting page 113.

cdavis5e:
> I think in https://developer.apple.com/metal/Metal-Shading-Language-Specification.pdf They describe what seems to be a solution to that, if I'm not mistaken, starting page 113.

I see what you're saying, but the problem with that is that there cannot be more than one `[[stage_in]]` parameter in a shader function. Presumably, the two structures, `FragmentInput0` and `FragmentInput1`, are for two different fragment shaders. Good try, though.

> The reason Apple's DXIL conversion can work is because they bypass the language they expose to developers and emit IR directly. That's why they can make it work, but we cannot.

I'm starting to wonder if MoltenVK really _should_ decompile to HLSL instead of MSL and use Apple's D3D shader converter.  Supporting tessellation would be a problem with that route, however, since AFAIK SPIRV-Cross does not yet support HLSL output from tessellation shaders.

I've also seriously considered building a shader converter of my own that takes SPIR-V and emits AIR directly. Then we could sidestep all the silly restrictions that Apple places on MSL. But that's a lot more work--I'd be starting from scratch.

rcaridade145:
```
struct main0_in
{
    float3 gl_BaryCoordEXT [[barycentric_coord, center_perspective]];
    float3 gl_Position [[position]]; // Fragment's position
};
```
Given this  can't we use the z component to manually calculate

HansKristian-Work:
I fixed Centroid/Sample decorations on bary coords now, so I don't think there is anything further that can be done before MSL improves.

PR #2143, opened 2023-05-02
HLSL: Add support to preserve (RW)StructuredBuffer resources.

LukasBanana:
This PR adds support to the HLSL backend to preserve `RW`-/`StructuredBuffer` resources when cross-compiling HLSL-to-HLSL. UE5 does this to support modern HLSL2021 features such as templates across all platforms including the D3D11 RHI where DXBC is the only supported shader bytecode but FXC does not support those features. Translating `StructuredBuffer` to `ByteAddressBuffer` breaks the resource binding. These changes make use of the `UserTypeGOOGLE` extension to preserve the original resource type. This feature is disable by default and can be enabled via the `preserve_structured_buffers` option.

### DIFF
--- a/spirv_glsl.cpp
+++ b/spirv_glsl.cpp
@@ -9606,6 +9606,9 @@ string CompilerGLSL::access_chain_internal(uint32_t base, const uint32_t *indice
 	bool pending_array_enclose = false;
 	bool dimension_flatten = false;
 
+	// If we are translating access to a structured buffer, the first subscript '._m0' must be hidden
+	bool hide_first_subscript = count > 1 && is_user_type_structured(base);
+
 	const auto append_index = [&](uint32_t index, bool is_literal, bool is_ptr_chain = false) {
 		AccessChainFlags mod_flags = flags;
 		if (!is_literal)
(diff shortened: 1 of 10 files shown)


spirv_hlsl.cpp
```diff
@@ -4969,6 +4988,12 @@ void CompilerHLSL::emit_access_chain(const Instruction &instruction)
 
 		auto *backing_variable = maybe_get_backing_variable(ops[2]);
 
+		if (is_user_type_structured(backing_variable->self))
```
HansKristian-Work, line 4991: backing_variable may be nullptr here.

spirv_hlsl.cpp
```diff
@@ -2603,11 +2603,30 @@ void CompilerHLSL::emit_buffer_block(const SPIRVariable &var)
...
+			else
+			{
+				// Otherwise, this StructuredBuffer only has a basic subtype, e.g. StructuredBuffer<int>
+				return this->type_to_glsl(parent_type);
+			}
+		};
+
+		std::string type_name;
+		if (is_user_type_structured(var.self))
+			type_name = join(is_readonly ? "" : "RW", "StructuredBuffer<", to_structuredbuffer_subtype_name(type), ">");
```
HansKristian-Work, line 2624: What about ROV structured buffers?

spirv_glsl.cpp
```diff
@@ -9606,6 +9606,9 @@ string CompilerGLSL::access_chain_internal(uint32_t base, const uint32_t *indice
 	bool pending_array_enclose = false;
 	bool dimension_flatten = false;
 
+	// If we are translating access to a structured buffer, the first subscript '._m0' must be hidden
+	const bool hide_first_subscript = count > 1 && is_user_type_structured(base) ? 1 : 0;
```
HansKristian-Work, line 9610: Seems more elegant to just `indices++; count--;` here and jump ahead in the type hierarchy instead of adding deeper branching later in the code.

spirv_hlsl.cpp
```diff
@@ -6693,3 +6718,15 @@ bool CompilerHLSL::builtin_translates_to_nonarray(spv::BuiltIn builtin) const
...
 }
+
+bool CompilerHLSL::is_user_type_structured(uint32_t id) const
+{
+	if (hlsl_options.preserve_structured_buffers)
+	{
+		// Compare left hand side of string only as these user types can contain more meta data such as their subtypes,
+		// e.g. "structuredbuffer:int"
+		const std::string &user_type = get_decoration_string(id, DecorationUserTypeGOOGLE);
+		return user_type.compare(0, 16, "structuredbuffer") == 0 || user_type.compare(0, 18, "rwstructuredbuffer") == 0;
```
HansKristian-Work, line 6729: Is there any risk of DXC optimizing away members of the struct? spirv-opt has been known to do this lately where the struct may be laid out as:

```
struct Foo
{
    [offset = 0] int a;
    [offset = 12] int d;
};
```

where the original was something like

```
struct Foo
{
   int a, b, c, d;
};
```
but b and c were unused. We'd need to make sure somehow that the layout works out.

spirv_glsl.cpp
HansKristian-Work, line 9610: Does this handle array of SSBO? I think it's the second index that would have to be skipped in that case.

HansKristian-Work, line 9610: Need to verify that back to back AccessChain works. E.g. AccessChain into an array of SSBOs, then access chain into a particular member, or access chain down to a vector, then access chain down to a scalar for example.

HansKristian-Work (changes requested):
Comments.

spirv_glsl.cpp
LukasBanana: Can you please provide a suggestion how jumping to the type hierarchy should look like since there are several variables that track certain states such as `row_major_matrix_needs_conversion`?
While the first subscript must be hidden in the output, the type of `._m0` must still be picked up during the loop:
```cpp
type = &get<SPIRType>(type->member_types[index]);
```

HansKristian-Work: Ok, I didn't consider that. I guess this can be kept then.

spirv_hlsl.cpp
LukasBanana: There is indeed a SPIRV-Tools pass that removes unused struct members, but it's only part of `-Os` optimization when you optimize for size. In our fork of DXC, we added this particular pass to the performance optimization, i.e. `-O2` etc. but we only apply it on the struct for the special case of `$Globals` cbuffer. If someone applies that optimization pass on structs used for structured buffers they would not only have to retain the offsets but also have potentially a last dummy member or otherwise the struct size would be incorrect if the last element is removed. I assume that optimization would also cause problems in GLSL.

HansKristian-Work: Inner structs don't support layout(offset) in GL unfortunately, so it is a problem there as well, but I haven't seen that reported as an issue yet. I'd rather not have to support all of that (MSL buffer packing is a nightmare) just to handle this scenario. If your use case can avoid making SPIRV-Cross' life difficult, that would be preferable.

We'd probably need to just run the "does this struct conform to HLSL structured buffer rules" to sanity check that nothing weird is going to happen. Likely a new buffer layout type ala HLSLCbuffer would have to be added for that.

spirv_glsl.cpp
LukasBanana: I updated the PR and this works for the following example now as well:
```hlsl
struct Data {
  float4 Color;
};
StructuredBuffer<Data> Colors[2];
float4 PSMain() : SV_Target {
  return Colors[1][3].Color;
}
```

HansKristian-Work:
I think the main concern now is some test cases. Given it's an opt-in feature, I'm not too concerned if we don't have them.

LukasBanana:
I added two new tests in SPIR-V text form in a separate commit: One for structured buffers and one for array or structured buffers.

PR #2144, opened 2023-05-06
upload binaries as artifacts

Trass3r:


HansKristian-Work:
Missing the CLA. I think uploading artifacts is reasonable, but I'm not sure how useful they are, at least the statically linked archives are basically unusable on their own.

Trass3r:
Done. Please merge by squashing.

PR #2148, opened 2023-05-12
MSL: Fix dref lod workaround on combined texture/samplers

etang-cw:
Example broken shader (compile with `--msl --msl-version 20300 --msl-sample-dref-lod-array-as-grad`): [shader-fs-090d3829caba1bbc.spv.zip](https://github.com/KhronosGroup/SPIRV-Cross/files/11457235/shader-fs-090d3829caba1bbc.spv.zip)


HansKristian-Work (approved):
LGTM if CI passes.

PR #2153, opened 2023-05-19
[WIP] MSL: Prevent RAW hazards on read_write textures

etang-cw:
MSL makes no guarantees on coherency of reads from textures, even if written earlier by the same thread.  This is isn't explicitly stated, but implied by section 6.12.17 of the [MSL spec](https://developer.apple.com/metal/Metal-Shading-Language-Specification.pdf), since the `fence()` method would be completely useless otherwise:
> The texture `fence()` member function ensures that writes to the texture by a thread become visible to subsequent reads from that texture by the same thread (the thread that is performing the write). Texture types (including texture buffers) that you can declare with the `access::read_write` attribute support the `Fence` function.

On the other hand, [GLSL makes stronger guarantees](https://www.khronos.org/opengl/wiki/Memory_Model#Incoherent_memory_access):
> If you perform an incoherent memory write, the value written will always be visible for reading. But only through that particular variable and only within the shader invocation that issued the write. You need not do anything special to make this happen. However, it is possible that, between writing and reading, another invocation may have stomped on that value. So long as that is not the case, reading it will produce the value you have written.

This is currently WIP pending the following things, which I'd like some input on:
- [ ] Adjust the pass to only add `fence()`s when needed (e.g. four reads in a row should only need one fence, and a shader that reads in the beginning and writes at the end should need no fences).
  - I'm not super familiar with writing optimization passes like this.  I was thinking something like this, but would like input on whether there's a better way of going about this
    1. Track three bits for each pair of (readwrite texture, block): input_clean, texture_used, output_clean
    2. Run over each block and set input_clean => true, texture_used => if any reads/writes are done to the texture and output_clean => false if the last action is a write
    3. For every block with output_clean == false, go to all its successors, set their input_clean to false, and if !texture_used, set their output_clean to false and add them to the processing list if it was previously true.
    4. Go through each block and use input_clean to mark the individual read operations as needing or not needing fences
- [x] Verify that SPIRV makes the same guarantees as GLSL.  I looked at the SPIRV spec, but couldn't find anything specific.  I'm also not very good at reading these kinds of spec documents.
- [ ] Maybe add tests to the Vulkan CTS?  I tried running `vk-default/glsl.txt` and `vk-default/image.txt`, and couldn't find any tests this fixes, are there any others I should try?  If there aren't any, is this something that should be added to the CTS?

### DIFF
--- a/reference/opt/shaders-msl/asm/comp/image-load-store-short-vector.invalid.asm.comp
+++ b/reference/opt/shaders-msl/asm/comp/image-load-store-short-vector.invalid.asm.comp
@@ -8,6 +8,7 @@ using namespace metal;
 static inline __attribute__((always_inline))
 void _main(thread const uint3& id, texture2d<float, access::read_write> TargetTexture)
 {
+    TargetTexture.fence();
     float2 loaded = TargetTexture.read(uint2(id.xy)).xy;
     float2 storeTemp = loaded + float2(1.0);
     TargetTexture.write(storeTemp.xyyy, uint2((id.xy + uint2(1u))));
--- a/reference/opt/shaders-msl/desktop-only/frag/image-ms.desktop.frag
+++ b/reference/opt/shaders-msl/desktop-only/frag/image-ms.desktop.frag
@@ -5,6 +5,7 @@ using namespace metal;
 
 fragment void main0(texture2d_ms<float> uImageMS [[texture(0)]], texture2d_array<float, access::read_write> uImageArray [[texture(1)]], texture2d<float, access::write> uImage [[texture(2)]])
 {
+    uImageArray.fence();
     uImage.write(uImageMS.read(uint2(int2(1, 2)), 2), uint2(int2(2, 3)));
     uImageArray.write(uImageArray.read(uint2(int3(1, 2, 4).xy), uint(int3(1, 2, 4).z)), uint2(int3(2, 3, 7).xy), uint(int3(2, 3, 7).z));
 }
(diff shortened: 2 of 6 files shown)


etang-cw:
[I ran a small benchmark with an ROV shader](https://gist.github.com/etang-cw/f82c3dadb962931bb0c4d2acce44c2ac) (it was the first thing I could think of that could be negatively affected by this), and there's a 30% hit for inserting an unnecessary fence before the read, and a 45% hit for inserting after the write.  Hopefully if your shader has more ALU work it'll be smaller.  Based on this, I'm sticking with the pre-read barriers, though I don't really know what other common RW texture uses there are.  The game I found this in was using a (through vkd3d) RW texture buffer as a counter array (please use a local variable and write it once at the end...), which also has an equal number of reads vs writes, but maybe there are common uses that involve more reads than writes?  Apple's new DXIL to AIR shader converter also inserts before each read.

It doesn't seem like I'll get around to writing that optimization pass any time soon, so I think I'll just leave it as is, with an option to disable it in case it hurts anyone's stuff too badly.

alyssarosenzweig:
> If this has to be done, the only reasonable way to make this work is to emit fence after each write() if the image is declared as read_write. There's no way it's anything other than a reordering barrier, so it should be fine.

It's not just a reordering barrier, it's a big hammer of a cache flush because image writes on AGX are not coherent with image reads (which use the texture cache). I have no idea why Apple thought it'd be a good idea to expose that detail in Metal instead of handling this in the compiler like Mesa does.

HansKristian-Work:
> > If this has to be done, the only reasonable way to make this work is to emit fence after each write() if the image is declared as read_write. There's no way it's anything other than a reordering barrier, so it should be fine.
> 
> It's not just a reordering barrier, it's a big hammer of a cache flush because image writes on AGX are not coherent with image reads (which use the texture cache). I have no idea why Apple thought it'd be a good idea to expose that detail in Metal instead of handling this in the compiler like Mesa does.

... what? I'm baffled by this design.

HansKristian-Work (approved):
Given you've tested perf here, I guess this is fine.

PR #2157, opened 2023-05-28
MSL: fix function constant deduplication misfire

rcombs:
Previously, if a constant without `DecorationSpecId` occurred before the constant with `DecorationSpecId` 0, it would be inserted into the `unique_func_constants` map with key 0 (the default return value from `get_decoration`). This prevented us from ever emitting the declaration with `[[function_constant(0)]]`, which produced some bizarre MSL compilation errors.

Instead, we now only insert into `unique_func_constants` when we know we're going to emit a `[[function_constant(…)]]` line.

This regressed in 41007cdc7d9c3a1f9b61e307d149f94a68fa3c9c.


HansKristian-Work (approved):
LGTM if CI passes.

Issue #2117, opened 2023-03-03
Support static linking in pkg-config


kasper93:
Currently only `spirv-cross-c-shared.pc` is provided for shared library. Could add a `spirv-cross-c.pc` with proper `pkg-config --static` support.

Thanks,
Kacper

HansKristian-Work:
A PR would be helpful.

PR #2159, opened 2023-05-30
Static pkg-config support

qyot27:
Addresses issue #2117 to the extent that a .pc file is created and installed for static builds.

I'm not sure if anything actually needs to be added to `Libs.private` or `Requires.private` to align with @kasper93's comment about `pkg-config --static`, as there's nothing in my build notes indicating I'd ever had to mess with those fields after-the-fact.  I was overly cautious about removing the `sharedlibdir` references in the static .pc.in template, so if those aren't needed, it's easy enough to remove them.


HansKristian-Work (approved):
LGTM if CI passes. Have some trivial nits I'll apply after merge.

PR #2168, opened 2023-06-18
HLSL:  flip-vert-y support for mesh-shader

Try:
This PR adds missing functionality to #1974

This PR tracks `OpAccessChain` that point to eider:
* `gl_MeshVerticesEXT[].gl_Position`
* `gl_MeshVerticesEXT[].gl_Position.y`

and wraps relevant stores with newly introduced function `spvFlipVertY`.
This is very different form usual flip-y, because `gl_Position` is write-only in spec

### DIFF
--- a/spirv_hlsl.cpp
+++ b/spirv_hlsl.cpp
@@ -2254,6 +2254,20 @@ void CompilerHLSL::emit_resources()
 		end_scope();
 		statement("");
 	}
+
+	if (is_mesh_shader && options.vertex.flip_vert_y)
+	{
+		statement("float4 spvFlipVertY(float4 v)");
+		begin_scope();
+		statement("return float4(v.x, -v.y, v.z, v.w);");
+		end_scope();
+		statement("");
+		statement("float spvFlipVertY(float v)");
+		begin_scope();
+		statement("return -v;");
+		end_scope();
+		statement("");
+	}
 }
 
 void CompilerHLSL::emit_texture_size_variants(uint64_t variant_mask, const char *vecsize_qualifier, bool uav,
@@ -2482,6 +2496,30 @@ void CompilerHLSL::analyze_meshlet_writes(uint32_t func_id, uint32_t id_per_vert
 				{
 					bool already_declared = false;
 					auto builtin_type = BuiltIn(get_decoration(var->self, DecorationBuiltIn));
+
+					if (op == OpAccessChain && i.length > 4)
+					{
+						auto &type = get_variable_element_type(*var);
+						auto *c = maybe_get<SPIRConstant>(ops[4]);
+						if (c != nullptr && c->scalar() == 0 &&
+						    has_member_decoration(type.self, 0, DecorationBuiltIn) &&
+						    get_member_decoration(type.self, 0, DecorationBuiltIn) == spv::BuiltInPosition)
+						{
+							const SPIRType &dst_type = get_pointee_type(ops[0]);
+							if (dst_type.vecsize == 4)
+							{
+								// full vec4 write
+								access_to_meshlet_position.insert(ops[1]);
+							}
+							else if (i.length > 5)
+							{
+								// gl_Position.y
+								auto *cY = maybe_get<SPIRConstant>(ops[5]);
+								if (cY != nullptr && cY->scalar() == 1)
+									access_to_meshlet_position.insert(ops[1]);
+							}
+						}
+					}
 
 					uint32_t var_id = var->self;
 					if (var->storage != StorageClassTaskPayloadWorkgroupEXT &&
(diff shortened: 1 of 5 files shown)


spirv_hlsl.cpp
```diff
@@ -4950,6 +4988,15 @@ void CompilerHLSL::write_access_chain(const SPIRAccessChain &chain, uint32_t val
 void CompilerHLSL::emit_store(const Instruction &instruction)
 {
 	auto ops = stream(instruction);
+	if (options.vertex.flip_vert_y && access_to_meshlet_position.count(ops[0]) != 0)
+	{
+		auto lhs = to_dereferenced_expression(ops[0]);
+		auto rhs = to_pointer_expression(ops[1]);
```
HansKristian-Work, line 4994:
```suggestion
		auto rhs = to_unpacked_expression(ops[1]);
```

spirv_hlsl.cpp
```diff
@@ -2483,6 +2497,30 @@ void CompilerHLSL::analyze_meshlet_writes(uint32_t func_id, uint32_t id_per_vert
 					bool already_declared = false;
 					auto builtin_type = BuiltIn(get_decoration(var->self, DecorationBuiltIn));
 
+					if (op == OpAccessChain && i.length > 4)
+					{
+						auto &type = get_variable_element_type(*var);
+						auto *c = maybe_get<SPIRConstant>(ops[4]);
+						if (c != nullptr && c->scalar() == 0 &&
+						    has_member_decoration(type.self, 0, DecorationBuiltIn) &&
```
HansKristian-Work, line 2505: This does not consider that you can have plain Position variables.

This check is also too naive. A more correct solution involves:

- An access chain that is created from another access chain will miss this check. It should inherit the access_to_meshlet_position in this case. Test case: Access chain into Position, then access chain into scalar.
- A more correct solution should traverse the access chain and insert the unordered_set when it comes across anything related to positions.
- OpCopyObject needs to inherit the flag.
- Missing OpInBoundsAccessChain

HansKristian-Work (changes requested):
I'm not sure this is desirable to add.

Mesh shaders are only supported in the very latest D3D12, and that already supports doing Y-flip natively, similar to Vulkan where the viewport can implement any Y-flip just fine.

See https://learn.microsoft.com/en-us/windows/win32/api/d3d12/ns-d3d12-d3d12_feature_data_d3d12_options13.

Try:
> See https://learn.microsoft.com/en-us/windows/win32/api/d3d12/ns-d3d12-d3d12_feature_data_d3d12_options13.

Tested locally with IntellUHD, RTX3070 - no support for negative viewport (and options13 overall).
(latest game driver for nvidia, not broken driver for intel)

> An access chain that is created from another access chain will miss this check. It should inherit the access_to_meshlet_position in this case. Test case: Access chain into Position, then access chain into scalar.

I've understood spirv-spec, that `OpAccessChain` should start with variable, but OK.

Would it make sense than to refactor spirv-cross code to generalize usage of `SPIRAccessChain`? 
Currently `SPIRAccessChain` exists only for storage-buffers, and `nullptr` for any else types of chains. This will help me to fully avoid combined-access-chains/opcopy/etc, and do all things in one go at `OpStore`

HansKristian-Work:
> I've understood spirv-spec, that OpAccessChain should start with variable, but OK.

AccessChain can start with any pointer. You can access chain, then further access chain.

> Tested locally with IntellUHD, RTX3070 - no support for negative viewport (and options13 overall).
(latest game driver for nvidia, not broken driver for intel)

You probably need AgilitySDK for this. Any D3D12 app using Mesh shaders will use that.

Try:
> You probably need AgilitySDK for this. Any D3D12 app using Mesh shaders will use that

Going thru their documentation now - seem to be dll hell... Dunno, if it's reasonable to ship game with.
So far cleaning up `SPIRAccessChain` code looks as better option - code cleaning + flip-y becomes trivial

HansKristian-Work:
> Going thru their documentation now - seem to be dll hell... Dunno, if it's reasonable to ship game with.

Basically every new game these days ships with it and you have no choice but to do so if you want anything resembling new features in D3D12.

HansKristian-Work:
>  So far cleaning up SPIRAccessChain code looks as better option - code cleaning + flip-y becomes trivial

SPIRAccessChain is not used for stage IO. It's only used for SSBOs really.

Try:
> Basically every new game these days ships with it and you have no choice but to do so if you want anything resembling new features in D3D12

Raytracing, bindless and mesh-shader do not require anything, other than a reasonably updated driver on user side.

> SPIRAccessChain is not used for stage IO. It's only used for SSBOs really.

Yes, that what I mean - in spirv-cross code access-chain applicable only for SSBO, yet in SPIRV it's used for every type of access

HansKristian-Work:
Generally access chains just collapse into SPIRExpression, which is what it should be. SPIRAccessChain is an edge case which explicitly bans to_expression since the caller needs to inspect the access chain to make sense of it. I would be against trying to slam everything into SPIRAccessChain if that's what you're thinking of. That is really not needed to fix whatever you need in this PR.

Try:
> Generally access chains just collapse into SPIRExpression, which is what it should be

Looking into `SPIRExpression` - this doesn't seem to be right tool eider. Basically `SPIRExpression` is a string wrapper, that doesn't carry much of useful information.

HansKristian-Work:
Right, but you have all information you need when parsing OpAccessChain, there is no need to inspect the access chain after the fact. It's just a matter of going through the indices and resolving types until you find the case you're looking for, then just inherit that "is_builtin_position" state.

PR #2207, opened 2023-10-03
Improve CI build times

etang-cw:
- Caches externals so they don't get rebuilt every time
- Parallelizes build

macOS build time goes from 30m to 15m (no cache) / 7m (with cache)
In addition, when the cache is available, you should get feedback on compilation errors within a minute or so


HansKristian-Work (approved):
I cannot verify if this is correct, but if it works it works I guess. Can always revert if things start breaking.

PR #2208, opened 2023-10-04
Fix declaration of `gl_SampleMaskIn` in HLSL backend.

LukasBanana:
This might not be the right place, but I wanted to start with a proposal to fix an issue with the cross-compilation of `gl_SampleMask` (GLSL) to `SV_Coverage` (HLSL).

I saw that SPIRV-Cross is already handling the case that `gl_SampleMask` is an array of bits while the corresponding `SV_Coverage` in HLSL is a scalar bitmask. However, the necessary information for that built-in type gets lost in `OpLoad` instructions because that decoration is not propagated to the resulting variable. I am not sure if this solution would break other cases that need the built-in types. A better solution would probably be to mark such variables in how their underlying type has to be interpreted for non-GLSL backends instead of copying the built-in decoration.

The issue can be observed with this simple HLSL fragment shader, compile it with DXC, and compile it back to HLSL with SPIRV-Cross:
```hlsl
float4 PSMain(uint mask : SV_Coverage) : SV_Target {
  return (float4)mask;
}
```
Permalink on shader-playground: https://shader-playground.timjones.io/7131a3e1959885f2afb4b8f897dadc2b

### DIFF
--- a/reference/opt/shaders-hlsl/frag/sample-mask-in-and-out.frag
+++ b/reference/opt/shaders-hlsl/frag/sample-mask-in-and-out.frag
@@ -1,5 +1,5 @@
-static int gl_SampleMaskIn;
-static int gl_SampleMask;
+static uint gl_SampleMaskIn[1];
+static uint gl_SampleMask[1];
 static float4 FragColor;
 
 struct SPIRV_Cross_Input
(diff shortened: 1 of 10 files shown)


HansKristian-Work:
The more proper solution I think is to keep the array-ness in the global static variables. I see the problem is that it's copying SampleMask array by value, which escapes access chain analysis.

E.g.:

static int gl_SampleMaskIn[1];
gl_SampleMaskIn[0] = stage_input.gl_SampleMaskIn;

and we can then get rid of the hack that is SampleMask handling in non-GLSL backends.

LukasBanana:
@HansKristian-Work I updated the PR. This works too and is less intrusive, but I had to add another function in the GLSL backend to allow overriding built-in types since `gl_SampleMaskIn` is a `uint` in HLSL but `int` in GLSL.

LukasBanana:
@HansKristian-Work don't quite understand how the logs from the failed CIS builds are related to these changes:
```
20: Test timeout computed to be: 10000000
20: Metal SDK is not present.
20: 
20: Failed to check if MSL -std=macos-metal2.2 is not supported. It probably is not.
20: Failed to check if MSL -std=macos-metal2.3 is not supported. It probably is not.
20: Failed to check if MSL -std=macos-metal2.4 is not supported. It probably is not.
20: Failed to check if MSL -std=metal3.0 is not supported. It probably is not.
```
Could it be a volatile error that disappears if you re-run those builds? Perhaps my force-push caused the trouble.

LukasBanana:
@HansKristian-Work It took me a few force-push attempts until I found all existing test cases, but everything is in the green now and the aforementioned example still works with this change.

PR #2209, opened 2023-10-06
Remove disabled MSVC warnings

bghgary:
[BinSkim](https://github.com/microsoft/binskim) is flagging these warnings as problematic. They don't seem to need to be disabled anymore.


HansKristian-Work:
Both warnings are observed in the CI build, so these disables are intentional.

bghgary:
I'll have to look for the warnings. I get that they are intentional, but they are potential security risks so will try to fix the warnings instead of disabling.

PR #2211, opened 2023-10-09
Add Natvis file with debug information for VectorView and SmallVector.

LukasBanana:
This Natvis file greatly improves the debuggability in VisualStudio. See attached screenshots of how `SmallVector` can be debugged with and without this Natvis file.

Debugging `input_variables` **with** Natvis:

SPIRV-Cross_WithNatvis

Debugging `input_variables` **without** Natvis:

SPIRV-Cross_WithoutNatvis


HansKristian-Work (approved):
I have no idea if this is correct, but can't hurt either.

PR #2213, opened 2023-10-10
Remove disabled MSVC warnings and fix the warnings

bghgary:
[BinSkim](https://github.com/microsoft/binskim) is flagging these warnings as problematic as they may indicate security issues. This change includes fixes for the warnings.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -142,7 +142,7 @@ if (CMAKE_COMPILER_IS_GNUCXX OR ((${CMAKE_CXX_COMPILER_ID} MATCHES "Clang") AND 
 	endif()
 elseif (MSVC)
 	# AppVeyor spuriously fails in debug build on older MSVC without /bigobj.
-	set(spirv-compiler-options ${spirv-compiler-options} /wd4267 /wd4996 $<$<CONFIG:DEBUG>:/bigobj>)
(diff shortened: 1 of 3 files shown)


HansKristian-Work:
This introduces a build failure. Both of these "fixes" are not relevant, and clearly not fixing any security issue (potential or otherwise). I'm not interested contorting stuff further to please MSVC's weird ideas of "secure" code here.

bghgary:
I did mark this a draft so that I can fix the build issues. If this is not going to be accepted even if I fix the build issues, I will stop contributing. We will look for a different solution. I was hoping to prevent other people from having this issue. It would also be good if we can have a discussion instead of shutting down PRs blankly. There are already some warnings being disabled that deal with these kind of issues in the code:
- https://github.com/KhronosGroup/SPIRV-Cross/blob/724433d72925f54682ae637c8bc4f4b7e83c4409/spirv_cross_c.cpp#L55-L58
- https://github.com/KhronosGroup/SPIRV-Cross/blob/724433d72925f54682ae637c8bc4f4b7e83c4409/tests-other/c_api_test.c#L3C1-L5
- etc.

Having a blanket disable warnings in the compiler will disable all warnings of that type from the compiler which may find issues that have potential to cause security issues. This is the reason why BinSkim is complaining. I agree these specific ones are probably not issues though.

PR #2214, opened 2023-10-11
MSL: Use powr instead of pow.

### LINKED ISSUE #2206: MSL: Imprecise GLSLstd450Pow translation
GLSL power function is undefined for negative _x_ so it should be translated to `metal::powr`, not `metal::pow`. For example `metal::pow(-2, 2)` is 4 while `metal::powr(-2, 2)` is NaN.

HansKristian-Work:
powr's additional UB rules match SPIR-V.

@billhollings @cdavis5e should verify this does not break CTS.

Fix #2206.


HansKristian-Work:
Ping @billhollings @cdavis5e.

billhollings (approved):
> Ping @billhollings @cdavis5e.

Sorry for the delay...and thanks for the reminder.

Just ran the relevant CTS on MoltenVK. 

All is well.

PR #2217, opened 2023-10-14
MSL: Workaround Metal 3.1 regression bug on recursive input structs.

billhollings:
Metal 3.1 introduced a Metal regression bug which causes an infinite recursion crash during Metal's analysis of an entry point input structure that itself contains internal recursion. This patch works around this by replacing the recursive input declaration with a alternate variable of type void*, and then casting to the correct type at the top of the entry point function.

- Add `CompilerMSL::Options::replace_recursive_inputs` to enable replacing recursive input.
- Add `Compiler::type_contains_recursion()` to determine if a struct contains internal recursion, and add custom Decorations to mark such structs, to short-cut future similar checks.
- Replace recursive input struct declarations with void*, and emit a recast to correct type at top of entry function.
- Add unit test.
- `Compiler::type_is_top_level_block()` remove hardcode reference to spirv_cross namespace, as it interferes with configurable namespaces (unrelated).

@HansKristian-Work This Metal regression surfaced quite recently, as macOS Sonoma was rolled out. This is a fairly simple patch fix that, if at all possible, it would be great to get this pulled in by [Mon Oct 16](https://github.com/KhronosGroup/MoltenVK/issues/2030), so it can be included in the upcoming Vulkan SDK release.

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -677,6 +677,7 @@ struct CLIArguments
 	bool msl_check_discarded_frag_stores = false;
 	bool msl_sample_dref_lod_array_as_grad = false;
 	bool msl_runtime_array_rich_descriptor = false;
+	bool msl_replace_recursive_inputs = false;
 	const char *msl_combined_sampler_suffix = nullptr;
 	bool glsl_emit_push_constant_as_ubo = false;
 	bool glsl_emit_ubo_as_plain_uniforms = false;
@@ -867,6 +868,7 @@ static void print_help_msl()
 	                "\t\tUses same values as Metal MTLArgumentBuffersTier enumeration (0 = Tier1, 1 = Tier2).\n"
 	                "\t\tNOTE: Setting this value no longer enables msl-argument-buffers implicitly.\n"
 	                "\t[--msl-runtime-array-rich-descriptor]:\n\t\tWhen declaring a runtime array of SSBOs, declare an array of {ptr, len} pairs to support OpArrayLength.\n"
+			        "\t[--msl-replace-recursive-inputs]:\n\t\tWorks around a Metal 3.1 regression bug, which causes an infinite recursion crash during Metal's analysis of an entry point input structure that itself contains internal recursion.\n"
 	                "\t[--msl-texture-buffer-native]:\n\t\tEnable native support for texel buffers. Otherwise, it is emulated as a normal texture.\n"
 	                "\t[--msl-framebuffer-fetch]:\n\t\tImplement subpass inputs with frame buffer fetch.\n"
 	                "\t\tEmits [[color(N)]] inputs in fragment stage.\n"
@@ -1233,6 +1235,7 @@ static string compile_iteration(const CLIArguments &args, std::vector<uint32_t> 
 		msl_opts.sample_dref_lod_array_as_grad = args.msl_sample_dref_lod_array_as_grad;
 		msl_opts.ios_support_base_vertex_instance = true;
 		msl_opts.runtime_array_rich_descriptor = args.msl_runtime_array_rich_descriptor;
+		msl_opts.replace_recursive_inputs = args.msl_replace_recursive_inputs;
 		msl_comp->set_msl_options(msl_opts);
 		for (auto &v : args.msl_discrete_descriptor_sets)
 			msl_comp->add_discrete_descriptor_set(v);
@@ -1792,6 +1795,8 @@ static int main_inner(int argc, char *argv[])
 	});
 	cbs.add("--msl-runtime-array-rich-descriptor",
 	        [&args](CLIParser &) { args.msl_runtime_array_rich_descriptor = true; });
+	cbs.add("--msl-replace-recursive-inputs",
+	        [&args](CLIParser &) { args.msl_replace_recursive_inputs = true; });
 	cbs.add("--extension", [&args](CLIParser &parser) { args.extensions.push_back(parser.next_string()); });
 	cbs.add("--rename-entry-point", [&args](CLIParser &parser) {
 		auto old_name = parser.next_string();
--- a/spirv_cross.cpp
+++ b/spirv_cross.cpp
@@ -1048,7 +1048,7 @@ ShaderResources Compiler::get_shader_resources(const unordered_set<VariableID> *
 	return res;
 }
 
-bool Compiler::type_is_top_level_block(const spirv_cross::SPIRType &type) const
+bool Compiler::type_is_top_level_block(const SPIRType &type) const
 {
 	if (type.basetype != SPIRType::Struct)
 		return false;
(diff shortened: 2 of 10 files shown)


cdavis5e (approved):
LGTM. I have one small comment, though. And I'd probably wait for H-K to give his OK before merging.

HansKristian-Work:
Deadline for the next SDK has already passed, btw.

main.cpp
```diff
@@ -1792,6 +1796,8 @@ static int main_inner(int argc, char *argv[])
 	});
 	cbs.add("--msl-runtime-array-rich-descriptor",
 	        [&args](CLIParser &) { args.msl_runtime_array_rich_descriptor = true; });
+	cbs.add("--msl-replace-recursive-inputs",
+			[&args](CLIParser &) { args.msl_replace_recursive_inputs = true; });
```
HansKristian-Work, line 1800: Indentation is off.

spirv_cross.cpp
```diff
@@ -5465,6 +5465,52 @@ void Compiler::analyze_interlocked_resource_usage()
 	}
 }
 
+// Helper function
+bool Compiler::check_internal_recursion(SPIRType &type, std::unordered_set<uint32_t> &checked_ids)
+{
+	// If this is not the base type (eg. it's a pointer or array), tunnel down
+	if (type.parent_type)
```
HansKristian-Work, line 5472: It's more correct to look at type.self here. It will point to the non-pointer / non-array type.

spirv_cross.cpp
```diff
@@ -5465,6 +5465,52 @@ void Compiler::analyze_interlocked_resource_usage()
...
+		return check_internal_recursion(get<SPIRType>(type.parent_type), checked_ids);
+
+	if ( !(type.basetype == SPIRType::Struct) )
+		return false;
+
+	if (has_extended_decoration(type.self, SPIRVCrossDecorationNonRecursiveStruct))
```
HansKristian-Work, line 5478: Do we *really* need two more decorations here? It bloats memory usage in the meta structs. These decorations are only used in local analysis, can just keep this information in a map while emitting entry point.

spirv_cross.cpp
```diff
@@ -5465,6 +5465,52 @@ void Compiler::analyze_interlocked_resource_usage()
...
 }
 
+// Helper function
+bool Compiler::check_internal_recursion(SPIRType &type, std::unordered_set<uint32_t> &checked_ids)
+{
+	// If this is not the base type (eg. it's a pointer or array), tunnel down
+	if (type.parent_type)
+		return check_internal_recursion(get<SPIRType>(type.parent_type), checked_ids);
+
+	if ( !(type.basetype == SPIRType::Struct) )
```
HansKristian-Work, line 5475: !=

spirv_cross.hpp
```diff
@@ -1145,6 +1145,8 @@ class Compiler
 	bool has_extended_member_decoration(uint32_t type, uint32_t index, ExtendedDecorations decoration) const;
 	void unset_extended_member_decoration(uint32_t type, uint32_t index, ExtendedDecorations decoration);
 
+	bool check_internal_recursion(SPIRType &type, std::unordered_set<uint32_t> &checked_ids);
```
HansKristian-Work, line 1148: const SPIRType

spirv_msl.hpp
```diff
@@ -505,6 +505,13 @@ class CompilerMSL : public CompilerGLSL
 		// Note: Only Apple's GPU compiler takes advantage of the lack of coherency, so make sure to test on Apple GPUs if you disable this.
 		bool readwrite_texture_fences = true;
 
+		// Metal 3.1 introduced a Metal regression bug which causes infinite recursion during 
+		// Metal's analysis of an entry point input structure that is itself recursive. Enabling
+		// this option will replace the recursive input declaration with a alternate variable of
+		// type void*, and then cast to the correct type at the top of the entry point function.
+		// The bug has been reported to Apple, and will hopefully be fixed in future releases.
+		bool replace_recursive_inputs = false;
```
HansKristian-Work, line 513: Does this really need to be an option? Seems like everyone needs this workaround. Linked list in buffers isn't exactly ... common.

spirv_cross.cpp
```diff
@@ -1048,7 +1048,7 @@ ShaderResources Compiler::get_shader_resources(const unordered_set<VariableID> *
 	return res;
 }
 
-bool Compiler::type_is_top_level_block(const spirv_cross::SPIRType &type) const
+bool Compiler::type_is_top_level_block(const SPIRType &type) const
```
HansKristian-Work, line 1051: Pushed a fix for this. Silly IDE auto-added it.

HansKristian-Work (changes requested):

HansKristian-Work:
If MoltenVK has a different SDK deadline from SPIRV-Cross and it can reference a different SPIRV-Cross commit locally, it would make sense to just push up a frozen branch that MVK can reference in the short-term. SPIRV-Cross upstream shouldn't need to block what ships in MVK SDKs.

KarenGhavam-lunarG:
If I can get a new spirv cross commit on Monday morning we can use that for the sdk

main.cpp
```diff
@@ -867,6 +868,8 @@ static void print_help_msl()
 	                "\t\tUses same values as Metal MTLArgumentBuffersTier enumeration (0 = Tier1, 1 = Tier2).\n"
 	                "\t\tNOTE: Setting this value no longer enables msl-argument-buffers implicitly.\n"
 	                "\t[--msl-runtime-array-rich-descriptor]:\n\t\tWhen declaring a runtime array of SSBOs, declare an array of {ptr, len} pairs to support OpArrayLength.\n"
+			        "\t[--replace-recursive-inputs]:\n\t\tWorks around a Metal 3.1 regression bug, which causes an infinite recursion"
+			        "crash during Metal's analysis of an entry point input structure that itself contains internal recursion.\n"
```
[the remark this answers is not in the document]
billhollings: Not sure what the indentation should be. It seems to be the same as some other options that have longer text. To be safe, I've moved it to a single line. I also noticed I had actually missed the msl-` prefix on this help output too. I've added that now.

main.cpp
billhollings: Not in my IDE. Must just be a tab-spacing issue. I've copied the spacing from the line above.

spirv_cross.cpp
billhollings: Fixed.

spirv_cross.hpp
billhollings: Fixed. And in `type_contains_recursion()`.

spirv_msl.hpp
billhollings: I made it an option, because it does not need to be invoked before macOS 14.0 or iOS 17.0 runtimes, _**and**_  I am hoping it should have a short shelf-life, if Apple responds to the regression bug (which by their feedback, I believe they will). So, it may not be needed again soon, and hopefully it will only apply to a short window of OS versions. 

MoltenVK tracks its need, and sets it, based on OS release. 

Invoking it only when it is needed makes sense, as it makes the MSL shaders less readable when it is invoked. And I wasn't sure that other SPIRV-Cross users would need it.

Although we _**might**_ try to internalize this into SPIRV-Cross based on MSL version, it's really tied to the Metal pipeline compiler version itself, not the MSL language version pushed to SPIRV-Cross.

BTW...given it's obscurity, I did contemplate ignoring it and waiting for Apple to fix it, but it does block CTS testing, which significantly hampers MoltenVK development.

spirv_cross.cpp
billhollings: I took `if (type.parent_type)` from `CompilerMSL::mark_as_packable()`. Maybe it's not needed at all?

Removed it here.

spirv_cross.cpp
billhollings: Good point. When I put it together, I wasn't sure how often it might be called, and I looked for a short-circuit to always recursing. 

I've added a `set` to hold the info now, and populate it with one round of recursion tests.

billhollings:
> If MoltenVK has a different SDK deadline from SPIRV-Cross and it can reference a different SPIRV-Cross commit locally, it would make sense to just push up a frozen branch that MVK can reference in the short-term. SPIRV-Cross upstream shouldn't need to block what ships in MVK SDKs.

All good points.

I was unaware that SPIRV-Cross had an SDK delivery date at all. MoltenVK typically uses the SPIRV-Cross main HEAD at time of MoltenVK SDK release, and this has worked so far.

I guess having a MoltenVK branch on SPIRV-Cross would give us options when something like this comes up, without rushing the main branch. However, this situation has proven extremely rare so far, and maintaining a separate MoltenVK branch feels like it could potentially become a maintenance nightmare for all of us down the road, if it deviates far from the main branch, by accumulating updates that don't quickly make it into the main branch.

If we can fit this into `main` by Monday this time, that would be best. If not, I can create that SPIRV-Cross branch as an emergency temporary measure for this release, but I'd want to merge it back into the main branch soon.

Either way, we can have further discussions about whether a dual-branch is needed, and how it can be made effective.

spirv_cross.cpp
billhollings: Okay. I wonder why this PR is not now choking on it?

PR #2227, opened 2023-11-13
Workaround for generating bool uniforms in ESSL1

### LINKED ISSUE #2223: uniforms of type bool are transpiled as uint when targeting #version 100
As the title says. This results in an exception

```
Unsigned integers are not supported on legacy ESSL.
```

To reproduce:

```
#version 310 es

layout(binding = 0) uniform Foo {
  bool bar;
} foo;

void main() {}
```

Disabling the relevant checks in spirv_glsl.cpp showed that `bar` was indeed changed into a `uint`.

elizagamedev:
Fixes #2223.

### DIFF
--- a/spirv_glsl.cpp
+++ b/spirv_glsl.cpp
@@ -15051,7 +15051,11 @@ string CompilerGLSL::flags_to_qualifiers_glsl(const SPIRType &type, const Bitset
 	{
 		auto &execution = get_entry_point();
 
-		if (flags.get(DecorationRelaxedPrecision))
+		if (type.basetype == SPIRType::UInt && is_legacy()) {
+			// HACK: This is a bool. See comment in type_to_glsl().
+			qual += "lowp ";
+		}
+		else if (flags.get(DecorationRelaxedPrecision))
 		{
 			bool implied_fmediump = type.basetype == SPIRType::Float &&
 			                        options.fragment.default_float_precision == Options::Mediump &&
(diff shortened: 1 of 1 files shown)


spirv_glsl.cpp
```diff
@@ -15051,7 +15051,11 @@ string CompilerGLSL::flags_to_qualifiers_glsl(const SPIRType &type, const Bitset
 	{
 		auto &execution = get_entry_point();
 
-		if (flags.get(DecorationRelaxedPrecision))
+		if (type.basetype == SPIRType::UInt && is_legacy()) {
+			// HACK: This is a bool. See comment in type_to_glsl().
+			qual += "lowp ";
+		}
+		else if (flags.get(DecorationRelaxedPrecision))
```
HansKristian-Work, line 15058:
```suggestion
		if (type.basetype == SPIRType::UInt && is_legacy())
		{
			// HACK: This is a bool. See comment in type_to_glsl().
			qual += "lowp ";
		}
		else if (flags.get(DecorationRelaxedPrecision))
```

spirv_glsl.cpp
```diff
@@ -15051,7 +15051,12 @@ string CompilerGLSL::flags_to_qualifiers_glsl(const SPIRType &type, const Bitset
 	{
 		auto &execution = get_entry_point();
 
-		if (flags.get(DecorationRelaxedPrecision))
+		if (type.basetype == SPIRType::UInt && is_legacy())
```
HansKristian-Work, line 15054:
```suggestion
		if (type.basetype == SPIRType::UInt && is_legacy_es())
```

Issue #2132, opened 2023-03-31
generate the smallest float/double that preserves the binary representation


pixelflinger:
Currently spirv-cross uses `%.32g` by default to format `floats`, this can be overridden with `SPIRV_CROSS_FLT_FMT` at compile time. When embedding shaders with binaries, reducing their binary footprint can be important.

What really matters is to preserve the  binary representation of the float in the SPIRV file. 32 digits is overkill for `floats` and `doubles`. We could use respectively 9 and 17 digits, which is always enough, however, this would still generate strings larger than needed.

Ideally, we'd want to use the smallest string that preserves the binary representation. 

For instance, I have tested the following implementation for `floats` only:

```
static inline std::string convert_to_smallest_string(float f, char locale_radix_point) {
    char buf[16]; // e.g.: -0.12345678e-12, 16 bytes needed
    float r;
    for (int i = 1; i < 9; i++) {
        sprintf(buf, "%.*g", i, f);
        sscanf(buf, "%f", &r);
        if (r == f) {
            break;
        }
    }
    fixup_radix_point(buf, locale_radix_point);

    // Ensure that the literal is float.
    if (!strchr(buf, '.') && !strchr(buf, 'e'))
        strcat(buf, ".0");

    return { buf };
}
```

For `doubles` and `long doubles` it might better to binary-search the right format length.

HansKristian-Work:
> We could use respectively 9 and 17 digits, which is always enough

This approach failed in one of my tests, but maybe I did something wrong. Either way, shortening the string would be great, it's been a long standing todo that just never bubbled up as something critical, but we need to have a strategy on how we're going to verify it if we're going to implement this. For FP32 we can roundtrip every possible value in a reasonable amount of time, and we should be able to test a reasonable subset of FP64 as well.

pixelflinger:
@HansKristian-Work there is some good information here (as to why `%.9g` works for floats):
https://randomascii.wordpress.com/2013/02/07/float-precision-revisited-nine-digit-float-portability

PR #2230, opened 2023-11-22
Allow to customize float literals in output

chirsz-ever:
This PR does not change the current behavior. It only works when invoking `set_float_formatter()` with initiative.

I use `std::shared_ptr` to store a `FloatFormatter` virtual base object, because it can use customized deleter (compare to `std::unique_ptr`). Would it be better to just use a raw pointer `FloatFormatter*` and require the user to keep it valid?

Related issues: #2132 #1599

### DIFF
--- a/spirv_common.hpp
+++ b/spirv_common.hpp
@@ -295,6 +295,20 @@ inline std::string convert_to_string(double t, char locale_radix_point)
 	return buf;
 }
 
+#if defined(__clang__) || defined(__GNUC__)
+#pragma GCC diagnostic pop
+#elif defined(_MSC_VER)
+#pragma warning(pop)
+#endif
+
+class FloatFormatter
+{
+public:
+	virtual ~FloatFormatter() = default;
+	virtual std::string format_float(float value) = 0;
+	virtual std::string format_double(double value) = 0;
+};
+
 template <typename T>
 struct ValueSaver
 {
@@ -317,12 +331,6 @@ struct ValueSaver
 	T &current;
 	T saved;
 };
-
-#if defined(__clang__) || defined(__GNUC__)
-#pragma GCC diagnostic pop
-#elif defined(_MSC_VER)
-#pragma warning(pop)
-#endif
 
 struct Instruction
 {
--- a/spirv_glsl.hpp
+++ b/spirv_glsl.hpp
@@ -1039,7 +1039,7 @@ protected:
 	std::unordered_set<LocationComponentPair, InternalHasher> masked_output_locations;
 	std::unordered_set<uint32_t> masked_output_builtins;
 
-	FloatFormatter* float_formatter = nullptr;
+	FloatFormatter *float_formatter = nullptr;
 	std::string format_float(float value) const;
 	std::string format_double(double value) const;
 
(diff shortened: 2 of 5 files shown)


spirv_glsl.hpp
```diff
@@ -287,6 +288,13 @@ class CompilerGLSL : public Compiler
 	void mask_stage_output_by_location(uint32_t location, uint32_t component);
 	void mask_stage_output_by_builtin(spv::BuiltIn builtin);
 
+	// Allow to control how to format float literals in the output.
+	// Set to "nullptr" to use the default "convert_to_string" function.
+	void set_float_formatter(std::shared_ptr<FloatFormatter> formatter)
```
HansKristian-Work, line 293: Ownership is not necessary here, just use a plain pointer.

HansKristian-Work (changes requested):
As mentioned, this is simple enough, see comments though.

spirv_common.hpp
```diff
@@ -295,6 +295,20 @@ inline std::string convert_to_string(double t, char locale_radix_point)
...
+#pragma GCC diagnostic pop
+#elif defined(_MSC_VER)
+#pragma warning(pop)
+#endif
+
+class FloatFormatter
+{
+public:
+	virtual ~FloatFormatter() = default;
+	virtual std::string format_float(float value) = 0;
```
HansKristian-Work, line 308: Is one needed for fp16, or will that hit the format_float part?

chirsz-ever: Currently `float16_t` will be converted to `float` before be converted to string:

https://github.com/KhronosGroup/SPIRV-Cross/blob/42aac916ab5db0cbaf55c2d41f4d063f4ce3955a/spirv_glsl.cpp#L5910-L5913

I think `format_float16` is not needed.

spirv_glsl.hpp
```diff
@@ -287,6 +287,14 @@ class CompilerGLSL : public Compiler
 	void mask_stage_output_by_location(uint32_t location, uint32_t component);
 	void mask_stage_output_by_builtin(spv::BuiltIn builtin);
 
+	// Allow to control how to format float literals in the output.
+	// Set to "nullptr" to use the default "convert_to_string" function.
+	// It needs to ensure that the pointer is valid before the compiler object destructs.
+	void set_float_formatter(FloatFormatter* formatter)
```
HansKristian-Work, line 293:
```suggestion
	void set_float_formatter(FloatFormatter *formatter)
```

spirv_glsl.hpp
```diff
@@ -287,6 +287,14 @@ class CompilerGLSL : public Compiler
 	void mask_stage_output_by_location(uint32_t location, uint32_t component);
 	void mask_stage_output_by_builtin(spv::BuiltIn builtin);
 
+	// Allow to control how to format float literals in the output.
+	// Set to "nullptr" to use the default "convert_to_string" function.
+	// It needs to ensure that the pointer is valid before the compiler object destructs.
```
HansKristian-Work, line 292:
```suggestion
	// This handle is not owned by SPIRV-Cross and must remain valid until compile() has been called.
```

spirv_glsl.hpp
```diff
@@ -1031,6 +1039,10 @@ class CompilerGLSL : public Compiler
 	std::unordered_set<LocationComponentPair, InternalHasher> masked_output_locations;
 	std::unordered_set<uint32_t> masked_output_builtins;
 
+	FloatFormatter* float_formatter = nullptr;
```
HansKristian-Work, line 1042:
```suggestion
	FloatFormatter *float_formatter = nullptr;
```

PR #2237, opened 2023-12-04
MSL: Fix patch vertex count with raw buffer tese input.

cdavis5e:
We can no longer rely on the `patch_control_point<>` array being present, so the best we can do is use the value given us at compile time.

This was an oversight on my part when I initially implemented the raw-buffer tessellation evaluation input mode. The lack of tests for the `PatchVertices` built-in almost certainly contributed, so I fixed that in this patch.

Fixes the test
`dEQP-VK.tessellation.shader_input_output.patch_vertices_in_tes`. This is the last failing test under `dEQP-VK.tessellation`.

### DIFF
--- a/reference/opt/shaders-msl/tese/read-patch-vertices-in-func.raw-tess-in.tese
+++ b/reference/opt/shaders-msl/tese/read-patch-vertices-in-func.raw-tess-in.tese
@@ -1,0 +1,18 @@
+#include <metal_stdlib>
+#include <simd/simd.h>
+
+using namespace metal;
+
+struct main0_out
+{
+    float4 gl_Position [[position]];
+};
+
+[[ patch(quad, 0) ]] vertex main0_out main0(uint gl_PrimitiveID [[patch_id]])
+{
+    main0_out out = {};
+    uint gl_PatchVerticesIn = 0;
+    out.gl_Position = float4(float(gl_PatchVerticesIn), 0.0, 0.0, 1.0);
+    return out;
+}
+
--- a/spirv_msl.cpp
+++ b/spirv_msl.cpp
@@ -13679,14 +13679,31 @@ void CompilerMSL::fix_up_shader_inputs_outputs()
 				break;
 			case BuiltInPatchVertices:
 				if (is_tese_shader())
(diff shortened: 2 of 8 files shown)


spirv_msl.cpp
```diff
@@ -13679,14 +13679,31 @@ void CompilerMSL::fix_up_shader_inputs_outputs()
...
-						statement(builtin_type_decl(bi_type), " ", to_expression(var_id), " = ",
-						          to_expression(patch_stage_in_var_id), ".gl_in.size();");
-					});
+				{
+					if (msl_options.raw_buffer_tese_input)
+					{
+						entry_func.fixup_hooks_in.push_back(
+						    [=]() {
+							    statement(builtin_type_decl(bi_type), " ", to_expression(var_id), " = ",
+							              get_entry_point().output_vertices, ";");
```
HansKristian-Work, line 13688: output_vertices? Is that correct? Is it intended that this execution mode is overridden by MoltenVK?

HansKristian-Work: I wouldn't mind just adding a set_patch_vertex_inputs() as a separate function to avoid any odd "reinterpretation" of execution modes.

reference/opt/shaders-msl/tese/read-patch-vertices-in-func.raw-tess-in.tese
```diff
@@ -0,0 +1,18 @@
...
+
+struct main0_out
+{
+    float4 gl_Position [[position]];
+};
+
+[[ patch(quad, 0) ]] vertex main0_out main0(uint gl_PrimitiveID [[patch_id]])
+{
+    main0_out out = {};
+    uint gl_BuiltIn_14 = 0;
```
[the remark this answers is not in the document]
cdavis5e: I'm actually kinda amazed that we were not really handling `gl_PatchVerticesIn` at all. For completeness, I've added a GLSL test as well.

spirv_msl.cpp
HansKristian-Work: Ok.

HansKristian-Work (approved):

Issue #2145, opened 2023-05-07
Metal output does not compile in sample using PhysicalStorageBuffer pointers


Hugobros3:
MoltenVK fails to run the [aobench](https://github.com/Hugobros3/shady/tree/master/samples/aobench) sample from shady. I tracked down the issue and was able to minimize the source code down to this:
```
; SPIR-V
; Version: 1.5
; Generator: Khronos; 35
; Bound: 5550
; Schema: 0
               OpCapability Int8
               OpCapability Int64
               OpCapability Int64
               OpCapability Shader
               OpCapability PhysicalStorageBufferAddresses
               OpExtension "SPV_KHR_physical_storage_buffer"
               OpMemoryModel PhysicalStorageBuffer64 Simple
               OpEntryPoint GLCompute %main "main" %globals
               OpExecutionMode %main LocalSize 16 16 1

               OpDecorate %ptr_uchar ArrayStride 8
               OpDecorate %ptr_uint ArrayStride 8
               OpDecorate %ptr_array_t ArrayStride 8
               OpDecorate %array_t ArrayStride 1
               OpDecorate %struct_t Block
               OpMemberDecorate %struct_t 0 Offset 0
               OpMemberDecorate %struct_t 1 Offset 8
               OpDecorate %ptr_struct ArrayStride 32
               OpDecorate %globals DescriptorSet 0
               OpDecorate %globals Binding 0

       %void = OpTypeVoid
      %uchar = OpTypeInt 8 0
       %uint = OpTypeInt 32 0
      %ulong = OpTypeInt 64 0
       %bool = OpTypeBool

   %ulong_12 = OpConstant %ulong 12
     %uint_0 = OpConstant %uint 0
     %uint_1 = OpConstant %uint 1
     %uint_2 = OpConstant %uint 2
   %uchar_69 = OpConstant %uchar 69
   %ulong_16 = OpConstant %ulong 16

   %ptr_uint = OpTypePointer PhysicalStorageBuffer %uint
  %ptr_uchar = OpTypePointer PhysicalStorageBuffer %uchar

    %array_t = OpTypeArray %uchar %ulong_12
%ptr_array_t = OpTypePointer PhysicalStorageBuffer %array_t

   %struct_t = OpTypeStruct %ptr_uchar %ptr_array_t
 %ptr_struct = OpTypePointer StorageBuffer %struct_t

    %void_fn = OpTypeFunction %void
[...]

PR #2238, opened 2023-12-04
Add an 'op' operand to SPIRType

Hugobros3:
Continuing the discussion started in https://github.com/KhronosGroup/SPIRV-Cross/issues/2145#issuecomment-1545564097:

The current design of `SPIRType` in SPIRV-Cross is "flat": a given instance of SPIRType provides information on the number of stars (point depth), array size, vector size and even matrix size all in the same struct, in pseudocode:

`Type(vectorwidth=2, ptr=1, as=Input, base=Int, bitwidth=32)`

This is a departure from most compilers and the SPIR-V specification where a complex type is built up from a base type by applying a number of constructors:

`PtrType(as=Input, pointee=VectorType(width=2, element_t=IntType(bitwidth=32)))`

The flat design in use currently is however problematic:
 * Complex types such as structures cannot effectively be flattened - so the actual implementation uses a mixture of both designs
 * Pattern matching on type constructors is done in a ad-hoc way by looking at fields, for example `type.array.empty()`. This is unsafe because a pointer to an array passes this test too (1)
 * There is no way to disambiguate pointer-to-array and array-of-pointers. The former patterns historically were absent from Vulkan flavours of SPIR-V due to the initially restrictive capabilities, but with the addition of physical storage buffers this is no longer tenable.
 
 (1) And in fact, a number of places in the implementation make such shortcuts, as I found out over the last few days.

This patch attempts to bring `SPIRType` back in line with this more traditional approach by introducing an `op` field to SPIRType, which corresponds to the SPIR-V opcode for the top-level type expression of any given type. This operand is set when constructing a type, either when parsing it or when created "inline".

Since there is still no validation that the `op` matches the various flattened fields, great care was taken to comb over the existing code and correctly set the operand wherever types are created. In the future it would be a great addition to move away from creating `SPIRType`s inline and instead make proper constructors, that would also enable a smoother migration fully away from the flat design.

The tests all pass with no regressions currently

### DIFF
--- a/spirv_msl.cpp
+++ b/spirv_msl.cpp
@@ -486,6 +486,7 @@ void CompilerMSL::build_implicit_builtins()
 
 				SPIRType vec4_type_ptr { spv::Op::OpTypePointer };
 				vec4_type_ptr = vec4_type;
+				vec4_type_ptr.op = spv::Op::OpTypePointer;
 				vec4_type_ptr.pointer = true;
 				vec4_type_ptr.pointer_depth++;
 				vec4_type_ptr.parent_type = type_id;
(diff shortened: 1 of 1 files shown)


HansKristian-Work:
There are CI failures on MSL. The test files need to be updated.

PR #2247, opened 2023-12-21
HLSL: fix packing of structs

Try:
It seems there was a typo in calculation of `begin_word`/`end_word`, causing some shaders in opengothic not to work


HansKristian-Work:
I reproduced the issue, but the fix seems very off to me. I'll look into it.

HansKristian-Work:
Fixed on main.

PR #2257, opened 2024-01-06
MSL: Add support for SPV_EXT_integer_dot_product

spnda:
As a dot product requires a reduce_add function I added that with overloads for all integer vector types. I don't quite like it, but there's no way around that using templates or something like `__builtin_reduce_add` which would usually be available with Clang but is not accepted by the Metal compiler.

I think this should cover all of the extra cases, the type widening to match the return type and the 32-bit integer unpacking.

### DIFF
--- a/reference/shaders-msl/comp/integer-dot-product.comp
+++ b/reference/shaders-msl/comp/integer-dot-product.comp
@@ -1,0 +1,44 @@
+#pragma clang diagnostic ignored "-Wmissing-prototypes"
+
+#include <metal_stdlib>
+#include <simd/simd.h>
+
+using namespace metal;
+
+template <typename T>
+T reduce_add(vec<T, 2> v) { return v.x + v.y; }
+template <typename T>
+T reduce_add(vec<T, 3> v) { return v.x + v.y + v.z; }
+template <typename T>
+T reduce_add(vec<T, 4> v) { return v.x + v.y + v.z + v.w; }
+
+struct InOut
+{
+    uint4 x;
+    uint4 y;
+    int result;
+};
+
+struct InOut2
+{
+    uint x;
+    uint y;
+    uint result;
+};
+
+struct InOut3
+{
+    ushort4 x;
+    ushort4 y;
+    int acc;
+    int result;
+};
+
+kernel void main0(device InOut& comp [[buffer(0)]], device void* spvBufferAliasSet0Binding1 [[buffer(1)]])
+{
+    device auto& comp2 = *(device InOut2*)spvBufferAliasSet0Binding1;
+    device auto& comp3 = *(device InOut3*)spvBufferAliasSet0Binding1;
+    comp.result = reduce_add(int4(comp.x) * int4(comp.y));
+    comp2.result = uint(reduce_add(uchar4(as_type<uchar4>(comp2.x)) * uchar4(as_type<uchar4>(comp2.y))));
+    comp3.result = addsat(reduce_add(int4(comp3.x) * int4(comp3.y)), comp3.acc);
+}
--- a/spirv_msl.cpp
+++ b/spirv_msl.cpp
@@ -7443,6 +7443,22 @@ void CompilerMSL::emit_custom_functions()
 			statement("");
 			break;
 
+		case SPVFuncImplReduceAdd:
+			// Metal doesn't support __builtin_reduce_add or simd_reduce_add, so we need this.
+			// Metal also doesn't support the other vector builtins, which would have been useful to make this a single template.
+
+			statement("template <typename T>");
+			statement("T reduce_add(vec<T, 2> v) { return v.x + v.y; }");
+
+			statement("template <typename T>");
+			statement("T reduce_add(vec<T, 3> v) { return v.x + v.y + v.z; }");
+
+			statement("template <typename T>");
+			statement("T reduce_add(vec<T, 4> v) { return v.x + v.y + v.z + v.w; }");
+
+			statement("");
+			break;
+
 		default:
 			break;
 		}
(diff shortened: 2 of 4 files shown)


spirv_msl.cpp
```diff
@@ -9626,6 +9656,75 @@ void CompilerMSL::emit_instruction(const Instruction &instruction)
...
+		if ((input_type2.basetype == SPIRType::BaseType::UInt || input_type2.basetype == SPIRType::BaseType::Int) && input_type2.vecsize == 1)
+		{
+			string type = opcode == OpSDot ? "char4" : "uchar4";
+			vec2input = join("as_type<", type, ">(", to_expression(vec2), ")");
+		} else
+			vec2input = to_expression(vec2);
+
+		auto& type = get<SPIRType>(result_type);
+		auto input_width = input_type1.vecsize == 1 ? 4 : input_type1.vecsize;
+		string exp = join("reduce_add(", type_to_glsl(type), std::to_string(input_width), "(", vec1input, ") * ", type_to_glsl(type), std::to_string(input_width), "(", vec2input, "))");
```
HansKristian-Work, line 9688: std::to_string(input_width) is not needed in join(). Just use the integer directly.

spirv_msl.cpp
```diff
@@ -9626,6 +9656,75 @@ void CompilerMSL::emit_instruction(const Instruction &instruction)
...
+		emit_op(result_type, id, exp, should_forward(vec1) && should_forward(vec2));
```
HansKristian-Work, line 9689: Need inherit_expression_dependencies here.

reference/shaders-msl/comp/integer-dot-product.comp
```diff
@@ -0,0 +1,50 @@
...
+char reduce_add(char4 x) { return x.x + x.y + x.z + x.w; }
+uchar reduce_add(uchar2 x) { return x.x + x.y; }
+uchar reduce_add(uchar3 x) { return x.x + x.y + x.z; }
+uchar reduce_add(uchar4 x) { return x.x + x.y + x.z + x.w; }
+short reduce_add(short2 x) { return x.x + x.y; }
+short reduce_add(short3 x) { return x.x + x.y + x.z; }
+short reduce_add(short4 x) { return x.x + x.y + x.z + x.w; }
+ushort reduce_add(ushort2 x) { return x.x + x.y; }
+ushort reduce_add(ushort3 x) { return x.x + x.y + x.z; }
+ushort reduce_add(ushort4 x) { return x.x + x.y + x.z + x.w; }
```
HansKristian-Work, line 19: Can this really not be made into a template?

```
template <typename T, int count>
T reduce_add2(vector<T, count> v) { return v.x + v.y; }
```
etc. Maybe it can be specialized for count 2, 3, and 4, but probably not worth the hassle.

reference/shaders-msl/comp/integer-dot-product.comp
```diff
@@ -0,0 +1,50 @@
...
+{
+    uint x;
+    uint y;
+    uint result;
+};
+
+kernel void main0(device InOut& comp [[buffer(0)]], device InOut2& comp2 [[buffer(1)]])
+{
+    comp.result = reduce_add(int4(comp.x) * int4(comp.y));
+    comp2.result = uint(reduce_add(uchar4(as_type<uchar4>(comp2.x)) * uchar4(as_type<uchar4>(comp2.y))));
```
HansKristian-Work, line 49: Missing saturating tests.

spirv_msl.cpp
```diff
@@ -9626,6 +9656,75 @@ void CompilerMSL::emit_instruction(const Instruction &instruction)
...
+		if ((input_type2.basetype == SPIRType::BaseType::UInt || input_type2.basetype == SPIRType::BaseType::Int) && input_type2.vecsize == 1)
+		{
+			string type = opcode == OpSDotAccSat ? "char4" : "uchar4";
+			vec2input = join("as_type<", type, ">(", to_expression(vec2), ")");
+		} else
+			vec2input = to_expression(vec2);
+
+		auto& type = get<SPIRType>(result_type);
+		auto input_width = input_type1.vecsize == 1 ? 4 : input_type1.vecsize;
+		string exp = join("addsat(reduce_add(", type_to_glsl(type), std::to_string(input_width), "(", vec1input, ") * ", type_to_glsl(type), std::to_string(input_width), "(", vec2input, ")), ", to_expression(acc), ")");
```
HansKristian-Work, line 9723: The result type cannot be used directly I think. If the opcode is SDotAccSat, you have to force the addition to happen in int, not uint. SPIR-V can mix and match like that. UDotAccSat seems to have language forcing signedness of 0, so that's probably fine as-is ...

HansKristian-Work (changes requested):
Misc comments.

spirv_msl.cpp
```diff
@@ -9626,6 +9656,75 @@ void CompilerMSL::emit_instruction(const Instruction &instruction)
...
+		uint32_t result_type = ops[0];
+		uint32_t id = ops[1];
+		uint32_t vec1 = ops[2];
+		uint32_t vec2 = ops[3];
+
+		auto& input_type1 = expression_type(vec1);
+		auto& input_type2 = expression_type(vec2);
+
+		string vec1input, vec2input;
+		if ((input_type1.basetype == SPIRType::BaseType::UInt || input_type1.basetype == SPIRType::BaseType::Int) && input_type1.vecsize == 1)
```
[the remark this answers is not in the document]
spnda: Do I just do `ops[5] == DotProductInput4x8BitPacked`? Or do I need to cast it or convert it using `get` first?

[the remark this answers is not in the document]
HansKristian-Work: It's a literal, but it's also an optional argument, so need to check length of the opcode first.

spirv_msl.cpp
spnda: Yes, I think you're right. I guess I'll just have to add some logic to generate "uint4" when `type.width == 32` and `input_type1.vecsize == 4` and its `OpUDot`. When it's `OpSDot` it'll always be `int4` regardless of the return type. Do I need to take anything else into account? Though interesting that the Vulkan CTS tests don't test if that is handled correctly...

spnda: Oh wait you're talking about the saturated addition? The SPIR-V spec says `The type of Accumulator must be the same as Result Type.`. So I'm pretty sure it can be used.

I thought you were talking about the result type of the component-wise multiplication. I'm not entirely sure, but I thought if its OpSDot and the result type is uint, the multiplication and reduce_add would still have to be done on signed integers, no? So me casting it always to the vector version of the result type could possibly be wrong because I'm possibly not sign-extending, right?

HansKristian-Work: Using signedness in OpTypeInt is somewhat optional in SPIR-V and only really applies in specific circumstances.

spnda: That would require a `as_type<int4>(x)` then, no?

billhollings:
In addition to the `reduce_add()` template function, is there any value in adding something like a `dot()` template function for integer vectors, so that the resulting MSL code looks both like the equivalent Metal float vector functions, and like the equivalent GLSL output, and is therefore more readable?

spnda:
> In addition to the `reduce_add()` template function, is there any value in adding something like a `dot()` template function for integer vectors, so that the resulting MSL code looks both like the equivalent Metal float vector functions, and like the equivalent GLSL output, and is therefore more readable?

based on a basic version of the function, perhaps. Now that the function is getting a bit more complex, especially around the signed/unsigned conversion and the unpacking of integers, I'm not sure I fully see the gain. Most of the logic that exists for the SPIR-V opcodes to be implemented correctly would have to happen outside of the function due to it having to know more about the context.

Mostly, I am referencing what @HansKristian-Work noticed about the unsigned -> signed integer conversion. I'll have to use codegen to effectively add a `as_type<int4>(uint4(x))` when an unsigned integer is provided for a signed integer parameter. I don't think that'd really be possible for a generic function. But this still applies for unpacking too, when the generated code will have to do `as_type<uchar4>(x)` and whether it casts to char4 or uchar4 is depending on the context, again.

Still, sure, I could add something simple like this:
```cpp
template <typename T>
T sdot(vec<T, N> vec1, vec<T, N> vec2) {
    return reduce_add(vec1 * vec2);
}
```
but that would cover a single case. Using functions would require us to write many different generic functions to handle the many different scenarios given. So, apart from added code size and more templates for the Metal compiler to go through, what does this gain over the inline implementation? I don't think it's a good idea.

spirv_msl.cpp
cdavis5e: Nah, just a regular functional cast will do.

spnda: I'm still not sure I can wrap my head around the entirety of this issue tbh. However, I've changed the code to always cast to the result type but in the case of OpSDot and the first vector of OpSUDot it'll cast to the signed version thereof. I think that solves the issue?

spirv_msl.cpp
```diff
@@ -9626,6 +9656,75 @@ void CompilerMSL::emit_instruction(const Instruction &instruction)
...
+		{
+			string type = opcode == OpSDot || opcode == OpSUDot ? "char4" : "uchar4";
+			vec1input = join("as_type<", type, ">(", to_expression(vec1), ")");
+		} else
```
cdavis5e, line 9676: Minor note, but in this project, the style is to put `if`/`else` on different lines from the braces.

If you haven't already, you should install `clang-format(1)`. It comes with a Git extension command (`git-clang-format(1)`) that can run `clang-format` over only your changes.

spnda: CLion already auto formats using clang-format for me generally. I already changed this earlier with the first block of code, but didn't in the second.

spirv_msl.cpp
```diff
@@ -9626,6 +9642,109 @@ void CompilerMSL::emit_instruction(const Instruction &instruction)
...
+	case OpSDot:
+	case OpUDot:
+	case OpSUDot:
+	{
+		uint32_t result_type = ops[0];
+		uint32_t id = ops[1];
+		uint32_t vec1 = ops[2];
+		uint32_t vec2 = ops[3];
+
+		auto& input_type1 = expression_type(vec1);
```
HansKristian-Work, line 9654:
```suggestion
		auto &input_type1 = expression_type(vec1);
```

spirv_msl.cpp
```diff
@@ -9626,6 +9642,109 @@ void CompilerMSL::emit_instruction(const Instruction &instruction)
...
+			else
+				SPIRV_CROSS_THROW("Packed vector formats other than 4x8Bit for integer dot product is not supported.");
+		}
+		else
+		{
+			vec1input = to_expression(vec1);
+			vec2input = to_expression(vec2);
+		}
+
+		auto& type = get<SPIRType>(result_type);
```
HansKristian-Work, line 9677:
```suggestion
		auto &type = get<SPIRType>(result_type);
```

spirv_msl.cpp
```diff
@@ -9626,6 +9642,109 @@ void CompilerMSL::emit_instruction(const Instruction &instruction)
...
+		auto result_type_cast = join(type_to_glsl(type), input_size);
+
+		// When the opcode specifies signed integers, we always cast to the signed integer type, regardless of the output type.
+		string_view type_cast1(result_type_cast);
```
HansKristian-Work, line 9681:
```suggestion
		string type_cast1 = result_type_cast;
```

spirv_msl.cpp
```diff
@@ -9626,6 +9642,109 @@ void CompilerMSL::emit_instruction(const Instruction &instruction)
...
+		if (type_cast1[0] == 'u' && (opcode == OpSDot || opcode == OpSUDot))
+			type_cast1 = type_cast1.substr(1);
+
+		string_view type_cast2(result_type_cast);
```
HansKristian-Work, line 9685:
```suggestion
		string type_cast2 = result_type_cast;
```

spirv_msl.cpp
```diff
@@ -9626,6 +9642,109 @@ void CompilerMSL::emit_instruction(const Instruction &instruction)
...
+	case OpUDotAccSat:
+	case OpSUDotAccSat:
+	{
+		uint32_t result_type = ops[0];
+		uint32_t id = ops[1];
+		uint32_t vec1 = ops[2];
+		uint32_t vec2 = ops[3];
+		uint32_t acc = ops[4];
+
+		auto& input_type1 = expression_type(vec1);
```
HansKristian-Work, line 9706:
```suggestion
		auto &input_type1 = expression_type(vec1);
```

spirv_msl.cpp
```diff
@@ -9626,6 +9642,109 @@ void CompilerMSL::emit_instruction(const Instruction &instruction)
...
+		auto& input_type2 = expression_type(vec2);
```
HansKristian-Work, line 9707:
```suggestion
		auto &input_type2 = expression_type(vec2);
```

spirv_msl.cpp
```diff
@@ -9626,6 +9642,109 @@ void CompilerMSL::emit_instruction(const Instruction &instruction)
...
+			else
+				SPIRV_CROSS_THROW("Packed vector formats other than 4x8Bit for integer dot product is not supported.");
+		}
+		else
+		{
+			vec1input = to_expression(vec1);
+			vec2input = to_expression(vec2);
+		}
+
+		auto& type = get<SPIRType>(result_type);
```
HansKristian-Work, line 9730:
```suggestion
		auto &type = get<SPIRType>(result_type);
```

spirv_msl.cpp
```diff
@@ -9626,6 +9642,109 @@ void CompilerMSL::emit_instruction(const Instruction &instruction)
...
+		auto result_type_cast = join(type_to_glsl(type), input_size);
+
+		string_view type_cast1(result_type_cast);
```
HansKristian-Work, line 9733:
```suggestion
		string type_cast1 = result_type_cast;
```

spirv_msl.cpp
```diff
@@ -9626,6 +9642,109 @@ void CompilerMSL::emit_instruction(const Instruction &instruction)
...
+		if (type_cast1[0] == 'u' && (opcode == OpSDotAccSat || opcode == OpSUDotAccSat))
+			type_cast1 = type_cast1.substr(1);
+
+		string_view type_cast2(result_type_cast);
```
HansKristian-Work, line 9737:
```suggestion
		string type_cast2 = result_type_cast;
```

spirv_msl.cpp
```diff
@@ -9626,6 +9642,109 @@ void CompilerMSL::emit_instruction(const Instruction &instruction)
...
+		if (type_cast2[0] == 'u' && opcode == OpSDotAccSat)
+			type_cast2 = type_cast2.substr(1);
+
+		string exp = join("addsat(reduce_add(", std::string(type_cast1), "(", vec1input, ") * ", std::string(type_cast2), "(", vec2input, ")), ", to_expression(acc), ")");
```
HansKristian-Work, line 9741:
```suggestion
		string exp = join(result_type_cast, "(addsat(reduce_add(", std::string(type_cast1), "(", vec1input, ") * ", std::string(type_cast2), "(", vec2input, ")), ", to_expression(acc), "))");
```

HansKristian-Work (changes requested):

HansKristian-Work:
Merged this locally. Fixing misc issues locally while adding more test coverage as I go.

HansKristian-Work:
On master.

PR #2267, opened 2024-01-17
Added support ccache for speedup compilation on Linux

GermanAizek:
@HansKristian-Work more information:
https://ccache.dev/
https://en.wikipedia.org/wiki/Ccache

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -88,6 +88,12 @@ else()
 	message(STATUS "SPIRV-Cross: Git not found, using unknown build version.")
 endif()
 
+find_program(CCACHE_FOUND ccache)
+if (CCACHE_FOUND)
+    set_property(GLOBAL PROPERTY RULE_LAUNCH_COMPILE ccache)
+    set(ENV{CCACHE_SLOPPINESS} pch_defines,time_macros)
+endif ()
+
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -88,6 +88,12 @@ else()
 	message(STATUS "SPIRV-Cross: Git not found, using unknown build version.")
 endif()
 
+find_program(CCACHE_FOUND ccache)
+if (CCACHE_FOUND)
+    set_property(GLOBAL PROPERTY RULE_LAUNCH_COMPILE ccache)
```
cdavis5e, line 93: After the `find_program()` above, the variable `CCACHE_FOUND` will contain the full pathname of the `ccache` program. You should probably use it, instead of assuming that `ccache` will be in the user's `PATH`.

cdavis5e (changes requested):
Historically, CMake build systems that use `ccache(1)` automatically have caused problems for me, e.g. when installing ports from source in MacPorts. Then again, I suppose it's easy enough to work around that with `-DCCACHE_FOUND=NO`.

HansKristian-Work:
Is there really a point adding this? This isn't a massive project where ccache really helps.

PR #2275, opened 2024-02-03
Raytracing support for SPIR-V to HLSL

calper-ql:


calper-ql:
Selected the wrong upstream :/ sorry..

PR #2292, opened 2024-03-07
MSL: Add support for overlapping bindings.

js6i:
This adds support for bindings which share the same DescriptorSet/Binding pair.

The motivating example is vkd3d, which uses overlapping arrays of resources to emulate D3D12 descriptor tables. The generated MSL argument buffer only includes the first resource (in this example 't0'):
```
struct spvDescriptorSetBuffer2
{
    array<texture2d<float>, 499968> t0 [[id(0)]];
    // Overlapping binding: array<texture3d<float>, 499968> t2 [[id(0)]];
};
```
When t2 is referenced, we cast the instantiated member:
```
float4 r1 = spvDescriptorSet2.t0[_79].sample(...);
float4 r2 = (*(constant array<texture3d<float>, 499968>*)&spvDescriptorSet2.t0)[_97].sample(...);
```

### DIFF
--- a/reference/opt/shaders-msl/comp/overlapping-bindings.msl31.argument.argument-tier-1.decoration-binding.device-argument-buffer.texture-buffer-native.comp
+++ b/reference/opt/shaders-msl/comp/overlapping-bindings.msl31.argument.argument-tier-1.decoration-binding.device-argument-buffer.texture-buffer-native.comp
@@ -1,0 +1,35 @@
+#include <metal_stdlib>
+#include <simd/simd.h>
+
+using namespace metal;
+
+constant uint3 gl_WorkGroupSize [[maybe_unused]] = uint3(1u);
+
+struct spvDescriptorSetBuffer0
+{
+    array<texture_buffer<uint, access::write>, 8> u0 [[id(0)]];
+};
+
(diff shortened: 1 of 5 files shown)


cdavis5e:
Honestly, I'm just amazed the compiler will let you do this, and yet _won't_ let you use unions in an argument buffer struct.

js6i:
Yeah..
I don't understand what the CI compiler is complaining about. It builds cleanly here with arm64-apple-darwin23.3.0 and gcc 13.2.0-4ubuntu3.

HansKristian-Work:
> argument_buffer_device_storage_mask

That one should be handled explicitly on the outside I think, since it's not part of type system at all. When you emit the alias you look up descriptor set and check which address space it has.

> Are those the kind of tests you want to see?

Yes.

> Translates to this:

Looks fine to me?

js6i:
> Looks fine to me?

You're right, for some reason I assumed the [[]] annotation always goes at the end..

js6i:
I added a couple hacks that probably aren't great.

HansKristian-Work:
Merged locally and improving things before merging upstream. Think I figured out how to make this neater.

PR #2297, opened 2024-03-15
Support cases of LUTs which are not function local.

js6i:
Prevent cases where arrays that are globally defined constants are redeclared on stack. On Intel macs, declaring a large, statically initialized spvUnsafeArray on stack may cause an internal compiler error.


HansKristian-Work:
Missing test. Also, isn't this a case where NonWritable decoration is used? A global variable that is never written to only makes sense if it has a NonWritable + initializer.

HansKristian-Work:
I think I get what this is trying to do now. We already have code in place to handle Private LUT for CFGs with one function, but this extends that idea to multi-function.

PR #2304, opened 2024-04-01
Use recursive template for spvArrayCopy

etang-cw:
I realize it's unlikely to have a >6-deep array, but the limitation was still bothering me

There's a comment on the old code saying you couldn't use recursive templates, but I couldn't get the recursive template to break on anything I tried, including Xcode 8 targeting macOS 10.11, macOS 10.13's online compiler, or Xcode 14 targeting iOS 8.  Do you remember why you originally put in that comment?  If so, maybe we should add where it fails to that comment so it can be removed in the future if support for whatever it fails on is ever dropped.


HansKristian-Work (approved):
Seems fine to me.

PR #2306, opened 2024-04-03
MSL: Implement support for EXT_mutable_descriptor_type and general aliasing with argument buffers

HansKristian-Work:
Here be dragons, illegal pointer casts, and all sorts of fiendish constructions.

Verified that MSL codegen should work. texture types are represented as pointer to internal texture type in AIR.

Supersedes #2292.


billhollings:
> Here be dragons

Nice one. Gave me a chuckle. 😉

Issue #2198, opened 2023-09-13
[mvk-error] SPIR-V to MSL conversion error: Argument buffer resource base type could not be determined.


pjako:
_[I was told to repost this](https://github.com/KhronosGroup/MoltenVK/issues/2016#issuecomment-1716428535)_
After I updated to the latest version of MVK I git the following error:

[mvk-error] SPIR-V to MSL conversion error: Argument buffer resource base type could not be determined. When padding argument buffer elements, all descriptor set resources must be supplied with a base type by the app.
[mvk-error] VK_ERROR_INVALID_SHADER_NV: Fragment shader function could not be compiled into pipeline. See previous logged error.`

The shader is written in hlsl, here are the relevent inputs:
[[vk::push_constant]]
struct rx_pushConstants {
uint instanceRefIndicies[6];
} rx_pushConstants;

[[vk::binding(0, 0)]] ByteAddressBuffer rx_staticResBlocks;
[[vk::binding(1, 0)]] ByteAddressBuffer rx_dynamicResBlocks;
[[vk::binding(2, 0)]] Texture2D rx_textures[];
[[vk::binding(3, 0)]] SamplerState rx_samplers[];

I use dxc to compile hlsl to spirv and MVK_CONFIG_USE_METAL_ARGUMENT_BUFFERS = 1

SpirvCross may does not like the ByteAddressBuffer? Is there a way around this problem without breaking my code?

Thanks for your help!

tkchanat:
I am having a similar problem with StructuredBuffer. Could it be related to `CompilerMSL::add_msl_resource_binding` failing to recognize their base type?

pjako:
> I am having a similar problem with StructuredBuffer. Could it be related to CompilerMSL::add_msl_resource_binding failing to recognize their base type?

Looks like it

rcoreilly:
I'm experiencing the same problem, specifically for use of a StructuredBuffer -- other tests without that work fine.  The sdk release version 261.1 works fine, but 268.1 breaks, so hopefully that helps to identify what changed to cause the problem?

rcoreilly:
[...]

PR #2320, opened 2024-05-06
MSL: Fix argument buffer padding.

billhollings:
- After padding a resource, retrieve either the next padded slot, or the next resource, in order to correctly offset the slot index.

Fixes issue #2198.

### DIFF
--- a/spirv_msl.cpp
+++ b/spirv_msl.cpp
@@ -18291,11 +18291,14 @@ void CompilerMSL::analyze_argument_buffers()
 					default:
 						break;
 					}
+
+					// After padding, retrieve the resource again. It will either be more padding, or the actual resource.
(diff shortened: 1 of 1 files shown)


HansKristian-Work:
I cherry-picked this and pushed to main.

billhollings:
> I cherry-picked this and pushed to main.

Thanks! 

I'll move MoltenVK to that for the SDK release. 

BTW...for future reference, although nice, it's not always necessary for you to rush to do that prior to an SDK release. MoltenVK retrieves based on a commit hash, so the branch hash is just as easy for MoltenVK to pull in as a `main` hash. The important thing of course is to make sure the branch changes (or something with the same functionality) make it into `main` at some point shortly afterwards though.

Issue #2322, opened 2024-05-10
MSL: Option to enforce fragment execution when side effects are present before discard


aitor-lunarg:
Details can be found here: https://github.com/KhronosGroup/MoltenVK/issues/2233

In short, Metal will incorrectly discard fragments with side effects under certain circumstances prematurely even when the side effect takes place before discard. The conditions are the following:
- Fragment will always be discarded
- Pre fragment depth fails fails
- Modifies depth value for a constant value in the fragment shader. This constant value will also fail the depth test.

To circumvent this issue, we need to trick Metal into thinking that not all fragment executions may be discarded

HansKristian-Work:
Fixed on main.

PR #2323, opened 2024-05-10
Add option to enforce fragment execution with side effects in MSL

aitor-lunarg:
Metal will incorrectly discard fragments with side effects under certain circumstances prematurely. The conditions are the following:
 - Fragment will always be discarded
 - Pre fragment depth fails fails
 - Modifies depth value for a constant value in the fragment shader. This constant value will also fail the depth test.

However, Metal will also discard the fragment even if it has operations with side effects inside the fragment shader before the discard operation.

Vulkan states the graphics pipeline to execute in the following order:
 - Pre fragment depth test (cannot discard here due to modifying depth value in fragment shader)
 - Fragment shader (where the depth is modified)
 - Post fragment depth test (where fragment should be discarded)

Therefore, we need to enforce fragment shader execution and not let Metal discard the fragment before that for such cases. This change adds an option to provide such utility.

Fixes https://github.com/KhronosGroup/SPIRV-Cross/issues/2322

If there's a better approach to accomplish this, I'm happy to investigate!


aitor-lunarg:
Closing in favor of a new upcoming PR

PR #2324, opened 2024-05-20
Add option to enforce fragment execution with side effects in MSL

### LINKED ISSUE #2322: MSL: Option to enforce fragment execution when side effects are present before discard

aitor-lunarg:
Fixes #2322

Metal will incorrectly discard fragments with side effects under certain circumstances prematurely. The conditions are the following:
 - Fragment will always be discarded after side effect operation
 - Pre fragment depth fails
 - Modifies depth value for a constant value in the fragment shader. This constant value will also fail the depth test.

However, Metal will also discard the fragment even if it has operations with side effects inside the fragment shader before the discard operation.

Vulkan states the graphics pipeline to execute in the following order:
 - Pre fragment depth test (cannot discard here due to modifying depth value in fragment shader)
 - Fragment shader (where the depth is modified and fragment discarded)
 - Post fragment depth test

Therefore, we need to enforce fragment shader execution and not let Metal discard the fragment before that for such cases. This change adds an option to provide such utility.


HansKristian-Work:
Merged to master.

HansKristian-Work:
Super cursed that we have to add a million workarounds for their broken shader compiler though ._.

PR #2325, opened 2024-05-20
MSL: Support ConstOffsets on image gather.

billhollings:
- Add `TextureFunctionArguments::has_array_offsets`.
- Add support functions `spvGatherConstOffsets()` & `spvGatherCompareConstOffsets()`.
- Add `SPVFuncImpl::SPVFuncImplGatherConstOffsets` & `SPVFuncImplGatherCompareConstOffsets`.
- Add unit tests.

@cdavis5e Please review too.

### DIFF
--- a/reference/opt/shaders-msl/frag/gather-compare-const-offsets.frag
+++ b/reference/opt/shaders-msl/frag/gather-compare-const-offsets.frag
@@ -1,0 +1,90 @@
+#pragma clang diagnostic ignored "-Wmissing-prototypes"
+#pragma clang diagnostic ignored "-Wmissing-braces"
+
+#include <metal_stdlib>
+#include <simd/simd.h>
+
+using namespace metal;
+
+template<typename T, size_t Num>
+struct spvUnsafeArray
+{
+    T elements[Num ? Num : 1];
+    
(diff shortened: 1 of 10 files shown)


cdavis5e (approved):
I still think doing four separate gathers feels like too much work, since 75% of the work will be thrown away. But, I don't see a better way forward here.

Just one small nitpick...

billhollings:
> I still think doing four separate gathers feels like too much work, since 75% of the work will be thrown away. But, I don't see a better way forward here.

I agree. It seems a waste. But as you've pointed out elsewhere, it needs to be a sample, rather than a read, to get the behavior at the edges right. FWIW...CTS actually uses direct reads when creating the reference image on the CPU, but has logic to handle the address modes, and -0.5-pixel skew when determining the correct pixel to read.

I supposed `spvGatherConstOffsets()` could do a bounds check on the coord value, and, if it's within bounds, use 4 reads, instead of gathers. Perhaps a future enhancement. But, on the other hand, gathers should be fairly efficient, cache-wise, so it might not cost much more to pull in 4 pixel values compared to 1. And functionally, we're in pretty niche-territory anyway.

PR #2328, opened 2024-05-21
Dynamic buffer cleanup

HansKristian-Work:
This was never legal in any API and I'm not sure why I ever bothered.


cdavis5e:
Actually, I believe that was me. I remember writing that comment...

PR #2331, opened 2024-05-21
MSL: Add option to force depth write in fragment shaders

aitor-lunarg:
Metal writes to the depth/stencil attachment before fragment shader execution if the execution does not modify the depth value. However, Vulkan expects the write to happen after fragment shader execution. To circumvent the issue we add a simple depth passthrough if the user opts in. Only required when the depth/stencil attachment is used as input attachment at the same time. It seems Metal does not correctly detect the dependency.

Required for https://github.com/KhronosGroup/MoltenVK/issues/2232

### DIFF
--- a/spirv_msl.hpp
+++ b/spirv_msl.hpp
@@ -529,6 +529,13 @@ public:
 		// with side effects. Provided as an option hoping Metal will fix this issue in the future.
 		bool force_fragment_with_side_effects_execution = false;
 
+		// If set, adds a depth pass through statement to circumvent the following issue:
+		// When the same depth/stencil is used as input and depth/stencil attachment, we need to
+		// force Metal to perform the depth/stencil write after fragment execution. Otherwise,
+		// Metal will write to the depth attachment before fragment execution. This happens
+		// if the fragment does not modify the depth value.
+		bool input_attachment_is_ds_attachment = false;
+
 		bool is_ios() const
 		{
 			return platform == iOS;
@@ -1094,6 +1101,7 @@ protected:
 	uint32_t builtin_stage_input_size_id = 0;
 	uint32_t builtin_local_invocation_index_id = 0;
 	uint32_t builtin_workgroup_size_id = 0;
+	uint32_t builtin_frag_depth_id = 0;
 	uint32_t swizzle_buffer_id = 0;
 	uint32_t buffer_size_buffer_id = 0;
 	uint32_t view_mask_buffer_id = 0;
@@ -1190,6 +1198,7 @@ protected:
 	bool needs_subgroup_size = false;
 	bool needs_sample_id = false;
 	bool needs_helper_invocation = false;
+	bool writes_to_depth = false;
 	std::string qual_pos_var_name;
 	std::string stage_in_var_name = "in";
 	std::string stage_out_var_name = "out";
--- a/main.cpp
+++ b/main.cpp
@@ -681,6 +681,7 @@ struct CLIArguments
 	bool msl_replace_recursive_inputs = false;
 	bool msl_readwrite_texture_fences = true;
 	bool msl_agx_manual_cube_grad_fixup = false;
+	bool msl_input_attachment_is_ds_attachment = false;
 	const char *msl_combined_sampler_suffix = nullptr;
 	bool glsl_emit_push_constant_as_ubo = false;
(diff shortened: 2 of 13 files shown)


spirv_msl.hpp
```diff
@@ -529,6 +529,13 @@ class CompilerMSL : public CompilerGLSL
 		// with side effects. Provided as an option hoping Metal will fix this issue in the future.
 		bool force_fragment_with_side_effects_execution = false;
 
+		// If set, adds a depth pass through statement to circumvent the following issue:
+		// When the same depth/stencil is used as input and depth/stencil attachment, we need to
+		// force Metal to perform the depth/stencil write after fragment execution. Otherwise,
+		// Metal will write to the depth attachment before fragment execution. This happens
+		// if the fragment does not modify the depth value.
+		bool input_attachment_is_ds_attachment = false;
```
HansKristian-Work, line 537: How does this interact with early-fragment-tests? It should be possible to use early-fragment-tests and reading depth as input attachment where there are no writes.

Also, I'm a little confused. Why can there be a hazard here? It should not be possible to read depth as a texture without flushing the render pass first. Which GPUs are affected by this?

HansKristian-Work (changes requested):

spirv_msl.hpp
aitor-lunarg: > How does this interact with early-fragment-tests?

That's a good question for which I don't have an answer for at the moment. I need to verify how Metal behaves with the early-fragment-tests option, since the CTS test variants this fix is for does not test early-fragment-tests. There's 2 possible situations that come to mind:
 1. This change is not required for early-fragment-tests. Easy to handle with a check.
 2. Same issue is present with early-fragment-tests. This would mean that we may have to disable early-fragment-tests in the shader to enforce this fix if the user opted in.

> Why can there be a hazard here?

Maybe I didn't explain myself correctly, let me provide a quick explanation on the investigation I did. Let me know if there's something I missed:

The CTS tests mentioned in the MoltenVK issue do the following:
 1. vkCmdDraw -> Renders to color and depth
 2. vkCmdPipelineBarrier (with correct parameters to flush the attachment between post fragment depth test stage to next draw calls pre fragment depth test stage. Post fragment depth test stage is where the depth write happens according to Vulkan)
 3. vkCmdDraw -> Reads from input attachment (which is also the depth attachment) and renders to color and depth

These tests fail with Metal. From my investigation and conclusions which may not be 100% accurate due to not having visibility on Metal drivers, I believe they write to the depth before fragment execution. Unsure if this is Metal trying to be smart and optimize depth operations to pre fragment execution, or if there's any other thing going on behind the scenes. The depth values read from the input attachment are the ones that should be post draw call 3 instead of being the post draw call 1. These same tests do not fail if we do the same with a color attachment.

I initially thought the tests were not valid, but after going in circles for some time I reached the conclusion they were valid: https://gitlab.khronos.org/Tracker/vk-gl-cts/-/issues/5087

> Which GPUs are affected by this?

Tested with M3 (unsure if other chip families have this issue)

HansKristian-Work: You're sampling from a texture here, so how is it possible that you end up sampling a depth value that was written? The tile has not been flushed out to memory yet, and Metal does not support depth-stencil fetch from tile iirc.

HansKristian-Work: What do you do on a subpass barrier or BY_REGION self barrier for input + depth? Do you flush the render pass and start it again?

aitor-lunarg: > Metal does not support depth-stencil fetch from tile iirc.

It does not, it's only supported for color attachments.

> What do you do on a subpass barrier or BY_REGION self barrier for input + depth? Do you flush the render pass and start it again?

MoltenVK will restart the render pass, so in the CTS test case, whenever there's a barrier, the current render pass will be terminated and a new one started. Then Metal is in charge of the synchronization (execution in record order and dependencies checked by Metal), which will flush the previous render pass before starting the next.

> depth write -> image is updated -> fragment shader is run

I believe we reached the same conclusion `Metal trying to be smart and optimize depth operations to pre fragment execution`. But without visibility on Metal drivers is hard be sure 100%, there may be something else missing.

HansKristian-Work (approved):

PR #2339, opened 2024-06-10
MSL: Fixes for constexpr sampler use with argument buffers.

billhollings:
Constexpr samplers are defined as local variables, but were treated as held within an argument buffer.

- `CompilerMSL::to_sampler_expression()` support constexpr samplers when using argument buffers, plus refactor to minimize generating expression text that may not be used.

- Handle padding around multi-plane images that require multiple textures. Only check for padding on the first plane, but include plane count in total argument buffer slots consumed. 
  
I experimented with a unit test, but couldn't figure out how to add the ability to define a constexpr sampler into the API in a straightforward manner.


HansKristian-Work (approved):
LGTM

PR #2343, opened 2024-06-14
MSL: Support descriptor sets with recursive content when using argument buffers.

billhollings:
When using argument buffers, handle descriptor set entry points with recursive content, similar to discrete entry points with recursive content.

- For descriptor sets entry points with recursive content, add descriptor set to recursive_inputs, and create a local var for it.

- For recursive entry points that are contained in a descriptor set argument buffer, don't add entry point to recursive_inputs, or create a local var for that content entry point.

- Add test shader.

### DIFF
--- a/spirv_msl.cpp
+++ b/spirv_msl.cpp
@@ -14059,8 +14059,8 @@ void CompilerMSL::fix_up_shader_inputs_outputs()
 			}
 		}
 
-		if ( !msl_options.argument_buffers &&
-			msl_options.replace_recursive_inputs && type_contains_recursion(type) &&
+		if (!msl_options.argument_buffers &&
+		     msl_options.replace_recursive_inputs && type_contains_recursion(type) &&
 		    (var.storage == StorageClassUniform || var.storage == StorageClassUniformConstant ||
 		     var.storage == StorageClassPushConstant || var.storage == StorageClassStorageBuffer))
 		{
@@ -18578,7 +18578,6 @@ void CompilerMSL::analyze_argument_buffers()
 				statement(addr_space, " auto& ", buffer_name, " = *(", addr_space, " ", buffer_type_name, "*)", buffer_name, "_vp;");
 			});
 		}
-
 	}
 }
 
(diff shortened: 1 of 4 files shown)


spirv_msl.cpp
```diff
@@ -14053,7 +14059,8 @@ void CompilerMSL::fix_up_shader_inputs_outputs()
 			}
 		}
 
-		if (msl_options.replace_recursive_inputs && type_contains_recursion(type) &&
+		if ( !msl_options.argument_buffers &&
+			msl_options.replace_recursive_inputs && type_contains_recursion(type) &&
```
HansKristian-Work, line 14063:
```suggestion
		if (!msl_options.argument_buffers &&
		     msl_options.replace_recursive_inputs && type_contains_recursion(type) &&
```

spirv_msl.cpp
```diff
@@ -18559,6 +18568,17 @@ void CompilerMSL::analyze_argument_buffers()
...
+		if (msl_options.replace_recursive_inputs && type_contains_recursion(buffer_type))
+		{
+			recursive_inputs.insert(type_id);
+			auto &entry_func = this->get<SPIRFunction>(ir.default_entry_point);
+			auto addr_space = get_argument_address_space(buffer_var);
+			entry_func.fixup_hooks_in.push_back([this, addr_space, buffer_name, buffer_type_name]() {
+				statement(addr_space, " auto& ", buffer_name, " = *(", addr_space, " ", buffer_type_name, "*)", buffer_name, "_vp;");
+			});
+		}
+
```
HansKristian-Work, line 18581:
```suggestion
		}
```

spirv_msl.cpp
```diff
@@ -18559,6 +18568,17 @@ void CompilerMSL::analyze_argument_buffers()
...
 	}
```
HansKristian-Work, line 18582:
```suggestion
			});
		}
	}
```

HansKristian-Work (approved):
LGTM. Another silly workaround for the pile :(

PR #2329, opened 2024-05-21
MSL: Improve handling of padded descriptors with argument buffers

HansKristian-Work:
Fix some issues with aliasing, and also ensures that declared array sizes always match the layout.

Fix #2318.

PR #2347, opened 2024-06-19
MSL: Support a runtime array with dynamic offset in an argument buffer.

billhollings:
- Modify `get_resource_array_size()` to retrieve size of descriptors in an argument buffer from resource_bindings (reversing change made in b236352).
- Retrieve size of runtime array from `get_resource_array_size()`.

@HansKristian-Work This PR reverses the b236352 commit from PR #2329. It was not clear why the restriction from that commit is required (what is the ABI referenced there?), and it means that variables held within an argument buffer descriptor set cannot have their sizes established by the app. This is particularly troublesome for SPIR-V runtime arrays, whose size must be set by the app.

Several hundred CTS tests use a runtime array in SPIR-V to read from a fixed-size array defined in a Vulkan descriptor set.

If the restriction in b236352 is still required, perhaps this PR could add a parameter to `msl_options` to control when is it required.

### DIFF
--- a/spirv_msl.cpp
+++ b/spirv_msl.cpp
@@ -1361,14 +1361,14 @@ void CompilerMSL::emit_entry_point_declarations()
 
 		if (is_array(type))
 		{
-			if (!type.array[type.array.size() - 1])
-				SPIRV_CROSS_THROW("Runtime arrays with dynamic offsets are not supported yet.");
-
 			is_using_builtin_array = true;
 			statement(get_argument_address_space(var), " ", type_to_glsl(type), "* ", to_restrict(var_id, true), name,
 			          type_to_array_glsl(type, var_id), " =");
 
-			uint32_t array_size = to_array_size_literal(type);
+			uint32_t array_size = get_resource_array_size(type, var_id);
+			if (array_size == 0)
+				SPIRV_CROSS_THROW("Size of runtime array with dynamic offset could not be determined from resource bindings.");
+
 			begin_scope();
 
 			for (uint32_t i = 0; i < array_size; i++)


spirv_msl.cpp
```diff
@@ -201,17 +201,14 @@ bool CompilerMSL::is_var_runtime_size_array(const SPIRVariable &var) const
...
+	if (array_size)
 		return array_size;
 
-	StageSetBinding tuple = { get_entry_point().model, desc_set,
+	StageSetBinding tuple = { get_entry_point().model,
+		                      get_decoration(id, DecorationDescriptorSet),
 		                      get_decoration(id, DecorationBinding) };
 	auto itr = resource_bindings.find(tuple);
-	return itr != end(resource_bindings) ? itr->second.first.count : array_size;
+	return itr != end(resource_bindings) ? itr->second.first.count : 0;
```
HansKristian-Work, line 211: IIRC, the problem here is mutable descriptor types or aliases where two aliases have different descriptor counts. If I recall correctly, the code will be screwed up if you try to alias resources with different array sizes. I don't understand what this revert is accomplishing.

A runtime array where VkPipelineLayout has a non-variable count must always be allocated with the size of the descriptor set layout. It should always be safe to use that array size in the shader.
For variable count, there is no such problem.

HansKristian-Work: > Several hundred CTS tests use a runtime array in SPIR-V to read from a fixed-size array defined in a Vulkan descriptor set.

Yes, and existing code will resize that runtime array in SPIR-V to match the pipeline layout. If anything, it should be more robust, since our [1] unsized array hack is technically UB.

HansKristian-Work: If you intended to revert, can you make that a proper revert-commit? That way I don't have to review extra to see if it's a true revert.

billhollings: You are correct, sir. Existing code works.

Somewhere in my testing, I got mixed up with `get_resource_array_size()`, but it was misdirected.

I've force-pushed the reversion of my changes there.

HansKristian-Work (approved):
LGTM

spirv_msl.cpp
HansKristian-Work: The new commit is good, ye. Thanks for verifying.

PR #2351, opened 2024-07-10
Stale access permissions for overlapping descriptors

js6i:


js6i:
Looks like there's a dependency in the other direction after all in the other tests sets I forgot about.. I'll look for a less invasive solution, maybe a force_recompile() will do.

spirv_msl.cpp
```diff
@@ -18167,6 +18167,9 @@ void CompilerMSL::emit_argument_buffer_aliased_descriptor(const SPIRVariable &al
 	}
 	else
 	{
+		if (ir.meta[aliased_var.self].decoration.qualified_alias != name)
```
HansKristian-Work, line 18170:
```suggestion
		// This alias may have already been used to emit an entry point declaration. If there is a mismatch, we need a recompile.
		// Moving this code to be run earlier will also conflict,
		// because we need the qualified alias for the base resource,
		// so forcing recompile until things sync up is the least invasive method for now.
		if (ir.meta[aliased_var.self].decoration.qualified_alias != name)
```

HansKristian-Work:
I tried a cleaner method, but that doesn't work either, since we depend on the qualified alias of the base resource. Guess this is the best we can do for now.

HansKristian-Work (approved):

PR #2360, opened 2024-08-05
Fix volatile helper invocation in non-main functions

etang-cw:
One of our tests only passed because it operated out of the entry point and would fail if you tried to do the same thing from any other function.  This makes it work no matter where it is.

~~Note: I couldn't think of any better names for the new variable than just adding an underscore.  If anyone else has a better name, please use it.~~

### DIFF
--- a/reference/shaders-msl-no-opt/frag/volatile-helper-invocation.msl23.spv16.frag
+++ b/reference/shaders-msl-no-opt/frag/volatile-helper-invocation.msl23.spv16.frag
@@ -1,3 +1,5 @@
+#pragma clang diagnostic ignored "-Wmissing-prototypes"
+
 #include <metal_stdlib>
 #include <simd/simd.h>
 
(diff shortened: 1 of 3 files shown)


spirv_msl.cpp
```diff
@@ -271,13 +271,14 @@ void CompilerMSL::build_implicit_builtins()
 	     active_input_builtins.get(BuiltInInstanceIndex) || active_input_builtins.get(BuiltInBaseInstance));
 	bool need_local_invocation_index = msl_options.emulate_subgroups && active_input_builtins.get(BuiltInSubgroupId);
 	bool need_workgroup_size = msl_options.emulate_subgroups && active_input_builtins.get(BuiltInNumSubgroups);
+	bool needs_helper_invocation_ = needs_helper_invocation || active_input_builtins.get(BuiltInHelperInvocation);
```
HansKristian-Work, line 274: Why do you need to create a new variable here? I would expect that needs_helper_invocation could just be set to true if active_input_builtins has HelperInvocation. What's the rationale for doing it this way?

reference/shaders-msl-no-opt/frag/volatile-helper-invocation.msl23.spv16.frag
```diff
@@ -8,18 +10,24 @@ struct main0_out
     float FragColor [[color(0)]];
 };
 
+static inline __attribute__((always_inline))
+void func(thread float& FragColor, volatile thread bool& gl_HelperInvocation)
```
HansKristian-Work, line 14: volatile thread seems like it's asking for trouble with spilling. volatile thread is somewhat meaningless. I know this comes up because in later SPIR-V discard + helper lane is done through volatile StorageClassInput.

HansKristian-Work (changes requested):
Nits.

spirv_msl.cpp
etang-cw: Mostly just me being scared to touch global variables.

Looked through all the uses and almost all of them used `needs_helper_invocation || active_input_builtins.get(BuiltInHelperInvocation)` anyways, so I switched it over and cleaned those up while I was at it.

reference/shaders-msl-no-opt/frag/volatile-helper-invocation.msl23.spv16.frag
etang-cw: Do you think we should just avoid `volatile` on `thread` variables in general?

HansKristian-Work (approved):
Seems okay.

PR #2377, opened 2024-09-08
GLSL: Convert switch blocks to if statements in legacy desktop versions

flafflar:
Legacy versions of GLSL do not support switch statements, but the compiler did not take this into account and produced switch statements in versions before 1.30. Since the compiler already supported converting switch statements to ifs for the legacy ES versions, I just replaced the `is_legacy_es()` checks with `is_legacy()` checks.

### DIFF
--- a/spirv_glsl.cpp
+++ b/spirv_glsl.cpp
@@ -17608,7 +17608,7 @@ void CompilerGLSL::emit_block_chain(SPIRBlock &block)
 
 		if (!collapsed_switch)
 		{
-			if (block_like_switch || is_legacy_es())
+			if (block_like_switch || is_legacy())
 			{
 				// ESSL 1.0 is not guaranteed to support do/while.
 				if (is_legacy_es())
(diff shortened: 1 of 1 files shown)


spirv_glsl.cpp
```diff
@@ -17712,7 +17712,7 @@ void CompilerGLSL::emit_block_chain(SPIRBlock &block)
 
 		if (!collapsed_switch)
 		{
-			if (block_like_switch && !is_legacy_es())
+			if ((block_like_switch || is_legacy()) && !is_legacy_es())
```
HansKristian-Work, line 17715: This is weird. Can you explain why this isn't just `if (block_like_switch && !is_legacy())`?

HansKristian-Work (changes requested):
comment

spirv_glsl.cpp
flafflar: In legacy ES, do/while loops are not supported, so a for loop is inserted around the if statements. In legacy desktop, a do/while loop is inserted. Therefore, if the version is a legacy one AND not ES, a `while(false)` needs to be inserted.

HansKristian-Work (approved):

PR #2379, opened 2024-09-11
Reduce recompile count with texture atomics

etang-cw:
Avoid waiting until a run where we're not forcing a recompile to force another recompile

Prevents going over the maximum recompile count on [shaders like this](https://github.com/user-attachments/files/16966591/recompile.spv.zip)

### DIFF
--- a/spirv_msl.cpp
+++ b/spirv_msl.cpp
@@ -10303,6 +10303,13 @@ void CompilerMSL::emit_atomic_func_op(uint32_t result_type, uint32_t result_id, 
 	{
 		auto obj_expression = to_expression(obj);
 		auto split_index = obj_expression.find_first_of('@');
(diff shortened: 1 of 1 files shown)


spirv_msl.cpp
```diff
@@ -10303,6 +10303,13 @@ void CompilerMSL::emit_atomic_func_op(uint32_t result_type, uint32_t result_id,
 	{
 		auto obj_expression = to_expression(obj);
 		auto split_index = obj_expression.find_first_of('@');
+		bool needs_reinterpret = opcode == OpAtomicUMax || opcode == OpAtomicUMin || opcode == OpAtomicSMax || opcode == OpAtomicSMin;
+		needs_reinterpret &= type.basetype != expected_type;
+		SPIRVariable* backing_var = nullptr;
```
HansKristian-Work, line 10308:
```suggestion
		SPIRVariable *backing_var = nullptr;
```

Issue #2385, opened 2024-10-05
 Compute shader compile fails with "Compiler encountered an internal error."


squidbus:
Re-posting from https://github.com/KhronosGroup/MoltenVK/issues/2363 as this is specific to shader translation.

The below compute shader and pipeline are producing the following error from `vkCreateComputePipelines` in MoltenVK:
```
[mvk-error] VK_ERROR_INITIALIZATION_FAILED: Compute pipeline compile failed (Error code 3):
Compiler encountered an internal error.
```

The crash report for MTLCompilerService says the following:
```
unable to legalize instruction: %62:_(s32), %63:_(s1) = 145 %59:_, %55:_
Context:
%62:_(s32), %63:_(s1) = 145 %59:_, %55:_
%59:_(s32) = 181 %58:_(s32)
%55:_(s32) = 90 %57:_(p64) :: (dereferenceable load (s32) from `i32 addrspace(64)* bitcast (i8 addrspace(64)* getelementptr inbounds ([28 x i8], [28 x i8] addrspace(64)* @memorycache0, i32 0, i32 8) to i32 addrspace(64)*)`, !tbaa !50, !alias.scope !54, !noalias !61, addrspace 64)
%58:_(s32) = 19 $r10
%57:_(p64) = 193 %7:_, %17:_(s64)
%7:_(p64) = 71 @memorycache0
%17:_(s64) = 120 i64 8
(in function: agc.main)
```

Compute shader (SPIR-V and generated MSL included): https://gist.github.com/squidbus/50273067f4bd89752473c04683524a93

Pipeline:
* Push Descriptor Layout:
  * Uniform Buffer at Binding 0
  * Uniform Texel Buffer at Binding 1
  * Storage Image at Binding 2
* Push Constants: Offset = 0, Size = 104

System: M3 Max w/ macOS 15.1 Beta 5 (and macOS 15)

This seems to be coming from the last line with the image write, as if I replace the coordinates argument with a fixed zero vector it compiles. Note that there is no error if I use the command line shader compiler, only when compiling a pipeline at runtime.

It looks like this is a Metal compiler bug (which I will try to report upstream - although I'm not sure how much attention feedback reports get), but is this something that could be worked around in SPIR-V to MSL translation?

HansKristian-Work:
This isn't exactly actionable for me as-is. Replacing coordinates with 0 is surely not a "fix" for anything? Can you propose a concrete compiler workaround?

[...]

PR #2392, opened 2024-10-14
MSL: Work around internal compiler error from mulhi.

squidbus:
Proposed fix for https://github.com/KhronosGroup/SPIRV-Cross/issues/2385

This introduces a helper function to perform `Op*MulExtended` instructions. The helper takes the following template parameters:
* The type of the result container.
* The type of the operands and result container fields. This is needed because `OpSMulExtended` can be used with unsigned types but will perform signed multiplication. The inputs are cast to signed to perform correct multiplication, and the results need to be cast back to unsigned afterwards in order to satisfy Metal/C++ rules for initializer list types.
* The type the function will use to multiply arguments, based on the instruction.

Updated the tests for these changes, and the shader in the linked issue now compiles without error.

### DIFF
--- a/reference/opt/shaders-msl/asm/comp/uint_smulextended.asm.comp
+++ b/reference/opt/shaders-msl/asm/comp/uint_smulextended.asm.comp
@@ -1,7 +1,15 @@
+#pragma clang diagnostic ignored "-Wmissing-prototypes"
+
 #include <metal_stdlib>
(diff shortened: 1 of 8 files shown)


HansKristian-Work:
On a quick glance this is fine, it can also be considered a cleanup, which is nice. I'll look at it in more detail tomorrow.

PR #2400, opened 2024-10-28
MSL: added mesh-shader support

Try:
Rework of old https://github.com/KhronosGroup/SPIRV-Cross/pull/2074
Also made some changes to make diff smaller and also introduced `SetMeshOutputsEXT`  wrapper, for sake of code-gen clarity.

Only mesh stage, for the sake of code-review. Task will be separated PR.
In tests `taskPayloadSharedEXT` is replaced to `const`; other than that test-cases are copied from HLSL

~Note: shader-playground is down today - wasn't able to fully test output.~
Tested shaders via shader-playground


HansKristian-Work:
Looking at this locally now. Just going to do minor cleanups, but this is a far more palatable PR than the first one and it can probably land without too much controversy.

PR #2401, opened 2024-10-30
Land PR 2400.

HansKristian-Work:


spirv_msl.cpp
```diff
@@ -18799,6 +19103,210 @@ void CompilerMSL::emit_block_hints(const SPIRBlock &)
...
+				          to_member_name(type_vert, index), ".y);", "    // Invert Y-axis for Metal");
+			}
+		}
+		statement("spvMesh.set_vertex(spvVI, spvV);");
+		end_scope();
+	}
+
+	if (mesh_out_per_primitive != 0 || builtin_mesh_primitive_indices_id != 0)
+	{
+		auto &type_prim = get<SPIRType>(mesh_out_per_primitive);
```
Try, line 19235: This line causes crash, if shader has no per-primitive outputs

HansKristian-Work: Right ...

HansKristian-Work: fixed.

PR #2405, opened 2024-11-05
ci: Build dependency for Github Actions

spencer-lunarg:
Recently we have found productivty slow down in other KhronosGroups repos because it can take hours for CI to run due to running out of Github Actions minutes. We have found that MacOS/iOS are charged as 10x the minute cost which really starts to add up. The goal is to not have those expensive CI actions run unless we know other things work first.

(There is an internal issue on this, but for short term, I have just been apply these types of YAML chanegs aroudn the repos https://gitlab.khronos.org/khronos-general/khronos-issues/-/issues/127)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -69,6 +69,8 @@ jobs:
         run: ctest --verbose -C Release
 
   build-mobile:
+    # iOS is 10x expensive to run on GitHub machines, so only run if we know something else passed as well
+    needs: build
     name: "Build ${{ matrix.sdk }}"
     strategy:
       matrix:


HansKristian-Work:
If mac/ios is charged at 10x rate, should we remove the recently added iphoneio/iphonesimulator/appletvos/appletvsimulator builds?

spencer-lunarg:
> should we remove the recently added iphoneio/iphonesimulator/appletvos/appletvsimulator builds?

My suggestion is maybe keep, in a few months if you find that they never once fail where only they fail and everything else is caught, then remove

Another option is to use a `if: github.ref == 'refs/heads/main'` and only do when you merge and not on each PR run

PR #2406, opened 2024-11-05
MSL: Pad array elements in Metal argument buffer when shader declares scalar.

billhollings:
A shader may declare and access the first element of a descriptor array as a scalar, which throws off the remaining structure of a Metal Argument Buffer. Add a padding member to the shader Metal argument buffer struct to consume the descriptor elements beyond the first.

- `CompilerMSL::analyze_argument_buffers()`: Track when a descriptor array is declared as a scalar in shader, and add a subsequent member to the Metal argument buffer struct to consume the elements beyond the first.
- `CompilerMSL::get_resource_array_size()`: Ensure variable id is valid.

### DIFF
--- a/spirv_msl.cpp
+++ b/spirv_msl.cpp
@@ -202,6 +202,9 @@ uint32_t CompilerMSL::get_resource_array_size(const SPIRType &type, uint32_t id)
 {
 	uint32_t array_size = to_array_size_literal(type);
 
+	if (id == 0)
+		return array_size;
+
 	// If we have argument buffers, we need to honor the ABI by using the correct array size
 	// from the layout. Only use shader declared size if we're not using argument buffers.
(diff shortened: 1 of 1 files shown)


spirv_msl.cpp
```diff
@@ -18844,12 +18850,19 @@ void CompilerMSL::analyze_argument_buffers()
...
 					rez_bind = get_argument_buffer_resource(desc_set, next_arg_buff_index);
+					prev_was_scalar_on_array_offset = 0;
 				}
 
+				uint32_t count = rez_bind.count;
+
+				// If the current resource is an array in the descriptor, but is a scalar
+				// in the shader, only the first element will be consumed. The next pass
+				// will add a padding member to consume the remaining array elements.
+				if(count > 1 && type.array.empty())
```
HansKristian-Work, line 18861:
```suggestion
				if (count > 1 && type.array.empty())
```

PR #2412, opened 2024-11-11
GLSL: Add option to emulate alpha test

rdb:
Adds an option to add an emulated alpha test for back-ends not supporting it, which I imagine is a common enough manipulation that it might be worth putting it in SPIRV-Cross, similar to the coordinate system fix-ups. Always tests against the output with location 0, if present. Adds a uniform float variable called `SPIRV_Cross_AlphaTestRef` for the reference value. Alpha test is done at every return point from the main entry point.

Example generated shader:
```glsl
#version 460

uniform float SPIRV_Cross_AlphaTestRef;
layout(location = 4) uniform sampler2D p4;
layout(location = 5) uniform vec4 p5;

layout(location = 0) out vec4 o0;
layout(location = 0) in vec3 i1_0;
layout(location = 1) in vec4 i1_1;

void main()
{
    o0 = textureProj(p4, i1_0);
    o0 += p5;
    o0 *= i1_1;
    if (o0.a <= SPIRV_Cross_AlphaTestRef) discard;
}
```

I can work on unit tests if you like, but I wanted confirmation first whether you think this is in-scope for SPIRV-Cross.

If not, then maybe we can discuss the best way I can hook into SPIRV-Cross to add the test on the engine side (eg. by allowing a custom name for the main entry point so I can concatenate a wrapper).

### DIFF
--- a/spirv_glsl.cpp
+++ b/spirv_glsl.cpp
@@ -3563,6 +3563,12 @@ void CompilerGLSL::emit_resources()
 		fixup_implicit_builtin_block_names(execution.model);
 		break;
 
+	case ExecutionModelFragment:
+		if (options.fragment.emulate_alpha_test_func != Options::Always &&
+		    options.fragment.emulate_alpha_test_func != Options::Never)
+			statement("uniform float SPIRV_Cross_AlphaTestRef;");
+		break;
+
 	default:
 		break;
 	}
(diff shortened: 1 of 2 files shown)


HansKristian-Work:
This is too gross to upstream.

> emulated alpha test for back-ends not supporting it

Fixed function alpha testing does not exist in any modern API. What is the use case here?

> allowing a custom name for the main entry point so I can concatenate a wrapper

That is simple enough. A simple option to replace the GLSL entry point name with something else isn't too invasive. It won't work for HLSL/MSL due to the native of those languages where they must consume all arguments, but if this is GLSL only, then ... sure.

rdb:
Clear.

GL compat profile still has fixed function alpha testing that can be enabled (functioning like this PR), typically the driver injects the extra kill into the shader, the use case for this was emulating that behaviour in core profile / GLES to achieve feature parity between our renderers.

Issue #2420, opened 2024-11-19
MSL: divergence in different code stages when relying on DecorationInvariant


mabaro:
The current implementation relies on adding the dependency chain of the **invariant** variable (i.e, position) to the **forced_temporary** group. This already splits a lot of operations reducing the chances of divergencies, but it's not enough for complex setups.
In our use case, the **spirv-opt** output has divergencies on different stages, splitting one particular operation on one stage (i.e.,  __961 - (_958 * dot(_961, _958))__) into temporaries, while keeping it inlined on the other stage.
When passing the produced spirv through **spirv-cross** the **invariant** decoration current behaviour keeps the divergences generating glitches.
The only solution I came up to have some sort of working code is:
- propagate decorationInvariant through the dependency chain
- use decorationInvariant in the same sense as decorationNoContraction in emit_binary_op/emit_binary_op_func (i.e., the latter is used by spirv_msl when outputting spvXXX functions)

This is for sure not a complete solution, and I don't know if that approach could even be considered adequated for merging at certain point, or for further investigation.

I'd appreciate any hints/opinions.
Cheers.

mabaro:
related post -> https://github.com/KhronosGroup/SPIRV-Cross/pull/1022

mabaro:
This pull request contains the changes that fix the mentioned divergence problem by having all operations involved being forced into temporaries, assuming the Metal compiler will do a proper job optimizing.

https://github.com/KhronosGroup/SPIRV-Cross/pull/2424

HansKristian-Work:
Ok, I think I grok the problem now.

If a dependent variable was marked forced temporary in one translation unit, but not the other, the tracking of the dependency chain breaks. I think I see a better solution though. I'll try that.

PR #2424, opened 2024-12-09
compute full dependency chain when invariant position is present

mabaro:
Related issue:
https://github.com/KhronosGroup/SPIRV-Cross/issues/2420


mabaro:
These changes enforce computing the whole dependency chain for all operations when invariant position decorator is present.

HansKristian-Work:
I pushed a more correct fix I think. Let me know if that resolves your issue.

HansKristian-Work:
Superseded, so closing.

PR #2427, opened 2024-12-11
Fix invariant propagation

mabaro:
GLSL/SPIRV output seems to work fine (i.e., no problems/glitches detected when using _invariant_), thus only for MSL output invariant position dependencies are forced into temporaries.

### DIFF
--- a/spirv_glsl.cpp
+++ b/spirv_glsl.cpp
@@ -11766,13 +11766,13 @@ void CompilerGLSL::disallow_forwarding_in_expression_chain(const SPIRExpression 
 	// Allow trivially forwarded expressions like OpLoad or trivial shuffles,
 	// these will be marked as having suppressed usage tracking.
 	// Our only concern is to make sure arithmetic operations are done in similar ways.
-	if (expression_is_forwarded(expr.self) && !expression_suppresses_usage_tracking(expr.self) &&
-	    forced_invariant_temporaries.count(expr.self) == 0)
+	if( ( forced_invariant_temporaries.count( expr.self ) == 0 ) &&
+		( backend.force_invariant_temporaries || ( expression_is_forwarded( expr.self ) && !expression_suppresses_usage_tracking( expr.self ) ) ) )
 	{
 		force_temporary_and_recompile(expr.self);
 		forced_invariant_temporaries.insert(expr.self);
 
-		for (auto &dependent : expr.expression_dependencies)
+		for (auto &dependent : expr.invariance_dependencies)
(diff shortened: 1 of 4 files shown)


mabaro:
@HansKristian-Work These bits were missing from my previous pull request in order to make it work for MSL

spirv_glsl.cpp
```diff
@@ -11766,8 +11766,8 @@ void CompilerGLSL::disallow_forwarding_in_expression_chain(const SPIRExpression
 	// Allow trivially forwarded expressions like OpLoad or trivial shuffles,
 	// these will be marked as having suppressed usage tracking.
 	// Our only concern is to make sure arithmetic operations are done in similar ways.
-	if (expression_is_forwarded(expr.self) && !expression_suppresses_usage_tracking(expr.self) &&
-	    forced_invariant_temporaries.count(expr.self) == 0)
+	if( ( forced_invariant_temporaries.count( expr.self ) == 0 ) &&
+		( backend.force_invariant_temporaries || ( expression_is_forwarded( expr.self ) && !expression_suppresses_usage_tracking( expr.self ) ) ) )
```
HansKristian-Work, line 11770: Can you explain what the real difference is here? The only difference I can see is that the previous code has an escape hatch for expression_suppresses_usage_tracking.

Is the fix here that for MSL, we should ignore that?

HansKristian-Work:
Please make sure to sign the CLA, otherwise I cannot merge anything.

spirv_glsl.cpp
mabaro: Without these **forced_invariant_temporaries** the invariant propagation  (i.e., forcing temporaries) stops before reaching the end of the dependency chain.
The intent of this option is to force the whole dependency chain into temporaries, only for MSL.
GLSL/SPIRV output (i.e, android) has not shown this kind of problems so far, thus I'd leave that codepath unchanged unless you think that would also make sense.

HansKristian-Work: Oh, I see now.

HansKristian-Work: Does https://github.com/KhronosGroup/SPIRV-Cross/pull/2428 resolve the issue?

mabaro: wrt #2428:
On a first check on a single shader abusing CompilerGLSL::flags_to_qualifiers_glsl() to output _INV_ tag it propagates nicely until the beginning of main().
Full rebuild seems good as well, no glitches.

HansKristian-Work:
Merged as part of #2428.

Issue #2435, opened 2025-01-20
MSL: Fragment shader that discards unconditionally does not compile.


squidbus:
Ran into an interesting case with a emulator shader translator where it was outputting a shader that branches over its whole logic to a `OpDemoteToHelperInvocation` at the end.

The SPIR-V at the end looks like this:
```
       %2613 = OpLabel
               OpDemoteToHelperInvocation
               OpBranch %195
        %195 = OpLabel
               OpUnreachable
               OpFunctionEnd
```

When translated to MSL, the shader results in a main function with the following structure:
```
fragment main0_out main0(main0_in in [[stage_in]], ...)
{
    main0_out out = {};
    ...
    if (false)
    {
        ...
        return out;
    }
    discard_fragment();
}
```

This results in the following MSL compile error:
```
program_source:1199:1: error: control reaches end of non-void fragment function [-Werror,-Wmtl-shader-return-type]
}
^
```

~~Now, I am not sure whether this is actually how the shader was supposed to be translated from its original assembly to SPIR-V.~~ I've looked into this further and the shader does in fact just block its logic behind an always-false condition and discard. I do believe the resulting SPIR-V is a valid shader with no errors from `spirv-val`, so I think cases like this where a shader discards unconditionally should produce valid MSL shaders. I believe adding a dummy return after the discard can fix the issue, see another case of this for reference: https://github.com/gfx-rs/wgpu/issues/4458

squidbus:
I saw there's similar handling for blocks ending in `Kill` already and it just needs to be extended to `Unreachable`, so I added that here: https://github.com/KhronosGroup/SPIRV-Cross/pull/2436

PR #2436, opened 2025-01-20
MSL: Terminate function with return value using return if ending in unreachable.

squidbus:
Fix for https://github.com/KhronosGroup/SPIRV-Cross/issues/2435.

If a function with return value ends in unreachable, insert a return statement to avoid MSL compile errors. This condition already exists for blocks ending in Kill, it just needs to be extended to cover Unreachable as well.

### DIFF
--- a/shaders-msl/asm/frag/unreachable-return.msl23.spv14.asm.frag
+++ b/shaders-msl/asm/frag/unreachable-return.msl23.spv14.asm.frag
@@ -1,0 +1,78 @@
+; SPIR-V
+; Version: 1.5
+; Generator: Khronos Glslang Reference Front End; 11
+; Bound: 46
+; Schema: 0
+               OpCapability Shader
+               OpCapability DemoteToHelperInvocation
+               OpExtension "SPV_EXT_demote_to_helper_invocation"
+          %1 = OpExtInstImport "GLSL.std.450"
+               OpMemoryModel Logical GLSL450
+               OpEntryPoint Fragment %main "main" %gl_FragCoord %frag_clr %buff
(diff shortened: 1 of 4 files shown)


squidbus:
I had included some updates to other test files from what the test script generated, but for some reason they're just generating different output on my machine than CI. So I removed them.

shaders-msl/asm/frag/unreachable-return.msl23.spv14.asm.frag
```diff
@@ -0,0 +1,78 @@
...
+         %27 = OpAccessChain %_ptr_Function_int %frag_coord %uint_0
+         %28 = OpLoad %int %27
+         %29 = OpIAdd %int %25 %28
+               OpStore %buff_idx %29
+               OpStore %frag_clr %34
+         %41 = OpLoad %int %buff_idx
+         %44 = OpAccessChain %_ptr_StorageBuffer_int %buff %int_0 %41
+               OpStore %44 %int_1
+               OpDemoteToHelperInvocation
+               OpUnreachable
```
HansKristian-Work, line 77: This OpUnreachable is completely bogus fwiw. Having Unreachable post-dominate the entry block means that the entire shader is UB to execute, but ... whatever.

squidbus: Yes I had the same suspicion and will probably have to visit the generation that results in this. But it seems to function elsewhere (I think just because discarding eliminates potential side effects?) and I figured also that, while the spec says that executing OpUnreachable is UB, there is no indication that it should fail to compile.

PR #2442, opened 2025-02-04
MSL: Expose information about specialization constants - macro mapping

dboyan:
Specialization constant may be translated into either function constant or macro in MSL. MoltenVK requires different processing for the two cases. We expose the list of constant ids and their corresponding macro names so that MoltenVK can use them properly support specialization on those macros.

References:
https://github.com/KhronosGroup/MoltenVK/pull/2434
https://github.com/KhronosGroup/MoltenVK/issues/2423#issuecomment-2601916375

### DIFF
--- a/spirv_glsl.hpp
+++ b/spirv_glsl.hpp
@@ -297,6 +297,9 @@ public:
 		float_formatter = formatter;
 	}
 
+	// Returns the macro name corresponding to constant id
+	std::string constant_value_macro_name(uint32_t id) const;
+
 protected:
 	struct ShaderSubgroupSupportHelper
 	{
@@ -679,7 +682,6 @@ protected:
 	                                    const SmallVector<uint32_t> &indices);
 	void emit_block_chain(SPIRBlock &block);
 	void emit_hoisted_temporaries(SmallVector<std::pair<TypeID, ID>> &temporaries);
-	std::string constant_value_macro_name(uint32_t id);
 	int get_constant_mapping_to_workgroup_component(const SPIRConstant &constant) const;
 	void emit_constant(const SPIRConstant &constant);
 	void emit_specialization_constant_op(const SPIRConstantOp &constant);
--- a/spirv_msl.cpp
+++ b/spirv_msl.cpp
@@ -8022,14 +8022,18 @@ void CompilerMSL::emit_specialization_constants_and_structs()
 				else if (has_decoration(c.self, DecorationSpecId))
 				{
 					// Fallback to macro overrides.
+					uint32_t constant_id = get_decoration(c.self, DecorationSpecId);
 					c.specialization_constant_macro_name =
-					    constant_value_macro_name(get_decoration(c.self, DecorationSpecId));
+					    constant_value_macro_name(constant_id);
 
 					statement("#ifndef ", c.specialization_constant_macro_name);
 					statement("#define ", c.specialization_constant_macro_name, " ", constant_expression(c));
 					statement("#endif");
 					statement("constant ", sc_type_name, " ", sc_name, " = ", c.specialization_constant_macro_name,
 					          ";");
+					
+					// Record the usage of macro
+					constant_macro_ids.insert(constant_id);
 				}
 				else
 				{
@@ -19165,6 +19169,11 @@ void CompilerMSL::set_combined_sampler_suffix(const char *suffix)
 const char *CompilerMSL::get_combined_sampler_suffix() const
 {
 	return sampler_name_suffix.c_str();
+}
+
+bool CompilerMSL::specialization_constant_is_macro(uint32_t const_id) const
+{
+	return constant_macro_ids.find(const_id) != constant_macro_ids.end();
 }
 
 void CompilerMSL::emit_block_hints(const SPIRBlock &)
--- a/spirv_msl.hpp
+++ b/spirv_msl.hpp
@@ -1176,7 +1176,7 @@ protected:
 	std::set<std::string> pragma_lines;
 	std::set<std::string> typedef_lines;
 	SmallVector<uint32_t> vars_needing_early_declaration;
-	std::set<uint32_t> constant_macro_ids;
+	std::unordered_set<uint32_t> constant_macro_ids;
 
 	std::unordered_map<StageSetBinding, std::pair<MSLResourceBinding, bool>, InternalHasher> resource_bindings;
 	std::unordered_map<StageSetBinding, uint32_t, InternalHasher> resource_arg_buff_idx_to_binding_number;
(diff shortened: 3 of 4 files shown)


dboyan:
I have updated the PR to finalize the API and clean up unnecessary code in previous versions. It does not change the generated MSL code but rather provides some extra information for users to know the handling of specialization constants in MSL to properly implement specialization.

The user of the API is: https://github.com/KhronosGroup/MoltenVK/pull/2441 (The API is stable and most likely won't change regardless of the user implementation)

Please let me know if there is concern over the change or amendment needed.

spirv_glsl.hpp
```diff
@@ -297,6 +297,9 @@ class CompilerGLSL : public Compiler
 		float_formatter = formatter;
 	}
 
+	// Returns the macro name corresponding to constant id
+	std::string constant_value_macro_name(uint32_t id);
```
HansKristian-Work, line 301: Where is this implemented?

spirv_msl.cpp
```diff
@@ -19167,6 +19171,20 @@ const char *CompilerMSL::get_combined_sampler_suffix() const
 	return sampler_name_suffix.c_str();
 }
 
+uint32_t CompilerMSL::get_constant_macro_count() const
```
HansKristian-Work, line 19174: Is this interface really needed? What you need to know is if constant_id N maps to a macro or not as you're compiling a pipeline. If you don't intend to override it, there's no need for you to know what it is, so this interface feels redundant.

HansKristian-Work (changes requested):
Comments.

spirv_glsl.hpp
dboyan: This is an existing method, I just moved it from protected to public so that external user (MoltenVK) can query the macro name.

dboyan: I can make it const in the next round.

spirv_msl.cpp
dboyan: I guess this is needed from a practical point of view (due to how MoltenVK is implemented). There, `CompilerMSL` lives in a separate shader converter module and is destroyed immediately after shader converter finishes its job. Rather, the result and related states are packed in a data structure to be used later for the pipeline.

So the problem is, at shader conversion stage, we don't know a priori the number of const ids. When compiling the pipeline, `CompilerMSL` is gone. So the current way seems the only viable choice.

I'm aware that the API design is kinda bound to how MotlenVK works. But on the other hand, I don't particularly find this design harmful, and I don't intend to drastically change how MoltenVK works at the current stage.

HansKristian-Work: > So the problem is, at shader conversion stage, we don't know a priori the number of const ids. When compiling the pipeline, CompilerMSL is gone. So the current way seems the only viable choice.

I mean, wouldn't it make more sense to be able to query "is constant ID a macro?"
It's definitely more natural that getting a full list of spec IDs, then searching through that array to find if the ID pipeline cares about is in there.

spirv_glsl.hpp
dboyan: Now the method is const.

spirv_msl.cpp
dboyan: > I thought MoltenVK already dealt with this case of spec constants being demoted to defines ... ?

Well. It only *partially* solves the problem. It demotes it into macros and stops there, without implementing specialization semantic over macro (so those demoted "specialization constants" always stick to their default values). Nobody discovers that this is a issue until recently...

dboyan: > wouldn't it make more sense to be able to query "is constant ID a macro?"

I agree. I believe that was what I did [in the MoltenVK part](https://github.com/KhronosGroup/MoltenVK/blob/ffdccabaf102448ec99340f3913aa1d387af9888/MoltenVK/MoltenVK/GPUObjects/MVKShaderModule.mm#L86-L88), it relies on the data structure constructed from the information we provide here. Maybe I'm wrong but I don't think having an API here to query if each constant id is a macro is most efficient and helpful way to construct the data structure for use in other part (I mean related to pipeline) of moltenvk.

dboyan: To fill in some context, here is the planned way to consume the API in MoltenVK:
1. https://github.com/KhronosGroup/MoltenVK/blob/ffdccabaf102448ec99340f3913aa1d387af9888/MoltenVKShaderConverter/MoltenVKShaderConverter/SPIRVToMSLConverter.cpp#L373
2. https://github.com/KhronosGroup/MoltenVK/blob/ffdccabaf102448ec99340f3913aa1d387af9888/MoltenVKShaderConverter/MoltenVKShaderConverter/SPIRVToMSLConverter.cpp#L555-L570

Note that when the API is used in the shader converter, it has no clue about what the pipeline would be like, neither can it know what specialization constant are in the shader (I know lacking compiler component to understand or manipulate shader is a shortcoming of the current architecture, but that's what we have for now). So I decided to retrieve the information in the current fashion.

I'm open to improve the API if a better alternative exists. But I personally doesn't see one.

cdavis5e: What I did in the past in this situation is create a `CompilerReflection` object and query that for data like this. It might be more ergonomic to add a method to `CompilerReflection` to query this information.

dboyan: Using `CompilerReflection` can be a plan. I realized that we can indeed use it to query the number of spec constants, and potentially other information we need down the road. But I have a concern. It is not trivial to decide whether a spec constant is to be demoted into a macro, and the logic is bound to MSL (it actually relies on `msl_options`). I'm not sure doing it in `CompilerReflection` is the best idea. I'd say I still prefer using an API to query the needed information up front, immediately after conversion here.

Meanwhile, I find that `get_constant_macro_count` can be avoided. There's already `get_specialization_constants` method, so here we only need to expose a simple query on whether a constant id is a macro without the need for counting and storing a list. I'll update that part later.

dboyan: Got rid of it in the updated version. Now we only expose a straightforward interface about "whether a specialization constant is a macro". The moltenvk counterpart is updated accordingly.

I think we can stick to the current form even if we want to further refine the moltenvk part.

spirv_msl.hpp
```diff
@@ -1171,6 +1176,7 @@ class CompilerMSL : public CompilerGLSL
 	std::set<std::string> pragma_lines;
 	std::set<std::string> typedef_lines;
 	SmallVector<uint32_t> vars_needing_early_declaration;
+	std::set<uint32_t> constant_macro_ids;
```
HansKristian-Work, line 1179: This should be unordered_set. There is no point in forcing ordering here.

HansKristian-Work, line 1179:
```suggestion
	std::unordered_set<uint32_t> constant_macro_ids;
```

HansKristian-Work (approved):

Issue #2443, opened 2025-02-04
Support for VK_KHR_zero_initialize_workgroup_memory


squidbus:
This extension allows initializing workgroup memory in shaders using a null constant. It is also required for Vulkan 1.3 support in MoltenVK.

What seems to be missing currently:
* Cannot zero-initialize workgroup memory; seems to be unimplemented and blocked by this conditional: https://github.com/KhronosGroup/SPIRV-Cross/blob/1a7b7ef6de02cf6767e42b10ddad217c45e90d47/spirv_glsl.cpp#L15776

* The CTS tests for this use specialization constants for array sizes, which is also unsupported with `OpConstantNull`: https://github.com/KhronosGroup/SPIRV-Cross/blob/1a7b7ef6de02cf6767e42b10ddad217c45e90d47/spirv_cross_parsed_ir.cpp#L1053-L1054

dboyan:
Actually I also happened looking into it recently. Your observation is accurate. I will share my current thinking and attempts here.

> Cannot zero-initialize workgroup memory; seems to be unimplemented and blocked by this conditional

For the first one, I would say to keep the current logic for non-workgroup memory. But for workgroup (i.e., shared in GLSL, threadgroup in MSL) memory with initializer, we add a simplified logic as I believe workgroup memory can only be initialized with OpConstantNull (Reference: [Validation logic in SPIRV-Tools as of now](https://github.com/alan-baker/SPIRV-Tools/blob/126c518aa78fc1972eed30f09bfed6f77b74c2d2/source/val/validate_memory.cpp#L590)).

Now the question becomes how to initialize the memory in different shading languages. In GLSL this is simple, we can just use [GL_EXT_null_initializer](https://github.com/KhronosGroup/GLSL/blob/main/extensions/ext/GL_EXT_null_initializer.txt) and initialize everything to `{ }`.

In MSL, this becomes trickier. I just attempted the cheapest way out, namely trying to use initializer on threadgroup variables (as MSL follows C++14 it *should* support this, although I have no idea how they map it into implementation). I initialize array or composite types with `{ }` and other types with their usual initializers. MSL compiler seemed happy with my logic and surprisingly it passed a bunch of CTS tests. However, there are 19 failing tests (iirc). I haven't got the time to look into the details of those passing and failing tests so I'm not sure if the correctness of this approach can be fixed by adding a barrier or if this approach is fundamentally flawed.

[...]

PR #2444, opened 2025-02-08
GLSL/MSL: Support VK_KHR_zero_initialize_workgroup_memory

dboyan:
The extension allows initializing workgroup memory with OpConstantNull, and is a part of Vulkan 1.3.

In GLSL, the support is realized via GL_EXT_null_initializer. While in MSL, we have to generate code snippets to let the threads initialize the memory. Currently I don't have a corresponding implementation for HLSL due to my lack of knowledge.

When connected with MoltenVK, the implementation can pass all Vulkan CTS compute tests enabled by the extension when an independent fix for specialization constants (https://github.com/KhronosGroup/MoltenVK/pull/2434) is applied. The MoltenVK code (with downstream fix) is available at: https://github.com/dboyan/MoltenVK/tree/zero_init_wg_mem-specfix (need to manually specify SPIRV-Cross tree with this PR branch when using `fetchDependencies` while building).

Vulkan CTS compute cases before enabling VK_KHR_zero_initialize_workgroup_memory:
```
  Passed:        81/45160 (0.2%)
  Failed:        0/45160 (0.0%)
  Not supported: 45079/45160 (99.8%)
```

Vulkan CTS compute cases after enabling VK_KHR_zero_initialize_workgroup_memory (with https://github.com/KhronosGroup/MoltenVK/pull/2434 applied):
```
  Passed:        712/45160 (1.6%)
  Failed:        0/45160 (0.0%)
  Not supported: 44448/45160 (98.4%)
```

Closes: #2443

### DIFF
--- a/spirv_cross.cpp
+++ b/spirv_cross.cpp
@@ -4911,13 +4911,16 @@ void Compiler::make_constant_null(uint32_t id, uint32_t type)
 		uint32_t parent_id = ir.increase_bound_by(1);
 		make_constant_null(parent_id, constant_type.parent_type);
 
-		if (!constant_type.array_size_literal.back())
-			SPIRV_CROSS_THROW("Array size of OpConstantNull must be a literal.");
-
-		SmallVector<uint32_t> elements(constant_type.array.back());
-		for (uint32_t i = 0; i < constant_type.array.back(); i++)
+		// The array size of OpConstantNull can be either literal or specialization constant.
+		// In the latter case, we cannot take the value as-is, as it can be changed to anything.
+		// Rather, we assume it to be *one* for the sake of initializer.
+		bool is_literal_array_size = constant_type.array_size_literal.back();
+		uint32_t count = is_literal_array_size ? constant_type.array.back() : 1;
+		SmallVector<uint32_t> elements(count);
+		for (uint32_t i = 0; i < count; i++)
 			elements[i] = parent_id;
-		set<SPIRConstant>(id, type, elements.data(), uint32_t(elements.size()), false);
+		auto &constant = set<SPIRConstant>(id, type, elements.data(), uint32_t(elements.size()), false);
+		constant.is_null_array_specialized_length = !is_literal_array_size;
 	}
 	else if (!constant_type.member_types.empty())
 	{
--- a/spirv_glsl.cpp
+++ b/spirv_glsl.cpp
@@ -681,6 +681,8 @@ string CompilerGLSL::compile()
 	backend.requires_relaxed_precision_analysis = options.es || options.vulkan_semantics;
 	backend.support_precise_qualifier =
 			(!options.es && options.version >= 400) || (options.es && options.version >= 320);
+	backend.constant_null_initializer = "{ }";
+	backend.requires_matching_array_initializer = true;
 
 	if (is_legacy_es())
 		backend.support_case_fallthrough = false;
@@ -5901,6 +5903,11 @@ string CompilerGLSL::constant_expression(const SPIRConstant &c,
 	if (is_pointer(type))
 	{
 		return backend.null_pointer_literal;
+	}
+	else if (c.is_null_array_specialized_length && backend.requires_matching_array_initializer)
+	{
+		require_extension_internal("GL_EXT_null_initializer");
+		return backend.constant_null_initializer;
 	}
 	else if (!c.subconstants.empty())
 	{
--- a/spirv_glsl.cpp
+++ b/spirv_glsl.cpp
@@ -15812,13 +15812,24 @@ string CompilerGLSL::variable_decl(const SPIRVariable &variable)
 		else if (options.force_zero_initialized_variables && type_can_zero_initialize(type))
 			res += join(" = ", to_zero_initialized_expression(get_variable_data_type_id(variable)));
 	}
-	else if (variable.initializer && !variable_decl_is_remapped_storage(variable, StorageClassWorkgroup))
+	else if (variable.initializer)
 	{
-		uint32_t expr = variable.initializer;
-		if (ir.ids[expr].get_type() != TypeUndef)
-			res += join(" = ", to_initializer_expression(variable));
-		else if (options.force_zero_initialized_variables && type_can_zero_initialize(type))
-			res += join(" = ", to_zero_initialized_expression(get_variable_data_type_id(variable)));
+		if (!variable_decl_is_remapped_storage(variable, StorageClassWorkgroup))
+		{
+			uint32_t expr = variable.initializer;
+			if (ir.ids[expr].get_type() != TypeUndef)
+				res += join(" = ", to_initializer_expression(variable));
+			else if (options.force_zero_initialized_variables && type_can_zero_initialize(type))
+				res += join(" = ", to_zero_initialized_expression(get_variable_data_type_id(variable)));
+		}
+		else
+		{
+			// Workgroup memory requires special handling. First, it can only be Null-Initialized.
+			// GLSL will handle this with null initializer, while others require more work after the decl
+			require_extension_internal("GL_EXT_null_initializer");
+			if (!backend.constant_null_initializer.empty())
+				res += join(" = ", backend.constant_null_initializer);
+		}
 	}
 
 	return res;
@@ -16579,6 +16590,12 @@ void CompilerGLSL::emit_function(SPIRFunction &func, const Bitset &return_flags)
 			// Comes from MSL which can push global variables as local variables in main function.
 			add_local_variable_name(var.self);
 			statement(variable_decl(var), ";");
+
+			// "Real" workgroup variables in compute shaders needs extra caretaking.
+			// They need to be initialized with an extra routine as they come in arbitrary form.
+			if (var.storage == StorageClassWorkgroup && var.initializer)
+				emit_workgroup_initialization(var);
+
 			var.deferred_declaration = false;
 		}
 		else if (var.storage == StorageClassPrivate)
@@ -16683,6 +16700,10 @@ void CompilerGLSL::emit_fixup()
 		if (options.vertex.flip_vert_y)
 			statement("gl_Position.y = -gl_Position.y;");
 	}
+}
+
+void CompilerGLSL::emit_workgroup_initialization(const SPIRVariable &)
+{
 }
 
 void CompilerGLSL::flush_phi(BlockID from, BlockID to)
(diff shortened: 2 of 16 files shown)


spirv_msl.cpp
```diff
@@ -272,7 +272,9 @@ void CompilerMSL::build_implicit_builtins()
 	    (active_input_builtins.get(BuiltInVertexId) || active_input_builtins.get(BuiltInVertexIndex) ||
 	     active_input_builtins.get(BuiltInBaseVertex) || active_input_builtins.get(BuiltInInstanceId) ||
 	     active_input_builtins.get(BuiltInInstanceIndex) || active_input_builtins.get(BuiltInBaseInstance));
-	bool need_local_invocation_index = (msl_options.emulate_subgroups && active_input_builtins.get(BuiltInSubgroupId)) || is_mesh_shader();
+	bool need_local_invocation_index =
```
HansKristian-Work, line 275: This seems like bogus. Can you prove this is needed with a test?

spirv_msl.cpp
```diff
@@ -1765,6 +1767,9 @@ string CompilerMSL::compile()
 		emit_resources();
 		emit_function(get<SPIRFunction>(ir.default_entry_point), Bitset());
 
+		// Compilation may update the required builtins, update here just in case
+		build_implicit_builtins();
```
HansKristian-Work, line 1771: Please prove this is needed.

spirv_glsl.cpp
```diff
@@ -15773,13 +15774,20 @@ string CompilerGLSL::variable_decl(const SPIRVariable &variable)
...
 	}
-	else if (variable.initializer && !variable_decl_is_remapped_storage(variable, StorageClassWorkgroup))
+	else if (variable.initializer)
 	{
-		uint32_t expr = variable.initializer;
-		if (ir.ids[expr].get_type() != TypeUndef)
-			res += join(" = ", to_initializer_expression(variable));
-		else if (options.force_zero_initialized_variables && type_can_zero_initialize(type))
-			res += join(" = ", to_zero_initialized_expression(get_variable_data_type_id(variable)));
+		if (!variable_decl_is_remapped_storage(variable, StorageClassWorkgroup)) {
```
HansKristian-Work, line 15779: Braces go on next line.

spirv_msl.cpp
```diff
@@ -19389,6 +19394,70 @@ void CompilerMSL::emit_mesh_tasks(SPIRBlock &block)
...
+		statement(join(to_name(var.self), " = ", to_initializer_expression(var), ";"));
+		end_scope();
+		if (is_boolean) {
+			type.basetype = SPIRType::Boolean;
+		}
+	} else {
+		// Otherwise, we use a loop to cooperatively initialize the memory within the group
+
+		// First, we define a few variable names;
+		string var_name = to_name(var.self);
```
HansKristian-Work, line 19425: This really needs test output to be included.

spirv_cross.cpp
```diff
@@ -4910,11 +4910,13 @@ void Compiler::make_constant_null(uint32_t id, uint32_t type)
...
-		if (!constant_type.array_size_literal.back())
-			SPIRV_CROSS_THROW("Array size of OpConstantNull must be a literal.");
-
-		SmallVector<uint32_t> elements(constant_type.array.back());
-		for (uint32_t i = 0; i < constant_type.array.back(); i++)
+		// The array size of OpConstantNull can be either literal or specialization constant.
+		// In the latter case, we cannot take the value as-is, as it can be changed to anything.
+		// Rather, we assume it to be *one* for the sake of initializer.
+		bool is_literal_array_size = constant_type.array_size_literal.back();
+		uint32_t count = is_literal_array_size ? constant_type.array.back() : 1;
```
HansKristian-Work, line 4917: Does this work only because we kinda throw away the array size of an OpConstantNull later when initializing the threadgroup variable anyway?

dboyan: It is the case for shared (threadgroup) variable, but not for other kinds. In those cases, the initializer list is still used. In fact, I'm not 100% sure if the handling is correct for those cases where initializer is used (although I hope it is if the languages act like C/C++). At least Apple's MSL compiler does not complain about it. For GLSL, I'm not an expert to say if it is right. We could again use `GL_EXT_null_initializer` to eliminate doubt but it seems to require some extra work I don't quite know how. I tried to set the array size to 0, but it seems to turn the array into a scalar value.

spirv_glsl.cpp
dboyan: Fixed

spirv_msl.cpp
dboyan: Two tests added to reflect the two cases of this function.

spirv_msl.cpp
dboyan: I changed how things work after discovering a few issues.
* I added a class member `needs_workgroup_zero_init`, initialized in a new analyzing pass before code generation. It clarifies the condition here and avoids calling `build_implicit_builtins` in recompile (actually we can't repeatedly call it without preventing it from adding some redundant builtins anyway)
* The initialization routines is limited to "real" workgroup variables, not those remapped to workgroup variables. I discovered that at least those remapped from tesc seem to be properly initialized already and the extra routine would cause troubles unless some more work that complicates the logic, which is unnecessary after all.

The new test `shaders-msl/comp/shared-zero-init-simple.comp` shows the case where `gl_LocalInvocationIndex` is introduced solely by the initialization routine.

spirv_msl.cpp
dboyan: This is not needed anymore.

spirv_cross.cpp
dboyan: I discovered the logic this part is a little problematic for glsl. I'm adding a test and trying to resolve it.

spirv_glsl.cpp
```diff
@@ -15805,13 +15806,21 @@ string CompilerGLSL::variable_decl(const SPIRVariable &variable)
...
-		else if (options.force_zero_initialized_variables && type_can_zero_initialize(type))
-			res += join(" = ", to_zero_initialized_expression(get_variable_data_type_id(variable)));
+		if (!variable_decl_is_remapped_storage(variable, StorageClassWorkgroup))
+		{
+			uint32_t expr = variable.initializer;
+			if (ir.ids[expr].get_type() != TypeUndef)
+				res += join(" = ", to_initializer_expression(variable));
+			else if (options.force_zero_initialized_variables && type_can_zero_initialize(type))
+				res += join(" = ", to_zero_initialized_expression(get_variable_data_type_id(variable)));
+		} else {
```
HansKristian-Work, line 15818: Brace on next line.

spirv_glsl.cpp
```diff
@@ -16572,6 +16581,12 @@ void CompilerGLSL::emit_function(SPIRFunction &func, const Bitset &return_flags)
 			// Comes from MSL which can push global variables as local variables in main function.
 			add_local_variable_name(var.self);
 			statement(variable_decl(var), ";");
+
+			// "Real" workgroup variables in compute shaders needs extra caretaking.
+			// They need to be initialized with an extra routine as they come in arbitrary form.
+			if (var.storage == StorageClassWorkgroup && var.initializer) {
```
HansKristian-Work, line 16587: Drop braces here.

spirv_msl.cpp
```diff
@@ -19397,6 +19421,68 @@ void CompilerMSL::emit_mesh_tasks(SPIRBlock &block)
...
+			type.basetype = SPIRType::Short;
+		}
+		statement("if (gl_LocalInvocationIndex == 0)");
+		begin_scope();
+		statement(join(to_name(var.self), " = ", to_initializer_expression(var), ";"));
+		end_scope();
+		if (is_boolean) {
+			type.basetype = SPIRType::Boolean;
+		}
+	} else {
```
HansKristian-Work, line 19446: Brace on next line.

spirv_msl.cpp
```diff
@@ -19397,6 +19421,68 @@ void CompilerMSL::emit_mesh_tasks(SPIRBlock &block)
...
+
+	begin_scope();
+
+	if (type.array.empty() && type.member_types.empty())
+	{
+		// For simple shared variables, we just initialize it in thread 0 of the block
+		// We use short to represent bool for threadgroup variable to workaround compiler bug,
+		// so we do a temporary fixup here. Alas. (see the type_to_glsl method)
+		bool is_boolean = type.basetype == SPIRType::Boolean;
+		if (is_boolean) {
```
HansKristian-Work, line 19436: Drop brace.

spirv_msl.cpp
```diff
@@ -19397,6 +19421,68 @@ void CompilerMSL::emit_mesh_tasks(SPIRBlock &block)
...
+
+		// First, we define a few variable names;
+		string var_name = to_name(var.self);
+		string var_ptr_name = join(var_name, "_ptr");
+		string var_size_name = join(var_name, "_sz");
+		string var_pos_name = join(var_name, "_pos");
+		string var_stride_name = join(var_name, "_stride");
+		string var_ptr2_name = join(var_name, "_ptr2");
+
+		statement("threadgroup uint *", var_ptr_name, " = (threadgroup uint *)&", var_name, ";");
```
HansKristian-Work, line 19457: Gross, but it's certainly not worse than the alternative ...

spirv_msl.cpp
```diff
@@ -19397,6 +19421,68 @@ void CompilerMSL::emit_mesh_tasks(SPIRBlock &block)
...
+		// so we do a temporary fixup here. Alas. (see the type_to_glsl method)
+		bool is_boolean = type.basetype == SPIRType::Boolean;
+		if (is_boolean) {
+			type.basetype = SPIRType::Short;
+		}
+		statement("if (gl_LocalInvocationIndex == 0)");
+		begin_scope();
+		statement(join(to_name(var.self), " = ", to_initializer_expression(var), ";"));
+		end_scope();
+		if (is_boolean) {
```
HansKristian-Work, line 19443: Drop brace.

spirv_cross.cpp
dboyan: Apparently GLSL doesn't like it when initializer and the corresponding array having different number of elements, which will *certainly* happen for (non-shared) null constant array of specialized dimensions. Although it is a rare case, I decided to try my best to generate valid code with the least level of invasive change. I decided to only apply `GL_EXT_null_initializer` on `OpConstantNull` arrays with non-literal length (tracked with a flag at creation), and it is limited to the GLSL backend. A glsl test is also introduced.

spirv_msl.cpp
dboyan: This is usually done at IR-level transformation rather than in the source. As I couldn't find a reliable way to let apple's compiler initialize threadgroup variables, I have to do it myself. My code is inspired by the NIR lowering pass in mesa. Thankfully msl is flexible enough to allow this in a straightforward manner.

HansKristian-Work (approved):
LGTM

PR #2452, opened 2025-03-03
MSL: Treat pointer to vector as scalar type when bitcasting

### LINKED ISSUE #2451: MSL: Accessing buffer address as vector pointer not compiling
I have some shaders compiled with DirectXShaderCompiler to SPIR-V, then used with SPIRV-Cross via MoltenVK. These shaders use buffer addresses in push constants to access certain constants. When a shader tries to access a vector type via the address using `vk::RawBufferLoad`, it fails to compile the SPIR-V to MSL.

Example shader (SPIR-V disassembly, generated GLSL, generated MSL): https://gist.github.com/squidbus/eb704f77173928b00b1b1820e6b110ee

Original HLSL: https://github.com/hedge-dev/UnleashedRecomp/blob/main/UnleashedRecomp/gpu/shader/gamma_correction_ps.hlsl

Resulting errors:
```
program_source:26:43: error: as_type cast from 'ulong' (aka 'unsigned long') to 'device int2 *' is not allowed
    int2 _55 = int2(gl_FragCoord.xy) - (*(as_type<device int2*>(reinterpret_cast<ulong>(g_PushConstants.SharedConstants + 16ul))));
                                          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
program_source:30:29: error: as_type cast from 'ulong' (aka 'unsigned long') to 'device int2 *' is not allowed
        _66 = any(_55 >= (*(as_type<device int2*>(reinterpret_cast<ulong>(g_PushConstants.SharedConstants + 24ul)))));
                            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
program_source:45:34: error: as_type cast from 'ulong' (aka 'unsigned long') to 'device float3 *' is not allowed
    float3 _81 = powr(_77.xyz, *(as_type<device float3*>(reinterpret_cast<ulong>(g_PushConstants.SharedConstants))));
                                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

dboyan:
Otherwise, they will be misregarded as vectors and can be used in as_type, which does not support pointer type as destination.

Fixes #2451

### DIFF
--- a/reference/opt/shaders-msl/asm/frag/bitcast-ptr-vec.argument-tier-1.msl23.asm.frag
+++ b/reference/opt/shaders-msl/asm/frag/bitcast-ptr-vec.argument-tier-1.msl23.asm.frag
@@ -1,0 +1,67 @@
+#pragma clang diagnostic ignored "-Wmissing-prototypes"
+
+#include <metal_stdlib>
+#include <simd/simd.h>
+
+using namespace metal;
+
+template<typename T>
+struct spvDescriptor
+{
+    T value;
+};
+
+template<typename T>
+struct spvDescriptorArray
+{
+    spvDescriptorArray(const device spvDescriptor<T>* ptr) : ptr(&ptr->value)
+    {
+    }
+    const device T& operator [] (size_t i) const
+    {
+        return ptr[i];
+    }
+    const device T* ptr;
+};
+
+struct type_ConstantBuffer_PushConstants
+{
+    ulong VertexShaderConstants;
+    ulong PixelShaderConstants;
+    ulong SharedConstants;
+};
+
+struct main0_out
+{
+    float4 out_var_SV_Target [[color(0)]];
+};
+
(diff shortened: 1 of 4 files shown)


cdavis5e (changes requested):
The change itself looks good, but needs a test case.

squidbus:
Can confirm this fixes the shader issue I had, thanks!

dboyan:
I added a test case taken from the original bug report. It's a little bit long but glslang doesn't generate the spirv I want and I'm not proficient enough to edit raw spirv assembly.

HansKristian-Work:
Seems correct, but it's using the wrong way to check for is_pointer. It's also done "wrong" in the surrounding code, so I'll just fix that as a follow-up.

PR #2454, opened 2025-03-19
MSL: Fix gather functions on Sequoia

etang-cw:
Sequoia's texture2d has a new template parameter (`template <typename T, access a, memory_coherence c, typename _Enable> struct texture2d : _texture2d<T, a, c>`), making the old template matching fail to match.  Switch to using decltype on the method we plan to call.

I also removed `spvForward`, as nothing really has custom copy/move constructors in MSL (and you can't easily use forwarding references due to the explicit address space qualifier requirement).


HansKristian-Work (approved):
Doesn't seem particularly controversial, LGTM.

PR #2464, opened 2025-04-11
MSL: Support broader tessellation I/O matching for VK_KHR_maintenance4.

billhollings:
Tessellation stages pass I/O via buffer content which may contain nested structs, which was not previously handled.

- Adjust the vector sizes of any nested struct members within tessellation input variables to match the vector sizes of the corresponding output variables from the previous pipeline stage.
- Update one corresponding test case.

### DIFF
--- a/spirv_msl.cpp
+++ b/spirv_msl.cpp
@@ -3756,6 +3756,20 @@ void CompilerMSL::add_variable_to_interface_block(StorageClass storage, const st
 		return;
 	}
 
+	// Tesselation stages pass I/O via buffer content which may contain nested structs.
+	// Ensure the vector sizes of any nested struct members within these input variables match
+	// the vector sizes of the corresponding output variables from the previous pipeline stage.
+	// This adjustment is handled here instead of ensure_correct_input_type() in order to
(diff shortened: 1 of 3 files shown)


spirv_msl.cpp
```diff
@@ -3951,6 +3965,42 @@ void CompilerMSL::add_variable_to_interface_block(StorageClass storage, const st
 	}
 }
 
+// Recursively iterate into the input struct type, and adjust the vecsize
+// of any nested members, based on location info provided through the API.
+// The location parameter is modified recursively.
+void CompilerMSL::ensure_struct_members_valid_vecsizes(SPIRType &struct_type, uint32_t &location) {
+	assert(struct_type.basetype == SPIRType::Struct);
```
HansKristian-Work, line 3972:
```suggestion
void CompilerMSL::ensure_struct_members_valid_vecsizes(SPIRType &struct_type, uint32_t &location)
{
	assert(struct_type.basetype == SPIRType::Struct);
```

HansKristian-Work (approved):

PR #2466, opened 2025-04-20
MSL: Add support for SPV_KHR_expect_assume

spnda:
Not sure about using the func impl thingy for defining macros, but this seems to work & compile fine. Couldn't think of a good test that actually had any codegen/performance improvements, but the LLVM IR does show that the Metal compiler is using the builtins at least to some degree.

Essentially, this just defines macros around the `__builtin_assume` and `__builtin_expect` builtins to be used for the two SPIR-V instructions, which are both builtins supported by Clang.

Relevant for KhronosGroup/MoltenVK#2490

### DIFF
--- a/spirv_msl.cpp
+++ b/spirv_msl.cpp
@@ -7869,6 +7869,26 @@ void CompilerMSL::emit_custom_functions()
 			statement("");
 			break;
 
+		case SPVFuncImplAssume:
+			statement_no_indent("#if !defined(SPV_ASSUME) && defined(__has_builtin)");
+			statement_no_indent("#if __has_builtin(__builtin_assume)");
+			statement_no_indent("#define SPV_ASSUME(x) __builtin_assume(x);");
+			statement_no_indent("#endif");
+			statement_no_indent("#if __has_builtin(__builtin_expect)");
+			statement_no_indent("#define SPV_EXPECT(x, y) __builtin_expect(x, y);");
+			statement_no_indent("#endif");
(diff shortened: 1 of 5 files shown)


spirv_msl.cpp
```diff
@@ -10222,6 +10242,27 @@ void CompilerMSL::emit_instruction(const Instruction &instruction)
 		break;
 	}
 
+	case OpAssumeTrueKHR:
+	{
+		auto condition = ops[0];
+		statement(join("SPV_ASSUME(", to_expression(condition), ")"));
```
HansKristian-Work, line 10248:
```suggestion
		statement(join("SPV_ASSUME(", to_unpacked_expression(condition), ")"));
```

spirv_msl.cpp
```diff
@@ -10222,6 +10242,27 @@ void CompilerMSL::emit_instruction(const Instruction &instruction)
...
+	}
+
+	case OpExpectKHR:
+	{
+		auto result_type = ops[0];
+		auto ret = ops[1];
+		auto value = ops[2];
+		auto exp_value = ops[3];
+
+		auto exp = join("SPV_EXPECT(", to_expression(value), ", ", to_expression(exp_value), ")");
```
HansKristian-Work, line 10259:
```suggestion
		auto exp = join("SPV_EXPECT(", to_unpacked_expression(value), ", ", to_unpacked_expression(exp_value), ")");
```

PR #2469, opened 2025-04-23
GLSL/MSL: Implement SPV_KHR_subgroup_rotate

squidbus:
Adds GLSL and MSL support for `SPV_KHR_subgroup_rotate`.
* For GLSL both clustered and non-clustered variants are supported.
* For MSL, clustered is not supported as I could not find an option available for this, but it is behind a separate Vulkan feature flag anyway.

### DIFF
--- a/spirv_msl.cpp
+++ b/spirv_msl.cpp
@@ -6914,6 +6914,36 @@ void CompilerMSL::emit_custom_functions()
 			statement("");
 			break;
 
+		case SPVFuncImplSubgroupRotate:
+			statement("template<typename T>");
+			statement("inline T spvSubgroupRotate(T value, ushort delta)");
+			begin_scope();
+			if (msl_options.use_quadgroup_operation())
+				statement("return quad_shuffle_rotate_down(value, delta);");
+			else
+				statement("return simd_shuffle_rotate_down(value, delta);");
+			end_scope();
+			statement("");
+			statement("template<>");
+			statement("inline bool spvSubgroupRotate(bool value, ushort delta)");
+			begin_scope();
+			if (msl_options.use_quadgroup_operation())
+				statement("return !!quad_shuffle_rotate_down((ushort)value, delta);");
+			else
+				statement("return !!simd_shuffle_rotate_down((ushort)value, delta);");
+			end_scope();
+			statement("");
+			statement("template<uint N>");
+			statement("inline vec<bool, N> spvSubgroupRotate(vec<bool, N> value, ushort delta)");
+			begin_scope();
+			if (msl_options.use_quadgroup_operation())
+				statement("return (vec<bool, N>)quad_shuffle_rotate_down((vec<ushort, N>)value, delta);");
+			else
+				statement("return (vec<bool, N>)simd_shuffle_rotate_down((vec<ushort, N>)value, delta);");
+			end_scope();
+			statement("");
+			break;
+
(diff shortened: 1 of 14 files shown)


HansKristian-Work (approved):

spirv_msl.cpp
```diff
@@ -6914,6 +6914,36 @@ void CompilerMSL::emit_custom_functions()
...
+			statement("template<uint N>");
+			statement("inline vec<bool, N> spvSubgroupRotate(vec<bool, N> value, ushort delta)");
+			begin_scope();
+			if (msl_options.use_quadgroup_operation())
+				statement("return (vec<bool, N>)quad_shuffle_rotate_down((vec<ushort, N>)value, delta);");
+			else
+				statement("return (vec<bool, N>)simd_shuffle_rotate_down((vec<ushort, N>)value, delta);");
+			end_scope();
+			statement("");
+			break;
```
spnda, line 6945: This implementation invokes undefined behavior, because (and this is a quote from the MSL spec):
> `T simd_shuffle_rotate_down(T data, ushort delta)`
Returns data from the thread whose SIMD lane ID is the sum of caller’s SIMD lane ID and delta. The value for delta needs to be the same for all threads in the SIMD-group. This function wraps values around the SIMD-group.

The point of this SPIR-V extension was to have `delta` be non-uniform, as the name of `OpGroupNonUniformRotateKHR` suggests.
> Return the Value of the [invocation](https://github.khronos.org/SPIRV-Registry/extensions/KHR/SPV_KHR_subgroup_rotate.html#Invocation) whose id within the group is calculated as follows:
LocalId = SubgroupLocalInvocationId if Execution is Subgroup or LocalInvocationId if Execution is Workgroup
RotationGroupSize = ClusterSize when ClusterSize is present, otherwise
RotationGroupSize = SubgroupMaxSize if the Kernel capability is declared and SubgroupSize if not.
Invocation ID = ( (LocalId + Delta) & (RotationGroupSize - 1) ) + (LocalId & ~(RotationGroupSize - 1))

The SPIR-V spec makes no requirement about `Delta` being the same for all lanes, while the Metal spec does. This implementation should not be used and shouldn't have been merged.

spnda (commented):
I think this commit should be reverted since its implementation is not correct and invokes undefined behavior, as far as I can tell.

spirv_msl.cpp
HansKristian-Work: > Delta must be dynamically uniform within Execution.

From SPIR-V spec.

HansKristian-Work: > OpGroupNonUniformRotateKHR

means non uniform execution (subgroups), not that arguments are non uniform.

spnda: Ah shit, yes you're right. I misunderstood the meaning of `dynamically uniform` here.

spnda: Sure, but technically UB means that it could work as you might expected :)

cdavis5e: I don't think it was _entirely_ a waste of time...

This _is_ what we need to do to support clustered rotations.

squidbus: Agreed, the clustered version of this will probably need simd_shuffle with manual calculations. Built-in path is only usable for non-clustered as far as I can tell.

PR #2470, opened 2025-04-23
Properly define SPV_EXPECT when builtin not available

spnda:
I noticed that `SPV_EXPECT` wasn't defined properly when the builtin wasn't available, and the guards for defining the macros weren't quite right.


spnda:
Thanks for the quick response!

PR #2472, opened 2025-04-24
MSL: Add support for DebugPrintf.

squidbus:
Translates `DebugPrintf` to `os_log_default.log`, provided in MSL 3.2.

I don't think this is needed for MoltenVK as the Vulkan Validation Layer is responsible for instrumenting debug prints into buffer writes, but may be useful for users translating SPIR-V for use with Metal directly.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -14,7 +14,7 @@ jobs:
     name: "Build ${{ matrix.platform }}"
     strategy:
       matrix:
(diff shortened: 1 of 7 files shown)


squidbus:
It seems that the version of macOS used in CI does not support MSL 3.2, causing the test to fail.

HansKristian-Work (approved):
Simple enough.

PR #2474, opened 2025-04-27
MSL: Add option to provide a default point size.

squidbus:
`VK_KHR_maintenance5` requires that `PointSize` default to `1.0` when not written in a shader. In order to support this requirement, this PR adds an option for MSL to emit a default `PointSize` value when it was not written, as according to Metal documentation the point size must be specified within the vertex shader.

I considered using the existing `enable_point_size_builtin` flag for this but it currently defaults to true. Changing this could negatively impact users relying on the default conversion behavior respecting `PointSize`, and leaving it means every shader emits a default value unless explicitly disabled. Thus I added a new flag to control emitting the default, and a new float option to set what that default is.

Have also added tests with this option where the point size is provided by the shader, meaning it should not be defaulted, and where the point size is not provided by the shader, meaning it should be.

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -684,6 +684,8 @@ struct CLIArguments
 	bool msl_input_attachment_is_ds_attachment = false;
 	bool msl_disable_rasterization = false;
 	bool msl_auto_disable_rasterization = false;
+	bool msl_enable_point_size_default = false;
+	float msl_default_point_size = 1.0f;
 	const char *msl_combined_sampler_suffix = nullptr;
 	bool glsl_emit_push_constant_as_ubo = false;
 	bool glsl_emit_ubo_as_plain_uniforms = false;
@@ -987,7 +989,8 @@ static void print_help_msl()
 	                "\t\tpartially transformed. This fixup gives correct results on Apple Silicon.\n"
 	                "\t[--msl-combined-sampler-suffix <suffix>]:\n\t\tUses a custom suffix for combined samplers.\n"
 	                "\t[--msl-disable-rasterization]:\n\t\tDisables rasterization and returns void from vertex-like entry points.\n"
-	                "\t[--msl-auto-disable-rasterization]:\n\t\tDisables rasterization if BuiltInPosition is not written.\n");
+	                "\t[--msl-auto-disable-rasterization]:\n\t\tDisables rasterization if BuiltInPosition is not written.\n"
+	                "\t[--msl-default-point-size <size>]:\n\t\tApplies a default value if BuiltInPointSize is not written.\n");
 	// clang-format on
 }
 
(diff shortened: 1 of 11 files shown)


main.cpp
```diff
@@ -1838,6 +1843,12 @@ static int main_inner(int argc, char *argv[])
 	cbs.add("--msl-input-attachment-is-ds-attachment", [&args](CLIParser &) { args.msl_input_attachment_is_ds_attachment = true; });
 	cbs.add("--msl-disable-rasterization", [&args](CLIParser &) { args.msl_disable_rasterization = true; });
 	cbs.add("--msl-auto-disable-rasterization", [&args](CLIParser &) { args.msl_auto_disable_rasterization = true; });
+	cbs.add("--msl-default-point-size",
+			[&args](CLIParser &parser)
+			{
+				args.msl_enable_point_size_default = true;
+				args.msl_default_point_size = static_cast<float>(parser.next_double());
+			});
```
HansKristian-Work, line 1851:
```suggestion
	cbs.add("--msl-default-point-size",
	               [&args](CLIParser &parser)
	               {
	                   args.msl_enable_point_size_default = true;
	                   args.msl_default_point_size = float(parser.next_double());
	               });
```

main.cpp
```diff
@@ -1838,6 +1843,12 @@ static int main_inner(int argc, char *argv[])
 	cbs.add("--msl-input-attachment-is-ds-attachment", [&args](CLIParser &) { args.msl_input_attachment_is_ds_attachment = true; });
 	cbs.add("--msl-disable-rasterization", [&args](CLIParser &) { args.msl_disable_rasterization = true; });
 	cbs.add("--msl-auto-disable-rasterization", [&args](CLIParser &) { args.msl_auto_disable_rasterization = true; });
+	cbs.add("--msl-default-point-size",
+	               [&args](CLIParser &parser)
+	               {
```
HansKristian-Work, line 1848: Seems like applying the style nit in web viewer didn't work <_< I'll do it locally before merging.

squidbus: I think I got what you wanted assuming it's like some of the other lines, fixed in commit.

spirv_msl.cpp
```diff
@@ -301,13 +305,30 @@ void CompilerMSL::build_implicit_builtins()
...
 		uint32_t workgroup_id_type = 0;
 
 		ir.for_each_typed_id<SPIRVariable>([&](uint32_t, SPIRVariable &var) {
 			if (var.storage != StorageClassInput && var.storage != StorageClassOutput)
 				return;
 			if (!interface_variable_exists_in_entry_point(var.self))
 				return;
+
+			auto &type = this->get<SPIRType>(var.basetype);
+			if (need_point_size && has_decoration(type.self, DecorationBlock))
```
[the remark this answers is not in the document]
squidbus: Slipped my mind there could be multiple ways to define, I added the built-in check below as well which should handle this.

spirv_cross_c.h
```diff
@@ -750,6 +750,8 @@ typedef enum spvc_compiler_option
 
 	SPVC_COMPILER_OPTION_MSL_AUTO_DISABLE_RASTERIZATION = 92 | SPVC_COMPILER_OPTION_MSL_BIT,
 
+	SPVC_COMPILER_OPTION_MSL_ENABLE_POINT_SIZE_DEFAULT = 93 | SPVC_COMPILER_OPTION_MSL_BIT,
```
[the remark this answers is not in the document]
squidbus: Done, thanks I didn't realize this was a thing.

HansKristian-Work (approved):

PR #2475, opened 2025-04-29
Use accurate negative OpSMod behavior.

squidbus:
Add option for providing `OpSMod` behavior accurate to the SPIR-V spec, as required by `VK_KHR_maintenance8`. Made it optional since it generates less clean output than just using the standard operator when this behavior is not needed, and Vulkan does not require it to be accurate without `maintenance8` enabled. Probably not the most ideal implementation of this but it’s the best I could figure out how to implement in this context.

CTS results for `dEQP-VK.spirv_assembly.*.maintenance8.*` using MoltenVK (unsupported tests are geometry shaders):
```
Test run totals:
  Passed:        24/28 (85.7%)
  Failed:        0/28 (0.0%)
  Not supported: 4/28 (14.3%)
  Warnings:      0/28 (0.0%)
  Waived:        0/28 (0.0%)
```


spirv_glsl.cpp
```diff
@@ -5701,6 +5700,17 @@ string CompilerGLSL::constant_op_expression(const SPIRConstantOp &cop)
 		GLSL_BOP(UGreaterThanEqual, ">=");
 		GLSL_BOP(SGreaterThanEqual, ">=");
 
+	case OpSMod:
+		if (options.enable_accurate_smod)
+		{
+			auto op0 = to_enclosed_expression(cop.arguments[0]);
```
HansKristian-Work, line 5706: Should be to_enclosed_unpacked_expression.

spirv_glsl.cpp
```diff
@@ -7507,6 +7517,42 @@ string CompilerGLSL::to_ternary_expression(const SPIRType &restype, uint32_t sel
 	return expr;
 }
 
+void CompilerGLSL::emit_signed_remainder(uint32_t result_type, uint32_t id, uint32_t left, uint32_t right,
+										 SPIRType::BaseType int_type, bool implicit_integer_promotion,
```
HansKristian-Work, line 7521: Style nit, align with spaces, not tab.

main.cpp
```diff
@@ -747,6 +747,7 @@ struct CLIArguments
 	bool use_420pack_extension = true;
 	bool remove_unused = false;
 	bool combined_samplers_inherit_bindings = false;
+	bool enable_accurate_smod = false;
```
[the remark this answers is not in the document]
squidbus: Okay, sounds good to me.

spirv_glsl.cpp
```diff
@@ -5701,6 +5700,14 @@ string CompilerGLSL::constant_op_expression(const SPIRConstantOp &cop)
 		GLSL_BOP(UGreaterThanEqual, ">=");
 		GLSL_BOP(SGreaterThanEqual, ">=");
 
+	case OpSMod:
+	{
+		auto op0 = to_enclosed_unpacked_expression(cop.arguments[0]);
```
HansKristian-Work, line 5705: This is problematic if op0 and op1 are complex expressions that get duplicated out. I think you need to ensure that the expression is read multiple times, so that the expressions can be placed in temporaries instead. track_expression_read() can be called here in addition to trigger this behavior.

spirv_glsl.hpp
```diff
@@ -703,6 +703,9 @@ class CompilerGLSL : public Compiler
 	bool should_dereference_caller_param(uint32_t id);
 	bool should_forward(uint32_t id) const;
 	bool should_suppress_usage_tracking(uint32_t id) const;
+	void emit_signed_remainder(uint32_t result_type, uint32_t id, uint32_t left, uint32_t right,
+	                           SPIRType::BaseType int_type, bool implicit_integer_promotion,
+	                           bool divisor_sign = false);
```
HansKristian-Work, line 708: No need to have default arguments here.

spirv_glsl.cpp
squidbus: Yeah I've been rather concerned by how the expressions get expanded out, I'll give that a try.

spirv_glsl.cpp
```diff
@@ -13468,7 +13482,7 @@ void CompilerGLSL::emit_instruction(const Instruction &instruction)
 		break;
 
 	case OpSMod:
-		GLSL_BOP_CAST(%, int_type);
+		emit_signed_remainder(ops[0], ops[1], ops[2], ops[3], int_type, implicit_integer_promotion, true);
```
HansKristian-Work, line 13485: If using Vulkan semantics, there should be no need to try to emulate the behavior.

squidbus: * There are some cases where `vulkan_semantics` is true like MSL where we do need this emulation.
* Do we need to worry about the correct behavior on Vulkan pre-maintenance8 regardless?

spirv_glsl.cpp
squidbus: It definitely seems a lot nicer now, thanks.

spirv_glsl.cpp
HansKristian-Work: - MSL backend should implement OpSMod separately.
- No.

squidbus: HLSL will also need this to be accurate, so we have to duplicate implementing this for every backend then? Would it be better to just have a separate flag to force it on anyway to keep things simple?

HansKristian-Work: You could have a backend flag for emulate_smod or something I guess.

squidbus: Okay, added that. GLSL with Vulkan semantics will no longer use the emulated path.

HansKristian-Work (approved):

PR #2488, opened 2025-05-26
glsl: Implement `SPV_EXT_replicated_composites`

theHamsta:
https://github.khronos.org/SPIRV-Registry/extensions/EXT/SPV_EXT_replicated_composites.html offers a streamlined way to initialize composites like vectors and matrices. Also, from extensions like `SPV_KHR_cooperative_matrix` or `SPV_NV_cooperative_vector`.

For constants, we piggyback on the existing logic for constant composites adding an additional bool to indicate whether our composites consists of multiple subconstants or a single replicated one.

In GLSL, most types support replication using their constructor. A notable exception to that are matrices like `mat4` which support `mat4(0.0)` but not `mat4(vec4(0.0))` how they are represented in SPIRV.

One motivation for this PR is to support construction from scalars for https://github.khronos.org/SPIRV-Registry/extensions/NV/SPV_NV_cooperative_vector.html

### DIFF
--- a/spirv_common.hpp
+++ b/spirv_common.hpp
@@ -1356,9 +1356,10 @@ struct SPIRConstant : IVariant
 
 	SPIRConstant() = default;
 
-	SPIRConstant(TypeID constant_type_, const uint32_t *elements, uint32_t num_elements, bool specialized)
+	SPIRConstant(TypeID constant_type_, const uint32_t *elements, uint32_t num_elements, bool specialized, bool replicated_ = false)
 	    : constant_type(constant_type_)
 	    , specialization(specialized)
+	    , replicated(replicated_)
 	{
 		subconstants.reserve(num_elements);
 		for (uint32_t i = 0; i < num_elements; i++)
@@ -1436,6 +1437,9 @@ struct SPIRConstant : IVariant
 
 	// For composites which are constant arrays, etc.
 	SmallVector<ConstantID> subconstants;
+
+	// Whether the subconstants are intended to be replicated (e.g. OpConstantCompositeReplicateEXT)
+	bool replicated = false;
 
 	// Non-Vulkan GLSL, HLSL and sometimes MSL emits defines for each specialization constant,
 	// and uses them to initialize the constant. This allows the user
--- a/spirv_glsl.cpp
+++ b/spirv_glsl.cpp
@@ -5940,6 +5940,32 @@ string CompilerGLSL::constant_expression(const SPIRConstant &c,
 		require_extension_internal("GL_EXT_null_initializer");
 		return backend.constant_null_initializer;
 	}
+	else if (c.replicated && type.op != spv::OpTypeArray)
+	{
+		if (type.op == spv::OpTypeMatrix)
+		{
+			uint32_t num_elements = type.columns;
+			// GLSL does not allow the replication constructor for matrices
+			// mat4(vec4(0.0)) needs to be manually expanded to mat4(vec4(0.0), vec4(0.0), vec4(0.0), vec4(0.0));
+			std::string res;
+			res += type_to_glsl(type);
+			res += "(";
+			for (uint32_t i = 0; i < num_elements; i++)
+			{
+				res += to_expression(c.subconstants[0]);
+				if (i < num_elements - 1)
+				{
+					res += ", ";
+				}
+			}
+			res += ")";
+			return res;
+		}
+		else
+		{
+			return join(type_to_glsl(type), "(", to_expression(c.subconstants[0]), ")");
+		}
+	}
 	else if (!c.subconstants.empty())
 	{
 		// Handles Arrays and structures.
@@ -5989,8 +6015,17 @@ string CompilerGLSL::constant_expression(const SPIRConstant &c,
 		}
 
 		uint32_t subconstant_index = 0;
-		for (auto &elem : c.subconstants)
+		size_t num_elements = c.subconstants.size();
+		if (c.replicated)
 		{
+			if (type.array.size() != 1) {
+				SPIRV_CROSS_THROW("Multidimensional arrays not yet supported as replicated constans");
+			}
+			num_elements = type.array[0];
+		}
+		for (size_t i = 0; i < num_elements; i++)
+		{
+			auto &elem = c.subconstants[c.replicated ? 0 : i];
 			if (auto *op = maybe_get<SPIRConstantOp>(elem))
 			{
 				res += constant_op_expression(*op);
@@ -6021,7 +6056,7 @@ string CompilerGLSL::constant_expression(const SPIRConstant &c,
 				}
 			}
 
-			if (&elem != &c.subconstants.back())
+			if (i != num_elements - 1)
 				res += ", ";
 
 			subconstant_index++;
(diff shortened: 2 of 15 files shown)


spirv_glsl.cpp
```diff
@@ -5940,6 +5940,33 @@ string CompilerGLSL::constant_expression(const SPIRConstant &c,
...
+				{
+					ss << ", ";
+				}
+			}
+			ss << ")";
+			return ss.str();
+		}
+		else
+		{
+			return join(type_to_glsl(type), "(", to_expression(c.subconstants[0]), ")");
```
cdavis5e, line 5967: It's worse than you realize.

`mat4(x)` doesn't do what you think it does. In GLSL, as well as in Metal, the single-scalar constructor does _not_ construct a matrix where _every_ element is `x`; instead, it constructs a matrix where only the diagonal elements are `x`. This happens to work for `mat4(0.0)`, because the other elements get set to zero regardless. Trying it with `mat4(1.0)`, though, does not result in a replicated composite, even with the `use_replicated_composites` pragma enabled. Unless this is the expected behavior of `SPV_EXT_replicated_composites`--and, based on my reading of the spec, there is no indication this is the case--this will break splatted matrices with nonzero arguments.

Since this case cannot be expressed at all in GLSL, a raw SPIR-V assembly shader is necessary in order to test this functionality.

shaders/comp/replicated-composites.spv16.vk.nocompat.comp
cdavis5e, line 1: Your change should work just as well for Metal, as is. If you place a similar file in the `shaders-msl` hierarchy, we can validate this is the case. Consider also doing this for HLSL. In both cases, though, the cooperative matrix stuff will need to be commented out, since neither the MSL nor HLSL compilers support that yet.

spirv_glsl.cpp
theHamsta: yes, I realized that this is behavior in GLSL. In SPIRV, a matrix is always a replicated vector.  So as you said `mat4(1.0)` will not result in replicated SPIRV (but `mat4(d,d,d,d)` for a vector `d` does).

I will add the SPIRV in addition to the GLSL, unless you think we should focus on SPIRV-only. I can change the GLSL to test `mat4(1.0,1.0,1.0,1.0)` instead

shaders/comp/replicated-composites.spv16.vk.nocompat.comp
theHamsta: I will still have to fix this
```
 11: /Users/runner/work/SPIRV-Cross/SPIRV-Cross/reference/shaders-msl/asm/comp/replicated-composites.spv16.asm.comp:16:23: error: variable 'uniform_float' is uninitialized when used here [-Werror,-Wuninitialized]
11:     float4 d = float4(uniform_float);
11:                       ^~~~~~~~~~~~~
11: /Users/runner/work/SPIRV-Cross/SPIRV-Cross/reference/shaders-msl/asm/comp/replicated-composites.spv16.asm.comp:15:24: note: initialize the variable 'uniform_float' to silence this warning
11:     float uniform_float;
11:                        ^
11:                         = 0.0
11: 1 error generated.
11: 
````

Comment (theHamsta) [reply shaders/comp/replicated-composites.spv16.vk.nocompat.comp:1]: I added the SPIRV from the GLSL shader for GLSL, Metal and HLSL. Since I could not get the compiler to generate a replicated array for dynamic values, I added this statement manually at the end of the SPIRV

File: spirv_glsl.cpp [hunk 2]
```diff
@@ -23,10 +23,13 @@
 
 #include "spirv_glsl.hpp"
 #include "GLSL.std.450.h"
+#include "spirv.hpp"
 #include "spirv_common.hpp"
+#include "spirv_cross_error_handling.hpp"
```
HansKristian-Work, line 28: Was this include added by autocomplete?

spirv_glsl.cpp
```diff
@@ -23,10 +23,13 @@
 
 #include "spirv_glsl.hpp"
 #include "GLSL.std.450.h"
+#include "spirv.hpp"
```
HansKristian-Work, line 26: Added by autocomplete?

spirv_glsl.cpp
```diff
@@ -23,10 +23,13 @@
...
 #include "spirv_common.hpp"
+#include "spirv_cross_error_handling.hpp"
 #include <algorithm>
 #include <assert.h>
 #include <cmath>
+#include <cstdint>
```
HansKristian-Work, line 32: Same

spirv_glsl.cpp
```diff
@@ -5989,8 +6018,16 @@ string CompilerGLSL::constant_expression(const SPIRConstant &c,
 		}
 
 		uint32_t subconstant_index = 0;
-		for (auto &elem : c.subconstants)
+		size_t num_elements = c.subconstants.size();
+		if (c.replicated) {
```
HansKristian-Work, line 6022: Braces go on next line.

spirv_glsl.cpp
```diff
@@ -15583,6 +15620,61 @@ void CompilerGLSL::emit_instruction(const Instruction &instruction)
 		break;
 	}
 
+	case OpCompositeConstructReplicateEXT:
+	{
+
```
HansKristian-Work, line 15625: Weird newline.

spirv_glsl.cpp
```diff
@@ -15583,6 +15620,61 @@ void CompilerGLSL::emit_instruction(const Instruction &instruction)
...
+		if (type.op == spv::OpTypeMatrix || type.op == spv::OpTypeArray)
+		{
+			// TODO: nested arrays
+			if (type.op == spv::OpTypeArray && type.array.size() != 1)
+			{
+				SPIRV_CROSS_THROW(
+				    "Multi-dimensional arrays currently not supported for OpCompositeConstructReplicateEXT");
+			}
+			uint32_t num_elements = type.op == spv::OpTypeMatrix ? type.columns : type.array[0];
+			spirv_cross::StringStream<> ss;
```
HansKristian-Work, line 15642: Don't use stringstream.

spirv_common.hpp
```diff
@@ -1353,9 +1353,10 @@ struct SPIRConstant : IVariant
 
 	SPIRConstant() = default;
 
-	SPIRConstant(TypeID constant_type_, const uint32_t *elements, uint32_t num_elements, bool specialized)
+	SPIRConstant(TypeID constant_type_, const uint32_t *elements, uint32_t num_elements, bool specialized, bool _replicated = false)
```
HansKristian-Work, line 1356: replicated_ not _replicated.

HansKristian-Work (changes requested):
Mostly nits I think.

spirv_glsl.cpp
theHamsta: removed

spirv_glsl.cpp
theHamsta: removed

spirv_glsl.cpp
theHamsta: removed now.

spirv_glsl.cpp
theHamsta: fixed

spirv_glsl.cpp
theHamsta: fixed

spirv_glsl.cpp
```diff
@@ -5940,6 +5940,32 @@ string CompilerGLSL::constant_expression(const SPIRConstant &c,
...
+			std::string res;
+			res += type_to_glsl(type);
+			res += "(";
+			for (uint32_t i = 0; i < num_elements; i++)
+			{
+				res += to_expression(c.subconstants[0]);
+				if (i < num_elements - 1)
+				{
+					res += ", ";
+				}
```
HansKristian-Work, line 5959:
```suggestion
					res += ", ";
```

HansKristian-Work (approved):

PR #2489, opened 2025-05-27
Draft: GLSL: implement `SPV_NV_cooperative_vector`

theHamsta:
This is on top of #2488 since replicated composites often occur in the initialization of cooperative vectors. This should stay in draft status until all issues with the previous PR are resolved.

I tried to roughly follow the implementation of `SPV_EXT_cooperative_matrix`. Though I might not have tracked  all dependencies of the instructions correctly. This is my first time working with spirv-cross, so some concepts are still unfamiliar to me. I tested the implementation on shaders in https://github.com/jeffbolznv/vk_cooperative_vector_perf

There is a roughly equivalent proposal for HLSL to which this SPV extension could be mapped: https://github.com/microsoft/hlsl-specs/blob/main/proposals/0026-hlsl-long-vector-type.md

### DIFF
--- a/spirv_glsl.cpp
+++ b/spirv_glsl.cpp
@@ -27,18 +27,17 @@
 #include "spirv_common.hpp"
 #include "spirv_cross_error_handling.hpp"
 #include <algorithm>
+#include <array>
 #include <assert.h>
 #include <cmath>
 #include <cstdint>
 #include <limits>
 #include <locale.h>
 #include <utility>
-#include <array>
 
 #ifndef _WIN32
 #include <langinfo.h>
 #endif
-#include <locale.h>
 
 using namespace spv;
 using namespace SPIRV_CROSS_NAMESPACE;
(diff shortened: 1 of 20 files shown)


theHamsta:
I also prepared a draft for https://registry.khronos.org/vulkan/specs/latest/man/html/VK_NV_cooperative_matrix2.html before working on coopvec-nv. But I discovered that interactions of  `VK_NV_cooperative_matrix2` with function calling might make it more complicated than coopvec-nv

spirv_parser.cpp
```diff
@@ -623,6 +623,23 @@ void Parser::parse(const Instruction &instruction)
...
+
+		type.basetype = SPIRType::CoopVecNv;
+		type.op = op;
+		type.coopVecNv.component_type_id = ops[1];
+		type.coopVecNv.component_count_id = ops[2];
+
+		// CoopVec-Nv can be used with integer operations like SMax where
+		// where spirv-opt does explicit checks on integer bitwidth
+		auto component_type = get<SPIRType>(type.coopVecNv.component_type_id);
+		type.width = component_type.width;
```
theHamsta, line 639: This could also be handled on the respective integer operations where it could defer to the bitwidth of the component type instead of the coop-vec

spirv_glsl.cpp
```diff
@@ -15497,6 +15527,112 @@ void CompilerGLSL::emit_instruction(const Instruction &instruction)
...
+		auto buf = ops[0];
+		auto offset = ops[1];
+		auto v1 = ops[2];
+		auto v2 = ops[3];
+		auto matrix_layout_id = ops[4];
+		auto matrix_iterpretation_id = ops[5];
+		auto matrix_stride_id = length >= 6 ? ops[6] : 0;
+		//void coopVecOuterProductAccumulateNV(const coopvecNV<T, M> v1, const coopvecNV<T, N> v2,
+		//T[] buf, uint offset, uint stride,
+		//int matrixLayout, int matrixInterpretation);
```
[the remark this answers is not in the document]
HansKristian-Work: Those comments should just be removed, yes.

spirv_glsl.cpp
```diff
@@ -48,6 +48,36 @@ enum ExtraSubExpressionType
 	EXTRA_SUB_EXPRESSION_TYPE_AUX = 0x20000000
 };
 
+const std::vector<std::tuple<uint32_t, const char *>> COOPVEC_COMPONENT_TYPE_NAMES = {
```
HansKristian-Work, line 51: This should just be a static const array of POD struct inside the function that uses it.
Having a global const of non-POD types requires global initialization which really should be avoided if we can help it.

HansKristian-Work (commented):
I'll just pull this in locally and fix what I need to fix.

PR #2490, opened 2025-05-28
MSL: Implement clustered subgroup operations for sizes other than 4.

cdavis5e:
This is required by Vulkan 1.4. Fortunately, I hit upon a way to do this with minimal overhead. The Metal compiler even inlines and unrolls the recursive template used here. I just hope that the Vulkan committee never agrees to mandate prefix op support--I have no idea how to do that one cheaply.

Unfortunately, the extra goop is needed to handle the case where one or more threads are inactive.

The function-like macro parameter used in `FUNC_SUBGROUP_CLUSTERED()` does work with Clang and GCC. I don't know if it works with MSVC.

Fix `LogicalAnd`, `LogicalOr`, and `LogicalXor` ops as well. These need a cast for the Metal compiler to allow this, just like with the shuffle and broadcast ops. The tests for these were marked "invalid," meaning no validation on the transpiled MSL shader was done. Is it any wonder no one caught this? It works now, though, so we can drop the `.invalid.` marker.


HansKristian-Work (approved):
Just going to assume that clustering code is correct. Otherwise, the code structure is fine.

cdavis5e:
> Just going to assume that clustering code is correct.

It is. I ran the CTS against it, on both Intel/AMD and Apple Silicon.

Test results for `dEQP-VK.subgroups.clustered.*`:
Before:
```
  Passed:           0/3244 (0.0%)
  Failed:         268/3244 (8.26%)
  Not supported: 2976/3244 (91.74%)
  Warnings:         0/3244 (0.0%)
  Waived:           0/3244 (0.0%)
```

After:
```
  Passed:         212/3244 (6.54%)
  Failed:          56/3244 (1.73%)
  Not supported: 2976/3244 (91.74%)
  Warnings:         0/3244 (0.0%)
  Waived:           0/3244 (0.0%)
```

PR #2495, opened 2025-06-11
Expose FPFastMathDefault to API execution mode query.

billhollings:


spirv_parser.cpp
```diff
@@ -381,13 +381,21 @@ void Parser::parse(const Instruction &instruction)
...
 		{
+		case ExecutionModeLocalSizeId:
 			execution.workgroup_size.id_x = ops[2];
 			execution.workgroup_size.id_y = ops[3];
 			execution.workgroup_size.id_z = ops[4];
-		}
+			break;
+
+		case ExecutionModeFPFastMathDefault:
+			execution.fp_fast_math_default_id = ops[3];
```
HansKristian-Work, line 393: I don't think this is a complete implementation. FPFastMathDefault takes both a type and mode ID to control behavior for specific types (?!), so I don't think this can be correct.

HansKristian-Work (changes requested):

spirv_parser.cpp
billhollings: Ah. Good point. In theory there could be more than one entry. 

I've modified to track multiple entries as a map.

For MSL compilation, only one set of fast-math flags can be applied at compile time, so I've combined them using accumulative AND'ing, so that only the fast-math flags that apply to all FP types will be applied to the overall compile.

billhollings:
Marked as WIP for now. I'm going to add handling for individual `FPFastMathMode` op decorations too.

spirv_msl.cpp
```diff
@@ -19861,6 +19861,28 @@ bool CompilerMSL::specialization_constant_is_macro(uint32_t const_id) const
...
+	for (auto &fp_pair : ep.fp_fast_math_defaults)
+		if (fp_pair.second)
+			accum_flags = and_if_not_zero(accum_flags, get<SPIRConstant>(fp_pair.second).scalar());
+
+	if (incl_ops)
+		for (auto &p_m : ir.meta)
+			if(p_m.second.decoration.decoration_flags.get(DecorationFPFastMathMode))
+				accum_flags = and_if_not_zero(accum_flags, p_m.second.decoration.fp_fast_math_mode);
+
+	return accum_flags;
```
HansKristian-Work, line 19883: Looking over the SPV_KHR_float_controls2 spec. I'm a little confused about how this deals with a global disable of FP math.

If I understand it, the absence of a global FP math default means the module defaults to normal SPIR-V behavior which is fast-math on by default unless disabled with NoContract. However, if the default behavior is set to 0, I think fast-math is just globally disabled. That also sort of seems to reset the flag tracking.

Also, I'm a little confused why this is queried by MoltenVK. Should SPIRV-Cross not deal with this on its own?

squidbus: One thing to note actually on the last point: assuming I’m reading the spec correctly, you may just be able to use a pragma in the file scope to set the fast math mode for the whole shader, instead of passing through to MoltenVK to set compiler options.

spirv_msl.hpp
```diff
@@ -542,6 +542,10 @@ class CompilerMSL : public CompilerGLSL
 		// The result can be queried with get_is_rasterization_disabled.
 		bool auto_disable_rasterization = false;
 
+		// Use Fast Math pragmas in MSL code, based on SPIR-V float controls and FP ExecutionModes.
+		// Requires MSL 3.2 or above, and has no effect with earlier MSL versions.
+		bool use_fast_math_pragmas = false;
```
squidbus, line 547: Is there a scenario where a user wouldn’t want their fast math flags translated?

spirv_msl.cpp
billhollings: @HansKristian-Work 

> > However, [SPV_KHR_float_controls2](https://github.khronos.org/SPIRV-Registry/extensions/KHR/SPV_KHR_float_controls2.html) turns that on its head, and explicitly disables Fast Math
> 
> It doesn't seem that simple, since this default only applies if you actually set any default mode. I guess to globally disable, you set the FP mode to 0 for e.g. Float32 once.
> 
> > Otherwise, if the current entry point _sets any FPFastMathDefault execution mode_ then all flags specified for any operand type or for the result type of the operation apply
> 
> My main concern here is that it doesn't seem like the code takes this into account, unless I'm missing something.

The logic was working, as interpreted in MoltenVK, which was doing the heavy-lifting of combining the various FP flags. But I appreciate that combining both FP Fast Math modes, and then "classic" execution modes makes it easier to interpret, and keeps it self-contained in SPIRV-Cross.

`get_fp_fast_math_flags()` now combines everything into one set of FP flags, and starts with full Fast Math as the default, then reducing based on FP and execution modes.

spirv_msl.hpp
billhollings: > Is there a scenario where a user wouldn’t want their fast math flags translated?

Actually, yes, it turns out. 

I could not find any way to actually enable the pragmas over the settings of `MTLCompileOptions`, which always overrode the pragmas, even when the `MTLCompileOptions` properties were not explicitly set. That seems bizarre, as one would expect some kind of hierarchy where the pragmas would temporarily override `MTLCompileOptions`, but I couldn't get it to work. maybe it's a short-term bug with the MSL compiler,

Because of this, MoltenVK will leave `use_fast_math_pragmas` disabled, and stick with manipulating `MTLCompileOptions`.

Even though MoltenVK isn't able to use this feature, I've left it in here in case there are other compile scenarios (like offline maybe) where this is operational, or in case whatever is blocking its use in MoltenVK is fixed.

squidbus: Ah, makes sense. Maybe MoltenVK could just skip compiler flags when MSL 3.2 is supported, but I suppose it makes sense to leave this option.

billhollings: > Ah, makes sense. Maybe MoltenVK could just skip compiler flags when MSL 3.2 is supported, but I suppose it makes sense to leave this option.

As I mentioned above, I tried that, but even leaving `MTLCompileOptions` unmodified still used the default values in `MTLCompileOptions` over the pragmas.

spirv_msl.cpp
```diff
@@ -5750,22 +5750,44 @@ void CompilerMSL::emit_header()
 {
 	// This particular line can be overridden during compilation, so make it a flag and not a pragma line.
 	if (suppress_missing_prototypes)
-		statement("#pragma clang diagnostic ignored \"-Wmissing-prototypes\"");
+		add_pragma_line("#pragma clang diagnostic ignored \"-Wmissing-prototypes\"");
```
HansKristian-Work, line 5753: Using add_pragma_line is somewhat questionable here. This will force a recompile, even before we have committed to emitting the header. This is wasteful, and I think this needs to fix add_pragma_line so that it doesn't force a recompile unless it's called after the header is committed. I think all uses of add_pragma happen before we emit the header, so maybe that path can simply be dropped.

HansKristian-Work: I'll look into cleaning that up.

PR #2544, opened 2025-09-14
Spirv cross for cpp17

ikhoeyZX:
migrate to spirv.hpp11
make spirv-cross compiled using std=c++17 (fix issue KhronosGroup#2523)
Need more help support for fix this in other platform
It work in my case (android, vita3k emulator)


HansKristian-Work:
An alternative approach I could support if you cannot split the translation units for whatever reason is a way to override the namespace used when including spirv-cross' spv headers to avoid conflicts like these as a pragmatic hack. There are similar considerations with SPIRV_CROSS_NAMESPACE already since linking conflicts is a thing.

PR #2549, opened 2025-09-25
CMake: enable C++17

corporateshark:
`SPIRV-Cross` de facto already relies on `C++17` features like the `[[maybe_unused]]` attribute.

If we don't enable `C++17` during compilation, we see warnings such as:
```
C5051: attribute [[maybe_unused]] requires at least '/std:c++17'; ignored.
```

Enabling C++17 resolves these warnings.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,8 +19,8 @@
 #  2. The MIT License, found at <http://opensource.org/licenses/MIT>.
 #
 
-cmake_minimum_required(VERSION 3.10)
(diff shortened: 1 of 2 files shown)


HansKristian-Work:
Relying on C++17 is unintended. Better fix is to remove the attribute. SPIRV-Cross targets C++11.

corporateshark:
@HansKristian-Work Sure! I removed `[[maybe_unused]]` completely.

Issue #2556, opened 2025-10-16
MSL: Request support new SPV_KHR_fma?


oscarbg:
Hi,
searching MSL pdf (https://developer.apple.com/metal/Metal-Shading-Language-Specification.pdf)in table 6.4 shows/seems MSL "fma" instruction is equivalent to new OpFmaKHR of SPV_KHR_fma,right?..

```
T fma(T a, T b, T c) Returns the correctly rounded floating-point
representation of the sum of c with the infinitely
precise product of a and b. Rounding of
intermediate products shall not occur. Edge
case behavior is per the IEEE 754-2008
standard.

```
if so should be easy to implement in SPIRV-Cross MSL backend?

thanks..

HansKristian-Work:
Feel free to PR it.

PR #2559, opened 2025-10-21
MSL: Support OpFmaKHR

squidbus:
Implements https://github.com/KhronosGroup/SPIRV-Cross/issues/2556

Comes in two commits:
* Roll dependencies up to latest to pull in `OpFmaKHR` support.
  * Fix HLSL functionality test which hit a new error in `spirv-val` related to decorated variable declaration order.
  * Fix some `GL_QCOM_image_processing` tests erroring during GLSL compile due to SPIR-V version being too low for extension.
  * Fix an MSL atomics test erroring during GLSL compile due to use of `gl_SemanticsRelaxed` with non-zero storage class semantic.
* Implement `OpFmaKHR` for MSL with test.

Have also ran CTS with this in MoltenVK:
```
./deqp-vk --deqp-case="dEQP-VK.spirv_assembly.instruction.compute.opfma.*"

Test run totals:
  Passed:        48/324 (14.8%)
  Failed:        0/324 (0.0%)
  Not supported: 276/324 (85.2%)
  Warnings:      0/324 (0.0%)
  Waived:        0/324 (0.0%)
```

### DIFF
--- a/reference/opt/shaders-msl/asm/comp/fma.spv16.asm.comp
+++ b/reference/opt/shaders-msl/asm/comp/fma.spv16.asm.comp
@@ -1,0 +1,20 @@
+#include <metal_stdlib>
+#include <simd/simd.h>
+
+using namespace metal;
+
+struct SSBO2
+{
+    float4 out_data[1];
+};
+
+struct SSBO
+{
+    float4 in_data[1];
+};
+
+kernel void main0(device SSBO2& _4 [[buffer(0)]], device SSBO& _6 [[buffer(1)]])
+{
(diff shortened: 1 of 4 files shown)


squidbus:
> and then seen you attach a spir-v test file: [shaders-msl/asm/comp/fma.spv16.asm.comp](https://github.com/KhronosGroup/SPIRV-Cross/pull/2559/commits/186a5f591a8f6d81066a53866ee8925a07c6b222#diff-e572ef7f45815e642ace604cb85a88fd70c3457e721efff3c0bdb10b6032963d) with: "Generator: Khronos Glslang Reference Front End; 11" so assuming you modified by hand the output glslang code to use SPV_KHR_fma, right?

Yes, it's hand-modified from compiled GLSL.

PR #2562, opened 2025-10-24
MSL: Also handle signedness mismatches in fragment outputs.

cdavis5e:
It turns out that the number of components is not the only place where Vulkan allows differences but Metal insists on strict equivalence. Vulkan _also_ allows the fragment shader and the attachment to differ in signedness--for example, rendering to an `R8G8B8A8_UINT` image where the fragment shader declares its output to be an `ivec4`.

To fix this, we must adjust the `[[color]]` output in the interface block to match the render target. Use the shader output specification mechanism to do this; also move the fragment component count here to save some memory at run time.


HansKristian-Work:
Based on this finding, I'm not sure where to go with this PR. If we're adding extra complexity and maintenance burden to support a case that only a bogus CTS test cares about, I'm a bit reluctant, but that's doesn't mean I'm blocking it outright. Awaiting a reply before this PR can progress.

aitor-lunarg:
So while the spec says it's undefined behavior, https://gerrit.khronos.org/c/vk-gl-cts/+/12319 forced all drivers to support this... My main concern is that because now all drivers have that undefined behavior defined, the workgroup could decide to actually make it defined I guess.

I've created an internal CTS issue for the public one to address this: https://gitlab.khronos.org/Tracker/vk-gl-cts/-/issues/6131

While KosmicKrisp works around this, it would be better if we didn't have to.

PR #2565, opened 2025-11-12
Add support for GL_EXT_ray_tracing_position_fetch

ChristianF-ARM:
Trying to convert SPIR-V using `SPV_KHR_ray_tracing_position_fetch` to GLSL currently fails with:
```
SPIRV-Cross threw an exception: Cannot resolve expression type.
```

This commit adds support for  `GL_EXT_ray_tracing_position_fetch` to improve this.


HansKristian-Work (approved):
LGTM

PR #2575, opened 2025-11-28
HLSL: Fix shader input variable for --remove-unused-variables

lance5:
Try to fix HLSL for --remove-unused-variables

### DIFF
--- a/spirv_hlsl.cpp
+++ b/spirv_hlsl.cpp
@@ -1709,7 +1709,7 @@ void CompilerHLSL::emit_resources()
 		ir.for_each_typed_id<SPIRVariable>([&](uint32_t, SPIRVariable &var) {
 			auto &type = this->get<SPIRType>(var.basetype);
 
-			bool is_hidden = is_hidden_input_variable(var);
+			bool is_hidden = is_hidden_vertex_input_variable(var);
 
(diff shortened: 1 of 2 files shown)


lance5:
Yes, this is a risk point
But if I don't do it this way, then when I use 'compiler. get_chader_sources' in the code to get Input,
The received 'stage_inputs' will be missing an' unused 'input. However, in actual HLSL, an' unused 'input is output
Ultimately, it led to inconsistent layout when calling D3D11 ` CreateInputLayout `
My idea is to divide it into two situations:
1. If the user uses' get-active_interacte_variables', then keep the obtained Input consistent with HLSL
2. If the user does not use 'get-active_interacte_variables', everything remains unchanged
And my fix is to maintain consistency in the first scenario
You can consider it.
Thank your are reading.

lance5:
Summary: Inconsistent behavior can be frustrating, and users who do not wish for IO mismatch should not use 'set_cabled_interacte_variables'

HansKristian-Work:
Applied a different fix locally, but this approach is fine.

lance5:
Sure, your repair is more professional than mine

HansKristian-Work:
Merged your first commit along my fixup, thanks.

PR #2605, opened 2026-03-03
Respect MemoryAccessVolatileMask on OpLoad to prevent forwarding

hughperkins:
### Problem

SPIRV-Cross has an expression forwarding optimization that inlines loaded values at each use site rather than materializing them into temporaries. For example, given:

```
%val = OpLoad %int %ptr
OpStore %dst1 %val
OpStore %dst2 %val
```

Forwarding produces:

```c
*dst1 = *ptr;
*dst2 = *ptr;
```

rather than:

```c
int val = *ptr;
*dst1 = val;
*dst2 = val;
```

This is done intentionally to produce cleaner, more readable generated code with fewer temporary variables. When a value is used only once, forwarding is always safe — it's the same single read, just without naming it. When a value is used multiple times, the duplicated dereferences look like extra memory I/O, but for normal storage buffer and uniform loads the downstream shader compiler (Metal, GLSL, etc.) can see that the pointer target hasn't been modified between uses and will CSE (common subexpression eliminate) them back into a single read. So in practice, forwarding doesn't cost extra I/O in the normal case.

However, for `PhysicalStorageBuffer` pointers (raw `device T*` in Metal, `buffer_reference` in GLSL), the downstream compiler cannot easily prove the pointer target is stable — especially when stores to the same address space happen between uses. If the `OpLoad` carries the `Volatile` memory access flag, the SPIR-V is explicitly stating that the memory location may change unpredictably. Duplicating the read by forwarding can then produce incorrect results: the two use sites may observe different values from what should be a single load.

### Fix

When processing an `OpLoad`, SPIRV-Cross already decides whether to forward the expression. This patch adds a check: if the load carries `MemoryAccessVolatileMask`, forwarding is disabled, forcing the value into a temporary. This ensures the memory is read exactly once, matching SPIR-V semantics.

### Test plan

- Added MSL regression test (`volatile-phys-buf-load-no-forward.asm.msl24.comp`): volatile load from a PhysicalStorageBuffer pointer used in two stores; reference output shows a temporary rather than duplicated dereferences
- Added Vulkan GLSL regression test (`volatile-phys-buf-load-no-forward.nocompat.vk.asm.comp`): same shader, verifying the GLSL backend also materializes a temporary
- Both tests fail (output differs from reference) without the fix and pass with it
- Ran full `shaders-msl-no-opt` and `shaders-no-opt` test suites; no regressions

### What the test does

The test is a minimal SPIR-V compute shader that:

1. Loads two raw GPU addresses (u64) from push constants — one for a source pointer, one for a destination pointer
2. Does a single `OpLoad` with `Volatile|Aligned` from the source pointer, producing one value
3. Stores that value to two locations: the destination pointer, and the destination pointer + 4 bytes

The reference output checks that SPIRV-Cross emits one read into a temporary and reuses it:

```c
int _22 = *(reinterpret_cast<device int*>(registers.addr));  // one read
*_21 = _22;                                                   // reuse
*(reinterpret_cast<device int*>(...)) = _22;                  // reuse
```

Without the fix, SPIRV-Cross forwards the load and emits two reads:

```c
*_21 = *_18;                                                  // read 1
*(reinterpret_cast<device int*>(...)) = *_18;                 // read 2
```

The test is a source-level regression test that verifies the shape of the generated code — it does not execute the shader at runtime.

### DIFF
--- a/reference/shaders-msl-no-opt/asm/comp/volatile-phys-buf-load-no-forward.asm.msl24.comp
+++ b/reference/shaders-msl-no-opt/asm/comp/volatile-phys-buf-load-no-forward.asm.msl24.comp
@@ -1,0 +1,19 @@
+#include <metal_stdlib>
+#include <simd/simd.h>
+
+using namespace metal;
+
+struct Registers
+{
+    ulong addr;
+    ulong addr2;
+};
+
+kernel void main0(constant Registers& registers [[buffer(0)]])
+{
+    device int* _21 = reinterpret_cast<device int*>(registers.addr2);
+    int _22 = *(reinterpret_cast<device int*>(registers.addr));
+    *_21 = _22;
+    *(reinterpret_cast<device int*>(reinterpret_cast<ulong>(_21) + 4ul)) = _22;
+}
+
--- a/reference/shaders-no-opt/asm/comp/volatile-phys-buf-load-no-forward.nocompat.vk.asm.comp.vk
+++ b/reference/shaders-no-opt/asm/comp/volatile-phys-buf-load-no-forward.nocompat.vk.asm.comp.vk
@@ -1,0 +1,30 @@
+#version 450
+#if defined(GL_ARB_gpu_shader_int64)
+#extension GL_ARB_gpu_shader_int64 : require
+#else
+#error No extension available for 64-bit integers.
+#endif
+#extension GL_EXT_buffer_reference2 : require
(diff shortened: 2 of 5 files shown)


HansKristian-Work (changes requested):
Blocking the forward is fine, but the comment is very wrong. Volatility is not ensured with this alone.

hughperkins:
Thanks!

HansKristian-Work (approved):
Seems fine as an incremental improvement.

Issue #1602, opened 2021-01-20
CUDA/OpenCL support


mbarriault:
Has there been any discussion regarding support for outputting CUDA and/or OpenCL C/C++ kernels? We have somewhere about 20-30 kernels currently with simultaneous implementations in CUDA and OpenCL C. The majority aren't particularly complicated and we anticipated no issues converting to GLSL for Vulkan. However, that will then be four implementations for each one (as at point no reason not to maintain a native MSL version), and would obviously be much simpler to have just one and use SIPRV-Cross (which is already planned for our rendering shaders).

The benefits of CUDA output a fairly self-explanatory. The Kernel execution mode might present a few challenges but certainly the GLCompute mode should map fairly well as a strict subset of functionality. The resulting code could then be compiled offline with nvcc or at runtime with NVRTC (which is what we do). Another substantial benefit is to be able to make use of the SPIR-V reflection API, as CUDA infamously has no reflection facilities whatsoever (I've written a few regular expressions to pluck out function definitions and arguments, but I don't exactly trust it).

For OpenCL the benefits aren't as obvious but still substantial. Primarily for targeting platforms that do not support OpenCL 2.1 (such as macOS or Nvidia) but also for running GLCompute kernels compiled from GLSL.

I've thought about taking on this project myself, but it's a substantial body of work I'm unsure if I have the resources for, and if there is interest would be better served as a community effort anyway. I imagine one could start with the MSL as a base, as it has a similar "C/C++"/physical memory style as CUDA/OpenCL. For example,

* A substantial amount of complexity is removed from no longer needing to handle graphics shaders. There's no input or output interfaces, no vertex attributes, no interpolation qualifiers, etc.
* CUDA doesn't have embedded/constexpr samplers, but OpenCL does.
* Specialization constants could map to template arguments in CUDA, or preprocessor macros in OpenCL C (OpenCL C++ has direct support).
* Buffers map pretty much identically as Metal, except CUDA omits address space.
* Textures map a little differently but not substantially so.
* Built-in types map almost identically (CUDA uses "unsigned" instead of 'u' prefix I think is about it).
* CUDA doesn't support swizzling, may have to omit a compile error if it's used.

Is this a realistic project to undertake? Has this been previously ruled out for some reason? Is there any documentation on creating new output implementations? If I do go down this road, would the project be willing to merge them, or would I be dealing with a fork situation?

HansKristian-Work:
[...]

PR #2610, opened 2026-03-17
Feature/opencl

gmeeker:
Implements the OpenCL side of #1602 (no CUDA support) and validates with clang. #1602 is pretty old so I'm curious if there's still interest in this, but I need OpenCL/CUDA and support for legacy reasons. While this backend is smaller due to being compute only, it requires some changes in spirv_glsl to deal with OpenCL being a C-based language. The tests have mostly been copied from MSL, with more tests to cover subgroup emulation. Nvidia and AMD never fully implemented the subgroup extensions this more complicated emulation is necessary.

Please let me know if you think this would be a useful backend and if you have any thoughts about the design. I'm looking at CUDA as well.


HansKristian-Work:
OpenCL support has come up before but it was rejected. There is absolutely no bandwidth to take on another backend. And seeing a 40 ksloc+ PR dumped on me without any prior discussion is absolutely not a great first impression.

gmeeker:
Thanks, I understand. We will go in a different direction.

PR #2611, opened 2026-03-29
Fix support of float atomics in threadgroup memory.

duburcqa:


spirv_msl.cpp
```diff
@@ -11347,6 +11098,20 @@ void CompilerMSL::emit_atomic_func_op(uint32_t result_type, uint32_t result_id,
 	else if (opcode == OpAtomicSMax || opcode == OpAtomicSMin)
 		expected_type = to_signed_basetype(type.width);
 
+	// Metal does not support float atomics in threadgroup memory.  When the
+	// atomic result type is integer but the pointee type is float (CAS-based
```
HansKristian-Work, line 11102: Does SPIRV-Cross do that CAS emulation? I'm not aware of that. I'm really missing a test case here. Just uploading some failing .spv is fine. I can commit that on the side if it's a hassle to run the test suite. I'm not convinced this would work as-is.

HansKristian-Work (changes requested):

spirv_msl.cpp
duburcqa: Thank you for taking the time to review my PR. Actually I moved CAS emulation in Quadrants itself to avoid having to modify SPIRV-Cross, I'm just replacing the original float atomics and the correct instruction to never generate code that would need this feature, so I don't need any changes in SPIRV-Cross.

For reference: https://github.com/Genesis-Embodied-AI/quadrants/commit/79ec04903881792aaa94295d1292addefceb5c44

PR #2615, opened 2026-04-24
HLSL: Zero-init OpUndef Phi incomings to avoid uninitialized reads

xen2:
When an `OpPhi` has `OpUndef` on an incoming edge, SPIRV-Cross emitted a read of the source variable. If that source was uninitialized too, FXC rejects the HLSL with `X4555` or `X4000`.

`OpUndef` allows any value per spec, so we substitute zero:

- `flush_phi`: emit `lhs = 0` when the Phi source is `OpUndef`.
- Loop variable declarations: add `= 0` when `static_expression` is `OpUndef`, via a new `undef_loop_variable_initializer_suffix` helper.

### Test

Added `shaders-hlsl-no-opt/asm/comp/phi-undef-loop.asm.comp` — a minimal SPIR-V loop with `OpUndef` as the preheader Phi incoming.

Without the patch, the generated HLSL is:

```hlsl
uint _19; // not set
for (uint _22 = 0u; _22 < 10u; _19++, _22++) // error X4555: cannot use casts on l-values (due to _19++ which is not set)
{
    _4.Store(0, _19);
}
```

With the patch, `_19` is declared as `uint _19 = 0u;` and FXC accepts it.

### DIFF
--- a/reference/shaders-hlsl-no-opt/asm/comp/phi-undef-loop.asm.comp
+++ b/reference/shaders-hlsl-no-opt/asm/comp/phi-undef-loop.asm.comp
@@ -1,0 +1,18 @@
+static uint _16;
+
+RWByteAddressBuffer _4 : register(u0);
+
+void comp_main()
+{
+    uint _19 = 0u;
+    for (uint _22 = 0u; _22 < 10u; _19++, _22++)
+    {
+        _4.Store(0, _19);
+    }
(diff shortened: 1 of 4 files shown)


xen2:
Still have issues in more complex case + tests failing, let me close that and reopen again once I can confirm it works better.

PR #2618, opened 2026-04-24
Use last rather than first dimension to check for runtime array

Toctave:
This fixes handling of multi-dimensional arrays with the last dimension being unspecified, as in this code:

```
layout(location = 0, binding = 0) buffer MyBuffer {
    float data[16][];
};
```


HansKristian-Work (approved):
Gotta love oneliners that fix things.

Issue #2622, opened 2026-05-06
HLSL: support SPIR-V library modules (no OpEntryPoint, only Export-decorated functions)


castano:
`spirv-cross --hlsl` cannot consume SPIR-V "library" modules, that is, modules with OpCapability Linkage, no OpEntryPoint, and one or more functions decorated with OpDecorate LinkageAttributes "<name>" Export. Such modules are produced by dxc -T lib_6_* and are useful for distributing pre-compiled shader libraries that callers can include from HLSL/GLSL source.

Today, parsing fails immediately:

SPIRV-Cross threw an exception: There is no entry point in the SPIR-V module.

at spirv_parser.cpp:155.

**Repro:**

```
// lib.hlsl
static const uint table[4] = { 10u, 20u, 30u, 40u };

export uint lookup(uint i) {
    return table[i];
}

dxc -T lib_6_3 -spirv -fspv-target-env=universal1.5 lib.hlsl -Fo lib.spv
spirv-cross --hlsl lib.spv
# SPIRV-Cross threw an exception: There is no entry point in the SPIR-V module.
```

(For the constant table to actually carry its values into the SPV, a separate DXC fix is needed tracked at [microsoft/DirectXShaderCompiler#8425](https://github.com/microsoft/DirectXShaderCompiler/issues/8425). The spirv-cross issue here is independent, even with that fix, spirv-cross can't process the module because there's no entry point.)

**Desired Behavior:**

When the module has no entry point but does have Export-decorated functions, emit each exported function as a top-level free function in the target language (HLSL/GLSL), with no [numthreads]/main() boilerplate. Output should be includable as a header by HLSL/GLSL source code.

Expected output for the repro:

```
static const uint table[4] = { 10u, 20u, 30u, 40u };

uint lookup(uint i) {
    return table[i];
}
```

**Motivation:**

I'd like to distribute the [Spark](https://ludicon.com/spark/) codecs as SPIR-V libraries that users can transform to HLSL/GLSL header-only libraries and include in their shader code.

**Proposed approach:**

[...]

PR #2623, opened 2026-05-06
Parser: support SPIR-V library modules without OpEntryPoint

castano:
A "library" SPIR-V module (produced by e.g. `dxc -T lib_6_*`) has  OpCapability Linkage, no OpEntryPoint, and one or more functions decorated with OpDecorate ... LinkageAttributes "<name>" Export. Today the parser rejects such modules with "There is no entry point in the SPIR-V module."

This change updates the parser to:

- During OpDecorate handling, collect the IDs of every Export-decorated symbol into ParsedIR::library_exports. The ID's kind is not yet known at that point (annotations precede definitions per the SPIR-V module layout), so this list tracks both functions and global variables.

- After parsing, populate ParsedIR::library_exported_functions as a typed view containing only the function IDs. Backends iterate this list. Future work could add `library_exported_variables`.

- If OpName was stripped (e.g. by spirv-opt --strip-debug), fall back to the linkage name so the function keeps its declared export name.

- After parsing, if there is no entry point but the module has exported functions, mark `ParsedIR::is_library_module` and designate the first exported function as the default entry point so analyses keyed on `default_entry_point` keep working unchanged. Modules with no exported functions still hit the original "no entry point" error.

This change sets the foundation for incoming changes and has no codegen impact on its own. Library modes now parse instead of throwing an error, but no output is generated yet. Modiels with OpEntryPoint behave identically.


castano:
This is the first set of changes to address issue https://github.com/KhronosGroup/SPIRV-Cross/issues/2622

HansKristian-Work (approved):

PR #2625, opened 2026-05-06
HLSL: emit library exports as free functions

castano:
PR #2623 and PR #2624 updated the parser to accept SPIR-V library modules and ensure per-function CFG analysis covers every exported function. This PR updates the HLSL backend to emit library exports as normal free functions instead of as the shader's entry point.

Stacked on #2624. This is step number 3 of the changes necessary to address issue https://github.com/KhronosGroup/SPIRV-Cross/issues/2622.

### DIFF
--- a/spirv_hlsl.cpp
+++ b/spirv_hlsl.cpp
@@ -3117,7 +3117,12 @@ uint32_t CompilerHLSL::input_vertices_from_execution_mode(SPIREntryPoint &execut
 
 void CompilerHLSL::emit_function_prototype(SPIRFunction &func, const Bitset &return_flags)
 {
-	if (func.self != ir.default_entry_point)
+	// In library mode default_entry_point points at the first exported
+	// function; treat every export as a normal function rather than as the
+	// shader's entry point.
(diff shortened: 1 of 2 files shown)


spirv_hlsl.cpp
```diff
@@ -7148,14 +7153,27 @@ string CompilerHLSL::compile()
...
 
-		emit_function(get<SPIRFunction>(ir.default_entry_point), Bitset());
-		emit_hlsl_entry_point();
+		if (ir.is_library_module)
+		{
+			// Emit each exported function as a normal free function.
+			// emit_function recursively emits callees, so internal helpers
+			// are picked up too.
+			for (auto export_id : ir.library_exported_functions)
+				emit_function(get<SPIRFunction>(export_id), Bitset());
```
HansKristian-Work, line 7162: Is there any potential problem if exports call into other exports? Would it get declared twice?

castano: No, `CompilerGLSL::emit_function` guards re-entry with `if (func.active) return;`. I added an additional test to make sure this was the case. I'll submit it as an additional PR.

HansKristian-Work (approved):

PR #2626, opened 2026-05-06
HLSL: regression test for library-mode emission

castano:
Add a SPIR-V-assembly regression input and reference output that exercise the library-mode HLSL backend introduced in #2623, #2624 and #2625.

The test module:
- has no OpEntryPoint
- declares OpCapability Linkage
- exports two functions (add_one, add_two) via OpDecorate ... LinkageAttributes ... Export
- has add_two call a private (non-exported) helper, exercising the recursive emit_function path that picks up internal helpers from each export's call tree.

The expected HLSL output is three free functions with their original declared names.

Stacked on https://github.com/KhronosGroup/SPIRV-Cross/pull/2625. This is step number 4 of the changes necessary to address issue https://github.com/KhronosGroup/SPIRV-Cross/issues/2622.

### DIFF
--- a/reference/shaders-hlsl/asm/lib/multi-export.asm.lib
+++ b/reference/shaders-hlsl/asm/lib/multi-export.asm.lib
@@ -1,0 +1,17 @@
+uint add_one(uint x)
+{
+    return x + 1u;
+}
+
+uint helper_add(uint a, uint b)
+{
+    return a + b;
+}
+
+uint add_two(uint y)
+{
+    uint _22 = y;
+    uint _23 = 2u;
+    return helper_add(_22, _23);
+}
+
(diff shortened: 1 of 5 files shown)


HansKristian-Work:
Are Linkage exports able to access global resources in any way, like global textures and such?

castano:
I have not exercised that yet, but I will need at some point. Spark has an option to bundle all the inline constants in a buffer, because some compilers have trouble with large blocks of inline constants. In that case the buffer will be exported as a global that the user needs to set up. I'll add the necessary functionality when I get there. I don't expect that to work at this point.

HansKristian-Work (approved):

PR #2628, opened 2026-05-06
GLSL: support SPIR-V library modules

castano:
Extends library-mode SPIR-V support to the GLSL backend, mirroring the HLSL emit and test changes from earlier PRs in this stack.

I think the main question is how to implement `emit_header()` I'm currently emitting a comment and skipping the `#version` and `#extension` directives, but as discussed in https://github.com/KhronosGroup/SPIRV-Cross/issues/2622 this may not be the best choice. I'm open to other suggestions.

I've also added two additional tests mirroring the HLSL tests. The main difference is the way these tests are validated by creating a minimal stub that includes the generated code.

Stacked on https://github.com/KhronosGroup/SPIRV-Cross/pull/2627. Addresses issue https://github.com/KhronosGroup/SPIRV-Cross/issues/2622.


HansKristian-Work:
Rebased it since there were conflicts.

PR #2632, opened 2026-05-11
Library modules export all global variables regardless of SPIR-V version.

castano:
Per discussion in https://github.com/KhronosGroup/SPIRV-Cross/issues/2622, in SPIR-V 1.4 global variables must be declared by the `OpEntryPoint`'s interface list. However, library modules do not have entry points, so default to the previous behavior and export all globals.


HansKristian-Work:
Shouldn't this be rebased? Having 10+ commits like this is just weird.

HansKristian-Work:
Anyway, it's trivial, so can take squash approach.

HansKristian-Work (approved):

castano:
The commit sequence is the result of stacked PRs, but I think for the most part can be ignored, merging the PR squashes the changes.

Issue #2630, opened 2026-05-11
GLSL: possible bug in opcode_is_precision_forwarding_instruction


castano:
[Infinite Flight](https://infiniteflight.com/) is using the [Spark](https://ludicon.com/spark) codecs in their most recent release. Their shader pipeline is a bit different to the one I've been testing, and this has surfaced some interesting bugs. This particular bug that I'm reporting here only prevents them from using the codecs with `mediump` types enabled, so it was not a blocker, but it prevents them from running the encoder at the best possible performance. I would like to fix that.

I was able to isolate the precise location where the bug occurred and why the code generation was wrong.

In one of the codecs I have a function declared like this:

```
mediump uint alu_encode_ept_q160(mediump uint x);
```

The result of this function is a `mediump uint` and is then fed to a `set_bits` helper function:
 
```
mediump uint ep0 = alu_encode_ept_q160(c0.x);
...
set_bits(block, 17, 6, ep0);
```

However, `set_bits` takes a `uint` argument:

```
 int set_bits(inout uvec4 block, int offset, int count, uint bits)
```

And internally does a shift:

```
        block.x |= uint(bits << uint(offset));
```

Note that in this case the offset is 17, which is OK if the result of the shift operation is in full precision. However, due to a combination of bugs in spirv-opt and spirv-cross, the type ends up being `mediump` and a shift greater than 16 is then undefined (on Mali this turns into a shift by 1). I'll report the spirv-cross bug separately, but the solution I came up with is to insert `OpCopyObject`'s to eliminate the `RelaxedPrecision` decoration status during some of the optimizations passes.

However, that solution alone is not enough, because spirv-cross incorrectly forwards the `RelaxedPrecision` decoration in that case. The solution I'm proposing is to simply remove `OpCopyObject` from `opcode_is_precision_forwarding_instruction`, even though that seems to defeat the purpose of the function. Based on the results I'm seeing, I think the comment is possibly wrong:

> SPIR-V doesn't require this, but it's somewhat implied it has to work this way, relaxed precision is only relevant when operating on the IDs, not when shuffling things around.

In this particular case, the main purpose of the `OpCopyObject` instruction was to eliminate the decoration, so this defeats that objective. Is there's some other way of preventing the propagation of `RelaxedPrecision` decorations without using an `OpCopyObject`?

HansKristian-Work:
[...]

PR #2633, opened 2026-05-11
GLSL: Add shifts and bit ops to sensitive operation list.

castano:
Another attempt at fixing https://github.com/KhronosGroup/SPIRV-Cross/issues/2630. Restore the behavior of `OpCopyObject` as a precision forwarding instruction and instead ensure shifts and bit opts are performed at the expected precision by adding them to the sensitive operation list so that they take the `analyze_precision_requirements` code path.


spirv_glsl.cpp
```diff
@@ -12761,7 +12773,7 @@ static bool opcode_is_precision_forwarding_instruction(Op op, uint32_t &arg_coun
 	case OpVectorExtractDynamic:
 	case OpSampledImage:
 	case OpImage:
-	// OpCopyObject intentionally excluded.
+	case OpCopyObject:
```
HansKristian-Work, line 12776: Why is this being reintroduced now where the previous commit didn't? The way I understand it, the root cause is that shifts and bitops weren't subject to precision.

shaders/asm/frag/op-copy-object-relaxed-precision.asm.frag
```diff
@@ -0,0 +1,74 @@
+; Verifies that OpCopyObject's result precision is determined by its OWN
+; RelaxedPrecision decoration, not by the operand's. The SPIR-V spec defines
+; OpCopyObject as: "Make a copy of Operand. There are no pointer dereferences
```
HansKristian-Work, line 3: I'm not sure it's correct to assume OpCopyObject would work like an "expanding" operation if it's not marked with relaxed precision.

E.g. a pattern of:

```
OpDecorate %narrow Relaxed
%narrow = OpSomeOp
%copy = OpCopyObject %narrow
%shift = OpShiftOp ... %copy
```

If %shift is not marked relaxed, I think the implication is that the operation happens at full precision, which means only at that point the relaxed input gets expanded. And that's what I understand this PR to do since I missed the cases of integer bitops.

spirv_glsl.cpp
castano: There are no changes on this line. You are looking at the commit that reverted my previous changes. If you look at the final changes the only difference is the addition of the OpShift and OpBit instructions to the sensitive operation list.

shaders/asm/frag/op-copy-object-relaxed-precision.asm.frag
castano: You are commenting on my previous attempt at solving this problem, which was to insert OpCopyObject instructions to narrow or widen the variables. I was coming at this problem thinking it was a spirv-opt issue, since the problem was introduced during ssa-rewrite and did not occur otherwise. I thought I could solve it with the insertion of the explicit copies, but was surprised to find spirv-cross forwarded the precision status to the destination defeating my attempts to address this. After taking a closer look at why spirv-cross behaved that way I realized you already had the machinery to handle this correctly, and that the only missing element was the addition of the shifts and bit ops to the precision-sensitive operation list, which is what this PR is proposing.

spirv_glsl.cpp
HansKristian-Work: Ok, I see you came to the conclusion in the issue. I review commit by commit rather than everything together, so that's why I got confused.

HansKristian-Work (approved):

Issue #669, opened 2018-08-29
Support gl_DrawID built-in (BuiltInDrawIndex) for MSL.


billhollings:
PR #668 added support for the SPV_KHR_shader_draw_parameters extension, but left out an implementation of the `gl_DrawID` built-in (BuiltInDrawIndex) for MSL.

Metal only supports a single draw when using indirect buffers, and repeated draw calls with increasing offsets are required to emulate the multiple draw calls of OpenGL & Vulkan.

The use of a buffer to hold the gl_DrawID value is likely the correct approach (ie `gl_DrawID [[buffer(n)]]`)...and needs to have an index that can be specified from the outside...similar to how other MSL resource indices are specified in the `CompilerMSL` config API.

denisk-dev:
Technically, this could be closed with the addition of #2634 . But could also wait until https://github.com/KhronosGroup/MoltenVK/pull/2742 is merged.

squidbus:
I believe this can be closed now.

PR #2634, opened 2026-05-13
MSL: add gl_DrawID (DrawIndex) support for MacOS

denisk-dev:
Related to these reported issues:

SPIRV-Cross: #669 
MoltenVK: [1743](https://github.com/KhronosGroup/MoltenVK/issues/1743)
 
Corresponding MoltenVK change:
https://github.com/denisk-dev/MoltenVK/tree/add-gl-draw-id

Kitten Space Agency (KSA) is using Vulkan and gl_DrawID is used in the rocket parts shaders.

### DIFF
--- a/reference/opt/shaders-msl/desktop-only/vert/shader-draw-parameters.desktop.for-tess.vert
+++ b/reference/opt/shaders-msl/desktop-only/vert/shader-draw-parameters.desktop.for-tess.vert
@@ -8,13 +8,13 @@ struct main0_out
     float4 gl_Position;
 };
 
-kernel void main0(uint3 gl_GlobalInvocationID [[thread_position_in_grid]], uint3 spvStageInputSize [[grid_size]], uint3 spvDispatchBase [[grid_origin]], device main0_out* spvOut [[buffer(28)]])
+kernel void main0(constant uint *gl_DrawID [[buffer(19)]], uint3 gl_GlobalInvocationID [[thread_position_in_grid]], uint3 spvStageInputSize [[grid_size]], uint3 spvDispatchBase [[grid_origin]], device main0_out* spvOut [[buffer(28)]])
 {
     device main0_out& out = spvOut[gl_GlobalInvocationID.y * spvStageInputSize.x + gl_GlobalInvocationID.x];
     if (any(gl_GlobalInvocationID >= spvStageInputSize))
         return;
     uint gl_BaseVertex = spvDispatchBase.x;
     uint gl_BaseInstance = spvDispatchBase.y;
-    out.gl_Position = float4(float(int(gl_BaseVertex)), float(int(gl_BaseInstance)), 0.0, 1.0);
+    out.gl_Position = float4(float(int(gl_BaseVertex)), float(int(gl_BaseInstance)), float(*gl_DrawID), 1.0);
 }
 
--- a/reference/opt/shaders-msl/desktop-only/vert/shader-draw-parameters.desktop.for-tess.vert
+++ b/reference/opt/shaders-msl/desktop-only/vert/shader-draw-parameters.desktop.for-tess.vert
@@ -8,13 +8,14 @@ struct main0_out
     float4 gl_Position;
 };
 
-kernel void main0(constant uint *gl_DrawID [[buffer(19)]], uint3 gl_GlobalInvocationID [[thread_position_in_grid]], uint3 spvStageInputSize [[grid_size]], uint3 spvDispatchBase [[grid_origin]], device main0_out* spvOut [[buffer(28)]])
+kernel void main0(constant uint* spvDrawIndex [[buffer(19)]], uint3 gl_GlobalInvocationID [[thread_position_in_grid]], uint3 spvStageInputSize [[grid_size]], uint3 spvDispatchBase [[grid_origin]], device main0_out* spvOut [[buffer(28)]])
 {
     device main0_out& out = spvOut[gl_GlobalInvocationID.y * spvStageInputSize.x + gl_GlobalInvocationID.x];
     if (any(gl_GlobalInvocationID >= spvStageInputSize))
         return;
     uint gl_BaseVertex = spvDispatchBase.x;
     uint gl_BaseInstance = spvDispatchBase.y;
-    out.gl_Position = float4(float(int(gl_BaseVertex)), float(int(gl_BaseInstance)), float(*gl_DrawID), 1.0);
+    uint gl_DrawID = *spvDrawIndex;
+    out.gl_Position = float4(float(int(gl_BaseVertex)), float(int(gl_BaseInstance)), float(int(gl_DrawID)), 1.0);
 }
 
--- a/reference/opt/shaders-msl/desktop-only/vert/shader-draw-parameters.desktop.vert
+++ b/reference/opt/shaders-msl/desktop-only/vert/shader-draw-parameters.desktop.vert
@@ -8,10 +8,11 @@ struct main0_out
     float4 gl_Position [[position]];
 };
 
-vertex main0_out main0(uint gl_BaseVertex [[base_vertex]], uint gl_BaseInstance [[base_instance]], constant uint *gl_DrawID [[buffer(19)]])
+vertex main0_out main0(constant uint* spvDrawIndex [[buffer(19)]], uint gl_BaseVertex [[base_vertex]], uint gl_BaseInstance [[base_instance]])
 {
     main0_out out = {};
-    out.gl_Position = float4(float(int(gl_BaseVertex)), float(int(gl_BaseInstance)), float(*gl_DrawID), 1.0);
+    uint gl_DrawID = *spvDrawIndex;
+    out.gl_Position = float4(float(int(gl_BaseVertex)), float(int(gl_BaseInstance)), float(int(gl_DrawID)), 1.0);
     return out;
 }
 
--- a/reference/opt/shaders-msl/desktop-only/vert/shader-draw-parameters.desktop.vert
+++ b/reference/opt/shaders-msl/desktop-only/vert/shader-draw-parameters.desktop.vert
@@ -8,10 +8,10 @@ struct main0_out
     float4 gl_Position [[position]];
 };
 
-vertex main0_out main0(uint gl_BaseVertex [[base_vertex]], uint gl_BaseInstance [[base_instance]])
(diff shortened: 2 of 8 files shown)


denisk-dev:
This is my first time contributing. Please, be gentle!

HansKristian-Work:
At minimum, add some tests demonstrating that the feature works.

denisk-dev:
Added the tests. Please, double-check that I updated the main correctly between my previous pull request and now.

As a note, I also do a full test with KSA to verify that it still runs and the rocket parts are showing up correctly.

reference/opt/shaders-msl/desktop-only/vert/shader-draw-parameters.desktop.vert
```diff
@@ -8,10 +8,10 @@ struct main0_out
     float4 gl_Position [[position]];
 };
 
-vertex main0_out main0(uint gl_BaseVertex [[base_vertex]], uint gl_BaseInstance [[base_instance]])
+vertex main0_out main0(uint gl_BaseVertex [[base_vertex]], uint gl_BaseInstance [[base_instance]], constant uint *gl_DrawID [[buffer(19)]])
```
HansKristian-Work: Why is DrawID a buffer. Is it not supported natively by Metal and it has to be emulated somehow?

spirv_msl.cpp
```diff
@@ -13821,7 +13835,9 @@ string CompilerMSL::member_attribute_qualifier(const SPIRType &type, uint32_t in
 				return string(" [[") + builtin_qualifier(builtin) + "]]";
 
 			case BuiltInDrawIndex:
-				SPIRV_CROSS_THROW("DrawIndex is not supported in MSL.");
+                if (msl_options.vertex_for_tessellation)
```
HansKristian-Work: Why no attribute for vertex for tessellation? I would imagine a similar hack is needed there too.

reference/opt/shaders-msl/desktop-only/vert/shader-draw-parameters.desktop.vert
HansKristian-Work: It would make more sense to me if this was not a pointer, but a direct argument, but I'd need to know what the emulation strategy for this feature is before we commit to a solution.

denisk-dev: From the #669 and some research, at least pre-Metal4 does not have a native DrawID. 

I originally had it as a direct argument, but (if I remember correctly) Metal complained that it needs to be a pointer. In the equivalent MoltenVK addition, we use drawIdx already available in the draw count loop. 

https://github.com/KhronosGroup/MoltenVK/pull/2742

spirv_msl.cpp
denisk-dev: KSA doesn't use DrawID with the tessellation shaders, so I couldn't confirm. I can comment out the "if" statement, if you think that's correct. I believe I would need to update the test case that runs with tessellation as well.

reference/opt/shaders-msl/desktop-only/vert/shader-draw-parameters.desktop.vert
cdavis5e: > Why is DrawID a buffer. Is it not supported natively by Metal and it has to be emulated somehow?

It isn't. If it were, I would've implemented it in #668.

denisk-dev: How are the changes looking for you? Do they make sense? I think I originally stumbled upon your commit from #668. It's a completely game-breaking issue for KSA right now, which is why I decided to try and implement it. Between this commit and the MoltenVK one, it seems to work for me and a couple other people from the KSA discord

HansKristian-Work: I don't have a problem with DrawID being loaded from a buffer if it has to be that way to be clear, I just want to make sure that the consumers of this API (i.e. MoltenVK) think this is the way forward.

denisk-dev: I haven't got any responses/comments on the MoltenVK side yet, possibly because it's currently failing the checks without this change in SPIRV-Cross

spirv_msl.cpp
cdavis5e: I believe H-K is correct here. It should work exactly the same in that case--all that option does is cause the vertex shader to be emitted as a compute `kernel` instead of a `vertex` function.

cdavis5e (changes requested):

reference/opt/shaders-msl/desktop-only/vert/shader-draw-parameters.desktop.vert
squidbus: Haven't touched the MVK side yet myself because it was waiting on this yes, the approach makes sense to me.

Also to be clear, Metal does not have multi-draw at all, hence why there's no native draw ID. The way MVK does it is to just do multiple individual draw calls in a loop.

spirv_msl.cpp
```diff
@@ -18120,7 +18136,18 @@ string CompilerMSL::builtin_to_glsl(BuiltIn builtin, StorageClass storage)
 			SPIRV_CROSS_THROW("BaseInstance requires Metal 1.1 and Mac or Apple A9+ hardware.");
 		}
 	case BuiltInDrawIndex:
-		SPIRV_CROSS_THROW("DrawIndex is not supported in MSL.");
+            // Not positive on the Metal version or if ios needs it
+            if (msl_options.supports_msl_version(1, 1) &&
+                // (msl_options.ios_support_base_vertex_instance || msl_options.is_macos()))
+                msl_options.is_macos())
```
squidbus: I don't think any of this conditional stuff here is necessary, it's a fully emulated parameter so we don't need to do support checks.

reference/opt/shaders-msl/desktop-only/vert/shader-draw-parameters.desktop.vert
HansKristian-Work: hhhhhhng. Well, ok. If this is considered the right approach, I'll review this.

reference/opt/shaders-msl/desktop-only/vert/shader-draw-parameters.desktop.for-tess.vert
```diff
@@ -8,13 +8,13 @@ struct main0_out
     float4 gl_Position;
 };
 
-kernel void main0(uint3 gl_GlobalInvocationID [[thread_position_in_grid]], uint3 spvStageInputSize [[grid_size]], uint3 spvDispatchBase [[grid_origin]], device main0_out* spvOut [[buffer(28)]])
+kernel void main0(constant uint *gl_DrawID [[buffer(19)]], uint3 gl_GlobalInvocationID [[thread_position_in_grid]], uint3 spvStageInputSize [[grid_size]], uint3 spvDispatchBase [[grid_origin]], device main0_out* spvOut [[buffer(28)]])
```
HansKristian-Work: There are no tests that attempt to read DrawID from inside a function. Usually the codegen changes for leaf functions.

E.g. `uint get_draw_id() { return gl_DrawID; }`

spirv_msl.cpp
```diff
@@ -282,6 +282,20 @@ void CompilerMSL::build_implicit_builtins()
 	needs_point_size_output =
 	    msl_options.enable_point_size_builtin && msl_options.enable_point_size_default &&
 	    entry_point_is_vertex();
+    
+    if (needs_draw_id_buffer_def) {
```
HansKristian-Work: This feels like it's in the wrong place, and I question where you got this code pattern from. Did you use other code as an example on how to do this? If so, where was it picked from?

spirv_msl.cpp
```diff
@@ -18120,7 +18134,9 @@ string CompilerMSL::builtin_to_glsl(BuiltIn builtin, StorageClass storage)
 			SPIRV_CROSS_THROW("BaseInstance requires Metal 1.1 and Mac or Apple A9+ hardware.");
 		}
 	case BuiltInDrawIndex:
-		SPIRV_CROSS_THROW("DrawIndex is not supported in MSL.");
+            // since it's fully emulated, no need to do the checks
+            needs_draw_id_buffer_def = true;
```
HansKristian-Work: This looks very wrong. Why is it flagging that buffer_def is needed after the functions have been declared? I would expect this to be flagged earlier, or maybe just checked with a is_active_builtin() or something like that when emitting workarounds.

spirv_msl.cpp
```diff
@@ -18456,7 +18472,7 @@ string CompilerMSL::builtin_type_decl(BuiltIn builtin, uint32_t id)
 	case BuiltInBaseInstance:
 		return "uint";
 	case BuiltInDrawIndex:
-		SPIRV_CROSS_THROW("DrawIndex is not supported in MSL.");
+        return "constant uint";
```
HansKristian-Work: constant uint feels like a smell, that's a storage qualifier which doesn't really belong here.

spirv_msl.cpp
```diff
@@ -18232,7 +18248,7 @@ string CompilerMSL::builtin_qualifier(BuiltIn builtin)
 	case BuiltInBaseInstance:
 		return "base_instance";
 	case BuiltInDrawIndex:
-		SPIRV_CROSS_THROW("DrawIndex is not supported in MSL.");
+        return string("buffer(") + std::to_string(msl_options.draw_id_buffer_index) + ")";
```
HansKristian-Work: This doesn't look consistent with the other code. I would expect the buffer-ness to be resolved when the function prototype is declared. Should this function even be called for DrawIndex if it's emulated like this?

HansKristian-Work (changes requested):
This feels subtly wrong in many ways, but it's been forever since I looked at this code, so I don't have an immediate suggestion on how to fix it. The feature is trivial enough I'll just try to do the fixes locally once I've re-groked how this stuff is supposed to be done.

HansKristian-Work (approved):

oscarbg:
@denisk-dev can share if can test/already tested "detroit become human" (the free demo is enough) with your moltenvk branch to see if it works? missing support for DrawIndex on MVK was the last known issue..

denisk-dev:
@oscarbg Let me update my MoltenVK repo to now point to the newly merged hash and you could give it a try? If not, I can look up that demo and see if I can run it

PR #2639, opened 2026-06-06
Add explicit FLOAT32, INT32, and UINT32 to MSLShaderVariableFormat

alexdremov:
This PR addresses failures in MoltenVK:
```
[mvk-error] VK_ERROR_INITIALIZATION_FAILED: Render pipeline compile failed (Error code 3):
   Fragment input(s) `user(locn10)` mismatching vertex shader output type(s) or not written by vertex shader.
```

MSLShaderVariableFormat previously only supported `ANY32`, forcing to default to UInt. This causes a Metal compiler crash (invalid return type or mismatching type) if the fragment shader actually expects a float input. By introducing explicit 32-bit types, MoltenVK can instruct SPIRV-Cross on the exact data type to use.

* Added MSL_SHADER_VARIABLE_FORMAT_FLOAT32, INT32, and UINT32 to the MSLShaderVariableFormat enum and the corresponding SPVC_MSL_SHADER_VARIABLE_FORMAT_* C-API definitions.
* Relaxed the dummy output generation condition in add_interface_block() to no longer strictly require capture_output_to_buffer. Previously, SPIRV-Cross would only generate missing interface variables if the pipeline was explicitly capturing output. This prevented MoltenVK from padding StageOut structs in standard rasterization pipelines, leading to linker failures when a Fragment shader expected an input that the Vertex shader didn't explicitly write.
* Replaced the capture_output_to_buffer requirement with `get_entry_point().model != ExecutionModelFragment`. This ensures that dummy output generation safely applies to all standard Vertex, Geometry, and Tessellation stages to satisfy downstream Metal linkers, while strictly preventing generic [[user(locn)]] variables from being illegally injected into the return structs of Fragment shaders.

**AI use disclosure**:
I am not a computer graphics expert and AI was used to help debug and prepare fixes for this issue. I confirm that I have manually tested this and a scene that was crashing before in "Detroit: Become Human" is not crashing after this fix.


spirv_msl.cpp
```diff
@@ -4648,7 +4657,7 @@ uint32_t CompilerMSL::add_interface_block(StorageClass storage, bool patch)
 		}
 	}
 
-	if (capture_output_to_buffer && storage == StorageClassOutput)
+	if (storage == StorageClassOutput && get_entry_point().model != ExecutionModelFragment)
```
alexdremov, line 4660: maybe `is_vertex_like_shader()` can be used here, I am not sure

cdavis5e:
This seems [awfully familiar...](/KhronosGroup/SPIRV-Cross/pull/2639)

alexdremov:
> This seems [awfully familiar...](/KhronosGroup/SPIRV-Cross/pull/2639)

Hey! Seems like you have linked this PR. Perhaps you meant something else.

spirv_msl.cpp
HansKristian-Work: This change makes no sense since the comments below explicitly talk about captured output, yes this removes that condition.

HansKristian-Work:
> MSLShaderVariableFormat previously only supported ANY32, forcing to default to UInt. This causes a Metal compiler crash (invalid return type or mismatching type) if the fragment shader actually expects a float input

?! I'm very confused. Please paste the vertex and fragment shaders in question that fail.

Is this attempting to workaround a broken game, or is this about correctness against Vulkan spec?

PR #2640, opened 2026-06-24
MSL: Add reversed depth viewport emulation

vkedwardli:
This add an opt-in MSL backend option to emulate reversed depth viewports by inverting clip-space Z in vertex-like shaders. 

The option originated from the MoltenVK workaround discussion in https://github.com/KhronosGroup/MoltenVK/pull/2760
On some affected Metal implementations, passing a reversed depth viewport directly to Metal can fail. The intended consumer is MoltenVK, which can normalize the Metal viewport depth range and ask SPIRV-Cross to emulate the missing reversed-depth transform in generated MSL

When enabled, SPIRV-Cross adds a `constant uint&` mask argument and emits:
```cpp
position.z = position.w - position.z
```
for viewports whose mask bit is set.

**Changes:**
- CompilerMSL::Options::emulate_reversed_depth_viewport
- CompilerMSL::Options::reversed_depth_viewport_buffer_index
- C API options for both fields
- MSL emission for the clip-space Z inversion in vertex-like shaders


spirv_cross_c.cpp
```diff
@@ -783,6 +783,14 @@ spvc_result spvc_compiler_options_set_uint(spvc_compiler_options options, spvc_c
 	case SPVC_COMPILER_OPTION_MSL_FORCE_FRAGMENT_WITH_SIDE_EFFECTS_EXECUTION:
 		options->msl.force_fragment_with_side_effects_execution = value != 0;
 		break;
+
```
HansKristian-Work: If you update the C API you have to bump the SPVC_C_API_VERSION_MINOR define and corresponding one in CMakeLists.txt.

spirv_msl.cpp
```diff
@@ -13544,6 +13549,19 @@ void CompilerMSL::emit_fixup()
 
 		if (is_vertex_like_shader() && !qual_pos_var_name.empty())
 		{
+			if (msl_options.emulate_reversed_depth_viewport)
+			{
+				if (qual_viewport_idx_var_name.empty())
```
HansKristian-Work, line 13554: This seems broken if the shader doesn't declare ViewportIndex on its own.

HansKristian-Work (changes requested):
Needs test coverage for cases with and without viewport index being declared in shader.

spirv_cross_c.cpp
vkedwardli: This option is mainly intended for Metal backend integrators like MoltenVK, so I’d rather avoid expanding the C API unless there’s a clear use case for it, removed the C API edits. 

Thank you for pointing this out

spirv_msl.cpp
vkedwardli: If the shader writes `ViewportIndex`, the emulation uses that value to select the mask bit. If it does not, the primitive uses viewport 0 by definition, so the generated code checks bit 0 directly.

I also added a comment to document that assumption.
`// If ViewportIndex is not written, the primitive uses viewport 0.`

vkedwardli:
Added tests for both cases now:
- no `ViewportIndex` declared, which should use mask bit 0
- `ViewportIndex` declared/written, which should select the mask bit from that value

Both references also verify that the generated MSL emits the clip-space Z inversion (`z = w - z`)

spirv_msl.cpp
HansKristian-Work: Right, ViewportIndex is an output, not an input, forgot about that ...

HansKristian-Work (approved):

PR #2649, opened 2026-07-13
MSL: Add ray tracing shader support

dttdrv:
Adds MSL lowering for Vulkan ray tracing stages using Metal visible function tables and intersection queries. Payload, callable, hit attribute, shader record, implicit buffer and acceleration structure metadata are carried across indirect stages. Ray query flags, shader binding table offsets and `SPV_KHR_ray_tracing_position_fetch` are covered.

The public C API and ABI are unchanged. Motion blur remains out of scope and would need a separate intersector path.

Debug, Release with warnings as errors, ASan/UBSan, namespace override and MSL-disabled builds pass. The normal and optimized MSL suites each pass 541/541, and all 18 locally runnable CTests pass. The six Metal compiler tests are unavailable with the local Command Line Tools setup.

MoltenVK integration: KhronosGroup/MoltenVK#2771


oscarbg:
small comment..
just seeing all SPIR-V raytracing extensions there is the reorder ext which isn't supported in Metal (and HW) yet.. and also the SPV_KHR_ray_tracing_position_fetch ext.. part of VK_KHR_ray_tracing_position_fetch..
curious if this functionality is available on MSL currently?

dttdrv:
@oscarbg Not directly. MSL can read per-primitive data stored in an acceleration structure, but it cannot read the original triangle positions itself. I tested this on M5 Pro and we can use per-primitive data to support position fetch for ray queries and pipeline hits, including AS copy and refit. It needs changes in both MoltenVK’s AS build path and SPIRV-Cross. Reorder is still unsupported.

dttdrv:
@HansKristian-Work, would you be willing to review this MSL ray tracing work when you have time? The local Release build and test suites pass. Thanks.

dttdrv:
I went back through the position fetch lowering and kept the public C API and ABI unchanged. Debug, Release, ASan/UBSan, namespace override and both MSL suites pass. MoltenVK #2771 now pins `62db8c83`

HansKristian-Work:
> would you be willing to review this MSL ray tracing work when you have time?

The change to MSL.cpp is like 3k in one go. That is not something you just throw up for review without discussing with upstream first and expect to glide through. I'm not sure when I will ever have time or capacity for that. Sorry, but I'm very tired of being the blocker for emulated feature number 132214234 in MoltenVK, maybe MVK needs to fork and maintain this stuff themselves.

dttdrv:
I went back through the compiler/runtime boundary. The MSL change combines generic ray-query fixes with MoltenVK-specific VFT dispatch and a host metadata ABI. That should not have been presented as one upstreamable compiler change without discussing the architecture first.

I am withdrawing this PR. MoltenVK #2771 will treat the full VFT lowering as an explicit experimental fork dependency pending MoltenVK maintainer acceptance, not as work SPIRV-Cross is expected to merge. The generic query issues—proceed operand handling, candidate-opaque inversion, low-eight-bit cull masks, and ray-flags state—will only be proposed separately after prior discussion.

PR #2650, opened 2026-07-18
MSL: add Vulkan mesh shader emulation

dttdrv:
Adds the SPIRV-Cross side of taskless Vulkan mesh shader support for MoltenVK #2777. It covers taskless payload remapping, primitive topology, output initialization, 64-bit varyings, position fixups, output spill and capture/replay, and dispatch base feedback.

The integration uses the existing C++ compiler API. The public C API and ABI are unchanged. Spill layout version 2 is the only accepted layout.

Debug and Release with warnings as errors pass. All 18 locally runnable CTests and the normal and optimized MSL suites pass. The six Metal compiler tests are unavailable with the local Command Line Tools setup.

MoltenVK #2777 pins `aaeadbdf`


dttdrv:
I went back through this before review. The added C API and ABI surface, unused result fields and old spill layout are gone, and there's a focused C++ test now. Debug, Release with warnings as errors, all runnable CTests and both MSL suites pass. MoltenVK #2777 pins `aaeadbdf`

HansKristian-Work:
Another 3K addition to MSL to ... emulate a feature a Metal already supports? No, no ...

dttdrv:
Looking back at this, I actually agree that this and #2649 should not have arrived at this size. I definitely could have taken a better route. I did not take into account that it is mostly one person reviewing these, so I apologize for that.

This still uses Metal’s native object and mesh stages rather than emulating mesh shading itself. The problem is that too much of the Vulkan taskless dispatch, spill and output handling ended up in SPIRV-Cross. I need to rethink that split before asking you to review it.

If you still want to help with this, I’ll send a short outline first. I can then keep any compiler changes small and separate, and move the MoltenVK-specific work out of this PR.

HansKristian-Work:
No, I don't want mesh shader emulation in upstream. If MoltenVK intends to make their life hard by emulating it, they need to maintain a downstream patch.

PR #2652, opened 2026-07-20
(GLSL) Support SPV_KHR_workgroup_memory_explicit_layout

baggins183:
Map SPV_KHR_workgroup_memory_explicit_layout to GL_EXT_shared_memory_block when emitting GLSL

### DIFF
--- a/spirv_cross.cpp
+++ b/spirv_cross.cpp
@@ -88,14 +88,18 @@ bool Compiler::variable_storage_is_aliased(const SPIRVariable &v)
 	bool image = type.basetype == SPIRType::Image;
 	bool counter = type.basetype == SPIRType::AtomicCounter;
 	bool buffer_reference = type.storage == StorageClassPhysicalStorageBuffer;
+	bool shared_block = type.storage == spv::StorageClassWorkgroup && uses_workgroup_memory_explicit_layout;
 
 	bool is_restrict;
 	if (ssbo)
 		is_restrict = ir.get_buffer_block_flags(v).get(DecorationRestrict);
+	else if (shared_block)
+		// When more than one shared block is present, all shared blocks must be decorated Aliased
+		is_restrict = !has_decoration(v.self, DecorationAliased);
 	else
 		is_restrict = has_decoration(v.self, DecorationRestrict);
 
-	return !is_restrict && (ssbo || image || counter || buffer_reference);
+	return !is_restrict && (ssbo || image || counter || buffer_reference || shared_block);
 }
 
 bool Compiler::block_is_control_dependent(const SPIRBlock &block)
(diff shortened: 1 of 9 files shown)


baggins183:
First time changing something in spirv-cross so I probably missed something, let me know. Is there anything needed for reflection?

spirv_cross.cpp
```diff
@@ -5266,6 +5273,21 @@ bool Compiler::type_is_opaque_value(const SPIRType &type) const
...
+	if (var.storage != StorageClassWorkgroup)
+		return false;
+	auto &type = get<SPIRType>(var.basetype);
+	auto *type_meta = ir.find_meta(type.self);
+	if (type_meta && type_meta->decoration.decoration_flags.get(DecorationBlock)) {
+		uses_workgroup_memory_explicit_layout = true;
+		return true;
+	}
+	// If one workgroup variable is a block, then all must be
+	assert(!uses_workgroup_memory_explicit_layout);
```
HansKristian-Work: Don't use asserts here. A SPIRV_CROSS_THROW is okay since it violates SPIR-V rules.

spirv_cross.cpp
```diff
@@ -5266,6 +5273,21 @@ bool Compiler::type_is_opaque_value(const SPIRType &type) const
 	                         type.basetype == SPIRType::Sampler || type.basetype == SPIRType::Tensor);
 }
 
+bool Compiler::check_is_shared_memory_block(const SPIRVariable &var)
+{
+	if (var.storage != StorageClassWorkgroup)
+		return false;
+	auto &type = get<SPIRType>(var.basetype);
+	auto *type_meta = ir.find_meta(type.self);
+	if (type_meta && type_meta->decoration.decoration_flags.get(DecorationBlock)) {
```
HansKristian-Work: Replace all of this with has_decoration().

spirv_glsl.cpp
```diff
@@ -723,6 +723,10 @@ void CompilerGLSL::find_static_extensions()
 	for (auto &ext : ir.declared_extensions)
 		if (ext == "SPV_NV_fragment_shader_barycentric")
 			barycentric_is_nv = true;
+
+	if (uses_workgroup_memory_explicit_layout) {
```
HansKristian-Work: Don't need braces here.

Also, requiring the extension here seems too early. It's only until codegen that you find out that you need to emit shared memory blocks. It would make more sense to just call require_extension_internal when emitting a shared block, and then there is nothing to worry about w.r.t. ordering.

HansKristian-Work (changes requested):
Some nits, but overall direction is sound.

baggins183:
Thanks. I removed uses_shared_memory_blocks because it was kind of pointless.

FYI I ran into some descriptor heap tests throwing errors (also happens on main). Also there was some arm tensor mismatch I just updated. Pretty sure tests are clean otherwise

baggins183:
shaders-no-opt/vulkan/frag/descriptor-heap.vk.nocompat.spv16.frag
and
shaders-no-opt/vulkan/frag/descriptor-heap.heap-legacy-mapping.vk.nocompat.spv16.frag

baggins183:
actually now that I think about it my glslangValidator is probably outdated

HansKristian-Work (approved):

PR #2654, opened 2026-07-23
Complete MSL lowering for geometry shaders and transform feedback

minamorl:
## Summary

This draft continues the geometry-shader and transform-feedback MSL work originally developed in the @js6i and @gpx1000 branches and used by the paired MoltenVK change.

- preserves the arrayed `gl_PerVertex` stage input used by geometry shaders
- copies position and user varyings into the mesh wrapper by input index
- keeps geometry access chains rooted at `in.gl_in[i]`
- restores pointer types required by transform-feedback buffer arguments
- emits the vertex/index hooks required by transform-feedback compute kernels without undeclared dispatch-base references
- captures every vertex in a triangle-list dispatch, including the middle vertex
- adapts transform-feedback layout queries to the current MSL backend API
- materializes flattened `gl_PerVertex` CullDistance output blocks so generated MSL is structurally valid

## Background

This work was resumed while rebasing older MoltenVK geometry-shader and transform-feedback WIP branches for an Apple-Silicon Vulkan stack used by a Zeno guest and DXVK. Compiler success alone was a false-green gate: the generated shaders could still consume the wrong geometry input or write incomplete transform-feedback output.

The fixes here preserve per-vertex array/pointer information and provide the hooks used by the MoltenVK execution and resource-binding side. They do not claim complete geometry-shader or CullDistance support.

## Paired change

The MoltenVK execution, resource-binding, feature-reporting, and pipeline-cache side is in KhronosGroup/MoltenVK#2786. That PR currently pins this branch at `3c567d2b554736fa4b86521bd14b303e225528bf`.

## Validation

- rebased from SPIRV-Cross `main` at `6c09849f`
- standalone SPIRV-Cross CLI rebuild at the pinned commit
- `cull-distance-block.vert` output matches its new MSL golden reference exactly
- integrated MoltenVK macOS private-API package build succeeds
- semantic geometry readback: asymmetric input vertices produce the expected output
- transform-feedback readback: three ordered `vec4` values match exactly; counter is 48 bytes
- paired host diagnostic result: **18 GREEN / 1 RED / 7 N/A**

The local Xcode installation does not currently include the separately downloadable Metal Toolchain component, so the new CullDistance golden was regression-compared but not compiled with the standalone `metal` command in this update.

## Known limitations

This remains a draft.

- Multi-stream geometry shaders are unsupported by the paired Metal mesh/XFB path.
- CullDistance block flattening now generates structurally valid MSL, but `[[user(cull0)]]` is not Vulkan's primitive CullDistance semantic. The paired MoltenVK implementation therefore reports `shaderCullDistance = false`.
- DXVK's exact FL 11_1 device request remains RED because it asks for unsupported `shaderCullDistance` and `geometryStreams`.
- Additional shader/topology coverage and upstream test integration remain necessary before merge readiness.

### DIFF
--- a/reference/shaders-msl/vert/cull-distance-block.vert
+++ b/reference/shaders-msl/vert/cull-distance-block.vert
@@ -1,0 +1,75 @@
+#pragma clang diagnostic ignored "-Wmissing-prototypes"
+#pragma clang diagnostic ignored "-Wmissing-braces"
+
+#include <metal_stdlib>
+#include <simd/simd.h>
+
+using namespace metal;
+
+template<typename T, size_t Num>
+struct spvUnsafeArray
+{
+    T elements[Num ? Num : 1];
+    
+    thread T& operator [] (size_t pos) thread
+    {
+        return elements[pos];
+    }
+    constexpr const thread T& operator [] (size_t pos) const thread
+    {
+        return elements[pos];
+    }
+    
+    device T& operator [] (size_t pos) device
+    {
+        return elements[pos];
+    }
+    constexpr const device T& operator [] (size_t pos) const device
+    {
+        return elements[pos];
+    }
+    
+    constexpr const constant T& operator [] (size_t pos) const constant
+    {
+        return elements[pos];
+    }
+    
+    threadgroup T& operator [] (size_t pos) threadgroup
+    {
+        return elements[pos];
+    }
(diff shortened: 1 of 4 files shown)


HansKristian-Work:
The previous iteration of incomplete GS/XFB hackery was rejected. What changes here warrant re-opening that old wound?

PR #2655, opened 2026-07-23
MSL: don't type a resource as depth/comparison from provably dead specialization branches

JackTYM:
## Problem

`CombinedImageSamplerUsageHandler` types a resource as Metal's `depth2d<T>` (comparison-only) the moment it sees *any* `OpImageSampleDrefImplicitLod`/etc. referencing it anywhere in a function — with no regard for whether that instruction sits in a branch that's actually reachable.

Shader compilers commonly emit a single shader that branches, based on a specialization-constant-derived condition, between a depth-compare sample and a regular sample of the *same* resource — e.g. to support a runtime-togglable comparison-sampler feature without compiling a separate shader per mode. This is legal, spec-compliant SPIR-V. When the specialization values supplied at compile time make the depth-compare branch provably dead, the resource is still force-typed `depth2d<T>` in the generated MSL. `depth2d` doesn't return real per-channel data the way `texture2d` does, so the regular sampling on the branch that actually executes comes back corrupted (in our case, a texture's alpha channel read back as a constant `1.0` instead of real per-texel data).

This is a known, previously-flagged gap — see the existing comment at `spirv_cross.cpp:1977-1979`:
> "Ideally, perhaps traverse the CFG instead of all blocks in order to eliminate dead blocks..."

## Fix

**Commit 1** adds a small, conservative constant-propagation utility (`evaluate_constant_scalar_expression`) that resolves a value through a bounded subset of pure integer/boolean opcodes, including inlining calls to simple (single-block, non-recursive) callee functions — the common shape of the small specialization-constant "getter" helpers these branches key off of. `find_specialization_dead_blocks()` uses this once per compile to determine which blocks are unreachable given the known specialization, and `CombinedImageSamplerUsageHandler` now skips instructions in those blocks entirely via a new `set_current_block()` override, rather than only affecting its direct callers.

**Commit 2** fixes a related inconsistency this surfaced under real-world testing: `to_function_name()`/`to_function_args()` decided whether to emit `.sample_compare(...)` (and its dref argument) purely from the *originating SPIR-V opcode*, independent of whether the resource was actually typed `depth2d<T>` in the output. Once resource typing correctly excludes dead-branch usage, a Dref instruction can remain textually present — on a path that's provably never taken — referencing a resource now typed `texture2d<T>`. `sample_compare()` doesn't exist on `texture2d<T>`, so without this fix Metal's own compiler rejected the generated source outright. Both call-name and dref-argument emission now consult `is_depth_image()`, the same source of truth `comparison_ids` already encodes.

## Testing

- New regression test: `shaders-msl-no-opt/asm/frag/spec-constant-dead-branch-comparison.asm.frag` — reproduces the pattern in isolation (a spec-constant-gated branch between a Dref sample and a regular sample of the same resource, with the default spec constant value making the Dref branch dead). Emits `depth2d<float>` without this change, `texture2d<float>` with it.
- Full `shaders-msl` and `shaders-msl-no-opt` suites, both with and without `--opt`, pass with no other changes — including the existing `image-type-normal-comparison-usage.asm.frag`, which covers the case where both sampling modes are genuinely live (a runtime, non-specialization condition), confirming this fix is correctly scoped to only the provably-dead case.
- Verified end-to-end against a real, affected shader outside this test suite: built into an actual MoltenVK, driving real Metal hardware, confirmed to produce correct output before/after.


denisk-dev:
Hi @JackTYM, I think I may be experiencing an issue with Kitten Space Agency (KSA), that might be somewhat related to your PR. KSA uses "globalTextures" (`layout (set = SET_TEXTURE, binding = 0) uniform texture2D globalTextures[];`) to hold both the "true" texture2d and the "true" depth2d. Then they have a Planet.frag, which ends up sampling both the "true" texture2d for planet textures and such, and "true" depth2d for shadows (
`float sampleShadowMap(ShadowUBO shadowUbo, vec2 uv, float depth, float bias)
{
    return texture(sampler2DShadow(globalTextures[shadowUbo.textureHandle], samplers[shadowUbo.samplerHandle]), vec3(uv, depth-bias));
}`)

From what I understand, having the sampler2DShadow use globalTextures[] causes globalTextures[] to be marked as float2d<array> (`constant auto &globalTextureArrays = reinterpret_cast<constant array<depth2d_array<float>, 1024> &>(spvDescriptorSet3.globalTextures);`), which as you mentioned in your PR, ends up losing 3 out of 4 texture2d values.

Short of asking developers to separate textures and shadows out of globalTextures[], is there anything that can be done on the SPIRV_Cross side?

Thanks!

P.S. just out of curiosity, I tested this commit against that situation, and it's still showing up as depth2d<float> unless I modify sampleShadowMap to not use sampler2DShadow)

JackTYM:
> Hi @JackTYM, I think I may be experiencing an issue with Kitten Space Agency (KSA), that might be somewhat related to your PR

Hey! Sadly, I don't think this is a related bug, my PR is specifically for a problem I ran into with CombinedImageSamplerUsageHandler looking for instructions in dead branches, while I believe your issue with sampleShadowMap would still be live code.

spirv_cross.cpp
```diff
@@ -2271,6 +2271,295 @@ uint32_t Compiler::evaluate_constant_u32(uint32_t id) const
...
+	{
+		if (!is_scalar(get<SPIRType>(c->constant_type)))
+			return false;
+		out_value = c->scalar();
+		return true;
+	}
+
+	if (const auto *cop = maybe_get<SPIRConstantOp>(id))
+	{
+		out_value = evaluate_spec_constant_u32(*cop);
```
HansKristian-Work: This seems questionable. Wouldn't this mean DCE is done based on the default spec constant value? I suppose it would work if user supplied the spec constants through the API, but that's not the default mode of operation.

HansKristian-Work:
Would it make more sense to just cast the texture type between depth and non-depth somehow at the sampling site if a mismatch is detected despite the normal analysis? DCE like this might help in some special cases, but given the example above with bindless, I think this is a dead end (pun intended). SPIR-V allows the same texture variable to be used in both contexts at the same time, and we don't have a good answer for that right now.

HansKristian-Work:
Here's an example shader that seems to compile just fine. While we cannot cast texture2d to depth2d directly, we can cast the pointer in a helper function.

```
#include <metal_stdlib>
#include <simd/simd.h>

using namespace metal;

struct main0_out
{
};

struct main0_in
{
    float4 vUV [[user(locn0)]];
};

template <typename T>
static inline depth2d<T> spvDepthCast(texture2d<T> t)
{
        return *reinterpret_cast<thread depth2d<T> *>(&t);
}

fragment main0_out main0(main0_in in [[stage_in]], texture2d<float> A [[texture(0)]], sampler ASmplr [[sampler(0)]])
{
    main0_out out = {};
    out.FragColor = float4(spvDepthCast(A).sample_compare(ASmplr, in.vUV.xyz.xy, in.vUV.xyz.z));
    return out;
}
```

```
; Function Attrs: argmemonly convergent nounwind readonly
define <{ <4 x float> }> @main0(<4 x float> %0, %struct._texture_2d_t addrspace(1)* nocapture readonly %1, %struct._sampler_t addrspace(2)* nocapture readonly %2) local_unnamed_addr #0 {
  %4 = bitcast %struct._texture_2d_t addrspace(1)* %1 to %struct._depth_2d_t addrspace(1)*
  %5 = shufflevector <4 x float> %0, <4 x float> undef, <2 x i32> <i32 0, i32 1>
  %6 = extractelement <4 x float> %0, i64 2
  %7 = tail call { float, i8 } @air.sample_compare_depth_2d.f32(%struct._depth_2d_t addrspace(1)* nocapture readonly %4, %struct._sampler_t addrspace(2)* nocapture readonly %2, i32 1, <2 x float> %5, float %6, i1 true, <2 x i32> zeroinitializer, i1 false, f
loat 0.000000e+00, float 0.000000e+00, i32 0) #1, !alias.scope !20
  %8 = extractvalue { float, i8 } %7, 0
  %9 = insertelement <4 x float> undef, float %8, i64 0
  %10 = shufflevector <4 x float> %9, <4 x float> undef, <4 x i32> zeroinitializer
  %11 = insertvalue <{ <4 x float> }> undef, <4 x float> %10, 0
  ret <{ <4 x float> }> %11
}
```

spirv_cross.cpp
JackTYM: No longer included in approach

JackTYM:
@HansKristian-Work I appreciate the review! Finally getting around to it.

> Would it make more sense to just cast the texture type between depth and non-depth somehow at the sampling site if a mismatch is detected despite the normal analysis? DCE like this might help in some special cases, but given the example above with bindless, I think this is a dead end (pun intended). SPIR-V allows the same texture variable to be used in both contexts at the same time, and we don't have a good answer for that right now.

I think this is a better solution, I've reworked the PR around the spvDepthCast suggestion instead of the dead-block DCE approach

Resources no longer get their declared type decided by usage or reachability at all. They always declare as their plain type (texture2d<T>, texturecube<T>, and the array variants), and each sample_compare() call site casts through spvDepthCast() right before the call, your reinterpret_cast pattern, with overloads added for the array/cube shapes. image_type_glsl() only looks at the SPIR-V type's own static depth flag now, so real depth attachments/render targets/subpass inputs are untouched

HansKristian-Work:
This seems like the ideal approach assuming it works in practice (type punning like this is dodgy, but we don't really have any option).

HansKristian-Work (approved):

amaiorano:
Hi, I'm seemingly hitting a problem with this change. We're getting errors now in the generated MSL like:

```msl
error: no matching function for call to 'spvDepthCast'
                            _10541 = half(spvDepthCast(spvDescriptorSet0.sampler0_shadowMap).sample_compare(spvDescriptorSet0.sampler0_shadowMapSmplr, _21065.xy, uint(rint(_21065.z)), _8776));
                                          ^~~~~~~~~~~~
<clipped>.metal:48:26: note: candidate template ignored: could not match 'texture2d' against 'depth2d_array'
static inline depth2d<T> spvDepthCast(texture2d<T> t)
                         ^
<clipped>.metal:54:32: note: candidate template ignored: could not match 'texture2d_array' against 'depth2d_array'
static inline depth2d_array<T> spvDepthCast(texture2d_array<T> t)
                               ^
<clipped>.metal:60:28: note: candidate template ignored: could not match 'texturecube' against 'depth2d_array'
static inline depthcube<T> spvDepthCast(texturecube<T> t)
                           ^
<clipped>.metal:66:34: note: candidate template ignored: could not match 'texturecube_array' against 'depth2d_array'
static inline depthcube_array<T> spvDepthCast(texturecube_array<T> t)
                                 ^
```

It looks like what's happening is that we're always casting the resource via `spvDepthCast`, even if it's already a depth type. Should it be only conditionally casting if it's not depth? Or maybe add no-op overloads of `svpDepthCast` for depth type inputs?

Maybe helpful: it seems that the resource is part of an argument buffer ("spvDescriptorSet0"), so maybe that's where the discrepency lies?

HansKristian-Work:
I cannot reproduce that. depth2d should never be declared anymore, so this issue should be impossible. Make a new issue and attach the failing SPIR-V.

amaiorano:
You are right, this was on our end. We're intercepting and modifying the generated MSL. Sorry about that!

PR #2663, opened 2026-08-09
MSL: Add ray tracing pipeline support

dttdrv:
This replaces #2649 with a much narrower compiler change.

The MoltenVK runtime stays in MoltenVK. This PR only adds the SPIR-V to MSL lowering needed for the Vulkan ray-tracing shader stages and their external runtime ABI. The path is disabled by default and is only enabled by an embedding runtime that provides that ABI.

The lowering covers ray-generation, miss, closest-hit, any-hit, intersection, and callable shaders, including payloads, hit attributes, callable data, shader records, helper functions, and the required ray-tracing instructions.

It also fixes existing MSL ray-query lowering for cull masks, candidate opacity, arrays, and helper functions. Ordinary non-ray shaders do not enter the new path. Position fetch is not included because Metal ray queries do not expose the original triangle positions.

The production compiler change is 495 additions and 62 deletions across five files. The public C API is unchanged. The focused compiler coverage passes, and the paired MoltenVK implementation builds and runs its Metal validation probes on my M5 Pro.

Paired MoltenVK work: https://github.com/KhronosGroup/MoltenVK/pull/2771


spirv_glsl.cpp
```diff
@@ -13001,6 +13001,16 @@ void CompilerGLSL::emit_instruction(const Instruction &instruction)
 		uint32_t ptr = ops[2];
 
 		flush_variable_declaration(ptr);
+		auto &type = get<SPIRType>(result_type);
```
HansKristian-Work, line 13004: Which problem is this solving? It doesn't seem like it's RT specific. If there are other bugs unrelated to RT I would expect it to be a separate PR to reduce the review burden.

spirv_glsl.cpp
```diff
@@ -19501,11 +19519,11 @@ BlockID CompilerGLSL::emit_block_chain_inner(SPIRBlock &block)
 	}
 
 	case SPIRBlock::IgnoreIntersection:
-		statement("ignoreIntersectionEXT;");
+		emit_ignore_intersection();
```
HansKristian-Work, line 19522: This is not how we do backend-specific opcodes. Just implement IgnoreIntersection in spirv_msl.cpp's emit_instruction().

spirv_glsl.hpp
```diff
@@ -873,8 +873,15 @@ class CompilerGLSL : public Compiler
 	virtual bool emit_array_copy(const char *expr, uint32_t lhs_id, uint32_t rhs_id,
 	                             StorageClass lhs_storage, StorageClass rhs_storage);
 	virtual void emit_block_hints(const SPIRBlock &block);
+	virtual void emit_ignore_intersection() { statement("ignoreIntersectionEXT;"); }
```
HansKristian-Work, line 876: Don't put code inline in the header like this.

spirv_msl.hpp
```diff
@@ -286,7 +288,7 @@ static const uint32_t kBufferSizeBufferBinding = ~(2u);
 // will start at max(kArgumentBufferBinding) + 1.
 static const uint32_t kArgumentBufferBinding = ~(3u);
 
-static const uint32_t kMaxArgumentBuffers = 8;
+static const uint32_t kMaxArgumentBuffers = 32;
```
HansKristian-Work, line 291: This is already changed upstream to 16. Don't use 32.

spirv_msl.hpp
```diff
@@ -312,6 +314,7 @@ class CompilerMSL : public CompilerGLSL
 		uint32_t shader_patch_output_buffer_index = 27;
 		uint32_t shader_tess_factor_buffer_index = 26;
 		uint32_t buffer_size_buffer_index = 25;
+		uint32_t acceleration_structure_address_table_buffer_index = 12;
```
HansKristian-Work, line 317: 12 ?!?

spirv_cross.cpp
```diff
@@ -694,6 +694,14 @@ static inline bool storage_class_is_interface(StorageClass storage)
 	case StorageClassStorageBuffer:
 		return true;
 
+	case StorageClassRayPayloadKHR:
```
HansKristian-Work, line 697: This is tautological. RT already requires SPIR-V 1.4, so the check is meaningless.
This seems like a general fix that should be a separate PR.

spirv_msl.cpp
```diff
@@ -10436,25 +10618,53 @@ void CompilerMSL::emit_instruction(const Instruction &instruction)
...
+		if (uses_ray_query_flags)
+			statement(ray_query_metadata_expression(ops[0]), ".flags = ", to_expression(ops[2]), ";");
 
 		statement(to_expression(ops[0]), ".reset(", "ray(", to_expression(ops[4]), ", ", to_expression(ops[6]), ", ",
-		          to_expression(ops[5]), ", ", to_expression(ops[7]), "), ", to_expression(ops[1]), ", ", to_expression(ops[3]),
+		          to_expression(ops[5]), ", ", to_expression(ops[7]),
+		          "), ", msl_options.acceleration_structure_descriptor_as_address ?
+		                       "*(device const raytracing::acceleration_structure<raytracing::instancing>*)" : "",
+		          to_expression(ops[1]),
+		          ", (", to_expression(ops[3]), " & 0xffu)",
```
HansKristian-Work, line 10661: Why mask with 0xff? If this is a general fix it should be made upstream as a separate fix.

spirv_msl.cpp
```diff
@@ -10509,11 +10723,11 @@ void CompilerMSL::emit_instruction(const Instruction &instruction)
 			emit_op(ops[0], ops[1], join(to_expression(ops[2]), ".get_committed_distance()"), false);
 		break;
 	case OpRayQueryGetIntersectionCandidateAABBOpaqueKHR:
-	{
-		flush_variable_declaration(ops[0]);
-		emit_op(ops[0], ops[1], join(to_expression(ops[2]), ".is_candidate_non_opaque_bounding_box()"), false);
+		flush_variable_declaration(ops[2]);
+		emit_op(ops[0], ops[1], join("!", to_expression(ops[2]), ".is_candidate_non_opaque_bounding_box()"), false);
```
HansKristian-Work, line 10727: This is already changed upstream.

CMakeLists.txt
```diff
@@ -547,6 +547,59 @@ if (SPIRV_CROSS_CLI)
 				target_link_libraries(spirv-cross-msl-ycbcr-conversion-test spirv-cross-c)
 				set_target_properties(spirv-cross-msl-ycbcr-conversion-test PROPERTIES LINK_FLAGS "${spirv-cross-link-flags}")
 
+				set(spirv-cross-rt-test-spirv)
```
HansKristian-Work, line 550: Why does it have to be done like this? I would expect normal tests here.

HansKristian-Work:
> This PR only adds the SPIR-V to MSL lowering needed for the Vulkan ray-tracing shader stages and their external runtime ABI.

I'm missing explanations on how exactly you expect this to work, what would code look like? I'm not going to dig through mountains of code to figure that out.

HansKristian-Work (changes requested):
There is a lot of stuff going on in here, and a lot of it seems like general changes that could be submitted separately. If you can, it's better to submit multiple PRs that accomplish goals on their own, rather than one big blob.

This PR on its own is still way too massive for me to review. I don't know Metal ray-tracing well, which makes reading this PR very hard for me. I have little interest in maintaining a very complex partial feature that will only benefit MVK emulation, so you'll have to make it as easy as possible to digest if you want to get this upstreamed.

dttdrv:
i've reworked this locally and separated the query, layout and interface-filtering fixes. the RT change uses `emit_instruction()` for MSL-specific operations, moves implementation out of headers, and uses normal shader fixtures.

the compiler emits visible functions; the runtime handles traversal and shader-table lookup. raygen passes its payload by reference, and the selected shader writes into that same payload. with the runtime definitions expanded, a miss shader looks like this:

```metal
[[visible]] void main0(thread spvRayInvocation& spvRay,
                      thread spvRayTracingState& spvRayState)
{
    *reinterpret_cast<thread float3*>(spvRay.data) = float3(1.0);
}
```

the runtime supplies these types and helpers. descriptor access uses `spvRayState.dispatch->descriptorSetAddresses`, which is also part of the proposed interface.

the normal tests pass. the 16 RT fixtures and two query fixtures also compile through Metal at MSL 3.0, with the runtime header supplied for RT.

PR #2664, opened 2026-08-18
MSL: Reserve Metal type names so resources cannot shadow them

darrinperry83:
A SPIR-V variable or struct member named after an MSL type emits a declaration that shadows the type, and the next declaration that uses that type fails to compile:

```
texture2d<float> sampler [[texture(0)]];
sampler samplerSmplr [[sampler(1)]];
// error: must use 'struct' tag to refer to type 'sampler' in this scope
// note: struct 'sampler' is hidden by a non-type declaration of 'sampler' here
```

This is reachable from ordinary shaders rather than a corner case. A combined image sampler always emits a `sampler <name>Smplr` parameter immediately after the texture, so a resource named `sampler` always breaks. The name need not be expressible in GLSL — SPIR-V carries arbitrary `OpName` strings, and `sampler` is a GLSL keyword, so this arrives from HLSL and other front-ends.

Found while running Blender's Vulkan backend on MoltenVK, where it failed pipeline creation with the error above.

### Fix

Reserve the type names emitted by `sampler_type()` and `image_type_glsl()`, so `replace_illegal_names()` renames colliding variables and struct members as it already does for other reserved words.

### The existing suite already contains this shadowing

The four changed references are `texture-access.swizzle` variants that declare a variable named `depth2d` of type `depth2d<float>`:

```
- depth2d<float> depth2d  [[texture(6)]]
+ depth2d<float> depth2d0 [[texture(6)]]
```

Those compiled only because the type name is not used again later in the same scope. `sampler` has no such luck, because the paired sampler parameter always follows.

### Testing

`./test_shaders.sh` passes with no unexpected deltas. The four reference changes above are the rename; all regenerated shaders compile with the Metal compiler.

A regression test is added as SPIR-V assembly (`shaders-msl/asm/frag/reserved-msl-type-names.asm.frag`) because `sampler` cannot be written in GLSL. It covers both emitters — `sampler` from `sampler_type()` and `depth2d` from `image_type_glsl()`.

### DIFF
--- a/reference/opt/shaders-msl/asm/frag/reserved-msl-type-names.asm.frag
+++ b/reference/opt/shaders-msl/asm/frag/reserved-msl-type-names.asm.frag
@@ -1,0 +1,22 @@
+#include <metal_stdlib>
+#include <simd/simd.h>
+
+using namespace metal;
+
+struct main0_out
+{
+    float4 FragColor [[color(0)]];
+};
+
+struct main0_in
+{
+    float2 vUV [[user(locn0)]];
+};
+
+fragment main0_out main0(main0_in in [[stage_in]], texture2d<float> sampler0 [[texture(0)]], texture2d<float> depth2d0 [[texture(1)]], sampler sampler0Smplr [[sampler(0)]], sampler depth2d0Smplr [[sampler(1)]])
+{
+    main0_out out = {};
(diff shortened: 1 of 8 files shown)


HansKristian-Work:
Seems sound. I'll update the CI conflicts.

HansKristian-Work (approved):

PR #2666, opened 2026-08-25
feat(msl): add depth clip enable emulation

anshulnoori:
Adds opt-in MSL support for Vulkan depth clip and clamp combinations that Metal cannot represent natively.

When enabled, SPIRV-Cross can:

- transform vertex or tessellation depth when clipping and clamping are both disabled;
- clamp fragment depth when clipping and clamping are both enabled;
- select per-viewport depth ranges through `ViewportIndex`;
- provide runtime state so dynamic changes do not require shader recompilation;
- preserve correct ordering with clip-space fixup, reversed viewports, and Metal OpenGL mode.

The depth-clip state buffer is emitted only when the shader stage needs. Existing output is unchanged.

This PR takes heavy inspiration from Mesa's KosmicKrisp implementation.

## Testing

- Warnings-as-errors and miscellaneous-warnings build: pass
- Shared build, install, and CTest: 23/23 pass
- MSL regression suite: 319/319 references match
- Apple Metal compilation: 286/286 eligible shaders compile
- New depth-clip fixtures: 12/12 match and compile

Not an expert so all advice/requested changes is appreciated!

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -678,6 +678,7 @@ struct CLIArguments
 	bool msl_check_discarded_frag_stores = false;
 	bool msl_force_fragment_with_side_effects_execution = false;
 	bool msl_emulate_reversed_depth_viewport = false;
+	bool msl_emulate_depth_clip_enable = false;
 	bool msl_sample_dref_lod_array_as_grad = false;
 	bool msl_runtime_array_rich_descriptor = false;
 	bool msl_replace_recursive_inputs = false;
@@ -983,6 +984,7 @@ static void print_help_msl()
 	                "\t\tHowever, Vulkan expects fragment shader to be executed since it cannot be discarded until the discard\n"
 	                "\t\tpresent in the fragment execution, which would also execute the operations with side effects.\n"
 	                "\t[--msl-emulate-reversed-depth-viewport]:\n\t\tEmulate reversed-depth viewports by inverting clip-space Z.\n"
+	                "\t[--msl-emulate-depth-clip-enable]:\n\t\tEmulate Vulkan depth clip and clamp combinations unsupported by Metal.\n"
 	                "\t[--msl-sample-dref-lod-array-as-grad]:\n\t\tUse a gradient instead of a level argument.\n"
 	                "\t\tSome Metal devices have a bug where the level() argument to\n"
 	                "\t\tdepth2d_array<T>::sample_compare() in a fragment shader is biased by some\n"
@@ -1301,6 +1303,7 @@ static string compile_iteration(const CLIArguments &args, std::vector<uint32_t> 
 		msl_opts.check_discarded_frag_stores = args.msl_check_discarded_frag_stores;
 		msl_opts.force_fragment_with_side_effects_execution = args.msl_force_fragment_with_side_effects_execution;
 		msl_opts.emulate_reversed_depth_viewport = args.msl_emulate_reversed_depth_viewport;
+		msl_opts.emulate_depth_clip_enable = args.msl_emulate_depth_clip_enable;
 		msl_opts.sample_dref_lod_array_as_grad = args.msl_sample_dref_lod_array_as_grad;
 		msl_opts.ios_support_base_vertex_instance = true;
 		msl_opts.runtime_array_rich_descriptor = args.msl_runtime_array_rich_descriptor;
@@ -1905,6 +1908,7 @@ static int main_inner(int argc, char *argv[])
 	cbs.add("--msl-check-discarded-frag-stores", [&args](CLIParser &) { args.msl_check_discarded_frag_stores = true; });
 	cbs.add("--msl-force-frag-with-side-effects-execution", [&args](CLIParser &) { args.msl_force_fragment_with_side_effects_execution = true; });
 	cbs.add("--msl-emulate-reversed-depth-viewport", [&args](CLIParser &) { args.msl_emulate_reversed_depth_viewport = true; });
+	cbs.add("--msl-emulate-depth-clip-enable", [&args](CLIParser &) { args.msl_emulate_depth_clip_enable = true; });
 	cbs.add("--msl-sample-dref-lod-array-as-grad",
 	        [&args](CLIParser &) { args.msl_sample_dref_lod_array_as_grad = true; });
 	cbs.add("--msl-no-readwrite-texture-fences", [&args](CLIParser &) { args.msl_readwrite_texture_fences = false; });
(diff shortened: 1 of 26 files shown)


main.cpp
```diff
@@ -983,6 +985,8 @@ static void print_help_msl()
 	                "\t\tHowever, Vulkan expects fragment shader to be executed since it cannot be discarded until the discard\n"
 	                "\t\tpresent in the fragment execution, which would also execute the operations with side effects.\n"
 	                "\t[--msl-emulate-reversed-depth-viewport]:\n\t\tEmulate reversed-depth viewports by inverting clip-space Z.\n"
+	                "\t[--msl-emulate-depth-clip-enable]:\n\t\tEmulate Vulkan depth clip and clamp combinations unsupported by Metal.\n"
+	                "\t[--msl-opengl-mode]:\n\t\tAdjust Vulkan clip-space output for Metal's OpenGL mode.\n"
```
HansKristian-Work: What is this?

spirv_msl.cpp
```diff
@@ -8441,6 +8446,17 @@ void CompilerMSL::emit_resources()
 // Emit declarations for the specialization Metal function constants
 void CompilerMSL::emit_specialization_constants_and_structs()
 {
+	if (needs_depth_clip_state_buffer())
+	{
+		statement("struct spvDepthClipState");
+		begin_scope();
+		statement("uint emulateViewportZ;");
+		statement("uint emulateDepthClamp;");
+		statement("float2 viewportDepthRanges[16];");
```
HansKristian-Work, line 8455: Does this represent the limit in Metal?

spirv_msl.cpp
```diff
@@ -13549,6 +13565,23 @@ void CompilerMSL::emit_fixup()
...
+				          ".w) * 0.5;       // Adjust clip-space for Metal");
+
+			if (msl_options.emulate_depth_clip_enable)
+			{
+				string viewport_idx =
+				    qual_viewport_idx_var_name.empty() ? "0" : join("uint(", qual_viewport_idx_var_name, ")");
+				statement("if (spvDepthClipState.emulateViewportZ != 0u)");
+				begin_scope();
+				statement("float2 spvViewportDepthRange = spvDepthClipState.viewportDepthRanges[", viewport_idx, "];");
+				statement(qual_pos_var_name, ".z = ", qual_pos_var_name,
```
HansKristian-Work, line 13579: I really don't think all of this emulated math is compliant with W-plane clipping ...

spirv_msl.cpp
```diff
@@ -13563,13 +13596,24 @@ void CompilerMSL::emit_fixup()
 				end_scope();
 			}
 
-			if (options.vertex.fixup_clipspace)
-				statement(qual_pos_var_name, ".z = (", qual_pos_var_name, ".z + ", qual_pos_var_name,
-						  ".w) * 0.5;       // Adjust clip-space for Metal");
+			if (msl_options.use_opengl_mode)
```
HansKristian-Work: I don't understand the motivation behind this option.

HansKristian-Work (changes requested):
Not too invasive, but I need answers to some questions.

spirv_msl.cpp
anshulnoori: MoltenVK enables the private `setOpenGLModeEnabled:` render-pass mode when `MVK_USE_METAL_PRIVATE_API=1`

In that mode MVK expects OpenGL clip-space depth `[-W, W]` instead of normal Metal depth `[0, W]` so MoltenVK only enables this option when that private render-pass mode is enabled.

I can change the flag name to make it more clear if needed, just didn't want to make it too verbose.

main.cpp
anshulnoori: Feature flags to enable the depth clip emulation and a OpenGL render pass that MoltenVK uses when ‘MVK_USE_METAL_PRIVATE_API=1’. Saw that other emulated features had a flag so wanted to follow the standard.

spirv_msl.cpp
anshulnoori: Yeah, two toggles, and then viewport depth ranges with the max length matching MoltenVK’s ‘kMVKMaxViewportScissorCount’

spirv_msl.cpp
anshulnoori: Uh, I copied it from KosmicKrisp's impl as best as I could so unsure about compliance.

Any suggestions?

spirv_msl.cpp
HansKristian-Work: But ... why? Vulkan uses [0, w] clip space. Why would MVK have any reason to enable OpenGL-style clip space?

spirv_msl.cpp
HansKristian-Work: Well, if it passes conformance, I guess I'll leave it. Still think it's broken, but I'm not going to fight it.

spirv_msl.cpp
anshulnoori: Not sure why it's there in MVK, I just added this flag in order to have wide support. I can do a bit more research into what it's used for, or we can remove it from this PR.

HansKristian-Work: I don't want to support something like this unless there's a clear rationale why. I would remove it from this PR (it's a separate thing), and if there is solid reason for supporting it, it can just be a different PR.

spirv_msl.cpp
squidbus: In KosmicKrisp the viewport Z transform on the shader side is only used when both clip and clamp are disabled (because to emulate this state we set the mode to clamp and viewport Z range to [0, 1] to avoid both clipping and clamping). The intended usage here should be the same.

spirv_msl.cpp
squidbus: > In that mode MVK expects OpenGL clip-space depth [-W, W] instead of normal Metal depth [0, W] so MoltenVK only enables this option when that private render-pass mode is enabled.

I think your understanding of this is not correct. OpenGL mode enables using certain features available through private APIs and does not change the depth range. I don't see a need for this to be introduced to SPIRV-Cross.

For reference, the way MoltenVK *does* handle [-W, W] clip space for the Vulkan extension is using the existing `fixup_clipspace` option, which should be applied before any viewport Z transform emulation (I haven't looked at exactly how they interact in this PR).

anshulnoori: Yeah I don't think I wrapped my head around how MVK handled it fully... Anyways, made the requested changes and force pushed.

anshulnoori:
@HansKristian-Work removed the OpenGL stuff, squashed it into the original commit, and force pushed.

HansKristian-Work (approved):
Ok, the code isn't invasive and roughly follows the other "fixup" code paths, so I'm okay with this.

PR #2669, opened 2026-09-01
MSL: Bump the descriptor set limit to 16.

### LINKED ISSUE #2662: MSL: More than 8 layout sets in Vulkan is not supported with Metal argument buffers
When more than 8 layout sets are used in a (comp) shader with Vulkan, SPRIV-Cross does not use another spvDescriptorSetBuffer for the 9th set, causing an error with a repeating buffer.

Kitten Space Agency (KSA) has introduced a shader with 9 layout sets, which now fails to load correctly. My question is if the `kMaxArgumentBuffers` should be based on Metal Tier1/Tier2, or if there is an intended workaround in place right now that is hitting a bug.

Shader portion causing an issue (unwrapped for ease of readability):

`
#define ATMOSPHERE_DATA_SET 8
layout(set = ATMOSPHERE_DATA_SET, binding = 0, std140) uniform AtmosphereDataUbo
{
    AtmosphereData atmosphereData;

    vec4 planetPositionAndPackedScreenStartCoords;

    int volumetricShadowsUseFlip;
    float volumetricShadowsOpacity;
    float distancePrecisionThreshold;
    float planetRadius;

    int distantPlanet;
    int atmosphereLutLayer;
    int pad0;
    int pad1;

} atmosphereDataUbo;
`

Error:  `error: cannot reserve 'buffer' resource location at index 1
kernel void main0(constant spvDescriptorSetBuffer0& spvDescriptorSet0 [[buffer(0)]], constant spvDescriptorSetBuffer1& spvDescriptorSet1 [[buffer(1)]], constant spvDescriptorSetBuffer2& spvDescriptorSet2 [[buffer(2)]], constant spvDescriptorSetBuffer3& spvDescriptorSet3 [[buffer(3)]], constant spvDescriptorSetBuffer4& spvDescriptorSet4 [[buffer(4)]], constant spvDescriptorSetBuffer5& spvDescriptorSet5 [[buffer(5)]], constant spvDescriptorSetBuffer6& spvDescriptorSet6 [[buffer(6)]], constant spvDescriptorSetBuffer7& spvDescriptorSet7 [[buffer(7)]], constant uint* spvDynamicOffsets [[buffer(30)]], constant AtmosphereDataUbo& atmosphereDataUbo [[buffer(1)]], constant CloudPassConsts& cloudPassConsts [[buffer(8)]], uint3 gl_GlobalInvocationID [[thread_position_in_grid]], uint3 gl_LocalInvocationID [[thread_position_in_threadgroup]])`

I have confirmed modifying max to: `static const uint32_t kMaxArgumentBuffers = 9;` fixes that issue. 

Tier 2 is technically allowed 31 argument buffers, if I'm not mistaken, but setting it to 31 causes a failure in MoltenVK due to:
`union MVKStageResourceBits {
	struct {
		MVKStaticBitSet<kMVKMaxTextureCount> textures;
		MVKStaticBitSet<kMVKMaxBufferCount>  buffers;
		MVKStaticBitSet<kMVKMaxSamplerCount> samplers;
		MVKStaticBitSet<kMVKMaxDescriptorSetCount> descriptorSetData;
	};
	MVKStaticBitSet<192> allBits;
	constexpr MVKStageResourceBits(): allBits{} {}
	// Note: Reset is done with a large memset over multiple StageResourceBits
};
static_assert(sizeof(MVKStageResourceBits) == sizeof(MVKStageResourceBits::allBits), "Make sure we can quickly process all bits");`
which ends up being 32!=24.

I'd like to know if `kMaxArgumentBuffers` should be moved into "public" and conditional on Tier1/Tier2, and also what value should be assigned if the chip is Tier2. Alternatively, if there is a workaround in place for more than 8 layout sets, why it's currently failing.

HansKristian-Work:
Fix #2662.

### DIFF
--- a/reference/shaders-msl-no-opt/comp/argument-buffer-set-limit.msl2.argument.argument-tier-1.comp
+++ b/reference/shaders-msl-no-opt/comp/argument-buffer-set-limit.msl2.argument.argument-tier-1.comp
@@ -1,0 +1,28 @@
+#include <metal_stdlib>
+#include <simd/simd.h>
+
+using namespace metal;
+
+struct SSBO
+{
+    float wr;
+};
+
+struct UBO
+{
+    float rd;
+};
+
+struct spvDescriptorSetBuffer15
+{
+    device SSBO* m_9 [[id(0)]];
+    texture2d<float> Samp [[id(1)]];
+    sampler SampSmplr [[id(2)]];
+    constant UBO* m_28 [[id(3)]];
+};
+
+kernel void main0(constant spvDescriptorSetBuffer15& spvDescriptorSet15 [[buffer(15)]])
+{
+    (*spvDescriptorSet15.m_9).wr = spvDescriptorSet15.Samp.sample(spvDescriptorSet15.SampSmplr, float2(0.5), level(0.0)).x + (*spvDescriptorSet15.m_28).rd;
+}
+
--- a/shaders-msl-no-opt/comp/argument-buffer-set-limit.msl2.argument.argument-tier-1.comp
+++ b/shaders-msl-no-opt/comp/argument-buffer-set-limit.msl2.argument.argument-tier-1.comp
@@ -1,0 +1,18 @@
+#version 450
+
+layout(set = 15, binding = 0) uniform UBO
+{
+	float rd;
+};
+
+layout(set = 15, binding = 1) buffer SSBO
+{
+	float wr;
+};
+
+layout(set = 15, binding = 2) uniform sampler2D Samp;
+
+void main()
+{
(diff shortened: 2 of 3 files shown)


denisk-dev:
@HansKristian-Work, thanks! Would that cause an issue with Metal Argument Buffers "Tier1", that has a hard limit of 8, if I'm not mistaken? I was thinking of doing a conditional, based on whether it's Tier1 or Tier2 for the kMaxArgumentBuffers (or equivalent, since that was supposed to be constant)

HansKristian-Work:
That's the client's job to ensure limits aren't breached. E.g. if this is MoltenVK, just don't expose more than 8 descriptor sets or whatever.

PR #2672, opened 2026-09-01
HLSL: Fix HLSL tests (FXC isn't running in CI anymore).

LukasBanana:
This PR fixes the HLSL tests on Windows. I noticed that the CI runs ignore FXC (not sure since when): https://github.com/KhronosGroup/SPIRV-Cross/actions/runs/33520477162/job/99898143627#step:10:6824

- Update HLSL references for 'replicated-composites.spv16.vk.asm.comp'; Their scalar-to-vector constructors are not supported by FXC nor DXC, it needs to be a type cast, e.g. `(float4)(0.0f)` instead of `float4(0.0f)`.
- Don't run FXC in test_shaders.py script for shader models > 5; Mesh shaders for instance are not supported by FXC.
- Fix ssbo-store-array.comp test shader; `Data d1` variable was read uninitialized and that fails with FXC while glslang seems to ignore it.
- There's one single `[[fallthrough]]` statement in the entire code base; ~Compile it out if target language is not C++17 or newer~ Replace it with `/* fallthrough */` (as the other places do) to fix warning in MSVC17.

### DIFF
--- a/spirv_glsl.cpp
+++ b/spirv_glsl.cpp
@@ -16636,12 +16636,11 @@ void CompilerGLSL::emit_instruction(const Instruction &instruction)
 		else
 		{
 			// HLSL needs to emit scalar-to-vector constructors as C-style type casts, e.g. `(float4)1.0` vs. `vec4(1.0)`.
-			const std::string ops_expr = to_expression(ops[2]);
 			if (!backend.use_constructor_splatting &&
 				type.vecsize > 1 && type.columns == 1 && is_scalar(get<SPIRType>(expression_type_id(ops[2]))))
-				rhs = join("(", type_to_glsl(type), ")", ops_expr);
+				rhs = join("(", type_to_glsl(type), ")", to_enclosed_expression(ops[2]));
 			else
-				rhs = join(type_to_glsl(type), "(", ops_expr, ")");
+				rhs = join(type_to_glsl(type), "(", to_expression(ops[2]), ")");
 		}
 		emit_op(result_type, id, rhs, true);
 		break;
--- a/spirv_glsl.cpp
+++ b/spirv_glsl.cpp
@@ -6391,7 +6391,13 @@ string CompilerGLSL::constant_expression(const SPIRConstant &c,
 		}
 		else
 		{
-			return join(type_to_glsl(type), "(", to_expression(c.subconstants[0]), ")");
+			// HLSL needs to emit scalar-to-vector constructors as C-style type casts, e.g. `(float4)1.0` vs. `vec4(1.0)`.
+			std::string subconst_expr = to_expression(c.subconstants[0]);
+			if (!backend.use_constructor_splatting &&
+				type.vecsize > 1 && type.columns == 1 && is_scalar(get<SPIRType>(expression_type_id(c.subconstants[0]))))
+				return join("(", type_to_glsl(type), ")", subconst_expr);
+			else
+				return join(type_to_glsl(type), "(", subconst_expr, ")");
 		}
 	}
 	else if (!c.subconstants.empty())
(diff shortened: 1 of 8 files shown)


spirv_hlsl.cpp
```diff
@@ -7167,6 +7167,7 @@ string CompilerHLSL::compile()
 	backend.explicit_struct_type = false;
 	backend.use_initializer_list = true;
 	backend.use_constructor_splatting = false;
+	backend.use_c_style_cast_for_scalar_to_vector = true;
```
HansKristian-Work: There is already the "use constructor splatting" option above that's intended to cover this case.

spirv_glsl.cpp
```diff
@@ -16629,7 +16639,11 @@ void CompilerGLSL::emit_instruction(const Instruction &instruction)
 		}
 		else
 		{
-			rhs = join(type_to_glsl(type), "(", to_expression(ops[2]), ")");
+			if (backend.use_c_style_cast_for_scalar_to_vector &&
```
HansKristian-Work: Just emit replicated arguments rather than the very goofy HLSL construct since this happens in common code. That way it aligns with use_constructor_splatting better.

spirv_glsl.cpp
```diff
@@ -8668,6 +8672,12 @@ bool CompilerGLSL::expression_is_non_value_type_array(uint32_t value_type_id, ui
 	       has_member_decoration(backed_type.self, 0, DecorationOffset);
 }
 
+bool CompilerGLSL::expression_is_scalar_type(uint32_t type_id) const
```
HansKristian-Work: Wouldn't is_scalar(type) be enough in the relevant code? Don't think we need a new special helper here.

spirv_msl.cpp
```diff
@@ -10070,7 +10070,7 @@ void CompilerMSL::emit_instruction(const Instruction &instruction)
 		case Dim1D:
 			if (!msl_options.texture_1D_as_2D)
 				SPIRV_CROSS_THROW("ImageQueryLod is not supported on 1D textures.");
-			[[fallthrough]];
+			/* fallthrough */
```
HansKristian-Work, line 10073: Why change this?

HansKristian-Work (changes requested):
Nits.

spirv_msl.cpp
HansKristian-Work: >  (as the other places do) to fix warning in MSVC17.

missed that. Ok, fine I guess.

spirv_glsl.cpp
LukasBanana: So you'd want something like `float4(1.0f, 1.0f, 1.0f, 1.0f)`? Wasn't there a case where some simple expressions are inlined (I dont quite recall, though). Either way, this could also happen for matrix types I believe, which could become very verbose, e.g. `float4x4(some_long_name_constant, some_long_name_constant, some_long_name_constant, some_long_name_constant, some_long_name_constant, ...)`. Do you want me to proceed with replacing this by just replicating the arguments?

HansKristian-Work: Ok, it's a bit of a wart, but I guess you can keep the HLSL-style splat then. Just add a comment that HLSL is the only relevant backend for this and use the existing backend bool. Can revisit as needed.

spirv_glsl.cpp
LukasBanana: Yes, that works too. I was looking for something like this but I wasn't sure about whether the type needs to be guarded by `maybe_get`. Since `expression_type_id()` always returns a type Id and throws otherwise, this should work, too.

spirv_glsl.cpp
LukasBanana: I am currently testing this with a slightly less verbose construct by emitting this as a C-style type cast without the extra parenthesis, e.g. `(float4)1.0f` instead of `(float4)(1.0f)`. Do you think this could cause issues later on, but I think the right hand side expression should always be a single token for scalar types, right? All tests are still passing with this adjustment.

spirv_hlsl.cpp
LukasBanana: Replace this with the existing `use_constructor_splatting`.

spirv_glsl.cpp
LukasBanana: Replace this helper function with the existing `is_scalar()`.

spirv_glsl.cpp
HansKristian-Work: Yes, I added a small fix that uses to_enclosed_expression for these cases.

HansKristian-Work (approved):

PR #2673, opened 2026-09-02
GLSL: Fix row_major matrix order for push_constant to struct translation

LukasBanana:
I stumbled across this issue when I started compiling some of my LLGL examples to all backends, including GLSL140 which translates `push_constant` blocks to struct uniforms.
This is the input source for the new test I added as part of this PR:
```hlsl
struct Matrix {
    float4x4 transform;
};

[[vk::push_constant]] Matrix matrix_constants;

float4 main(float4 in_position : POSITION) : SV_Position {
    return mul(matrix_constants.transform, in_position);
}
```

HLSL shaders with the `[[vk::push_constant]]` attribute that are compiled to SPIR-V, their matrix layout uses row_major by default but these push constants are translated to bare structs in GLSL for OpenGL semantics, which do not support `layout(row_major)` qualifiers. In this case, allow the respective expression in `member_is_non_native_row_major_matrix()` to be marked with the `need_transpose` bit.

The expected behavior is presented in the new test 'push-constant-row-major-matrix.asm.vert':
```glsl
#version 450
struct type_PushConstant_Matrix {
    mat4 transform;
};
uniform type_PushConstant_Matrix matrix_constants;
layout(location = 0) in vec4 in_var_POSITION;
void main() {
    gl_Position = matrix_constants.transform * in_var_POSITION;
}
```

Without this change, SPIRV-Cross compiles this to the wrong expression:
```glsl
#version 450
struct type_PushConstant_Matrix {
    mat4 transform;
};
uniform type_PushConstant_Matrix matrix_constants;
layout(location = 0) in vec4 in_var_POSITION;
void main() {
    gl_Position = in_var_POSITION * matrix_constants.transform; // <- Wrong without 'layout(row_major)'
}
```

### DIFF
--- a/spirv_glsl.cpp
+++ b/spirv_glsl.cpp
@@ -11307,7 +11307,13 @@ string CompilerGLSL::access_chain_internal(uint32_t base, const uint32_t *indice
 			else
 				physical_type = 0;
 
-			row_major_matrix_needs_conversion = member_is_non_native_row_major_matrix(*type, index);
+			// GLSL does not allow `layout(row_major)` qualifier inside bare struct declarations.
+			// Structs used as members of UBO/SSBO blocks can have layout qualifiers applied at the block level.
+			// Push constant blocks in OpenGL are also emitted as bare structs (without Block decoration in output).
+			auto *var = maybe_get_backing_variable(base);
+			const bool is_push_constant_emulated = !options.vulkan_semantics && var != nullptr && var->storage == StorageClassPushConstant;
+
+			row_major_matrix_needs_conversion = member_is_non_native_row_major_matrix(*type, index, is_push_constant_emulated);
 			type_id = type->member_types[index];
 			type = &get<SPIRType>(type->member_types[index]);
 		}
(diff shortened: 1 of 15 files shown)


spirv_glsl.cpp
```diff
@@ -11307,7 +11307,13 @@ string CompilerGLSL::access_chain_internal(uint32_t base, const uint32_t *indice
 			else
 				physical_type = 0;
 
-			row_major_matrix_needs_conversion = member_is_non_native_row_major_matrix(*type, index);
+			// GLSL does not allow `layout(row_major)` qualifier inside bare struct declarations.
+			// Structs used as members of UBO/SSBO blocks can have layout qualifiers applied at the block level.
+			// Push constant blocks in OpenGL are also emitted as bare structs (without Block decoration in output).
+			auto *var = maybe_get<SPIRVariable>(base);
```
HansKristian-Work: This seems dubious. Push constants for Vulkan target *is* emitted as a Block-like type that supports row-major. This check should be constrained to non-Vulkan GLSL. Also, maybe_get isn't enough, you want maybe_get_backing_variable to take care of chained OpAccessChain.

The test should be named `.asm.vk.vert` so that it generates both Vulkan GLSL and classic GLSL outputs.

HansKristian-Work (changes requested):
Comments.

spirv_glsl.cpp
LukasBanana: Ah thanks, I wasn't sure about the suffix.
- I pushed an update and now it produces two versions, one for OpenGL and one for Vulkan.
- I renamed the boolean blow to be more specific and added the vulkan semantic condition.
- I replaced `maybe_get` with `maybe_get_backing_variable` and ran the unit tests, all successful (except for the HLSL ones, for which I have the other PR).

HansKristian-Work (approved):

PR #2678, opened 2026-09-08
MSL: Fix layout of struct arrays whose ArrayStride exceeds the packed struct size

hmennen90:
## Problem

A std140 struct array such as

```glsl
struct SpotLight { vec3 position; float range; vec3 direction; float angle; vec3 color; float intensity; float penumbra; };
layout(std140, binding = 0) uniform UBO { SpotLight spot_lights[4]; int spot_light_count; vec3 albedo; float roughness; } ubo;
```

has `ArrayStride 64` but packs to 52 bytes in MSL. When the array is indexed dynamically and followed by further members, the emitted MSL is laid out wrong in one of two ways:

1. `SpotLight spot_lights[4]; char _m1_pad[48];` — native stride 52, so `spot_lights[1..3]` read from the wrong offsets. Cause: `mark_scalar_layout_structs()` runs before the element struct is packed by its own `align_struct()` pass, compares the ArrayStride against the *unpacked* size (64 == 64) and never marks the struct for `spvPaddedArrayElement`.
2. `spvPaddedArrayElement<SpotLight, 64> spot_lights[4]; char _m7_pad[48];` — happens when a later compilation pass does switch to the padded element (observed with a larger shader that triggers another recompile). The `SPIRVCrossDecorationPaddingTarget` recorded for the following member in the first pass (48 bytes from the 4×52 layout) is never cleared, so every member after the array is shifted by 48 bytes.

Both were found through a GLSL→SPIR-V→MSL pipeline where uniforms written at their SPIR-V offsets (`albedo`, `roughness`, …) were read back from the wrong location on Metal.

## Fix

* `align_struct()` re-checks the packed struct size after `ensure_member_packing_rules_msl()` and, when the declared ArrayStride is larger, sets `DecorationArrayStride` on the struct and requests `SPVFuncImplPaddedArrayElement` — the same mechanism `mark_scalar_layout_structs()` uses, just at a point where the packed size is known.
* `align_struct()` unsets `SPIRVCrossDecorationPaddingTarget` before recomputing it, so a stale value from an earlier pass cannot survive a recompile.

## Test

`shaders-msl/frag/struct-array-stride-padded-element.frag` (dynamically indexed struct array followed by int/vec3/float members) plus references. The full `--msl` and `--msl --opt` reference suites pass with no other reference changing.

### DIFF
--- a/reference/opt/shaders-msl/frag/struct-array-stride-padded-element.frag
+++ b/reference/opt/shaders-msl/frag/struct-array-stride-padded-element.frag
@@ -1,0 +1,52 @@
+#pragma clang diagnostic ignored "-Wmissing-prototypes"
+
+#include <metal_stdlib>
+#include <simd/simd.h>
+
+using namespace metal;
+
+template <typename T, int stride>
+struct spvPaddedArrayElement { T data; char padding[stride - sizeof(T)]; };
+
+struct SpotLight
+{
+    packed_float3 position;
+    float range;
+    packed_float3 direction;
+    float angle;
+    packed_float3 color;
+    float intensity;
+    float penumbra;
+};
+
+struct UBO
+{
+    spvPaddedArrayElement<SpotLight, 64> spot_lights[4];
+    int spot_light_count;
+    char _m2_pad[12];
+    packed_float3 albedo;
+    float roughness;
+    float alpha;
+};
+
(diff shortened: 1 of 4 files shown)


HansKristian-Work (approved):
Makes sense. Thanks.

PR #2681, opened 2026-09-17
MSL: Fix invalid pointer constructor when unrolling row-major column …

MouseChannel:
When unrolling a column load from a row-major matrix with a non-forwarded access chain (e.g., volatile loads or forced temporaries), `convert_row_major_matrix` was called with the pointer type instead of the value type. 

In MSL (which emits pointer types), this caused `type_to_glsl_constructor` to generate an invalid pointer constructor like `constant float3*(...)`.

What I did:

- Extract the value type using `get_pointee_type(exp_type)` before building the constructor expression in `convert_row_major_matrix`.
- Add MSL regression test in `shaders-msl-no-opt/asm/vert/row-major-column-load.asm.vert`.


HansKristian-Work (approved):

HansKristian-Work:
Solved it in a more direct way since pointer types should never have reached down here in the first place, but thanks for the find.

